PR #42, opened 2022-07-29
Add VS code settings.json file

### LINKED ISSUE #27: Common VSCode settings
How to setup the cod editor is a rather personal thing, but it is useful to have a common set of settings or tools adding the coding. For the `.vscode/settings.json` file, I would suggest to have the following:

```json
{
    "python.linting.pylintEnabled": true,
    "python.linting.flake8Enabled": true,
    "python.linting.mypyEnabled": true,
    "python.linting.enabled": true,
    "python.formatting.provider": "black",
    "editor.formatOnSave": true,
    "workbench.tree.indent": 24,
    "editor.rulers": [
        88
    ],
    "files.autoSave": "afterDelay",
    "python.testing.unittestEnabled": false,
    "python.testing.pytestEnabled": true,
    "python.testing.pytestArgs": [
        "-s",
        "tests"
    ],
    "workbench.editorAssociations": {
        "*.ipynb": "jupyter-notebook"
    },
    "rewrap.autoWrap.enabled": true,
    "editor.wordWrapColumn": 88,
```

The suitability of these options should be discussed and some of them might need further configuration in VSCode, like installing another extension or a dev-dependency.

alexdewar:
So I started out by adding the settings.json file suggested by @dalonsoa, but I've made
a few changes:
* Removed ``workbench.tree.ident`` and ``files.autosave`` settings as I think those
  aren't specific to this repo (users might want different settings)
* Remove options relating to line rewrapping (``black`` does this already)
* Remove ``pytest`` settings: this should be done in a project-wide config file (e.g.
  setup.cfg)
* Add ``--no-cov`` to ``pytest`` args otherwise breakpoints won't work (see https://github.com/microsoft/vscode-python/issues/693)

Closes #27.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,4 @@
 {
-    "python.linting.pylintEnabled": true,
     "python.linting.flake8Enabled": true,
     "python.linting.mypyEnabled": true,
     "python.linting.enabled": true,
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,24 @@
 {
-    "python.testing.pytestArgs": [
-        "."
+    "python.linting.pylintEnabled": true,
+    "python.linting.flake8Enabled": true,
+    "python.linting.mypyEnabled": true,
+    "python.linting.enabled": true,
+    "python.formatting.provider": "black",
+    "editor.formatOnSave": true,
+    "workbench.tree.indent": 24,
+    "editor.rulers": [
+        88
     ],
+    "files.autoSave": "afterDelay",
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    "python.testing.pytestArgs": [
+        "-s",
+        "tests"
+    ],
+    "workbench.editorAssociations": {
+        "*.ipynb": "jupyter-notebook"
+    },
+    "rewrap.autoWrap.enabled": true,
+    "editor.wordWrapColumn": 88
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,20 +5,16 @@
     "python.linting.enabled": true,
     "python.formatting.provider": "black",
     "editor.formatOnSave": true,
-    "workbench.tree.indent": 24,
     "editor.rulers": [
         88
     ],
-    "files.autoSave": "afterDelay",
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
+    // --no-cov needed so that breakpoints work: https://github.com/microsoft/vscode-python/issues/693
     "python.testing.pytestArgs": [
-        "-s",
-        "tests"
+        "--no-cov"
     ],
     "workbench.editorAssociations": {
         "*.ipynb": "jupyter-notebook"
     },
-    "rewrap.autoWrap.enabled": true,
-    "editor.wordWrapColumn": 88
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,8 +10,10 @@
     ],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    // --no-cov needed so that breakpoints work: https://github.com/microsoft/vscode-python/issues/693
     "python.testing.pytestArgs": [
+        "-s",
+        "tests",
+        // --no-cov needed so that breakpoints work: https://github.com/microsoft/vscode-python/issues/693
         "--no-cov"
     ],
     "workbench.editorAssociations": {


davidorme:
Just to note that my plan is _not_ to use the *.ipynb format for Jupyter notebooks. See: 
https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/develop/docs/source/development/jupyter_notebooks.md

I don't know if that changes the settings.

.vscode/settings.json
```diff
@@ -1,7 +1,20 @@
...
+    "python.linting.flake8Enabled": true,
+    "python.linting.mypyEnabled": true,
+    "python.linting.enabled": true,
+    "python.formatting.provider": "black",
+    "editor.formatOnSave": true,
+    "editor.rulers": [
+        88
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
```
dalonsoa, line 6: This should be here, otherwise VSCode testing does not know what to use for tests.

alexdewar:
Overview of VS code settings: https://code.visualstudio.com/docs/getstarted/settings

dalonsoa (changes requested):
The testing options are for VSCode. Just `setup.cfg` or `pyproject.toml` is not enough.

.vscode/settings.json
```diff
@@ -1,7 +1,20 @@
...
+    "python.linting.pylintEnabled": true,
+    "python.linting.flake8Enabled": true,
+    "python.linting.mypyEnabled": true,
+    "python.linting.enabled": true,
+    "python.formatting.provider": "black",
+    "editor.formatOnSave": true,
+    "editor.rulers": [
+        88
     ],
     "python.testing.unittestEnabled": false,
```
jaideep777, line 11: Do we want this to be 'true'?

.vscode/settings.json
```diff
@@ -1,7 +1,20 @@
...
-    "python.testing.pytestEnabled": true
+    "python.testing.pytestEnabled": true,
+    // --no-cov needed so that breakpoints work: https://github.com/microsoft/vscode-python/issues/693
```
jaideep777, line 13: Is this a comment?

.vscode/settings.json
alexdewar: It still is set to ``true``! See below

jaideep777 (changes requested):
Submitting a sample review.

.vscode/settings.json
alexdewar: I don't think so.

Basically there are multiple unit testing frameworks available in Python and I *think* you guys are using ``pytest``, rather than the built-in ``unittest`` one.

@davidorme Have I got that right?

jaideep777: Got it! So I am resolving this conversation.

alexdewar: > The testing options are for VSCode. Just `setup.cfg` or `pyproject.toml` is not enough.

So I *think* we can drop the ``pytest`` arguments ``-s tests`` because if the test directory is set appropriately in e.g. ``setup.cfg`` then I think VS code will detect it (and so the test tab will work properly) and if it's only set in ``settings.json`` then VS code's test tab will work, but confusingly, it won't work when you run ``pytest`` from the command line (or when the CI system runs it).

.vscode/settings.json
alexdewar: Yep!

alexdewar:
> The testing options are for VSCode. Just `setup.cfg` or `pyproject.toml` is not enough.

So I *think* we can drop the ``pytest`` arguments ``-s tests`` because if the test directory is set appropriately in e.g. ``setup.cfg`` then I think VS code will detect it (and so the test tab will work properly) and if it's only set in ``settings.json`` then VS code's test tab will work, but confusingly, it won't work when you run ``pytest`` from the command line (or when the CI system runs it).

dalonsoa:
Ok, we can leave it like that and change it if it does not work

.vscode/settings.json
```diff
@@ -1,7 +1,22 @@
 {
-    "python.testing.pytestArgs": [
-        "."
+    "python.linting.pylintEnabled": true,
```
alexdewar, line 2: It looks like ``pylint`` isn't in the ``poetry`` dev dependencies btw. Do we want to enable it or not?

I suggest we either don't use this linter or add it to the dependencies.

@dalonsoa @davidorme

alexdewar:
> Just to note that my plan is _not_ to use the *.ipynb format for Jupyter notebooks. See: https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/develop/docs/source/development/jupyter_notebooks.md
> 
> I don't know if that changes the settings.

We could remove the setting if it's not needed. It shouldn't break things though!

I leave that decision to the VR dev team. Would you like me to remove it or no? :smile:

.vscode/settings.json
dalonsoa: If we want to use it - and it seems we want, right @davidorme - it should be in the dependency list, yes.

.vscode/settings.json
dalonsoa: omg

dalonsoa (approved):
Keeping aside adding `pylint` to the dependency list, it looks good to me

jaideep777 (approved):

PR #59, opened 2022-08-12
Core test

### LINKED ISSUE #31: Add tests for functions/methods in grid.py
There is no testing of the functions in `grid.py`.  **Unit tests** should be added:

- [x] `CoreGrid` constructor
- [x] `make_square_grid`
- [x] `make_hex_grid` 
- [ ] `export_geojson` 
- [ ] Once it is extracted from where it is, `_feature`

vgro:
# Description

Added tests for CoreGrid constructor, make_square_grid, and make_hex_grid functions in core.py. 

Fixes #31 (partly) 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -1,10 +1,53 @@
-from virtual_rainforest.core.grid import CoreGridConfig
+import pytest
+from scipy.spatial.distance import euclidean
+
+from virtual_rainforest.core.grid import CoreGrid, CoreGridConfig
 
 
 def test_CoreGridConfig():
+    """test to Configure the `core.grid` module
+
+    This data class is used to setup the arrangement of grid cells to be used in
+    running a `virtual_rainforest` simulation.
+
+    Attrs:
+        grid_type:
+        cell_contiguity:
+        cell_area:
+        cell_nx:
+        cell_ny:
+    """
     obj_1 = CoreGridConfig()
     assert obj_1.grid_type == "square"
     assert obj_1.cell_contiguity == "rook"
     assert obj_1.cell_area == 100
     assert obj_1.cell_nx == 10
     assert obj_1.cell_ny == 10
+
+
+def test_make_square_grid():
+    """test Create a square grid by comparing eucledian distance between corner points
+
+    Args:
+        cell_area:
+        cell_nx:
+        cell_ny:
+    """
+    test = CoreGrid._make_square_grid(100, 10, 10)
+    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
+    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
+    assert distance == distance2
+
+
+def test_make_hex_grid():
+    """test Create a hex grid by comparing eucledian distance between corner points
+
+    Args:
+        cell_area:
+        cell_nx:
+        cell_ny:
+    """
+    test = CoreGrid._make_hex_grid(100, 10, 10)
+    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
+    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
+    distance == pytest.approx(2 * distance2)
--- a/tests/test_grid.py
+++ b/tests/test_grid.py
@@ -52,12 +52,20 @@ def test_make_square_grid():
 
     test = CoreGrid._make_square_grid(100, 10, 10)
 
-    # TODO check if correct number of points were created
+    # check if correct number of points were created
+    assert len(test.keys()) == 100
 
-    # TODO check that distance between neighboring points is euqal for all integers
-    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
-    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
-    assert distance == distance2
+    # check that distance between neighboring points is euqal for all integers
+    # (this works but is a bit long)
+    for x in range(9):
+        for y in range(9):
+            distance = euclidean(
+                test[f"{x}-{y}"]["poly"][0], test[f"{x}-{y}"]["poly"][2]
+            )
+            distance2 = euclidean(
+                test[f"{x}-{y}"]["poly"][1], test[f"{x}-{y}"]["poly"][3]
+            )
+            assert distance == pytest.approx(distance2)
     # TODO check that angle between points = 90 deg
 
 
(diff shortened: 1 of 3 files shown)


tests/test_grid.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        cell_area:
+        cell_nx:
+        cell_ny:
+    """
+    test = CoreGrid._make_square_grid(100, 10, 10)
+    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
+    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
+    assert distance == distance2
```
alexdewar, line 39: You might want to make this test ``approx`` too

jacobcook1995:
It seems like you are falling foul of the flake8's doctoring checks, so it isn't passing CI. Me and David turned off doc string checks for tests in `safedata_validator`, with the logic that tests are not public functions. I don't know if this is good/best practice though?

tests/test_grid.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Args:
+        cell_area:
+        cell_nx:
+        cell_ny:
+    """
+    test = CoreGrid._make_hex_grid(100, 10, 10)
+    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
+    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
+    distance == pytest.approx(2 * distance2)
```
alexdewar, line 53: Why ``2 * distance2``? I'm sure it's obvious but it might be useful to have a comment for those of us who can't visualise it :smile:

alexdewar:
@jacobcook1995 I think there are two problems here: one is that of missing docstrings, the other is of malformed docstrings. You might want to disable the former for tests (though personally I wouldn't -- you really should have a comments saying what each of your tests do anyway, so why not just make them docstrings?) but I think malformed docstrings are a code style problem like any other and should just be fixed.

Most of the "public" functions you have won't really be used by end users anyway, but it's still worth having docstrings for them as internal documentation. Same applies to tests.

tests/test_grid.py
vgro: ok, thank you

tests/test_grid.py
```diff
@@ -52,12 +52,20 @@ def test_make_square_grid():
...
-    # TODO check that distance between neighboring points is euqal for all integers
-    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
-    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
-    assert distance == distance2
+    # check if correct number of points were created
+    assert len(test.keys()) == 100
+
+    # check that distance between neighboring points is euqal for all integers
+    # (this works but is a bit long)
+    for x in range(9):
```
dalonsoa, line 60: It is not necessary to test every possible combination to have some reassurance that this is correct. What I would do is to pick 3 random values for x and y within the [0, 9] range and check if this condition is fulfilled for them. That would be 10 times faster and I would say it is good enough.

tests/test_grid.py
```diff
@@ -67,10 +75,25 @@ def test_make_hex_grid():
...
-    distance = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][1])
-    distance2 = euclidean(test["0-0"]["poly"][0], test["0-0"]["poly"][3])
-    distance == pytest.approx(2 * distance2)
+
+    # check if correct number of points were created
+    assert len(test.keys()) == 100
+
+    # check that distance between neighboring points is euqal for all integers
+    # (this works but is a bit long)
+    for x in range(9):
```
dalonsoa, line 85: As above.

tests/test_grid.py
```diff
@@ -0,0 +1,99 @@
...
+
+    obj_1 = CoreGridConfig()
+    assert obj_1.grid_type == "square"
+    assert obj_1.cell_contiguity == "rook"
+    assert obj_1.cell_area == 100
+    assert obj_1.cell_nx == 10
+    assert obj_1.cell_ny == 10
+
+
+def test_make_square_grid():
```
dalonsoa, line 47: Using the [`hypothesis`](https://hypothesis.readthedocs.io/en/latest/index.html) tool I mentioned will make this look like this:

```python
from hypothesis import given
from hypothesis.strategies import integers

@given(integers(min_value=0, max_value=9), integers(min_value=0, max_value=9))
def test_make_square_grid(x, y):
    from scipy.spatial.distance import euclidean  # type: ignore

    from virtual_rainforest.core.grid import CoreGrid

    test = CoreGrid._make_square_grid(100, 10, 10)

    # check if correct number of points were created
    assert len(test.keys()) == 100
    
    distance = euclidean(test[f"{x}-{y}"]["poly"][0], test[f"{x}-{y}"]["poly"][2] )
    distance2 = euclidean( test[f"{x}-{y}"]["poly"][1], test[f"{x}-{y}"]["poly"][3])
    assert distance == pytest.approx(distance2)
```
It is more concise that the loop version, more thorough than using just a few random points and way, way faster.

alexdewar:
It looks like the tests are failing on the CI system because ``scipy`` is not being installed (though presumably it is installed on your local machine).

You need to run ``poetry add scipy`` and commit the result.

tests/test_grid.py
```diff
@@ -55,22 +58,40 @@ def test_make_square_grid():
...
-
-def test_make_hex_grid():
-    """Test _make_hex_grid()."""
+    # check distance between corner points is the same
+    distance = euclidean(test[f"{x}-{y}"]["poly"][0], test[f"{x}-{y}"]["poly"][2])
+    distance2 = euclidean(test[f"{x}-{y}"]["poly"][1], test[f"{x}-{y}"]["poly"][3])
+    assert distance == pytest.approx(distance2)
+
+    # check angle between corner points is the same
+    from math import atan2, degrees, pi
```
dalonsoa, line 67: A simpler approach would be to calculate the inner product of the two segments and check if it is zero. So, if AB and AC are numpy arrays that go from point A to B and point A to C respectively, then they will form a ±90 deg angle if:

```python
assert np.inner(AB, AC) == pytest.approx(0)
```
If you need specifically the angle (which you do for the hexagonal one), you can follow an approach like this: https://www.atqed.com/numpy-vector-angle

tests/test_grid.py
vgro: I changed the approach so no longer relevant. But it is a good point to add comments here :-)

tests/test_grid.py
vgro: I implemented hypothesis, so no longer relevant. However, very useful to know that tests don't need to be so extensive :-)

tests/test_grid.py
vgro: I agree this looks much better. I implemented the change and added hypothesis to our package

tests/test_grid.py
```diff
@@ -0,0 +1,141 @@
...
+
+    obj_1 = CoreGridConfig()
+    assert obj_1.grid_type == "square"
+    assert obj_1.cell_contiguity == "rook"
+    assert obj_1.cell_area == 100
+    assert obj_1.cell_nx == 10
+    assert obj_1.cell_ny == 10
+
+
+@given(integers(min_value=0, max_value=9), integers(min_value=0, max_value=9))
```
dalonsoa, line 49: Tests in MacOS are failing for some uinknown reason - I can reproduce this in my laptop. To solve this, remove the deadline limitation for this test:

```python
from hypothesis import given, settings
...
@settings(deadline=None)
@given(integers(min_value=0, max_value=9), integers(min_value=0, max_value=9))
def test_make_square_grid(x, y):
    ...
```

vgro: It's the same for me, it passes all the tests on my laptop. I will try your suggestions

vgro: all tests pass now.

jacobcook1995 (approved):

tests/test_grid.py
vgro: I replaced the initial checks for angles with this approach, thanks for the suggestion

jacobcook1995 (approved):

PR #65, opened 2022-08-24
Adding design notes on core data setup and configuration

### LINKED ISSUE #64: Proposal for data configuration system and data generation
We need an initial implementation of how to configure and reference data within the simulation.

This should include the ability to reference spatial data in a variety of configurations (cell by cell or mapped to spatial categories) and should include the ability to use artificially generated data for testing.

davidorme:
This pull request provides a proposal for the data initialisation process.

Fixes #64 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/design/core.md
+++ b/docs/source/development/design/core.md
@@ -63,7 +63,7 @@ useful (and all of these could add a time dimension):
 
 #### Two dimensial spatial indexing
 
-- Simple indices (`idx_n` and `idx_y`) giving the index of a square grid. The data
+- Simple indices (`idx_x` and `idx_y`) giving the index of a square grid. The data
   should match the configured grid size and only really works for square grids (it could
   just about work for a hex grid with alternate offsets!).
 - Standard `x` and `y` coordinates (or `easting` and `northing`). This would require
@@ -83,7 +83,30 @@ useful (and all of these could add a time dimension):
 ### Defining mappings
 
 A user can optionally configure mappings, which are loaded and validated before any
-other data. These use one of the first three approaches above, which index _individual_
+other data. These _have_ to use one of the methods that provides values to every cell,
+so either of the 2D approaches or using `cell_id`. So for example, the array below might
+define an arrangement of a gradient between different levels of forest cover: matrix to
+logged to forest.
+
+```toml
+[core.data.forest_cover]
+values  = [['M', 'M', 'L'],
+          ['L', 'L', 'F'],
+          ['L', 'F', 'F']]
+```
+
+Then you could have a separate variable that uses that variable to map values per class
+onto the spatial grid. The config structure here is in _no way fixed_ - it is the
+concept that matters!
+
+```toml
+[core.data.soil_depth]
+values = {forest_cover.M=0.1, 
+          forest_cover.L=0.5, 
+          forest_cover.F=1.0}
+```
+
+These use one of the first three approaches above, which index _individual_
 cells, to provide the spatial layout of a categorical variable. For example, it could
 use `x` and `y` coordinates to map `habitat` with values `forest`, `logged_forest` and
 `matrix`.
@@ -104,8 +127,9 @@ var = habitat
 Once any mappings have been established, then the configuration defines the source file
 and variable name for required forcing variables. The dimension names are used to infer
 how to spatially index the data and we have a restricted set of dimension names that
-must be used to avoid ambiguity. We also need to have `time` and `depth` as reserved
-dimension names but the spatial indexing uses:
+must be used to avoid ambiguity. We also need to have other reserved dimension names
+(for example, `time`, `depth`, `height`) but spatial indexing expects the following
+dimension names to define particular indexing approaches:
 
 - `x`, `y`: use coordinates to match data to cells,
 - `idx_x`, `idx_y`: just use indices to match data to cells,
@@ -152,9 +176,6 @@ It would be good if these could be set via configuration but also use the same
 functionality to create a NetCDF output. That _should_ effectively be the same as
 configuring the data generator with a set random number generator seed.
 
-These could get arbitrarily complex - so at some point we should just say, if you want
-sufficiently complex generated data, just roll your own NetCDF files!
-
 I think this can basically just use all the options of `numpy.random`, possibly with the
 inclusion of interpolation along a time dimension at a given interval if a time axis is
 present.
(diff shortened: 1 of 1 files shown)


docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+`matrix`.
+
+When loading **data**, a variable could now use that mapping variable to unpack values
+for each of the three categories into cells.
+
+These are configured as an array of tables to support multiple mappings:
+
+```toml
+[[core.mappings]]
+file = /path/to/netcdf.nc
```
jacobcook1995, line 98: Here and in a few other places text is highlighted in red is that intentional?

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+
+If a variable only has a `time` or `depth` dimension, it is assumed to be spatially
+constant.
+
+These are configured directly to defined input slots. The config should also accept
+values directly to avoid having to create NetCDF files with trivial variables.
+
+```toml
+[core.data.precipitation]
+file = /path/to/netcdf.nc
```
jacobcook1995, line 123: Again is the red highlight intentional?

jacobcook1995 (commented):
Made a couple of comments, also am wondering whether this actually should close issue #64, which it currently will automatically when merged

docs/source/development/design/core.md
davidorme: Huh. I think that is just GitHub Markdown flavour not handling the `toml` section.

davidorme:
That's a good point about closing - I intended the issue just to provide the proposal for discussion, but it might as well serve to cover the implementation.

This would have been better as a GH discussion!

docs/source/development/design/core.md
jacobcook1995: Yeah I saw a similar weirdness on an issue on one of the toml repos, so guess it must be that

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+
+### Spatial indices
+
+Data is (always?) associated with a grid cell, so the system needs to be able to match
+data values to the cell id of grid cells. The following spatial indexing options seem
+useful (and all of these could add a time dimension):
+
+#### Two dimensial spatial indexing
+
+- Simple indices (`idx_n` and `idx_y`) giving the index of a square grid. The data
```
vgro, line 66: do you mean **idx_x** and idx_y?

davidorme: I do 😄

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+  dimension that covers all of the configure cell ids.
+
+- A **mapping** of particular attributes to sets of cells. The obvious use case here is
+  habitat type style data - for example different soil bulk densities or plant
+  communities. To unpack that a bit:
+
+### Defining mappings
+
+A user can optionally configure mappings, which are loaded and validated before any
+other data. These use one of the first three approaches above, which index _individual_
```
vgro, line 86: This is a bit unclear. You presented two approaches (2d or 1d), could you please clarify?

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+file = /path/to/netcdf.nc
+var = habitat
+```
+
+### Loading data
+
+Once any mappings have been established, then the configuration defines the source file
+and variable name for required forcing variables. The dimension names are used to infer
+how to spatially index the data and we have a restricted set of dimension names that
+must be used to avoid ambiguity. We also need to have `time` and `depth` as reserved
```
vgro, line 107: I think the 'depth' dimension needs a different name to better capture the above ground levels, or we might add a 'height' dimension, too.

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+
+[core.data.elevation]
+values = [[1,2,3], [2,3,4], [3,4,5]]
+```
+
+### Data Generator
+
+It seems useful to have a `DataGenerator` class that can be used via the configuration
+to provide random or constant data.
+
```
vgro, line 141: could you be more explicit what sort of data? e.g. line 146 'optionally some kind of variation'

docs/source/development/design/core.md
davidorme: Yup - basically you can't use a mapping to define a mapping. You can _only_ use 2d, or a 1d set of cell_ids.

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+- optionally a time axis,
+- optionally some kind of cycle,
+- optionally some kind of probability of different states.
+
+It would be good if these could be set via configuration but also use the same
+functionality to create a NetCDF output. That _should_ effectively be the same as
+configuring the data generator with a set random number generator seed.
+
+These could get arbitrarily complex - so at some point we should just say, if you want
+sufficiently complex generated data, just roll your own NetCDF files!
```
vgro, line 156: Could you please explain this a bit more? When would this apply?

docs/source/development/design/core.md
davidorme: Makes sense. I don't know if any of our input data will have height strata - there certainly will be _within_ the simulation, but we might do at a later stage.

docs/source/development/design/core.md
vgro: ok, then maybe change it to 'one of the ### two approaches above' ?

docs/source/development/design/core.md
vgro: Yes that is true. I just thought if we set up the grid we might as well include all dimensions.

docs/source/development/design/core.md
```diff
@@ -46,3 +46,130 @@ The system should have:
...
+
+- Simply using a `cell_id` dimension in the NetCDF file to match data to grid cells.
+  This is entirely agnostic about the grid shape - users just need to provide a
+  dimension that covers all of the configure cell ids.
+
+- A **mapping** of particular attributes to sets of cells. The obvious use case here is
+  habitat type style data - for example different soil bulk densities or plant
+  communities. To unpack that a bit:
+
+### Defining mappings
```
vgro, line 83: How is 'mapping' different from 'spacial indices'?

docs/source/development/design/core.md
davidorme: The model I have in my head is based around the `numpy.random` methods (https://numpy.org/doc/stable/reference/random/generator.html).

A user could provide:
* A constant (or constant grid), but this would just be set up as values.
* A set of values (single or grid) _with variation_. So here a DataGenerator might be:

```python
 ex1 = DataGenerator(loc=5, scale=2, distribution='normal')
 ex2 = DataGenerator(mean=5, sigma=2, distribution='lognormal')
```

* A time series of values with variation. Here, you'd need a time axis giving the temporal location of the sampling points, which could be interpolated if necessary. So for example, a 2 x 2 grid that increases in loc and scale over a year.

```python
loc = [[[1, 2],
        [3, 4]],
       [[2, 3],
        [4, 5]]]

scale = [[[1, 1],
          [1, 1]],
         [[1.2, 1.2],
          [1.2, 1.2]]]

time = ['2020-01-01', '2020-12-31']

ex3 = DataGenerator(loc = loc, scale=scale, method='normal', time=time, time_axis=2)
```

docs/source/development/design/core.md
davidorme: Hard to define! But if someone wanted to simulate a scenario of 5 drought years followed by 3 years of elevated rainfall then an El Niño and then a plague of frogs - that's probably outside the scope 😄

vgro (approved):
Looks very good, clear starting point to design the core set up and config. 
Maybe we can start a discussion for more details on data inputs types/format and grid shapes?

docs/source/development/design/core.md
davidorme: OK - you could use a 2D grid to define this arrangement of a gradient from matrix to logged to forest:
```toml
[core.data.habitat]
values  = [['M', 'M', 'L'],
          ['L', 'L', 'F'],
          ['L', 'F', 'F']]
```
Then you could have a separate variable that uses that mapping:

```toml
[core.data.soil_depth]
values = {M=0.1, L=0.5, F=1.0}
```
And we'd then unpack that to give the grid.

vgro (approved):
Thanks for adding examples and clarification ( I'm just making sure it's idiot proof ;-) ). I think the document is very clear now, ready to merge from my point of view 👍

jacobcook1995 (approved):

PR #69, opened 2022-09-02
Restructure of grid creation code

### LINKED ISSUE #28: Enable adding new grids
At the moment, there are only square and hex grids, but I can imagine - maybe erroneously - other kinds in the future. To enable adding new grids in an easy way, including by the user externally to `virtual_rainforest`, a method for registering valid grids should be put in place. 

The canonical - or at least one of the most popular ones - way of doing this in Python is using decorators. A couple of examples on using this pattern can be found in:

- [pycsvy/writers.py at develop · ImperialCollegeLondon/pycsvy (github.com)](https://github.com/ImperialCollegeLondon/pycsvy/blob/develop/csvy/writers.py)
- [solcore6/calculable_parameters.py at develop · qpv-research-group/solcore6 (github.com)](https://github.com/qpv-research-group/solcore6/blob/develop/solcore/parameter_sources/calculable_parameters.py)

And a fantastic crash course on decorators can be found here - read it: it is very likely you will use it somewhere in the project:

[Primer on Python Decorators – Real Python](https://realpython.com/primer-on-python-decorators/)

The steps are:

- [ ] Create a registry, typically just a simple dictionary. This should include an explicit signature for the function that will be registered.
- [ ] Create a decorator that registers a function within that registry. The function should follow the required signature.
- [ ] Use the registry and the input to call the relevant function.

Eg. For the case of the grid, the end code using the registry should look something like:

```python
grid = GRID_REGISTRY.get(config.grid_type, None)

if grid:
	self.cell_dict = grid(
	                cell_area=config.cell_area,
	                cell_nx=config.cell_nx,
	                cell_ny=config.cell_ny,
	                )
else:
	# Deal with the None case if not having a grid is invalid.
```

`pydantic` will only accept two explicit grid types, which makes expanding with new options in the future more difficult as the code needs to be modified in several places. It is more sustainable to use the registry and deal with the None case appropriately. When using `pydantic` there is a risk - obviously - of being pedantic and overuse the type checking. 

Registering a new grid will now look something like:

```python
@register_grid(name="triangular")
def make_triangular_grid(cell_area: float, cell_nx: int, cell_ny: int) -> dict:
	pass
```

Mind that this registration can be done *outside* of the virtual rainforest codebase so it makes it very, very easy to try out new things, including by end users.

In any case, making the grids should not be static methods but independent functions, simplifying their testing.

davidorme:
# Description

This PR refactors the grid code for three main reasons:

* Removing the dependency on `fiona` - while excellent, the `conda` dependent install for this is too high a cost. The `Shapely` package loses some IO options for a wide range of GIS formats but provides extensive geometry operations.
* Establishing `GRID_REGISTRY` and a decorator to populate it:  a user extendable dictionary of grid layouts.
* Provide a better signature for the main Grid (was CoreGrid) class that exposes the actual arguments and not a single `config` object.

The changes also then updates the existing `test_grid.py` to use the new structure, including moving imports outside of test functions. `pyproject.toml` and `poetry.lock` are updated to replace `fiona` (and tidy up a few misplaced / extraneous dev dependencies).

Fixes #28

## Type of change

## Key checklist

## Further checks


vgro:
Thanks for the update David. A few comments:
- great that you found a way to work around fiona, this will make it easier for windows user
-  I am not sure what the advantage of the grid registry and the decorator is compared to the previous approach, I personally find it easier to understand how things are nested by using the dot operator. Apart from that, it looks correct as far as I can tell
- I like the way you changed the Grid object, it is much more intuitive now (without the config)
From my side it looks all good, I think it would be good to wait for the experts' comments before merging :-)

tests/test_grid.py
```diff
@@ -6,103 +6,86 @@
...
+- TODO grid.Grid.dump and grid.Grid.dumps for0 geojson
 
 """
 
+import numpy as np
+import numpy.linalg as LA
 import pytest
 from hypothesis import given, settings
 from hypothesis.strategies import integers
+from scipy.spatial.distance import euclidean  # type: ignore
```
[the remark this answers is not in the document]
davidorme: That's interesting.  I like the clarity of seeing which bits of the package are being tested at the top of the file, but I think I can see a point for imports from the _package being tested_ inside the test function. If the import doesn't work then only the test using that import fails rather than the whole test file and the test is effectively also testing that the objects import correctly. However that does seem like quite a fundamental problem - if some imports are broken, the rest of the file completing is a bit secondary!

 For other imports, having them once at the top of the file seems cleaner?

davidorme:
@vgro The main advantage of the registry is that users can extend the set of grids. They could write:

```python
@register_grid('lspace')
def make_l_shaped_grid(...):
```

Honestly, once we've got the three regular polygons that tesselate, it isn't obvious what the use case would be, although I can sort of imagine a situation where someone uses an irregular polygon map. We might want to keep in mind use cases with focal areas of differing size and shape?

tests/test_grid.py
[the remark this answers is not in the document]
dalonsoa: I tend to import everything but pytest-spcecific stuff (and hypothesis) within each test to make sure everything is reset as much as possible in between tests. It makes it clearer which tests needs what package. And finally, also, facilitates mocking, as you cannot mock something if is already imported.

Having said that, I think it is fine to import general tools like numpy or scipy at the top.

tests/test_grid.py
```diff
@@ -6,103 +6,86 @@
...
-    """
-    from virtual_rainforest.core.grid import CoreGridConfig
-
-    obj_1 = CoreGridConfig()
-    assert obj_1.grid_type == "square"
-    assert obj_1.cell_contiguity == "rook"
-    assert obj_1.cell_area == 100
-    assert obj_1.cell_nx == 10
-    assert obj_1.cell_ny == 10
+from virtual_rainforest.core.grid import make_hex_grid, make_square_grid
```
dalonsoa, line 28: This should definitely go inside each test

tests/test_grid.py
```diff
@@ -6,103 +6,86 @@
...
-    distance2 = euclidean(test[f"{x}-{y}"]["poly"][1], test[f"{x}-{y}"]["poly"][3])
-    assert distance == pytest.approx(distance2)
+    # check number of vertices in outer ring is correct
+    xy = np.array(test[f"{x}-{y}"].exterior.coords)
+    assert xy.shape == (5, 2)
+
+    # check that distance between corner points is the same
+    # fmt: off
+    distance = euclidean(xy[0, ], xy[2, ])
+    distance2 = euclidean(xy[1, ], xy[3, ])
```
dalonsoa, line 48: This looks odd. Why not just do the following and remove the skip formatting tags? The result is the same, and seems a lot cleaner. 
```suggestion
    distance = euclidean(xy[0], xy[2])
    distance2 = euclidean(xy[1], xy[3])
```
And the same below.

tests/test_grid.py
```diff
@@ -6,103 +6,86 @@
...
 
-    import numpy as np
-
-    A = np.array(test[f"{x}-{y}"]["poly"][0])
-    B = np.array(test[f"{x}-{y}"]["poly"][1])
-    C = np.array(test[f"{x}-{y}"]["poly"][3])
-
-    assert np.inner(B - A, C - A) == pytest.approx(0)
+    # fmt: off
+    iprod = np.inner(xy[1, ] - xy[0, ], xy[3, ] - xy[0, ])
```
dalonsoa, line 57: Same here.

tests/test_grid.py
```diff
@@ -6,103 +6,86 @@
...
+    # check number of vertices in outer ring is correct
+    xy = np.array(test[f"{x}-{y}"].exterior.coords)
+    assert xy.shape == (7, 2)
 
-    # check that angle between points = 60 deg
-    import numpy as np
-    import numpy.linalg as LA
+    # check that distance between corner points is the same
+    # fmt: off
+    distance = euclidean(xy[0, ], xy[3, ])
```
dalonsoa, line 77: Same idea here

virtual_rainforest/core/grid.py
```diff
@@ -1,203 +1,370 @@
...
     """
 
-    def __init__(self, config: CoreGridConfig) -> None:
+    def decorator_register_grid(func: Callable):
+
+        # Add the grid type to the registry
+        if grid_type in GRID_REGISTRY:
+            LOGGER.warning(
+                "Grid type %s already exists and is being replaced", grid_type
+            )
```
dalonsoa, line 53: I would do this check outside of the inner decorator.

virtual_rainforest/core/grid.py
```diff
@@ -1,203 +1,370 @@
...
+        GRID_REGISTRY[grid_type] = func
+
+        # Pass the function through for use, using @wraps to expose the original
+        # function dunder attributes (__name__, __doc__ etc.)
+        @wraps(func)
+        def wrapper_register_grid(*args, **kwargs):
+
+            return func(*args, **kwargs)
+
+        return wrapper_register_grid
```
dalonsoa, line 63: Why not just returning `func`? This wrapping is needed when you are doing something to the input arguments of `func` or its output, but you want to keep the dunder attributes, as you explain. But that's not the case here. in other words, I think you can safely do:
```suggestion
        return func
```

virtual_rainforest/core/grid.py
```diff
@@ -1,203 +1,370 @@
...
+    """
+
+    # Create the polygon prototype object, with origin at 0,0 and area 1
+    # Note coordinate order is anti-clockwise - right hand rule.
+    prototype = Polygon([[0, 0], [1, 0], [1, 1], [0, 1], [0, 0]])
+
+    # Scale to requested size and origin
+    scale_factor = np.sqrt(cell_area)
+    prototype = scale(prototype, xfact=scale_factor, yfact=scale_factor, origin=(0, 0))
+    prototype = translate(prototype, xoff=xoff, yoff=yoff)
```
dalonsoa, line 93: This is pretty cool!

virtual_rainforest/core/grid.py
```diff
@@ -1,203 +1,370 @@
...
+        xoff: An offset to use for the grid origin in the X direction.
+        yoff: An offset to use for the grid origin in the Y direction.
+    """
+
+    # Note shapely.affinity.rotate for inversion
+
+    raise NotImplementedError()
+
+
+# @dataclass
```
dalonsoa, line 194: Remove wat is not used - in a separate commit, so it can be recovered easily if needed.

virtual_rainforest/core/grid.py
```diff
@@ -1,203 +1,370 @@
...
+    # Note coordinate order is anti-clockwise - right hand rule.
+    prototype = Polygon([[0, 0], [1, 0], [1, 1], [0, 1], [0, 0]])
+
+    # Scale to requested size and origin
+    scale_factor = np.sqrt(cell_area)
+    prototype = scale(prototype, xfact=scale_factor, yfact=scale_factor, origin=(0, 0))
+    prototype = translate(prototype, xoff=xoff, yoff=yoff)
+
+    # Tile the prototypes to create the grid
+    cell_dict = {}
```
dalonsoa, line 96: I wonder if a dict is the best option here (no idea - it depends on the end use case). Personally, I would use a nested list - given that numpy arrays won't work in this case - so I can do simple indexing like `cell[0][3]` without having to create a string key first. Probably a pandas DataFrame would work, so you can access independently whole rows and columns.

But again, it depends on the use case.


```

File: virtual_rainforest/core/grid.py [hunk 11]
```diff
@@ -1,203 +1,370 @@
...
-        cell_nx:
-        cell_ny:
-    """
 
-    grid_type: Literal["hex", "square"] = "square"
-    cell_contiguity: Literal["queen", "rook"] = "rook"
-    cell_area: float = 100
-    cell_nx: StrictInt = 10
-    cell_ny: StrictInt = 10
+def register_grid(grid_type: str):
```
dalonsoa, line 33: I think you need to be explicit here about the return type of this decorator and the inner one so the signature expected for the decorated function is clear.

dalonsoa (changes requested):
I have requested a few changes and made a couple of comments, but otherwise I think it looks cleaner and way easier to maintain and test than having an if statement with several branches.

tests/test_grid.py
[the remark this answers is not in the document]
davidorme: Makes sense. There's a bit of a grey line about which goes where and we should standardise that, but I can clearly see the argument for imports from within the project.

tests/test_grid.py
davidorme: So it is! That's an R/Python thing. If you give an R matrix a single index, then it is an index to a single element, iterating over the axes. I'd forgotten that it could return a slice in Python.

virtual_rainforest/core/grid.py
davidorme: Yeah - was overthinking it.

virtual_rainforest/core/grid.py
davidorme: I'm going to leave it as is for the moment. We need to think about how we structure spatial data internally, and I think there might be good reasons to flatten X and Y and just make this a flat list with a single integer index for location across cells. I agree that using numeric indices and not string keys is cleaner, but I'm going to wait until we have a decision there.

dalonsoa (approved):
All looks good to me, now!

PR #76, opened 2022-09-26
Feature/extend ci

### LINKED ISSUE #75: CI setup to include Windows and wider Python versions.
At the moment, the CI workflow is only testing the code on Mac and Ubuntu under Python 3.9. We definitely want this to cover Windows and more recent versions of Python.

davidorme:
# Description

This PR just extends the CI workflow to include Windows (since #69 removed the requirement for`fiona`) and adds Python 3.10 into the matrix. This is pinned at 3.10.6 because of a `mypy` issue - but a release of `mypy` to allow us to unpin to 3.10 is expected soon.

Fixes #75 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        python-version: [ "3.9" , "3.10" ]
+        python-version: [ "3.9" , "3.10.6" ]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jacobcook1995 (approved):

.github/workflows/ci.yml
```diff
@@ -18,8 +18,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, macos-latest ]
-        python-version: [ 3.9 ]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        python-version: [ "3.9" , "3.10.6" ]
```
dalonsoa, line 22: Do you really need to be that specific about the Python version here? What happens if you just use 3.10?

dalonsoa (approved):

.github/workflows/ci.yml
davidorme: For the moment it breaks because of `mypy` - see f3b5297 and https://github.com/python/mypy/issues/13627.

PR #78, opened 2022-09-26
Basic config system

### LINKED ISSUE #63: Set up a basic config system
We need a consistent way to configure runs of the model.

We ideally want a system that can read in multiple files, and consolidate them into a single configuration object. This would allow users to change small parts of the configuration without having to make edits to a single massive file. We also want to validate the configuration files to ensure that all expected config items are provided, that no items are duplicated, and that all items have the correct type. To do this we will define a `json` schema with the expected types, which will then used to validate the contents of a set of `toml` config files. If validation is successful a combined `toml` config should be saved for future reference, and a `ConfigObject` should be produced which can be used to start the process of initialising a set of model runs.

Considered using `strictyaml` rather than `toml`, but decided against as it seems less well supported. 

**Additional context**
At some point we need to define a command line option that allows the user to set the folder to look for config files in, but that is almost certainly beyond the scope of this issue.

jacobcook1995:
# Description

I've written a basic configuration system. The idea with this system is that it validates the most basic properties of the configuration settings, e.g. is the parameter specifying a path to some climate data a string, leaving more advanced/specific validation to downstream functions.

The system takes in a path to a user provided folder and then either loads in all `toml` files in this folder, or loads in a set of user specified `toml` files from this folder. Checks are then done to ensure that these files exist, are formatted correctly and that the lowest level tags are not defined more than once across the files. The contents of these files are then validated using [`json_schema`](https://python-jsonschema.readthedocs.io/en/stable/). These schema are defined in the `__init__`'s of each module (i.e. `plants` and `core`), and added to a schema registry. The relevant schema for a particular configuration setup are read from this registry and combined into a single schema. This combined schema enforces that properties not included in the schema are invalid. Provided this schema passes the quality/sanity checks, it is then used to validate the configuration data. If this passes the combined configuration data is written out as a `toml` file (adds a requirement for [`tomli_w`](https://pypi.org/project/tomli-w/)), and is also saved to a globally accessible dictionary (called `COMPLETE_CONFIG`).

This system has previously been discussed in #68. I've tried to take the discussion from there onboard but I'm sure there's more that needs doing/agreeing.

Fixes #63 

## Type of change

## Key checklist

## Further checks


virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,254 @@
...
+
+    if len(modules) != len(set(modules)):
+        LOGGER.critical(
+            f"The list of modules to configure given in the core configuration file "
+            f"repeats {len(modules) - len(set(modules))} names!"
+        )
+        return None
+
+    # Construct combined schema for all relevant modules
+    comb_schema: dict = {}
```
dalonsoa, line 188: For this you might want to consider [ChainMap](https://docs.python.org/3/library/collections.html#collections.ChainMap), as a more elegant solution to combine python dictionaries.

virtual_rainforest/plants/__init__.py
```diff
@@ -0,0 +1,43 @@
+from virtual_rainforest.core.config import register_schema
+
+
+@register_schema("plants")
+def schema():
```
dalonsoa, line 5: As above, I will definitely put this in a JSON file and read it from it.

dalonsoa (changes requested):
Overall, the purpose looks good as well as the overall architecture, but the implementation needs some refinements. I will review the tests once my comments are actioned (which might mean ignored, of course).

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,254 @@
...
+
+# Dictionary to store schema registry
+SCHEMA_REGISTRY: dict = {}
+
+
+def register_schema(module_name: str) -> Callable:
+    """Decorator function to add configuration schema to the registry."""
+
+    def wrap(func: Callable):
+        SCHEMA_REGISTRY[module_name] = func()
```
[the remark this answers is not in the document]
davidorme: Or overwrite existing `module name` and warn? That's what I'm currently using in the grid registry.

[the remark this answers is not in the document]
davidorme: I also think this would be a good place to validate the schema itself:

https://python-jsonschema.readthedocs.io/en/stable/api/jsonschema/protocols/#jsonschema.protocols.Validator.check_schema

I don't know whether we use a specific versioned validator or just the default, but not adding malformed schemas to the registry seems like a good built in developer safeguard.

tests/fixtures/all_config.toml
```diff
@@ -0,0 +1,14 @@
+[config.core]
```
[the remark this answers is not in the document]
jacobcook1995: Yeah I think this doesn't make sense, and adds needless complexity to the JSON schema to boot. I'll switch to the simpler naming scheme

virtual_rainforest/core/__init__.py
```diff
@@ -0,0 +1,67 @@
+from virtual_rainforest.core.config import register_schema
+
+
+# This is definitely a bit questionable, its basically here to make sure that the plants
+# module is imported somewhere when tests are run, this ensures that the plant schema is
+# added to testing. I don't think this is the best or right way to do something like
+# this, but couldn't come up with a better way to do it
+# TODO - Think of a better solution for this
```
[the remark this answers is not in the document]
davidorme: One design consideration here might be conditional loading of schemas. As `vr` gets more complex, we could end up with more and more schema, and loading and validating them on request via the user configuration rather than hard coding them all  via `__init__.py` might be an advantage. 

Another thing is that we probably want that system in `__init__.py` to automatically discover schemas, rather than extensions having to manually add schema loading for `my_new_plant_module`.

But - that is all longer term.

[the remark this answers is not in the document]
jacobcook1995: The test functions also do not use `schema` directly, they just need the schema registry to have be populated with the appropriate schema, so that the functions being tested can read the schema from there. So run into the same `flake8` pre-commit problem if I import there

virtual_rainforest/core/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+    from virtual_rainforest.plants import schema
+
+    return schema
+
+
+load_plants_schema()
+
+
+@register_schema("core")
+def schema():
```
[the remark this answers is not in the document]
dalonsoa: I really encourage you to use `Path` from [`pathlib`](https://docs.python.org/3/library/pathlib.html) instead of the functionality in `os` for path related operations. It is really the best practice nowadays and the preferred approach in the community. The above will look like:

```python
from pathlib import Path

config_schema = Path(__file__).parent.resolve() / "plant_schema.json"
```
which I hope you agree is more readable and cleaner. It's `glob` capabilities are also more easy to deal with for pattern matching filenames.

And yes, JSON file in the same directory than the relevant module.

virtual_rainforest/core/__init__.py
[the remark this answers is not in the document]
dalonsoa: That particular error `F401` is sensible to ignore as the import is for a very good reason. So in those cases you just add `noqa` at the end like:
```python
from virtual_rainforest.plants import schema  # noqa
```

virtual_rainforest/core/__init__.py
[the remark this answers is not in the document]
davidorme: I very much like that `/` operator!

davidorme (changes requested):
I agree with @dalonsoa - big picture looks good, but some tweaks needed

virtual_rainforest/core/config.py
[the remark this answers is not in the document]
jacobcook1995: I would probably lean towards using a critical error rather than a warning here. Basically think in cases where a `module name` is repeated non-sensical errors downstream are all but guaranteed if the configuration process is continued.

jacobcook1995:
I've attempted to address all the comments, and to extend the testing to cover the new (easier to test) structure. Let me know what you think

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def check_outfile(output_folder: str, out_file_name: str) -> None:
+    """Check that final output file is not already in the output folder.
+
+    Args:
+        output_folder: Path to a folder to output the outputted complete configuration
+            file to
+        out_file_name: The name to save the outputted complete configuration file under
+    """
```
dalonsoa, line 89:
```suggestion
    Args:
        output_folder: Path to a folder to output the outputted complete configuration
            file to
        out_file_name: The name to save the outputted complete configuration file under
    Raises:
        OSError: If the final output file already exist.
    """
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+                if len(toml_files) != 0:
+                    files.extend(toml_files)
+                else:
+                    empty_fold.append(path)
+            else:
+                files.append(Path(path))
+
+        else:
+            # Add missing path to list of missing paths
+            not_found.append(path)
```
dalonsoa, line 133: I think this can be un-nest a little bit:
```suggestion
        if p.is_dir():
            toml_files = list([f for f in p.glob("*.toml")])
            if len(toml_files) != 0:
                files.extend(toml_files)
            else:
                empty_fold.append(path)
        elif p.is_file():
            files.append(p)

        else:
            # Add missing path to list of missing paths
            not_found.append(path)
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    return None
+
+
+def collect_files(cfg_paths: list[str]) -> list[Path]:
+    """Collect all toml files from a user specified list of files and directories.
+
+    Args:
+        cfg_paths: A path or a set of paths that point to either configuration files, or
+            folders containing configuration files
+    """
```
dalonsoa, line 110: Add the possible exceptions this function might raise.

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+
+    # Preallocate file list
+    files = []
+    not_found = []  # Stores all invalid paths
+    empty_fold = []  # Stores all empty toml folders
```
dalonsoa, line 115: Not totally sure if these lists will store `Path`, `str` or a mixture of both. Maybe add some type annotations here?

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    """
+
+    # Preallocate container for file names and corresponding dictionaries
+    file_data: list[tuple[Path, dict]] = []
+    conflicts = []
+
+    # Load all toml files that we want to read from
+    for file in files:
+        # If not then read in the file data
+        with open(file, "rb") as f:
```
dalonsoa, line 175:
```suggestion
        with file.open("rb") as f:
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    """Load in a set of toml files checking that no tags are repeated.
+
+    This function also ensure that no tags are repeated across different toml files.
+
+    Args:
+        files: List of files to be read in and checked for overlapping tags
+    """
+
+    # Preallocate container for file names and corresponding dictionaries
+    file_data: list[tuple[Path, dict]] = []
```
dalonsoa, line 169: I'd make this a dictionary because then it is easier to handle and merge the items:
```suggestion
    file_data: dict[Path, dict] = {}
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+        with open(file, "rb") as f:
+            try:
+                toml_dict = tomllib.load(f)
+                # Check for repeated entries across previous nested dictionaries
+                for item in file_data:
+                    repeats = check_dict_leaves(item[1], toml_dict, [])
+                    for elem in repeats:
+                        conflicts.append((elem, file, item[0]))
+
+                file_data.append((file, toml_dict))
```
dalonsoa, line 184: if changed to a dict:
```suggestion
                file_data[file] = toml_dict
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    for file in files:
+        # If not then read in the file data
+        with open(file, "rb") as f:
+            try:
+                toml_dict = tomllib.load(f)
+                # Check for repeated entries across previous nested dictionaries
+                for item in file_data:
+                    repeats = check_dict_leaves(item[1], toml_dict, [])
+                    for elem in repeats:
+                        conflicts.append((elem, file, item[0]))
```
dalonsoa, line 182: If changed to a dict:
```suggestion
                for existing_file, existing_dict in file_data.items():
                    repeats = check_dict_leaves(existing_dict, toml_dict, [])
                    for elem in repeats:
                        conflicts.append((elem, file, existing_file))
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+        msg = "The following tags are defined in multiple config files:\n"
+        for conf in conflicts:
+            msg += f"{conf[0]} defined in both {conf[1]} and {conf[2]}\n"
+        msg = msg[:-1]
+        log_and_raise(msg, RuntimeError)
+
+    # Merge everything into a single dictionary
+    config_dict: dict = {}
+    for item in file_data:
+        dpath.util.merge(config_dict, item[1])
```
dalonsoa, line 204: And here is where we use ChainMap:

```suggestion
    from collections import ChainMap  # This imported at the top
    config_dict = ChainMap(*file_data.values())
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+def load_in_config_files(files: list[Path]) -> dict:
+    """Load in a set of toml files checking that no tags are repeated.
+
+    This function also ensure that no tags are repeated across different toml files.
+
+    Args:
+        files: List of files to be read in and checked for overlapping tags
+    """
```
dalonsoa, line 166: Add relevant execeptions here

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    for module in modules:
+        if module in SCHEMA_REGISTRY:
+            try:
+                m_schema = SCHEMA_REGISTRY[module]["properties"][module]
+            except KeyError as err:
+                log_and_raise(
+                    f"Schema for {module} module incorrectly structured, {err} key "
+                    f"missing!",
+                    KeyError,
+                )
```
dalonsoa, line 261: If schemas are checked when added to the registry, then this is not needed, rigth?

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+                comb_schema = SCHEMA_REGISTRY[module]
+            # Otherwise only save truncated part of the schema
+            else:
+                comb_schema["properties"][module] = m_schema
+                # Add module name to list of required modules
+                try:
+                    comb_schema["required"].append(module)
+                except KeyError:
+                    log_and_raise(
+                        f"The schema for {modules[0]} does not set the module as a "
```
dalonsoa, line 273:
```suggestion
                        f"The schema for {module} does not set the module as a "
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+                )
+            # Store complete schema if no previous schema has been added
+            if comb_schema == {}:
+                comb_schema = SCHEMA_REGISTRY[module]
+            # Otherwise only save truncated part of the schema
+            else:
+                comb_schema["properties"][module] = m_schema
+                # Add module name to list of required modules
+                try:
+                    comb_schema["required"].append(module)
```
dalonsoa, line 270: If `comb_schema["required"]` is a list, I don't see how this might fail. I don't think a `try... except` block is needed.

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+                    )
+        else:
+            log_and_raise(
+                f"Expected a schema for {module} module configuration, it was not "
+                f"provided!",
+                RuntimeError,
+            )
+
+    p_paths = []
+    p_len = len("/properties")
```
dalonsoa, line 285: Not needed.

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    # Recursively search for all instances of properties in the schema
+    for (path, value) in dpath.util.search(comb_schema, "**/properties", yielded=True):
+        if path == "properties":
+            p_paths.append("")
+        else:
+            size = len(path)
+            p_paths.append(path[: size - p_len])  # Strip out final "/properties"
```
dalonsoa, line 292:
```suggestion
        p_paths.append("" if path == "properties" else path.replace("/properties", ""))
```

virtual_rainforest/core/config.py
```diff
@@ -0,0 +1,362 @@
...
+    saved in toml format in the specified configuration file folder. This configuration
+    is finally used to populate the global `COMPLETE_CONFIG` dictionary.
+
+    Args:
+        cfg_paths: A path or a set of paths that point to either configuration files, or
+            folders containing configuration files
+        output_folder: Path to a folder to output the outputted complete configuration
+            file to
+        out_file_name: The name to save the outputted complete configuration file under.
+    """
```
dalonsoa, line 325: Add possible exceptions this will raise.

dalonsoa (changes requested):
Now it looks much, much better. Great work. I've made a bunch of suggestions about improving the code, mostly related to readability and conciseness. I might need to give another pass - specially to tests once you're done.

virtual_rainforest/core/config.py
jacobcook1995: The schema checking when adding to the registry just checks whether the schema is a valid JSON schema. This catches a bit more, i.e. whether the schema contains the keys we require which could be missing from an otherwise valid schema. I could move this check to the schema registry decorator if that's a better place for it?

dalonsoa: I think it makes sense. If the schema MUST have those keys, then the schema is just not valid without them, so it should not be registered to start with, right?

virtual_rainforest/core/config.py
jacobcook1995: At the moment no check is performed that the "required" tag is actually included in a schema, so if the first schema is missing it then this step will fail. This might be better handled by a check that the "required" key is included in every schema actually, particularly as schema loading order changing could errors to suddenly appear with the current setup

virtual_rainforest/core/logger.py
```diff
@@ -0,0 +1,31 @@
...
+It is very likely to be further extended in future.
+"""
+
+import logging
+from typing import Type
+
+LOGGER = logging.getLogger("virtual_rainforest")
+
+
+def log_and_raise(msg: str, exception: Type[Exception], extra: dict = None) -> None:
```
dalonsoa, line 16: I think you can do this in one go using `LOGGER.error`. See example https://stackoverflow.com/a/37992988
That would save this whole function.

tests/test_config.py
```diff
@@ -0,0 +1,314 @@
...
+def test_final_validation_log(caplog, expected_log_entries):
+    """Checks that validation passes as expected and produces the correct output."""
+
+    config.validate_config(["tests/fixtures"], out_file_name="complete_config")
+
+    # Remove generated output file
+    # As a bonus tests that output file was generated correctly + to the right location
+    Path("./complete_config.toml").unlink()
+
+    print(caplog)
```
dalonsoa, line 269: Provably needs to be removed. Print statements in tests are useful only for debug purposes

tests/test_config.py
```diff
@@ -0,0 +1,314 @@
...
+    """Check errors in finding module schema."""
+
+    # Check that find_schema fails as expected
+    with pytest.raises(expected_exception):
+        config.find_schema(config_dict)
+
+    log_check(caplog, expected_log_entries)
+
+
+@pytest.mark.parametrize(
```
dalonsoa, line 215: There is only one case... why do you parametrize?

tests/test_config.py
```diff
@@ -0,0 +1,314 @@
...
+            ["d1.d2.d3", "d1.d2.d4"],
+        ),
+    ],
+)
+def test_check_dict_leaves(d_a, d_b, overlap):
+    """Checks overlapping dictionary search function."""
+    assert overlap == config.check_dict_leaves(d_a, d_b, [])
+
+
+@pytest.mark.parametrize(
```
dalonsoa, line 40: There is only one case... why do you parametrize?

dalonsoa (approved):
Looks really good, now! You're a pro of parametrised testing! I've left a few comments, but no blockers so happy to approve.

tests/test_config.py
jacobcook1995: Ahh yeah I missed that!

virtual_rainforest/core/logger.py
jacobcook1995: Would it halt the execution of the program? Because I guess the advantage of defining this function is that it lets you log an error and halt the program with a single function call, rather than having to call `LOGGER.error` followed by `raise OSError` every time?

dalonsoa: You're right. It does not raise thee exception. So, good as it is!

tests/test_config.py
```diff
@@ -0,0 +1,314 @@
...
+            KeyError,
+        ),
+        (
+            "bad_module_3",
+            {"type": "object", "properties": {"bad_module_3": {}}},
+            KeyError,
+        ),
+    ],
+)
+def test_register_schema_errors(schema_name, schema, expected_exception):
```
jacobcook1995, line 305: @dalonsoa  I also had a quick question on this test. I could work out how to test that the decorator raised the correct error type in response to bad input, but couldn't work out how to check the error message in the log (which would contain info on what key was missing). I was wondering if you had any advice/pointers here?

virtual_rainforest/core/logger.py
davidorme: Phew! This function is lifted from another project that @jacobcook1995 and I work on 😄

tests/test_config.py
davidorme: You want to capture the exception:

```
with pytest.raises(expected_exception) as err:
```

And then you can use `str(err)`.

jacobcook1995: Hmmm `str(err)` doesn't seem to return a message here instead it returns `'<ExceptionInfo for raises contextmanager>'`, I think something about testing a decorator throws it off. `caplog` also didn't behave as I expected it to here

jacobcook1995: Comparing `caplog` works only if it's not contained by the `with:` statement, but that is a workable solution

davidorme (approved):

PR #83, opened 2022-10-11
Strict requirement for type hinting

jacobcook1995:
# Description

Based on the discussion in #60 it seems like a good idea to make type annotations mandatory. If this becomes a problem down the we can always switch to a less strict option, but if we are going to use a more strict option, switching to it down the line would be a nightmare (there already were a lot of extra `mypy` errors to fix with our currently pretty minimal code base).

I could find type stubs for either `shapely` or `dpath`, so turned off type checking when I imported them. I may not have been looking in the right places for them though. So, if anyone knows where they can be found, let me know.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/training/test_mfm.py
+++ b/docs/source/development/training/test_mfm.py
@@ -4,12 +4,12 @@ import pytest
 from mfm import TimesTable, my_float_multiplier, my_picky_float_multiplier
 
 
-def test_fm():
+def test_fm() -> None:
 
     assert 10 == my_float_multiplier(2, 5)
 
 
-def test_pfm():
+def test_pfm() -> None:
 
     with pytest.raises(ValueError) as err_hndlr:
 
@@ -18,7 +18,7 @@ def test_pfm():
     assert str(err_hndlr.value) == "Both x and y must be of type float"
 
 
-def test_pfm_fail():
+def test_pfm_fail() -> None:
 
     with pytest.raises(ValueError) as err_hndlr:
 
@@ -36,7 +36,7 @@ def test_pfm_fail():
         (-1.5, -3.0, 4.5),
     ],
 )
-def test_pfm_param_noid(x, y, expected):
+def test_pfm_param_noid(x: float, y: float, expected: float) -> None:
 
     assert expected == my_picky_float_multiplier(x, y)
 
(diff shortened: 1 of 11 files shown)


davidorme:
It looks like `shapely` is planning on implementing but isn't there yet (https://github.com/shapely/shapely/issues/721) but I can find nothing on `dpath`. Both of those could be suppressed at the project level in the `cfg` file, but we can look at that if we end up using these a lot in different modules, rather than just `grid` and `config`.

davidorme:
It looks like we should be able to use `conftest.py` to configure `mypy` to be less strict in the test suite than the main code? Do we really need to enforce typing on test functions? I guess it doesn't hurt, but it is finnicky.

jacobcook1995:
Yeah I wasn't sure whether or not requiring type hints on testing code is desirable or not, but that can be changed if we decide that it isn't

davidorme:
I have had a look at this, because my next PR (#84) is affected. We can include the code below in `tests/conftest.py` to make `mypy` more forgiving on the tests. 
```
def pytest_configure(config):
    plugin = config.pluginmanager.getplugin("mypy")
    plugin.mypy_argv.append("--allow-untyped-calls")
    plugin.mypy_argv.append("--allow-untyped-defs")
    plugin.mypy_argv.append("--allow-incomplete-defs")
```

I think this is probably worth doing - I guess it is a similar argument to whether or not you use docstrings.  I think docstrings are worthwhile because it is the 'right' place to comment on tests when needed, but I'm not sure we need to enforce typing on tests. @dalonsoa and @alexdewar ?

dalonsoa:
I think that being more relaxed with typing in the tests make sense, so if there is a way of removing the most annoying aspects of mypy there, as it seems there is, I would use it.

davidorme (changes requested):
OK. With that feedback, I'd say we roll back the typing added in `tests` and then add the local `pytest-mypy` plugin configuration to `conftest.py`.

alexdewar:
I agree. The only caveat I would add though is that if it ends up being a faff to exclude the ``tests`` folder from the mypy linting, you could just leave the type hints there. In practice this will just mean adding a bunch of ``-> None``s, which isn't that big a deal.

jacobcook1995:
Actually appears to be fairly straightforward to add module/folder level exclusions for `mypy` in `setup.cfg`

davidorme (approved):
That config solution looks good - a bit cleaner! I've applied this branch to #84 and it works, so I'm happy to approve this.

Issue #74, opened 2022-09-20
Get the website building and discuss structure


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
As it currently exists, the package documentation doesn't build anywhere, and doesn't have a clear structure.

**Describe the solution you'd like**
Would be good to setup a system to build the docs as a site, and in doing that we would have to agree upon a structure for the docs folders. We could probably build the site using [Read the Docs](https://readthedocs.org)

**Describe alternatives you've considered**
We don't necessarily need to do this soon, we can stick with fragmentary documentation for the time being, but we are definitely going to have to do this at some point.

PR #84, opened 2022-10-11
Updated grid implementation

### LINKED ISSUE #31: Add tests for functions/methods in grid.py

### LINKED ISSUE #37: Convert TODOs in grid.py to GitHub issues
These all look like they could be converted into issues:
```
 - set up neighbourhoods. ? store as graph (networkx - might only need a really
       lightweight graph description).
TODO - export of files to geojson
TODO - import of geojson grids? Way to link structured landscape into cells
```

If you're not sure that a feature will be needed/liked by others, that's ok. Raising it
as an issue just gives a central place for discussion about it to happen.

### LINKED ISSUE #79: Simplify grid cell indexing system

As noted [here](https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/69#discussion_r964375433), using a dictionary of grid cells and using grid position strings (e.g. `grid.cell_dict['1-1']`) as keys is complex (and likely slow).

In that comment @dalonsoa suggested nested lists and using nested integer indexing (so `grid.cell_list[x_idx][y_idx]`), but just having a single integer index may be even simpler (so `grid.cell_list[cell_id]`). So, a square grid might get cell ids:

```
6 7 8
3 4 5
0 1 2
```

Advantages:

* A single integer points to the cell - shorter code, as above.
* It is easier to map cells to neighbours: cell 4 has rook neighbours `[1, 3, 5, 7]` and not `[(1,0),(0,1),(2,1),(1,2)]`.
* It makes it easy to have spatial layouts that don't conform to a regular grid. Even with hex grids, the x indices are offset for alternate rows, but we could equally have 'grids' with cells of irregular size and shape (way down the development path!).

One open question is how to map cell id on to stored data on the grid. For the square grid - which is going to be our main use case - some operations could run most efficiently on a 2D grid (`numpy` and `scipy` - things like distance), so keeping spatial data as a 2D array (and providing a `grid.map` attribute that converts cell id to numpy indices) is simple. However data loading _could_ unravel all spatial data onto a single axis so that cell id is also the index onto the data.

davidorme:
# Description

This updates the grid system in line with the discussion in #79 and implements a first draft of neighbour lists and cell to cell distance calculations

Closes #79 

Also:

* Closes #37 
* Closes #31

## Type of change

## Key checklist

## Further checks


davidorme:
This clashes with #83 as it stands because the testing isn't typed. We should resolve that PR first.

davidorme:
That is now resolved - tests do not require mypy annotation and so this PR is in line with `develop`

docs/source/virtual_rainforest/core/grid.md
```diff
@@ -42,44 +41,92 @@ square_grid
...
+```{code-cell} ipython3
+square_grid.get_distances(45, square_grid.neighbours[45])
+```
+
+```{code-cell} ipython3
+hex_grid.get_distances([1, 40], hex_grid.neighbours[40])
+```
+
+```{code-cell} ipython3
+square_grid.set_distances()
```
vgro, line 121: I find 'set.distances()' a bit confusing here. At first, I thought that this modifies the extent of the grid. I had a look into the core code and from what I understand, this writes a matrix with distances on demand? maybe there might be a better name for this function or you could add a comment

virtual_rainforest/core/grid.py
```diff
@@ -64,7 +78,7 @@ def make_square_grid(
     cell_ny: int,
     xoff: float = 0,
```
vgro, line 79: What is the offset for if we ask the user to preprocess all inputs?

docs/source/virtual_rainforest/core/grid.md
```diff
@@ -19,20 +19,19 @@ kernelspec:
 This module is used to define the grid of cells used in a `virtual_rainforest`
 simulation. Square and hexagon grids are currently supported.
 
-```{code-cell}
+```{code-cell} ipython3
 import matplotlib.pyplot as plt
 
-from virtual_rainforest.core.grid import CoreGrid, CoreGridConfig
+from virtual_rainforest.core.grid import Grid
```
vgro, line 25: I like the simplification of the Grid class

docs/source/virtual_rainforest/core/grid.md
davidorme: Good point - `calculate_distances` would be better.

virtual_rainforest/core/grid.py
davidorme: If a user has data from a real location with real coordinates in a projected coordinate system, then forcing them to change the origin of those coordinates is not helpful. And we want to create outputs that preserve the actual coordinates too, so that output GIS files can be plotted correctly in the real location.

virtual_rainforest/core/grid.py
```diff
@@ -29,12 +33,22 @@
...
 
+class GridStructure(NamedTuple):
+    """Data structure to be returned from grid creator functions.
+
+    The contents should be two equal length lists giving unique integer ID values
+    and corresponding shapely.geometry.Polygon objects.
+    """
+
+    ids: list[int]
+    polygons: list[Polygon]
```
dalonsoa, line 44: Given that you need to ensure they have the same length, I would use a `dataclass` instead, which lets you do validation afterwards. Moreover, I will make it `frozen` and use `tuples` instead of `lists` if they are not meant to be modified afterwards. Something like:

```suggestion
from dataclasses import dataclass
...
@dataclass(frozen=True)
class GridStructure:
    """Data structure to be returned from grid creator functions.

    The contents should be two equal length lists giving unique integer ID values
    and corresponding shapely.geometry.Polygon objects.
    """

    ids: Tuple[int, ...]
    polygons: Tuple[Polygon, ...]
    
    def __post_init__(self):
        if len(self.ids) != len(self.polygons):
            raise ValueError("Useful description of error...")
```

virtual_rainforest/core/grid.py
```diff
@@ -85,19 +102,19 @@ def make_square_grid(
...
+            # Define the cell id as integer count starting lower left by row
+            cell_id = x_idx + (y_idx) * cell_nx
             # Store shifted geometry under cell id
-            cell_dict[cell_id] = translate(
+            cell_list[cell_id] = translate(
                 prototype, xoff=scale_factor * x_idx, yoff=scale_factor * y_idx
             )
 
-    return cell_dict
+    return GridStructure(ids=list(range(len(cell_list))), polygons=cell_list)
```
dalonsoa, line 117: Modify accordingly if change above implemented.

virtual_rainforest/core/grid.py
```diff
@@ -143,21 +166,21 @@ def make_hex_grid(
...
             # Store shifted geometry under cell id
-            cell_dict[cell_id] = translate(
+            cell_list[cell_id] = translate(
                 prototype,
                 xoff=2 * apothem * x_idx + apothem * (y_idx % 2),
                 yoff=1.5 * side_length * y_idx,
             )
 
-    return cell_dict
+    return GridStructure(ids=list(range(len(cell_list))), polygons=cell_list)
```
dalonsoa, line 183: Same here

virtual_rainforest/core/grid.py
```diff
@@ -232,14 +261,39 @@ def __init__(
...
+        grid_structure = creator(
             cell_area=self.cell_area,
             cell_nx=self.cell_nx,
             cell_ny=self.cell_ny,
             xoff=self.xoff,
             yoff=self.yoff,
         )
 
+        self.cell_id = grid_structure.ids
+        self.polygons = grid_structure.polygons
```
dalonsoa, line 273: I'm not sure of the utility of defining a dedicated class for grid structure and then, immediately after using it, get rid of it by unpacking its values. You could equally do:

```python
self.cell_id, self.polygons = creator(...)
```
With the same effect and less complexity.

dalonsoa (changes requested):
It looks good. I've just made a couple of comments in relation to the `GridStructure` class that need to be address - either way.

virtual_rainforest/core/grid.py
davidorme: Yup - that makes a lot of sense. These should indeed be frozen and equal length.

virtual_rainforest/core/grid.py
davidorme: I see the point about the transient class. I wanted a consistent return type that should be used by anything registered as a grid creator. I first went with a custom Type as a tuple of ids and polys, but switched to NamedTuple to use named attributes not indices.

We _could_ just use that GridStructure internally, but I was trying to avoid the extra attribute depth (`grid.id` not `grid.gridstructure.id`). The suggestion above is certainly more elegant.

dalonsoa: I'm not convinced - having a transient class is, in the end, increasing code complexity and overheads while types are totally transparent and innocuous at runtime. Specially being so shot lived! What's the downside of using a custom signature used by all grid creators?

```python
GRID_STRUCTURE_SIG = Tuple[Tuple[int, ...], Tuple[Polygon, ...]]
```

Moreover, you can modify the `register_grid` decorator such that the validation of the length of the tuples takes place there.

But, anyway, it's up to you. It is not a huge deal. In general, custom classes should only be used when there are obvious gains over using built-in (or third party) solutions out of the box. I don't see that here.

davidorme: That's pretty much exactly where I started 😄 and then then I got drawn into `object.id` access, which really isn't needed. The only useful thing that a custom `dataclass` _would_ uniquely allow is to populate a default `ids` sequence but that really doesn't seem worth the extra complexity! I will revert back to the custom sig.

davidorme: Is there any particular value in checking the tuples are of equal length in the `register_grid` decorator rather than in the `Grid` class? We'd have to wrap the function call to update the decorator and it's a one liner within `Grid`.

dalonsoa: No advantage, really

davidorme: Right. I've committed those changes to the wrong PR - they are now sitting on the end of #94.

Having said that, I pulled the most recent change on this branch (f57049fb881927e5ebd4753007d2c1ea4322925e) into #94 to get the web build working. I _think_ the net result is that we can merge #94 and then simply close this one. That's irritatingly messy, but probably easier than `revert`ing the last commit to #94 and moving that commit onto this PR?

davidorme: @dalonsoa - IIRC you had some `zsh` magic that looked like it was including colorised branch names in the shell prompt. What was doing that?

dalonsoa: This is what you're looking for, although I use `fish` instead of `zsh`:

- https://medium.com/@elviocavalcante/5-steps-to-improve-your-terminal-appearance-on-mac-osx-f58b20058c84
- https://ostechnix.com/oh-fish-make-shell-beautiful/

davidorme:
OK  - I have run `git revert HEAD` on #74 and then used `git cherry-pick` to bring those changes across here. Mostly as an exercise in figuring out how to do it 😄

This is much cleaner - I've added a test for the two current `Grid` init failure modes.

davidorme:
Also

* Closes #37 
* Closes #31

virtual_rainforest/core/grid.py
```diff
@@ -113,7 +106,7 @@ def make_square_grid(
                 prototype, xoff=scale_factor * x_idx, yoff=scale_factor * y_idx
             )
 
-    return GridStructure(ids=list(range(len(cell_list))), polygons=cell_list)
+    return (tuple(range(len(cell_list))), tuple(cell_list))
```
dalonsoa, line 109: You don't need the external parenthesis. A `return` statement with elements separated by commas automatically results in a tuple.
```suggestion
    return tuple(range(len(cell_list))), tuple(cell_list)
```

PR #86, opened 2022-10-12
Fixed mutable default argument problem

### LINKED ISSUE #85: Possible reuse of dict arg in ``check_dict_leaves()``
https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/9a406d77ae25554d6317e6986373d457161a9c31/virtual_rainforest/core/config.py#L98

I think you may have fallen foul of this common Python gotcha: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

In Python, if you define a default arg as an empty list then append to it, the next time you call that function the default arg itself will have been modified. It's an extremely odd behaviour and a common trap.

Might be worth checking to see if this mistake has been made elsewhere.

jacobcook1995:
# Description

This fixes the problem pointed out by Alex where I had used a mutable default argument, which is apparently a common python gotcha. I fixed it using the approach laid out [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). I've fixed this bug in the place Alex pointed out, and checked to see if I had made the same mistake elsewhere (I luckily hadn't).

Fixes #85

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -76,7 +76,7 @@ COMPLETE_CONFIG: dict = {}
 
 
 def check_dict_leaves(
-    d1: dict, d2: dict, conflicts: list = [], path: list = []
+    d1: dict, d2: dict, conflicts: list = None, path: list = None
 ) -> list:
     """Recursively checks if leaves are repeated between two nested dictionaries.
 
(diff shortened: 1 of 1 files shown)


davidorme:
I'm puzzled here that `mypy` doesn't insist on `Optional[list]` here?

alexdewar (approved):
LGTM!

alexdewar:
@davidorme That's a good spot! I think you're right, ``strict_optional = true`` is a more sensible setting -- otherwise any value can be implicitly `None`, which really isn't good from a type-checking perspective. It's handy to have the types annotated ``Optional`` as appropriate and helps ``mypy`` give better linting.

PR #89, opened 2022-10-14
Feature/strict optional

### LINKED ISSUE #88: Turn strict optional type checking on
At the moment strict optional type checking in `mypy` is turned off. This means that `None` is treated as a valid input type in all cases, which somewhat negates the point of type checking.

We can switch strict optional type checking on by setting `strict_optional = True` in `setup.cfg`. Then in cases where we want to allow `None` as a valid function input the type hint becomes e.g. `Optional[str]` rather than `str`.

jacobcook1995:
# Description

Turns on strict optional typing and turns off implicit optional typing, this means that every time that `None` is an acceptable alternative argument to a function, the type hint should make use of `Optional[]` from `typing`.

Fixes #88 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ docstring-convention = google
 [mypy]
 ignore_missing_imports = False
 strict_optional = True
+no_implicit_optional = True
 disallow_untyped_calls = True
 disallow_untyped_defs = True
 disallow_incomplete_defs = True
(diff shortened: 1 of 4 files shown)


davidorme (approved):
LGTM - good catch on the implicit optional as well.

PR #92, opened 2022-10-14
Feature/version fixup

### LINKED ISSUE #90: Package version is multiply defined
`poetry` provides the version tool to increment the package version in `pyproject.toml` but at the moment the version is also hardcoded in the `version.py` module.

**To Reproduce**

```
import virtual_rainforest.version as vrv
vrv.__version__
```
This is _not_ the same value as in `pyproject.toml`

There is a good solution [here](https://stackoverflow.com/questions/67085041/) to allow the package to report the version - either when installed by `poetry` in development mode or via `pip` - using only the single source in `pyproject.toml`

This should be implemented in `__init__.py`, so we can remove `version.py`. 

Whilst in `__init__.py`, there is also some rogue logger config commands that should be moved into `logger.py`.

davidorme:
# Description

The package version in `virtual_rainforest.__version__` now tracks the version set by `poetry` in `pyproject.toml`

Fixes #90 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,9 +15,9 @@ import sys
 
 sys.path.insert(0, os.path.abspath("../"))
 
-from virtual_rainforest import version as vr_version  # noqa: E402
+import virtual_rainforest as vr  # noqa: E402
 
-version = vr_version.__version__
+version = vr.__version__
 release = version
 
 # -- Project information -----------------------------------------------------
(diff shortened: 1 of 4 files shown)


dalonsoa (approved):
Nice!

dalonsoa:
But what about the documentation? Can you get the version in Sphinx the same way?

davidorme:
Oddly, that failure didn't happen when running `pytest` locally. I don't know why the local `mypy` didn't go looking in the docs tree. I've updated `conf.py` to use the new API for the version number and all should be well.

PR #93, opened 2022-10-18
Feature/default config

### LINKED ISSUE #87: Allowing default config values
At the moment the configuration system does not set default values, this means that if an item required for the configuration of a specific module is missing the configuration process fails. This is undesirable behaviour, particularly for users who are not familiar with the inner workings of the module in question.

Each module currently has an associated `JSONSchema` that specifies which configuration items the module expects, along with the expected types of these items. This can be extended so that where appropriate default values are set in the schema. Then if a value isn't provided, the default is added to the config instead. This final config will be saved so that configuration options that the simulation was run can still be determined regardless of whether default config values are changed down the line.

Alternatively these default values could be coded into the relevant functions. However, this would make it much harder to gather the default values in a single place for the purposes of documentation.

jacobcook1995:
# Description

Extends the configuration system to allow default values to be populated if the user doesn't provide particular tags. These defaults are populated from the `JSON Schema` for each module. This method of storing and populating defaults works well for simple defaults (e.g. single item or a list of items).

However, it doesn't work for more complex defaults (e.g. climate time series, tables of functional types). So, the system will need to be further extended to tackle these cases later. I will open an issue (or a discussion?) to track this

Fixes #87 

## Type of change

## Key checklist

## Further checks


jacobcook1995:
Hmmmm for some reason the CI checks have been stuck with the status "Expected — Waiting for status to be reported" since I opened this pull request. The "Actions" tab also seems to shows no indication of the CI being run

jacobcook1995:
I opened the pull request the way that I normally do, e.g. using the button on the GitHub page for the specific feature branch. The only difference I can think of is that I didn't run CI on the final commit to the branch (because it was an update to the documentation)

dalonsoa:
That is odd. Can you try to merge the latest version of develop and push the changes? Maybe GitHub runners are down for whatever reason...

dalonsoa:
Now it works... 🤷

jacobcook1995:
Currently not behind develop so can't merge it in, but made a very minor change to a description in one of the `JSON schema` to test if pushing starts the CI

davidorme (approved):
Looks good to me.

virtual_rainforest/core/config.py
```diff
@@ -236,6 +233,80 @@ def load_in_config_files(files: list[Path]) -> dict:
...
+        ):
+            yield error
+
+    return validators.extend(
+        validator_class,
+        {"properties": set_defaults},
+    )
+
+
+def add_core_defaults(config_dict: dict) -> dict:
```
alexdewar, line 272: I have a feeling that you could restrict your `dict` types somewhat more by specifying the key type: `dict[str, any]`

There are some other places where I think you could do this too

virtual_rainforest/core/config.py
```diff
@@ -236,6 +233,80 @@ def load_in_config_files(files: list[Path]) -> dict:
...
+            "Expected a schema for core module configuration, it was not provided!",
+            RuntimeError,
+        )
+
+    try:
+        ValidatorWithDefaults(core_schema).validate(config_dict)
+    except exceptions.ValidationError as err:
+        log_and_raise(
+            f"Validation of core configuration files failed: {err.message}",
+            RuntimeError,
```
alexdewar, line 304: I'd personally consider a custom exception type for this. It seems like there are several places where configuration-related exceptions can be raised and it'd be useful to let the caller check for this specifically.

virtual_rainforest/core/config.py
```diff
@@ -315,11 +386,42 @@ def construct_combined_schema(modules: list[str]) -> dict:
     return comb_schema
 
 
+def validate_with_defaults(config_dict: dict, comb_schema: dict) -> dict:
```
alexdewar, line 389: See comment above re `dict` types

virtual_rainforest/core/config.py
```diff
@@ -315,11 +386,42 @@ def construct_combined_schema(modules: list[str]) -> dict:
...
+    # Make a new validator that allows the addition of defaults
+    ValidatorWithDefaults = extend_with_default(Draft202012Validator)
+
+    # Validate the input configuration settings against the combined schema
+    # This step also adds in all default module configuration details
+    try:
+        ValidatorWithDefaults(comb_schema).validate(config_dict)
+    except exceptions.ValidationError as err:
+        log_and_raise(
+            f"Validation of configuration files failed: {err.message}", RuntimeError
```
alexdewar, line 414: See comment above re exceptions

virtual_rainforest/core/config.py
```diff
@@ -315,11 +386,42 @@ def construct_combined_schema(modules: list[str]) -> dict:
...
+        )
+
+    return config_dict
+
+
 def validate_config(
     cfg_paths: Union[str, list[str]],
```
alexdewar, line 421: You're using Python 3.9 right? If you're using Python 3.10 you can write this as `str | list[str]`

virtual_rainforest/core/config.py
```diff
@@ -315,11 +386,42 @@ def construct_combined_schema(modules: list[str]) -> dict:
...
     output_folder: str = ".",
     out_file_name: str = "complete_config",
-) -> None:
+) -> dict:
```
alexdewar, line 424: Another `dict`

virtual_rainforest/core/config.py
```diff
@@ -315,11 +386,42 @@ def construct_combined_schema(modules: list[str]) -> dict:
...
+            instance
+        comb_schema: Combined schema for all modules that are being configured
+
+    Raises:
+        RuntimeError: If the configuration files fail to validate against the JSON
+            schema
+    """
+
+    # Make a new validator that allows the addition of defaults
+    ValidatorWithDefaults = extend_with_default(Draft202012Validator)
```
alexdewar, line 406: Is there any reason you can't extend `Draft202012Validator` by inheritance (i.e. `class ValidatorWithDefaults(Draft202012Validator)`?

alexdewar: Going one step further, I'd be inclined to make this a member method of your `ValidatorWithDefaults` class.

Minor nit: I found the name a bit confusing. How about `validate_and_add_defaults()`? It's a bit cumbersome but conveys that `config_dict` is being modified

virtual_rainforest/core/config.py
```diff
@@ -315,11 +386,42 @@ def construct_combined_schema(modules: list[str]) -> dict:
...
+    # Validate the input configuration settings against the combined schema
+    # This step also adds in all default module configuration details
+    try:
+        ValidatorWithDefaults(comb_schema).validate(config_dict)
+    except exceptions.ValidationError as err:
+        log_and_raise(
+            f"Validation of configuration files failed: {err.message}", RuntimeError
+        )
+
+    return config_dict
```
alexdewar, line 417: Not sure why you're returning `config_dict` again. Isn't it modified in place?

virtual_rainforest/core/config.py
```diff
@@ -379,5 +475,5 @@ def validate_config(
     with open(f"{output_folder}/{out_file_name}.toml", "wb") as toml_file:
         tomli_w.dump(config_dict, toml_file)
 
-    # Populate the global config dictionary with the complete validated config
-    COMPLETE_CONFIG["config"] = config_dict
+    # Return the complete validated config
+    return config_dict
```
alexdewar, line 479: As above, don't bother returning `config_dict` again if it's actually modified in place.

If you *don't* want it to be modified in place, it might be worth having a test that checks whether this holds true.

alexdewar (changes requested):
I've made some small suggestions.

I like that you now have the defaults specified in the schema -- I think it's good to keep all that info in one place.

virtual_rainforest/core/config.py
jacobcook1995: Ahh right yes I wasn't sure how to type nested dictonaries, but `dict[str, any]` seems sensible

virtual_rainforest/core/config.py
jacobcook1995: Yeah we are, shame as it's a much nicer syntax

virtual_rainforest/core/config.py
jacobcook1995: Agree with the above comment, but in this case `config_dict` is created within the function, so I think it has to be returned to be accessible to downstream functions?

virtual_rainforest/core/config.py
```diff
@@ -236,6 +233,80 @@ def load_in_config_files(files: list[Path]) -> dict:
...
+
+    try:
+        ValidatorWithDefaults(core_schema).validate(config_dict)
+    except exceptions.ValidationError as err:
+        log_and_raise(
+            f"Validation of core configuration files failed: {err.message}",
+            RuntimeError,
+        )
+
+    return config_dict
```
alexdewar, line 317: Oops, *this* is where I meant to stick my comment re returning `config_dict`. I'm guessing the arg is actually being modified in place, right?

virtual_rainforest/core/config.py
jacobcook1995: Having a bit of difficultly with this one. Basically `jsonschema` already creates a class structure for validators, with the function `validators.extend` being used to create new class instances. Not really sure how to define my own `ValidatorWithDefaults` and add in `validate_and_add_defaults()` as a member function without overwriting the class structure defined by the `jsonschema` package? As an alternative I was thinking of keeping `validate_and_add_defaults()` where it is, but creating the Validator once at global scope. This stops `ValidatorWithDefaults` from having to be created multiple times

alexdewar: Okey doke. I didn't realise that's how ``jsonschema`` does things, but if that's the case then it's ok by me!

virtual_rainforest/core/config.py
```diff
@@ -280,42 +294,39 @@ def add_core_defaults(config_dict: dict) -> dict:
...
         ValidatorWithDefaults(core_schema).validate(config_dict)
     except exceptions.ValidationError as err:
         log_and_raise(
             f"Validation of core configuration files failed: {err.message}",
-            RuntimeError,
+            ConfigurationError,
         )
 
-    return config_dict
+    return None
```
alexdewar, line 318: This line is redundant.

virtual_rainforest/core/config.py
```diff
@@ -398,30 +411,28 @@ def validate_with_defaults(config_dict: dict, comb_schema: dict) -> dict:
...
         ValidatorWithDefaults(comb_schema).validate(config_dict)
     except exceptions.ValidationError as err:
         log_and_raise(
-            f"Validation of configuration files failed: {err.message}", RuntimeError
+            f"Validation of configuration files failed: {err.message}",
+            ConfigurationError,
         )
 
-    return config_dict
+    return None
```
alexdewar, line 428: Again, you don't need to ``return None``

alexdewar (changes requested):
There are a couple of unnecessary ``return None``s in there and otherwise I've just got one question. Apart from that, it all looks good!

alexdewar (approved):
That looks better.

Side note: I'm not sure user programs should really be raising ``OSError``s in general anyway. They tend to be thrown when e.g. syscalls fail etc., when it's actually the OS complaining.

PR #106, opened 2022-10-28
Basic definition and initialisation of models

### LINKED ISSUE #96: define basic Model class
Need to define a basic structure that is common across models so that the high level function that runs the complete `vr` model is working with the same structure across models/modules.

A base `Model` class that defines the basic init properties, and also the functions that will be used by the higher level function that runs the entire simulation. These functions will generally be overwritten for the individual module subclasses, but including them here forces us to use a consistent naming scheme.

Really think a central definition is necessary, as it's far easier to maintain than scattered module level definitions

jacobcook1995:
# Description

In order to enforce a consistent api across our models, I've added a `BaseModel` class which sets the core set of functions and attributes all our models should possess. As an example I've also defined a `SoilModel` class which inherits from `BaseModel` and overwrites one of the functions. I've written a function that performs checks and then initialises a `SoilModel` object. This function is then added to a registry of models using a decorator, this should allow all models to be initialised using the same api.

Fixes #96 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -5,7 +5,7 @@ define models based on the class defined in model.py
 """
 
 from contextlib import nullcontext as does_not_raise
-from logging import CRITICAL, INFO, WARNING
+from logging import CRITICAL, ERROR, INFO, WARNING
 
 import pytest
 from numpy import datetime64, timedelta64
@@ -41,47 +41,46 @@ def test_base_model_initialization(caplog, mocker):
 
 
 @pytest.mark.parametrize(
-    "no_layers,raises,expected_log_entries",
+    "no_layers,expected_log_entries",
     [
         (
             2,
-            does_not_raise(),
             (),
         ),
         (
             -2,
-            pytest.raises(ValueError),
             (
                 (
-                    CRITICAL,
+                    ERROR,
                     "There has to be at least one soil layer in the soil model!",
                 ),
             ),
         ),
         (
             2.5,
-            pytest.raises(TypeError),
             (
                 (
-                    CRITICAL,
+                    ERROR,
                     "The number of soil layers must be an integer!",
                 ),
             ),
         ),
     ],
 )
-def test_soil_model_initialization(caplog, no_layers, raises, expected_log_entries):
+def test_soil_model_initialization(caplog, no_layers, expected_log_entries):
     """Test `SoilModel` initialization."""
 
-    # Check whether initialising the model fails as expected
-    with raises:
-        model = SoilModel(timedelta64(1, "W"), no_layers)
+    # Initialize model
+    model = SoilModel(timedelta64(1, "W"), no_layers)
 
-        # In cases where it passes then checks that the object has the right properties
-        assert set(["setup", "spinup", "solve", "cleanup"]).issubset(dir(model))
-        assert model.name == "soil"
-        assert str(model) == "A soil model instance"
-        assert repr(model) == "SoilModel(update_interval = 1 weeks, no_layers = 2)"
+    # In cases where it passes then checks that the object has the right properties
+    assert set(["setup", "spinup", "solve", "cleanup"]).issubset(dir(model))
+    assert model.name == "soil"
+    assert str(model) == "A soil model instance"
+    assert (
+        repr(model)
+        == f"SoilModel(update_interval = 1 weeks, no_layers = {int(no_layers)})"
+    )
 
     # Final check that expected logging entries are produced
     log_check(caplog, expected_log_entries)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -1,0 +1,61 @@
+"""Test module for model.py (and associated functionality).
+
+This module tests the functionality of model.py, as well as other bits of code that
+define models based on the class defined in model.py
+"""
+
+from contextlib import nullcontext as does_not_raise
+from logging import CRITICAL
+
+import pytest
+from numpy import datetime64, timedelta64
+
+from virtual_rainforest.core.model import Model
+
+from .conftest import log_check
+
+
+@pytest.mark.parametrize(
+    "start_time,end_time,update_interval,raises,expected_log_entries",
+    [
+        (
+            datetime64("2022-10-26"),
+            datetime64("2052-10-26"),
+            timedelta64(1, "W"),
+            does_not_raise(),
+            (),
+        ),
+        (
+            datetime64("2052-10-26"),
+            datetime64("2022-10-26"),
+            timedelta64(1, "W"),
+            pytest.raises(ValueError),
+            (
+                (
+                    CRITICAL,
+                    "Model cannot end at an earlier time than it starts!",
+                ),
+            ),
+        ),
+    ],
+)
+def test_model_initialization(
+    caplog, start_time, end_time, update_interval, raises, expected_log_entries
+):
+    """Test `Model` initialization."""
+
+    # Check whether initialising the model fails as expected
+    with raises:
+        model = Model(start_time, end_time, update_interval)
+
+        # In cases where it passes then checks that the object has the right properties
+        assert set(["setup", "spinup", "solve", "cleanup"]).issubset(dir(model))
+        assert model.name == "base model"
+        assert str(model) == "A base model instance"
+        assert (
+            repr(model) == "Model(start_time=2022-10-26, end_time=2052-10-26,"
+            " update_interval=1 weeks)"
+        )
+
+    # Final check that expected logging entries are produced
+    log_check(caplog, expected_log_entries)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -10,7 +10,7 @@ from logging import CRITICAL, INFO, WARNING
 import pytest
 from numpy import datetime64, timedelta64
 
-from virtual_rainforest.core.model import BaseModel, register_model
+from virtual_rainforest.core.model import BaseModel
 from virtual_rainforest.soil.model import InitialisationError, SoilModel
 
 from .conftest import log_check
@@ -140,22 +140,14 @@ def test_soil_model_initialization(
 def test_register_model_errors(caplog):
     """Test that the schema registering models generates correct errors/warnings."""
 
-    @register_model("soil")
-    def to_be_decorated() -> SoilModel:
-        return SoilModel(
-            datetime64("2022-10-26"),
-            datetime64("2052-10-26"),
-            timedelta64(1, "W"),
-            2,
-        )
+    class NewSoilModel(BaseModel, model_name="soil"):
+        """Test class for use in testing __init_subclass__."""
 
-    to_be_decorated()
-
-    # Then check that the correct (critical error) log messages are emitted
+    # Then check that the correct (warning) log messages are emitted
     expected_log_entries = (
         (
             WARNING,
-            "Model type soil already exists and is being replaced",
+            "Model with name soil already exists and is being replaced",
         ),
     )
     log_check(caplog, expected_log_entries)
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -11,6 +11,7 @@ import pytest
 from numpy import datetime64, timedelta64
 
 from virtual_rainforest.core.model import BaseModel
+from virtual_rainforest.soil.model import SoilModel
 
 from .conftest import log_check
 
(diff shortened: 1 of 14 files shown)


virtual_rainforest/soil/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+    schema_file = Path(__file__).parent.resolve() / "soil_schema.json"
+
+    with schema_file.open() as f:
+        config_schema = json.load(f)
+
+    return config_schema
+
+
+@register_model("soil")
+def generate_soil_model(
```
dalonsoa, line 26: This is generally what is called a `factory` method of a class: a class method that takes a configuration - often read from disk - and creates the object of that class. 

So, if used, this function will become a class method of the main class below:

```python
class SoilModel(BaseModel):

    @classmethod
    def factory(cls, config: dict[str, Any]) -> SoilModel:
        # Same contents than now, but ending in:
        return cls(start_time, end_time, update_interval, no_layers)
```

If used, `factory` should become an abstract method (and a class method) of `BaseClass` so all subclasses need to implement their own version.

If used in combination with `__init_subclass__` above, then `register_model` is not needed, and the only difference if that, when initialising the model from the registry, it should use:

```python
MODEL_REGISTRY[`soil`].factory(config)
```

All models will be initialised in the same way, with the `factory` method.

I think it is neater to keep the factory alongside the class, but it is up to you.

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,67 @@
...
+    Args:
+        start_time: Point in time that the model simulation should be started.
+        end_time: Time that the model simulation should end
+        update_interval: Time to wait between updates of the model state.
+
+    Attributes:
+        name: Names the model that is described
+    """
+
+    name = "soil model"
```
dalonsoa, line 36: Maybe just `soil`, as it is clear it is a model, by definition...

virtual_rainforest/core/model.py
```diff
@@ -0,0 +1,103 @@
...
+        pass
+
+    def cleanup(self) -> None:
+        """Function to delete objects within the class that are no longer needed."""
+        pass
+
+    def __repr__(self) -> str:
+        """Represent a Model as a string."""
+        return (
+            f"BaseModel(start_time={self.start_time}, end_time={self.end_time}, "
```
dalonsoa, line 85: If you use the following, there's no need to overwrite this method in child classes:

```suggestion
            f"{self.__class__.__name__}(start_time={self.start_time}, end_time={self.end_time}, "
```

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,67 @@
...
+    # SHOULD WORK
+    def setup(self) -> None:
+        """Function to set up the soil model."""
+        for layer in range(0, self.no_layers):
+            LOGGER.info(f"Setting up soil layer {layer}")
+
+    def __repr__(self) -> str:
+        """Represent a Model as a string."""
+        return (
+            f"SoilModel(start_time={self.start_time}, end_time={self.end_time}, "
```
dalonsoa, line 48: See above

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,67 @@
...
+            f"update_interval={self.update_interval}, no_layers={self.no_layers})"
+        )
+
+    def __init__(
+        self,
+        start_time: datetime64,
+        end_time: datetime64,
+        update_interval: timedelta64,
+        no_layers: int,
```
dalonsoa, line 57: Given that this child class is changing the signature of the parent class - `BaseClass` has no `no_layers` -, I'd use in `*args, **kwargs` in the `__init__` method of the later so it is clear that child classes might have more positional and keyword arguments. I'd keep those two here, in case this class is, in time, subclassed with more specialised versions.

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,67 @@
...
+from numpy import datetime64, timedelta64
+
+from virtual_rainforest.core.logger import LOGGER, log_and_raise
+from virtual_rainforest.core.model import BaseModel
+
+
+class InitialisationError(Exception):
+    """Custom exception class for model initialisation failures."""
+
+    pass
```
dalonsoa, line 16: No needed
```suggestion
```

dalonsoa (changes requested):
I've done a first pass and point out a few changes to the construction of the `BaseClass` and children, which I think will make the code more concise, but overall it looks great!

virtual_rainforest/soil/model.py
jacobcook1995: This definitely saves a lot of effort in terms of not having to write a new `__repr__` for every new model instance. I was wondering though whether there's anyway to generate to the function signature automatically, as this changes for the inheriting classes? Or alternatively if using the `__repr__` to print the full function signature is not actually a sensible thing to do?

dalonsoa: I've never used it myself, but probably you could see if you can exploit `inspect.Signature` for this: https://docs.python.org/3/library/inspect.html#inspect.Signature

virtual_rainforest/soil/model.py
jacobcook1995: So `SoilModel.__init__` would take `*args, **kwargs` which would then be unpacked and used to init `BaseModel`? Or should every `.__init__` method take `*args, **kwargs` as arguments?

dalonsoa: At least the base model - even if they are not used there - so it becomes clear that the signature of child classes might have more arguments. Changing the signature of methods in child classes is generally considered bad practice. Constructors are slightly different because it is common to have child classes requiring different inputs than the base classes, but even in that case, it is good if it can be indicated that the signature could be fluid in children. See for example https://stackoverflow.com/a/64156492

davidorme: At some point in the program structure, we need to be unpacking chunks of the `config` into their appropriate models. Now, we could do that:

1.  by having those config options as arguments to the subclasses (and hence use `*args, **kwargs` in the base class and call a subclass using `Plants(**config['plants'])`) or,
2.  we could have the base class accept a `config` arguments so that there is always a single argument to accept a config object that then gets unpacked within `__init__` (so models are always created as `Plants(config=config['plants'])`.

Having typed that out - I don't much like the second option. The first is clearer because then we have to document the subclass `__init__` arguments in the class docstring, which is the natural place for them. The only advantage I can see is that some of these models could have very complex signatures?

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,67 @@
...
+        name: Names the model that is described
+    """
+
+    name = "soil"
+
+    def __init__(
+        self,
+        start_time: datetime64,
+        end_time: datetime64,
+        update_interval: timedelta64,
```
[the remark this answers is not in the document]
jacobcook1995: Agree on `end_time`. In the case of `start_time` we need something to initially populate `last_update` so that the `should_update` function can run properly. We could use 0.0 and just decouple model timing from overall simulation (`datetime64`) timing. I think converting between times would be a fairly big source of potential errors though. So think we should keep `start_time` as an input, even it is just used to give `last_update` an initial value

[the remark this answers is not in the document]
davidorme: Interesting point. I guess I had been thinking that `should_update` is initialised as `True` so that it **always** runs first time (and with an order defined by the model order in the config?). So when the simulation starts up:

* the `spinup` method is run for each model 
* then we start the timing loop with a single common timeline 
* each model gets run for the first time in the first pass and then sets when next to run it

Otherwise, we could conceivably have a model that is badly configured not to run until after it is needed.

dalonsoa (commented):
I cannot complete the review now, but I've just left a quick suggestion on the `repr` part. Will aim to finish by tomorrow morning.

virtual_rainforest/core/model.py
```diff
@@ -0,0 +1,116 @@
...
+    name = "base"
+    # TODO - Once higher level timing function is written use it to set this
+    last_update = datetime64("2000-01-01")
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        # Save args and kwargs, so that they can be found by the __repr__
+        self._args = args
+        self._kwargs = kwargs
+
+        self.update_interval: timedelta64 = args[0]
```
dalonsoa, line 52: This is very fragile - what if the first positional argument is not a date. Or if there are no positional arguments at all! 

Why did you remove the explicit arguments you had before? I think that was a better approach. Explicit is always better than implicit. And it also duplicates the storage of information, as the inputs are include din the explicit attributes and in args/kwargs. Just saving sometime when creating the `repr` string does not justify that, in my opinion. 

I'd suggest the following approach, in combination with the comments below: include an attribute storing the attribute strings that need to be included in `repr`, and use that.

```suggestion
    def __init__(self, update_interval: timedelta64, **kwargs: Any):
        self.update_interval = update_interval
        self._repr = ["update_interval"]
```

virtual_rainforest/core/model.py
```diff
@@ -0,0 +1,116 @@
...
+                func_sig += f"{arg},"
+            else:
+                func_sig += f" {arg},"
+
+        if len(self._kwargs) > 0:
+            # TODO - Work out how to print kwargs nicely, (once we have some)
+            func_sig += f" {self._kwargs}"
+        else:
+            # Remove final comma
+            func_sig = func_sig[:-1]
```
dalonsoa, line 110: A better version, following my last comment above:
```suggestion
        func_sig = ", ".join([f"{k} = {getattr(self, k)}" for k in self._repr])
```

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if args[1] < 1:
+            log_and_raise(
+                "There has to be at least one soil layer in the soil model!", ValueError
+            )
+        elif not isinstance(args[1], int):
+            log_and_raise("The number of soil layers must be an integer!", TypeError)
+
+        super(SoilModel, self).__init__(*args)
+        self.no_layers = args[1]
```
dalonsoa, line 51: Follows the comments above:

```suggestion
    def __init__(self, update_interval: timedelta64, no_layers: int, **kwargs: Any):
        no_layers = int(no_layers)
        if no_layers < 1:
            log_and_raise(
                "There has to be at least one soil layer in the soil model!", ValueError
            )

        super(SoilModel, self).__init__(update_interval, **kwargs)
        self.no_layers = no_layers
        self._repr.append("no_layers")
```

virtual_rainforest/soil/model.py
```diff
@@ -0,0 +1,108 @@
...
+
+        LOGGER.info(
+            "Information required to initialise the soil model successfully extracted."
+        )
+
+        return cls(update_interval, no_layers)
+
+    # THIS IS BASICALLY JUST A PLACEHOLDER TO DEMONSTRATE HOW THE FUNCTION OVERWRITING
+    # SHOULD WORK
+    def setup(self) -> None:
```
dalonsoa, line 96: Keeping aside this is a placeholder, what would be the purpose of the `setup` method? Why the model cannot be setup in the `factory` method or the `__init__` method? Seems redundant to me.

tests/test_models.py
```diff
@@ -0,0 +1,148 @@
...
+        (),
+    )
+
+
+@pytest.mark.parametrize(
+    "no_layers,raises,expected_log_entries",
+    [
+        (
+            2,
+            does_not_raise(),
```
dalonsoa, line 48: neat!

dalonsoa (changes requested):
I've made a few comments, but they are interrelated, so if you decide not to apply them, then there's little for you to change. It looks good!

virtual_rainforest/soil/model.py
dalonsoa: I see. I think it can be split in two - or more functions - to keep the separation of concerns clear. The question is if they should be public methods defined in `BaseModel` to be overwritten by children classes or if they can be internal, defined by each class as needed.

For the use case you describe, I think `setup` does not need to be in the `BaseModel` class and can be called internally by `factory` (or `__init__` as appropriate). The only condition is that the created `grid` need to be passed as an argument to `factory`, so the pools can be setup appropriately in each cell.

A different situation is if that setup step needs information from other models. In that case, probably it is indeed necessary some sort of `setup` method used by models to talk to each other at `t=0` to define their initial state before the simulation can start.

davidorme: I think it is the latter case. I can imagine some models might not need to talk to other models (animals?) but the setup for plants and abiotic is likely to involve discussion. So, I think having the public method which is called for all models at t=0 (and might just `pass`) works here.

dalonsoa: In that case I would not make it an `abstractmethod` but rather a normal one that, by default, does nothing, so models that do not need any setup can skip overwriting it.

dalonsoa (approved):
All good now!

PR #110, opened 2022-11-04
Update to CI

jacobcook1995:
# Description

I've updated a number of things about the CI setup. Let me know if any of them don't seem sensible.
1) I've added python version 3.11 to the CI, as it is now out and we intend to support it.
1) I've updated the poetry version used to v1.2.2, so that `poetry.lock`'s created using `poetry` v1.2 don't cause CI build failures.
1) Increased minimum mypy requirement (in `pyproject.toml`) from 0.961 to 0.981, as this version fixes [my-py issue #13627](https://github.com/python/mypy/issues/13627). This means that it is no longer necessary to pin the CI to use python version 3.10.6, so 3.10 can be used instead.
1) Updated `actions/checkout` and `pre-commit/action` to versions that use Node.js 16 rather than 12, [see](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/). This fixes the depreciation warning that was coming up for all successful action runs.

## Type of change

## Key checklist

## Further checks


davidorme (commented):
Those all seem sound to me. One thing is that `poetry 1.2` introduced a new way of grouping dependencies, so that you can specify extra dependencies in themed groups (so we could have a `docs` group and a `testing` group). That seems useful, so we might want to update `pyproject.toml` to use that.

jacobcook1995:
> Those all seem sound to me. One thing is that `poetry 1.2` introduced a new way of grouping dependencies, so that you can specify extra dependencies in themed groups (so we could have a `docs` group and a `testing` group). That seems useful, so we might want to update `pyproject.toml` to use that.

Yeah that seems like a good idea, but probably only once everyone has updated to using `poetry 1.2` (I didn't find updating all that straightforward). Should I make a separate issue for it?

davidorme:
Personally, I think it makes sense to update it in this PR, then it has all moved across to `poetry 1.2` (although I get that is backwards compatible). This is a tricky one - I'm not sure how this PR works across branches. 

The `poetry` environment can only be updated by individuals _manually_, I think? But the `pyproject.toml` does specify the version required for building the package. What I don't know is whether that setting is only used to build actual distributions (wheels etc) or whether it also affects the 'development' installation that is running when we develop the code. 

So if we merge this - do we then need to get the team to all update. And how does that play out for @dalonsoa and @alexdewar, who have other projects that might currently use older `poetry`.

Basically, I think we should do this but the results are not restricted to this repo?

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        python-version: [ "3.9" , "3.10.6" ]
+        python-version: [ "3.9" , "3.10", "3.11" ]
```
alexdewar, line 22: Python 3.11? That's brave!

Not a bad idea to try with the latest version of Python, but just don't be surprised if things don't work yet. It'll probably be months (years?) before all the common packages actually work with v3.11.

alexdewar (approved):

.github/workflows/ci.yml
jacobcook1995: Ahh right, I think this is maybe a coming from Julia thing where the updates have often fixed major problems with the language, so people switch pretty quickly. I guess we can keep 3.11 in the CI for now and then remove it if/when it becomes a problem?

jacobcook1995:
I've taken a shot at grouping the dependancies, let me know if they seem like sensible groupings. A fair number are ungrouped, but I guess groups will become more obvious as we develop more

.github/workflows/ci.yml
davidorme: I briefly added `3.11-dev`  (7c9ef442074040e78be5183e7a9dd5dd74311eb8) and the wheels came off _alarmingly_ quickly.

alexdewar: Sure, up to you. It's not a bad thing, but it just might mean you have to remove it from the CI again later -- not a big deal.

alexdewar:
I've only ever bothered using a separate group for developer dependencies and maybe another one for building the docs. My logic has always been that you want all of your devs to run ``pre-commit`` and ``pytest`` etc. but you likely won't need tools for building docs as they'll just be running on the CI.

It's very a personal choice though and the fact that `poetry` has added this capability suggests that other people want more granular groupings too.

davidorme (approved):
LGTM

PR #112, opened 2022-11-14
Draft high level function to setup and run simulation

### LINKED ISSUE #98: Function to initialise set of models
We need a function that, based on the config settings, selects which models to run, initialises them, and then returns them as something that can be further operated upon.

These models should probably be subclasses of the `Model` class suggested in issue #96. These will be defined within each module, and should generally overwrite a significant number of the base class functions. These should be initialised using information from the config. This should be done by a function called by the top level `run_vr` function, which should then return an object containing all models that are to be run as part of the simulation.

Feel they should share a common naming, as anything else adds maintenance complications.

### LINKED ISSUE #100: High level run_vr function
As Diego suggested we need a high level function to run the virtual rainforest model

This function should read in the data from the data system, run the configuration process, use this information to set up a set of models to run, then use a timing loop to solve the models over time, and finally save output data.

jacobcook1995:
# Description

I've written a draft of the high level function to setup and run the `virtual_rainforest` model. Most of this function is just a set of TODOs at the moment, but config validation and model configuration is currently performed. This involved defining functions to extract the models to be configured from the config object, and to actually configure a set of models which can then be used downstream.

While doing this I also changed error handling a bit to ensure that the user is provided with the maximum amount of useful information, e.g. report all models that fail to configure before ending the process, and reporting all schema validation errors rather than just the first.

Fixes #98
Fixes #100

## Type of change

## Key checklist

## Further checks


virtual_rainforest/main.py
```diff
@@ -0,0 +1,126 @@
...
+    LOGGER.info(f"Attempting to configure the following models: {model_list}")
+
+    # Make list of missing models, and return an error if necessary
+    miss_model = [model for model in model_list if model not in MODEL_REGISTRY.keys()]
+    if miss_model != []:
+        LOGGER.error(
+            f"The following models cannot be configured as they are not found in the "
+            f"registry: {miss_model}"
+        )
+        return None
```
davidorme, line 42: I don't immediately see the value in having `select_models` return None here, rather than than simply raising the error, which is currently deferred to `vr_run`? Just fail and have a simpler return value (no Optional).

virtual_rainforest/main.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Returns:
+        modules: A set of models to be configured
+    """
+
+    # Remove "core" from model list as it is not a model
+    if "core" in model_list:
+        model_list.remove("core")
+
+    LOGGER.info(f"Attempting to configure the following models: {model_list}")
```
davidorme, line 33: The optimal style for formatting logging messages seems to be to still use `%s` style formats - mostly because the error messages doesn't get built unless the message is emitted. Here, I think f-strings are always evaluated. The costs in this case (and many cases) are trivial but we should be consistent in what we use where. I prefer f-strings for general compactness and readability, but I've been using `%s` style for logging because of the deferred/non-evaluation.

https://docs.python.org/3/howto/logging.html#optimization

Thoughts?

virtual_rainforest/main.py
```diff
@@ -0,0 +1,126 @@
...
+    if any(model is None for model in models_cfd):
+        log_and_raise(
+            "Could not configure all the desired models, ending the simulation.",
+            InitialisationError,
+        )
+        return
+    else:
+        LOGGER.info(
+            "All models successfully configured, now attempting to initialise them."
+        )
```
davidorme, line 110: Similarly, why is this not in `configure models` (which is otherwise a one liner).

davidorme (changes requested):
The main structure looks good - some thoughts about where errors get raised and a minor point about string formatting in logging

virtual_rainforest/main.py
jacobcook1995: Yeah I agree, easier to understand the higher level function if a `try: except` block is used instead of `if is None:` as well

davidorme: Here, you don't even need the higher level `try` since you're `log_and_raise`-ing. You can just let select_models fail and do the `log_and_raise` there.

virtual_rainforest/soil/model.py
```diff
@@ -40,57 +36,67 @@ class SoilModel(BaseModel, model_name="soil"):
...
-            log_and_raise(
+        except (ValueError, pint.errors.DimensionalityError) as e:
+            valid_input = False
+            LOGGER.error(
                 f"Configuration types appear not to have been properly validated. This "
                 f"problem prevents initialisation of the soil model. The first instance"
-                f" of this problem is as follows: {str(e)}",
-                InitialisationError,
+                f" of this problem is as follows: {str(e)}"
             )
```
davidorme, line 87: To me, this check belongs in the `__init__` method. If someone calls `SoilModel('23 martian days)'` programatically, then that has to throw a `DimensionalityError`. I think this method should only handle unpacking `config` and feeding it into a class instance. So `KeyErrors` but not stuff which needs validating in `__init__`.

virtual_rainforest/soil/model.py
```diff
@@ -40,57 +36,67 @@ class SoilModel(BaseModel, model_name="soil"):
...
+            )
 
         super(SoilModel, self).__init__(update_interval, **kwargs)
         self.no_layers = int(no_layers)
         # Save variables names to be used by the __repr__
         self._repr.append("no_layers")
 
     @classmethod
-    def factory(cls, config: dict[str, Any]) -> SoilModel:
+    def factory(cls, config: dict[str, Any]) -> Optional[SoilModel]:
```
davidorme, line 54: I'm not sure I like using `None` to signal failure here. It would be clearer to have something explicit. My first thought would be to have `BaseModel.init_failed` attribute that could then be checked in a loop:

```
if any([m.init_failed for m in model_list]):
    raise_and_log("Model initialisation failure")
```

That would then have to be something that each `__init__` handles. Another option - which is probably clearer - is to leave Model `__init__` methods as things that _raise_ Exceptions and don't have to worry about setting flags. The programatic use fails at the point of failure, but the factory method traps those failure modes, converts the exception to a `log_and_raise` and returns a `FailedModel` or something like that? And then have:

```
if any([isinstance(m, FailedModel) for m in  model_list]):
        raise_and_log("Model initialisation failure")
```

That's not entirely clear - I think what I'm trying to do is to keep a separation between core functions (which fail by raising) and the wrapping for the UI, which is set up to defer failure in some locations to provide as much feedback as possible for a given stage.

virtual_rainforest/main.py
jacobcook1995: Does using f-strings prevent deferred evaluation? Nothing in that optimisation doc seems to suggest that they do, but I admitted don't have a firm grasp of how python strings work so might be missing something obvious

jacobcook1995: Yep definitely see the value in avoiding expensive calls for logging output that won't actually ever be emitted. However, with their example:

```python
if logger.isEnabledFor(logging.DEBUG):
    logger.debug('Message with %s, %s', expensive_func1(),
                                        expensive_func2())
```

I was wondering if the optimisation was really to do with not using an f-string or rather nesting the logging call within a conditional statement? Like would the below not accomplish the same thing (particularly if the function calls had arguments)? Though admittedly in a less compact way

```python
if logger.isEnabledFor(logging.DEBUG):
    output1 = expensive_func1()
    output2 = expensive_func2()
    logger.debug(f"Message with {output1}, {output2}")
```

virtual_rainforest/soil/model.py
jacobcook1995: So the `__init__` would take in e.g. `"23 martian days"` as a string, and then the conversion to a `Quantity` (and potential throwing of a `DimensionalityError`) would take place there?

virtual_rainforest/soil/model.py
jacobcook1995: Yeah I returned a `None` in these cases because I didn't want to return a `Model` object that could be further operated on. I think I'm more inclined by the option of creating a `FailedModel` class, as it's a bit clearer. As well as being extensible for other types of model failures (e.g. failures of initialisation from data)

virtual_rainforest/soil/model.py
davidorme: Oh - huh. Hadn't twigged that SoilModel is expecting an `np.timedelta64` and so this is specifically converting from config user friendliness to programming API.

davidorme: Shouldn't the missing keys stuff be handled by the schema validation though?

virtual_rainforest/main.py
davidorme: That would work as:

```
if logger.isEnabledFor(logging.DEBUG):
    logger.debug(f"Message with {expensive_func1()}, {expensive_func2()}")
```

The optimisation link isn't particularly clear - this one is better: https://klichx.dev/2019/11/09/f-strings-and-python-logging/

virtual_rainforest/soil/model.py
jacobcook1995: Yeah the key stuff should be handled by the schema validation when the whole process is run. I put in additional checking here to cover for cases where the user inadvertently modified the configuration, but I'm not 100% sure whether or not that additional checking is sensible

virtual_rainforest/main.py
jacobcook1995: Ahh right so by not using f-strings you can avoid the evaluations when they are not gated behind an `if` statement, which is very useful. Definitely think we should switch then

virtual_rainforest/soil/model.py
jacobcook1995: Do you reckon I should remove the missing key checking? And would you recommend moving the `np.timedelta64` conversion to the `__init__`?

virtual_rainforest/core/config.py
```diff
@@ -425,9 +436,20 @@ def validate_with_defaults(
...
+            # Construct details of the tag associated with the error
+            tag = ""
+            for k in error.path:
+                tag += f"[{k}]"
+            LOGGER.error("%s: %s" % (tag, error.message))
         log_and_raise(
-            f"Validation of configuration files failed: {err.message}",
+            "Validation of configuration files failed see above errors",
             ConfigurationError,
         )
```
dalonsoa, line 454: Given that this is done identically above - and it is more substantial that just raising an exception - could we put it inside a function, so it is easier to test independently and reuse, if needed, in other places?

virtual_rainforest/core/model.py
```diff
@@ -99,3 +103,41 @@ def __repr__(self) -> str:
     def __str__(self) -> str:
         """Inform user what the model type is."""
         return f"A {self.name} model instance"
+
+
+class FailedModel(BaseModel, model_name="failed"):
```
dalonsoa, line 108: I really don't see the point of this Failed model. If a model fails to be constructed, then an execution should be raised. If we want o collect these exception and let the show go on with the rest of the models until we have constructed all of them or just stop things there is a separate question, as we discussed the other day. But this model, in itself just a placeholder that does nothing, does not seem to me as the right way to go.

virtual_rainforest/main.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # Remove "core" from model list as it is not a model
+    if "core" in model_list:
+        model_list.remove("core")
+
+    LOGGER.info("Attempting to configure the following models: %s" % model_list)
+
+    # Make list of missing models, and return an error if necessary
+    miss_model = [model for model in model_list if model not in MODEL_REGISTRY.keys()]
+    if miss_model != []:
```
dalonsoa, line 42: Not sure if it is faster or not, but I usually prefer:

```suggestion
    if len(miss_model):
```

virtual_rainforest/main.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    # Use factory methods to configure the desired models
+    models_cfd = [model.factory(config) for model in model_list]
+
+    if any([isinstance(model, FailedModel) for model in models_cfd]):
+        log_and_raise(
+            "Could not configure all the desired models, ending the simulation.",
+            InitialisationError,
+        )
```
dalonsoa, line 75: I think that a cleaner approach to do this without the `FailModel` class would be something like:

```suggestion
    failed_models = []
    models_cfd = []
    for model in model_list:
        try:
            models_cfd.append(model.factory(config))
        except (RelevantException, OrListOfThem):
            failed_models.append(model.name)

    if len(failed_models):
        log_and_raise(
            f"Could not configure all the desired models, ending the simulation. These failed {failed_models}.",
            InitialisationError,
        )
```
Which also shows the advantage of indicating which models failed.

virtual_rainforest/main.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    # Remove "core" from model list as it is not a model
+    if "core" in model_list:
+        model_list.remove("core")
+
+    LOGGER.info("Attempting to configure the following models: %s" % model_list)
+
+    # Make list of missing models, and return an error if necessary
+    miss_model = [model for model in model_list if model not in MODEL_REGISTRY.keys()]
```
dalonsoa, line 41: If you create a new lust - a set, actually, and remove `core`, you don't need to check if it is there and you don't need to have a `deepcopy` in the calling function.

```suggestion
    model_list_ = set(model_list) - {"core"}

    LOGGER.info("Attempting to configure the following models: %s" % model_list_)

    # Make list of missing models, and return an error if necessary
    miss_model = [model for model in model_list_ if model not in MODEL_REGISTRY.keys()]
```

virtual_rainforest/main.py
```diff
@@ -0,0 +1,125 @@
...
+
+    Args:
+        cfg_paths: Set of paths to configuration files
+        output_folder: Folder to save combined configuration to
+        out_file_name: Name for the combined configuration file
+    """
+
+    config = validate_config(cfg_paths, output_folder, out_file_name)
+
+    model_list = select_models(deepcopy(config["core"]["modules"]))
```
dalonsoa, line 98: No need to deepcopy if you do what I suggest above.

virtual_rainforest/soil/model.py
```diff
@@ -40,57 +36,67 @@ class SoilModel(BaseModel, model_name="soil"):
...
+            )
 
         super(SoilModel, self).__init__(update_interval, **kwargs)
         self.no_layers = int(no_layers)
         # Save variables names to be used by the __repr__
         self._repr.append("no_layers")
 
     @classmethod
-    def factory(cls, config: dict[str, Any]) -> SoilModel:
+    def factory(cls, config: dict[str, Any]) -> Union[SoilModel, FailedModel]:
```
dalonsoa, line 54: A factory method should always return the initialised model or fail with an exception. Returning something different is a code smell and likely to result in issues. It is not needed, as I suggest above.

virtual_rainforest/soil/model.py
```diff
@@ -40,57 +36,67 @@ class SoilModel(BaseModel, model_name="soil"):
...
+            try:
+                LOGGER.info(
+                    "Information required to initialise the soil model successfully "
+                    "extracted."
+                )
+                return cls(update_interval, no_layers)
+            except InitialisationError:
+                return FailedModel()
+        else:
+            return FailedModel()
```
dalonsoa, line 99: Not needed, as I point out above.

dalonsoa (changes requested):
As usual, I've reviewed the new code and will work on the tests once this is sorted out.

Good first attempt as having something that runs!!

virtual_rainforest/main.py
davidorme: Yeah - much better. My brain was stuck in list comprehension mode, so wanting to end up with a list of objects that we could then scan for failure.

dalonsoa (approved):
Very nice! Nothing more to add!

virtual_rainforest/main.py
```diff
@@ -0,0 +1,128 @@
...
+    failed_models = []
+    models_cfd = []
+    for model in model_list:
+        try:
+            models_cfd.append(model.from_config(config))
+        except InitialisationError:
+            failed_models.append(model.name)
+
+    # If any models fail to configure inform the user about it
+    if len(failed_models):
```
davidorme, line 73: This is really picky and a general question for @dalonsoa and @alexdewar. 

The 'pythonic' approach here is `if failed_models:` because the empty list == False. I think there is a general preference for these 'implicit' tests (https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations).

I think this is cleaner but I tend to make an exception for `None` - `if var is None:` rather than `if var:` because it is usually in cases where I _explicitly_ want to check that something is not None rather than some other 'falsey' value.

We should aim for a common style here.

alexdewar: I personally prefer `if failed_models` over `if len(failed_models)`. I think you're right that consistency is the main thing though.

davidorme: I definitely prefer `if failed_models` - but I'm also trying to justify my liberal use of `if var is None:` elsewhere! 

There is an argument to be made for using explicit testing of truthy/falsey objects when they might take a range of falsey values, but ending up in that situation feels like a code smell. I guess I tend to use `if var is None:` to check for `var` being undefined (default state/optional), which I think of as being different from 'falsey' and so want to make that explicit.

dalonsoa: In my case, I would normally go for the more verbose `if len(failed_models) == 0` because it is totally explicit and self explanatory. I'm not sure why y suggested the shortened version...

Anyway, consistency, as you and Alex are suggesting, is the important bit: choose one style; stick with it.

davidorme: OK - so to avoid hanging this up too much, lets go with implicit as in the Google style guide, with an exception for `if var is None:` for explicit semantic testing for _undefined_ . I realise that's a split decision. 🍰  and 🍰 👉 😮 .

davidorme (approved):
LGTM.

Issue #107, opened 2022-10-28
Developer docs for v0.1


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
In developing the basic model framework for version 0.1 we are introducing a fair number of constraints on how lower level functions are written. This isn't really a problem for those of us involved in actively developing the systems (e.g. me and David), but others will have to be caught up.

**Describe the solution you'd like**
Extend the developer docs to explain how development of specific modules should be fitted into the broader framework. We probably need to think a bit about the best format this could take.

PR #115, opened 2022-11-17
Improved module docstrings

### LINKED ISSUE #105: Extend the logger and document it
At the moment the logger is not documented, mainly because what it is currently doing is very simple. As we develop v0.1 we will need to use more complex logging behaviours (e.g. nesting logging entries). This extension will need to be documented so that all developers know how to use it correctly/consistently.

When the logger is extended documentation really must be added.

jacobcook1995:
# Description

I've extended the module docstrings for the `core.logger`, `core.config`, `core.model` and  `soil.model` modules. The aim is that they give enough information that people weren't involved in building these systems (i.e. Vivienne and Taran) can understand how they are supposed to be used. So let me know if anything is unclear, or if you feel like anything needs further expansion.

I'm also far from an expert in using reStructuredText, so David let me know if any of my usage seems suspect.

In terms of viewing the docstrings it is probably best to make the docs and then view them in browser. This can be done by running `cd docs` followed by `make html`. If this doesn't work you might have to make sure that `Jupyter` is properly set up (see [here](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/develop/docs/source/development/jupyter_notebooks.md#jupyter-lab)).

Fixes #105
Also addresses #107, but doesn't completely resolve it.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/virtual_rainforest/core/config.md
+++ b/docs/source/virtual_rainforest/core/config.md
@@ -36,10 +36,10 @@ information cannot be repeated between files as there is no way to establish whi
 two values (of e.g. `core.grid.nx`) the user intended to provide. In this case, the
 module will throw critical error and the `virtual_rainforest` model will not configure.
 The user supplies a list of config files and/or folders to look for config files, within
-the supplied folders every `toml` file will be read in. An output folder and file name
-should also be provided, if a file already exists in the output folder that has the same
-name as the user provided output file name the configuration will critically fail. This
-happens in order to minimise the chance of significant confusion downstream.
+the supplied folders every `toml` file will be read in. Once the complete configuration
+is validated, it is saved as a single file using a user provided name and location. If a
+file already exists with this name at this location, the configuration process will
+critically fail. This minimises the chance of significant confusion downstream.
 
 ## Optional module loading
 
--- a/docs/source/virtual_rainforest/core/config.md
+++ b/docs/source/virtual_rainforest/core/config.md
@@ -1,4 +1,4 @@
-# The `core.config` module
+# The configuration module
 
 This module is used to configure a `virtual_rainforest` simulation run. This module
 reads in a set of configuration files written using `toml`. It is setup in such a way as
--- a/docs/source/virtual_rainforest/core/config.md
+++ b/docs/source/virtual_rainforest/core/config.md
@@ -35,11 +35,11 @@ input can be spread across an arbitrarily large number of config files. However,
 information cannot be repeated between files as there is no way to establish which of
 two values (of e.g. `core.grid.nx`) the user intended to provide. In this case, the
 module will throw critical error and the `virtual_rainforest` model will not configure.
-Config files are read from a folder that the user specifies, this can either be every
-`toml` file in the folder, or a user provided list of files. If a file exists in this
-folder that has the same name as the user provided output file name the configuration
-will critically fail, in order to minimise the chance of significant confusion
-downstream.
+The user supplies a list of config files and/or folders to look for config files, within
+the supplied folders every `toml` file will be read in. An output folder and file name
+should also be provided, if a file already exists in the output folder that has the same
+name as the user provided output file name the configuration will critically fail. This
+happens in order to minimise the chance of significant confusion downstream.
 
 ## Optional module loading
 
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -55,7 +55,10 @@ else:
     import tomli as tomllib
 
 SCHEMA_REGISTRY: dict = {}
-"""A registry for different module schema."""
+"""A registry for different module schema.
+
+:meta hide-value:
+"""
 
 
 class ConfigurationError(Exception):
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -2,10 +2,10 @@
 
 The `core.config` module is used to read in the various configuration files, validate
 their contents, and then configure a ready to run instance of the virtual rainforest
-model.
+model. The basic details of how this system functions can be found
+:ref:`here<virtual_rainforest/core/config:the configuration module>`.
 """
-# TODO - Think about how to expand the above docstring
-# Maybe just link the user to the already written config docs?
+# OKAY DECORATOR USAGE SHOULD BE DEMONSTRATED
 
 import sys
 from collections import ChainMap
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -2,10 +2,40 @@
 
 The `core.config` module is used to read in the various configuration files, validate
 their contents, and then configure a ready to run instance of the virtual rainforest
-model. The basic details of how this system functions can be found
+model. The basic details of how this system is used can be found
 :ref:`here<virtual_rainforest/core/config:the configuration module>`.
+
+When a new module is defined a `JSON` file should be written, which includes the
+expected configuration tags, their expected types, any value constraints, and where
+appropriate default values. This schema should be saved in the folder of the module that
+it relates to. In order to make this schema generally accessible to the `vr` package, it
+should then be added to the schema registry.  The
+:func:`~virtual_rainforest.core.config.register_schema` decorator is used for this
+purpose.
+
+Schema registration for each module takes place in the module's `__init__.py`. Here, a
+function is written that reads in the `JSON` file describing the schema. This function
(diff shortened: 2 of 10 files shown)


docs/source/virtual_rainforest/core/config.md
```diff
@@ -35,11 +35,11 @@ input can be spread across an arbitrarily large number of config files. However,
 information cannot be repeated between files as there is no way to establish which of
 two values (of e.g. `core.grid.nx`) the user intended to provide. In this case, the
 module will throw critical error and the `virtual_rainforest` model will not configure.
```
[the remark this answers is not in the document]
davidorme: Note that the `:role:` thing needs to vary with what is being pointed to: `:class:`,  `:meth:`, `:func:`, `:attr:` etc. etc.

davidorme (changes requested):
Some structural and formatting stuff. I haven't scanned content yet!

davidorme:
This is now built at: https://virtual-rainforest.readthedocs.io/en/feature-extend_module_docstrings

TaranRallings (approved):

docs/source/virtual_rainforest/core/config.md
```diff
@@ -35,11 +35,11 @@ input can be spread across an arbitrarily large number of config files. However,
...
-Config files are read from a folder that the user specifies, this can either be every
-`toml` file in the folder, or a user provided list of files. If a file exists in this
-folder that has the same name as the user provided output file name the configuration
-will critically fail, in order to minimise the chance of significant confusion
-downstream.
+The user supplies a list of config files and/or folders to look for config files, within
+the supplied folders every `toml` file will be read in. An output folder and file name
+should also be provided, if a file already exists in the output folder that has the same
+name as the user provided output file name the configuration will critically fail. This
```
vgro, line 41: The updated text reads better but could be simplified, for example 'if a file name already exists the configuration will fail'

virtual_rainforest/core/config.py
```diff
@@ -1,10 +1,42 @@
...
+************************************************** # noqa: D205
+
+The :mod:`core.config` module is used to read in the various configuration files,
+validate their contents, and then configure a ready to run instance of the virtual
+rainforest model. The basic details of how this system is used can be found
+:ref:`here<virtual_rainforest/core/config:the configuration module>`.
+
+When a new module is defined a ``JSON`` file should be written, which includes the
+expected configuration tags, their expected types, any value constraints, and where
+appropriate default values. This schema should be saved in the folder of the module that
```
vgro, line 11: there is a verb missing 'where appropriate default values??'

virtual_rainforest/core/config.py
```diff
@@ -1,10 +1,42 @@
...
+    def schema() -> dict:
+        schema_file = Path(__file__).parent.resolve() / "example_module_schema.json"
+
+        with schema_file.open() as f:
+            config_schema = json.load(f)
+
+        return config_schema
+
+It's important to note that the schema will only be added to the registry if the module
+``__init__`` is run. This means that somewhere in your path the module must be imported.
```
vgro, line 36: From just reading this section, it is not clear to me what 'the path' is. Is it pointing to the location of the vr package or the module?

virtual_rainforest/core/logger.py
```diff
@@ -1,10 +1,51 @@
...
+are displayed. In practice, we are likely to generally set the logging level to INFO so
+that DEBUG messages are suppressed, except when we are actively trying to debug the
+model.
+
+The logging module also defines a function that we should generally make use of to kill
+the simulation when something goes wrong. This function `log_and_raise()` raises an
+exception (of the developers choice) and adds a developer specified CRITICAL message to
+the log. This function is useful as it ensures that CRITICAL logging events are
+accompanied by a simulation ending exception. As such, this is probably the only means
+by which you should log a CRITICAL message.
```
vgro: This introduction is very good and clear 👍

virtual_rainforest/core/model.py
```diff
@@ -1,12 +1,42 @@
...
+************************************************** # noqa: D205
+
+The `core.model` module defines the api that all individual models (e.g. the soil model)
+should conform to. This consists of a class (`BaseModel`), which defines the expected
+functions. This class is an abstract base class, which is a class that is never intended
+to be instantiated itself but instead serves as a blueprint for inheriting classes. Some
+of its functions will be inherited and used by its child classes (e.g. `__repr__` and
+`__str__`), unless they are explicitly overwritten in the child class. However, it also
+possesses abstract methods (denoted by `@abstractmethod`), which are merely
+placeholders. For these abstract methods child classes have to overwrite the methods
```
vgro:
```suggestion
placeholders. For these abstract methods, child classes have to overwrite the methods
```
This reads a bit easier with a comma  :-)

vgro: Also, could you give an example for a case when a child overwrites an abstract method?

virtual_rainforest/core/model.py
```diff
@@ -1,12 +1,42 @@
...
+with new functions, otherwise class inheritance fails. This ensures that a consistent
+api (set of functions) is used across models, while also ensuring that functions don't
+default to an inappropriate generic behaviour due to a function not being defined for a
+particular model. Abstract classes should thus be used for functions that are always
```
vgro: do you mean 'abstract class' or 'abstract method' here?
Also, does that mean that every function we write in our modules needs to be a child of the abstract method class?

virtual_rainforest/core/model.py
```diff
@@ -1,12 +1,42 @@
...
+`__init_subclass__` function means that every child class is automatically added to the
+model registry (called `MODEL_REGISTRY`), but that in every case a `model_name` has to
+be provided to register it under. This model name should be unique. Although
+registration is automatic, it only happens when class definition happens, which requires
+the module which defines the child class to be imported somewhere. At the moment, we
+import all child models in the top level `__init__.py` to ensure that they are added to
+the registry.
+
+An example of `Model` inheritance from `BaseModel` can been seen in the `soil.model`
+module.
```
vgro, line 45: I think you did a very good job here, it could possibly be simplified or more concise in places, but this is really a minor critique.

virtual_rainforest/core/config.py
jacobcook1995: Yeah I tried to cram too much info into one list there, I'll separate it out a bit

virtual_rainforest/core/config.py
jacobcook1995: So basically so long as the module of interest (e.g. `virtual_rainforest.soil`) is imported somewhere in setting up your script to run the registry gets populated. This doesn't have to be direct and you can get away with importing something that imports something else which imports the `virtual_rainforest.soil` module. In this case, the registry would have been properly populated with the soil model.

Path is definitely not a good way to describe this though, I'm not sure how to phrase it. Does "chain of imports" make more sense, or should I aim for a more wordy description of the problem?

virtual_rainforest/core/model.py
jacobcook1995: I think the idea is that the high-level functions that setup, pinup and run the whole model should all follow a consistent pattern. Think these functions will call loads of module specific functions. I don't think there's any sense in the module specific functions being based on an abstract base class in general, as there won't be great similarities between modules at this level. Basically what `setup`, `spinup`, `solve` etc are there to do is to transition from the shared high level api to the module specific details, so as far as I'm aware I don't think there's any sense in lower level functions sharing a common framework/abc.

jacobcook1995: And yeah you are right that should be `Abstract methods`

jacobcook1995: Do you think I should add any of my above reply into the docstring?

virtual_rainforest/core/model.py
jacobcook1995: Tried to make it slightly less rambly

virtual_rainforest/core/model.py
vgro: Yes, I think it would be good if you could clarify this paragraph a bit more 👍

vgro: I think the more examples you can provide, the better. It is quite an abstract topic...

vgro (approved):

vgro (approved):
Thanks for clarifying these sections 👍

davidorme (approved):
LGTM

PR #117, opened 2022-11-21
Added basic entry point

### LINKED ISSUE #101: Command line entry point
Need to define an entry point so that the simulation can be run from the command line.

Write a client script that takes in a list of configuration files from the command line, configures the simulation and then runs the high level `vr_run` function with this configuration.

Considered writing multiple entry points, e.g. one for configuration, one for actually running. But seems simpler to have a single entry point now, and then refactoring it down the line as required.

jacobcook1995:
# Description

Added a `vr_run` entry point that allows the model to be run from the command line.

At the moment the testing of this entry point is pretty minimal, i.e. just tests that the command exists and that version info can be returned correctly. I'm not 100% on how appropriate unit testing is for very high level functions, but let me know your thoughts.

Fixes #101 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,14 +41,7 @@ The research team are supported by the Imperial College London
 Engineering](https://www.imperial.ac.uk/admin-services/ict/self-service/research-support/rcs/service-offering/research-software-engineering/)
 team.
 
-## Example usage
+## Model documentation
 
-An example configuration for the Virtual Rainforest model is provided in
-`data/example_config.toml`. A simulation using this configuration can be run through the
-command line by calling:
-
-```shell
-poetry shell
-poetry install
-vr_run data/example_config.toml
-```
+The full documentation for the Virtual Rainforest model can be viewed
+[here](https://virtual-rainforest.readthedocs.io/en/latest/).
--- a/README.md
+++ b/README.md
@@ -40,3 +40,15 @@ The research team are supported by the Imperial College London
 [Research Software
 Engineering](https://www.imperial.ac.uk/admin-services/ict/self-service/research-support/rcs/service-offering/research-software-engineering/)
 team.
+
+## Example usage
+
+An example configuration for the Virtual Rainforest model is provided in
+`data/example_config.toml`. A simulation using this configuration can be run through the
+command line by calling:
+
+```shell
+poetry shell
+poetry install
+vr_run data/example_config.toml
+```
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -6,6 +6,7 @@ defined in main.py that it calls.
 
 from contextlib import nullcontext as does_not_raise
 from logging import CRITICAL, ERROR, INFO
+from pathlib import Path
 
 import pytest
 
@@ -162,7 +163,7 @@ def test_vr_run_miss_model(mocker, caplog):
     mock_conf.return_value = {"core": {"modules": ["topsoil"]}}
 
     with pytest.raises(InitialisationError):
-        vr_run("tests/fixtures/all_config.toml", ".", "delete_me")
+        vr_run("tests/fixtures/all_config.toml", Path("./delete_me.toml"))
 
     expected_log_entries = (
         (INFO, "Attempting to configure the following models: ['topsoil']"),
@@ -193,7 +194,7 @@ def test_vr_run_bad_model(mocker, caplog):
     }
 
     with pytest.raises(InitialisationError):
-        vr_run("tests/fixtures/all_config.toml", ".", "delete_me")
+        vr_run("tests/fixtures/all_config.toml", Path("./delete_me.toml"))
 
     expected_log_entries = (
         (INFO, "Attempting to configure the following models: ['soil']"),
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -164,6 +164,8 @@ def test_vr_run_miss_model(mocker, caplog):
 
     with pytest.raises(InitialisationError):
         vr_run("tests/fixtures/all_config.toml", Path("./delete_me.toml"))
+        # If vr_run is successful (which it shouldn't be) clean up the file
+        Path("./delete_me.toml").unlink()
 
     expected_log_entries = (
         (INFO, "Attempting to configure the following models: ['topsoil']"),
@@ -195,6 +197,8 @@ def test_vr_run_bad_model(mocker, caplog):
 
     with pytest.raises(InitialisationError):
         vr_run("tests/fixtures/all_config.toml", Path("./delete_me.toml"))
+        # If vr_run is successful (which it shouldn't be) clean up the file
+        Path("./delete_me.toml").unlink()
 
     expected_log_entries = (
         (INFO, "Attempting to configure the following models: ['soil']"),
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -4,7 +4,6 @@ The `core.config` module is used to read in the various configuration files, val
 their contents, and then configure a ready to run instance of the virtual rainforest
 model.
 """
-# TODO - find config folder based on command line argument
 
 import sys
 from collections import ChainMap
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -178,38 +178,39 @@ def check_dict_leaves(
     return conflicts
 
 
-def check_outfile(output_folder: str, out_file_name: str) -> None:
+def check_outfile(merge_file_path: Path) -> None:
     """Check that final output file is not already in the output folder.
 
     Args:
-        output_folder: Path to a folder to output the outputted complete configuration
-            file to
-        out_file_name: The name to save the outputted complete configuration file under
+        merge_file_path: Path that merged config file is meant to be saved to
     Raises:
         ConfigurationError: If the final output directory doesn't exist, isn't a
-            directory, or the final output file already exist there.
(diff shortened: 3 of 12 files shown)


virtual_rainforest/core/config.py
```diff
@@ -187,16 +186,30 @@ def check_outfile(output_folder: str, out_file_name: str) -> None:
             file to
         out_file_name: The name to save the outputted complete configuration file under
     Raises:
-        ConfigurationError: If the final output file already exist.
+        ConfigurationError: If the final output directory doesn't exist, isn't a
+            directory, or the final output file already exist there.
```
alexdewar, line 190:
```suggestion
            directory, or the final output file already exists.
```

virtual_rainforest/entry_points.py
```diff
@@ -0,0 +1,67 @@
...
+
+    This program sets up and runs a simulation of the Virtual Rainforest model. At
+    present this is incomplete. At the moment a set of configuration files are read in
+    based on user supplied paths, these are converted to a single configuration file,
+    which is then output for further reference. This combined configuration is then used
+    to initialise a set of models.
+    """
+
+    # Check function docstring exists, as -OO flag strips docstrings I believe
+    desc = textwrap.dedent(_vr_run_cli.__doc__ or "Python in -OO mode: no docs")
```
alexdewar, line 28: Where did you find out that `-O0` strips docstrings? I'm assuming you're talking about C-flags here, in which case I'd be quite surprised: `-O0` means "disable compiler optimisations", so why this would strip docstrings out I have no idea. I'd be surprised if any level of optimisation removes docstrings.

alexdewar (approved):
Apart from one query and one suggestion, it looks good.

I'm assuming this works but haven't checked. I think you're right that there's not a lot of point in adding unit tests for higher-level functions: they should be for small units.

Is there an example config file in the repository? If not, I think it'd be worth adding one (it doesn't really matter what it does) along with a note in the readme saying how to run it with `vr_run`.

virtual_rainforest/entry_points.py
jacobcook1995: Yeah it seemed strange to me (and I must admit to not being 100% sure on what `python -OO` is/does), but `mypy` insisted that we provide an alternative for when `_vr_run_cli.__doc__ = None`. (This)[https://github.com/python/mypy/issues/9170] `mypy` issue suggested that this was to do with ensuring support for the `-OO` flag

jacobcook1995: Actually here's the details from `python --help`:
"
-O     : remove assert and __debug__-dependent statements; add .opt-1 before
         .pyc extension; also PYTHONOPTIMIZE=x
-OO    : do -O changes and also discard docstrings; add .opt-2 before
         .pyc extension
"
which makes sense as `-OO` explicitly discards dosctrings. Agree that removing docstrings seems like a very intense level of optimisation though

alexdewar: Oh, that's weird! Thanks for the info. I guess there's no choice then!

I assumed it was something to do with gcc's optimisation flags, but apparently not.

alexdewar: You can disable `mypy` for a single line by adding `# ignore: type` to the end, which is always an option.

davidorme (changes requested):
This looks good - some minor structural stuff, which is mostly my opinion!

virtual_rainforest/entry_points.py
davidorme: That's true - forgot that! The fix is easy enough here, I guess, even if it is madly obscure. We might want to start up a document on `mypy` strategy for the repo to cover what we are going to do in particular cases so that we can keep them consistent?

virtual_rainforest/entry_points.py
```diff
@@ -0,0 +1,67 @@
...
+    parser.add_argument(
+        "-o",
+        "--output_folder",
+        default=".",
+        type=str,
+        help="Folder that the output config file should be saved in.",
+        dest="output_folder",
+    )
+    parser.add_argument(
+        "-n",
```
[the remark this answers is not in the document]
jacobcook1995: Yeah I agree, that is a bit of a hangover from when we were reading files from a specified config folder. So this can probably all be collapsed into a single merge option

virtual_rainforest/entry_points.py
alexdewar: That's a good idea, though I'd probably keep it as general advice rather than having formal rules about it. It would be worth mentioning that `# ignore: type` exists and that it should only be used when necessary.

alexdewar: (PS - I find I mostly end up using it for bits of code where I know a value can't possibly be `None`, but `mypy` hasn't figured that out.)

davidorme: Yeah - that is very aggravating. I've just had a thing where I want `thing = None` and then a loop looks for match to set `thing  = 'str_value'`. Then I explicitly handle the `None` case and after that use the  found `str_value`  and `mypy` throws its toys. So, instead I start with `thing = ""`, which seems less clear to me.

alexdewar: Yeah -- less pythonic too.

You can also do this instead:
```py
thing: Optional[str] = None
# etc.
```

then `mypy` knows it is allowed to be `None` or a string.

dalonsoa: I'm sorry I'm late to this conversation, but why we need this command, to start with?

davidorme: Which command, Diego? The `textwrap.dedent` or the whole `_vr_run_cli()`?

dalonsoa: No, the `textwrap.dedent`. And, it is running VR with `-OO` an expected use case that needs this? Just curious - I'm not complaining about it.

README.md
```diff
@@ -40,3 +40,15 @@ The research team are supported by the Imperial College London
...
+
+An example configuration for the Virtual Rainforest model is provided in
+`data/example_config.toml`. A simulation using this configuration can be run through the
+command line by calling:
+
+```shell
+poetry shell
+poetry install
+vr_run data/example_config.toml
+```
```
davidorme, line 54: I don't think this is the right place for this. The README is very simple and this example is really out of context and hard to follow here. I think this needs to be in the `docs` and be the beginnings of a basic how to page.

Also the `poetry` installation is very much the developer mode. Most end users will `pip` it - and then there is the problem that the path of the config file will be within the local python library and not the working directory. There must be a canonical way to simplify user access to provided example files!

davidorme: If you do move this, a good time to add that RTD link 😄

tests/test_main.py
```diff
@@ -162,7 +163,7 @@ def test_vr_run_miss_model(mocker, caplog):
     mock_conf.return_value = {"core": {"modules": ["topsoil"]}}
 
     with pytest.raises(InitialisationError):
-        vr_run("tests/fixtures/all_config.toml", ".", "delete_me")
+        vr_run("tests/fixtures/all_config.toml", Path("./delete_me.toml"))
```
davidorme, line 167: What's this `delete_me.toml` file? Is it a real file? I don't understand how it is being used.

davidorme (changes requested):

README.md
davidorme: And promote @TaranRallings !

tests/test_main.py
jacobcook1995: It's a file path to write the complete configuration to, if it exists at the start then the test would fail, which is part of why I've given it a stupid name. The other reason I gave it a stupid name is that if for some reason the test doesn't fail and raise an `InitialisationError`, `vr_run` will actually write the config to that name. So I gave it that name during writing the unit tests to remind me to delete it when it got accidentally written out

README.md
jacobcook1995: Ahh yeah fair enough, I will move it to a basic how to page

tests/test_main.py
davidorme: Got it. That seems clunky - we don't want tests that pollute the file space on failure. I'm not sure of the best solution though - the obvious one is to use `pyfakefs` to set up a virtual file system to run tests in. That seems like a sledgehammer to crack this, but we might well make use of it later.

davidorme: Alternatives:

 -  the test could look for the file and delete it on the way out.
 - you could write to an explicit [tempfile](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp), but that is just moving the litter to somewhere we can't see it 😄

README.md
davidorme: And I think `install` should be before `shell` if you do hang on to to the `poetry` installation for the moment.

davidorme (changes requested):
Accidentally started a second review!

README.md
davidorme: When it is `pip` installed, _all_ of the package is (normally) installed to one of the `site-packages` locations (user or system). The `vr_run` command is installed so that it can be found from `PATH`, but the path to any specific file in the package is going to be obscure. I think the solution is that we want a `vr install_demo` command that can retrieve the config and some accompanying data files into a given location. That can programatically access the file locations and then additionally make sure that users play around with _copies_ of the files rather than corrupting the package data.

davidorme (approved):
LGTM

PR #121, opened 2022-11-28
Main timing loop

### LINKED ISSUE #103: Timing of overall simulation
The top level function to run the simulation needs to set up a timing loop in order to do so.

The timing settings should be read in from the configuration, this should be used to set up a while loop that handles the timing. Models should have been set up to only update a particular timing intervals

jacobcook1995:
# Description

I've added a timing loop to the main function which runs the Virtual Rainforest model. This loop controls which models should be updated on a given time step. I've also added a few functions which check that timing options work sensibly together, and warn the user when they probably don't. To facilitate all this the time check function defined in `BaseModel` was modified, and a new function was added which starts the timing.

One thing it would be useful to get feedback on is whether the way I divide models between `to_refresh` and `fixed` in `get_models_to_update` is efficient. Because at some point these are presumably going to be fairly large objects, so if I'm copying them in an inefficient manner the performance overhead might be quite large.

Anyway, let me know your thoughts. 

Fixes #103 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/model.py
+++ b/virtual_rainforest/core/model.py
@@ -100,7 +100,7 @@ class BaseModel(ABC):
     def should_update(self, current_time: datetime64) -> bool:
         """Determines whether a model should be updated for a specific time step."""
 
-        if current_time > self.last_update + self.update_interval:
+        if current_time >= self.last_update + self.update_interval:
             self.last_update = current_time
             return True
         return False
--- a/virtual_rainforest/core/model.py
+++ b/virtual_rainforest/core/model.py
@@ -45,8 +45,7 @@ class BaseModel(ABC):
     """
 
     name = "base"
-    # TODO - Once higher level timing function is written use it to set this
-    last_update = datetime64("2000-01-01")
+    last_update = None
 
     def __init__(self, update_interval: timedelta64, **kwargs: Any):
         self.update_interval = update_interval
@@ -87,7 +86,10 @@ class BaseModel(ABC):
     def should_update(self, current_time: datetime64) -> bool:
         """Determines whether a model should be updated for a specific time step."""
 
-        if current_time > self.last_update + self.update_interval:
+        if (
+            not self.last_update
+            or current_time > self.last_update + self.update_interval
+        ):
             self.last_update = current_time
             return True
         return False
--- a/virtual_rainforest/core/model.py
+++ b/virtual_rainforest/core/model.py
@@ -16,10 +16,14 @@ from typing import Any, Type
 
 from numpy import datetime64, timedelta64
 
-from virtual_rainforest.core.logger import LOGGER
+from virtual_rainforest.core.logger import LOGGER, log_and_raise
 
 MODEL_REGISTRY: dict[str, Type[BaseModel]] = {}
 """A registry for different models."""
+
+
+class ImproperFunctionCall(Exception):
+    """Custom exception class for cases where functions should not have been called."""
 
 
 class InitialisationError(Exception):
@@ -45,7 +49,6 @@ class BaseModel(ABC):
     """
 
     name = "base"
-    last_update = None
 
     def __init__(self, update_interval: timedelta64, **kwargs: Any):
         self.update_interval = update_interval
@@ -83,13 +86,21 @@ class BaseModel(ABC):
     def setup(self) -> None:
         """Function to use input data to set up the model."""
 
+    def start_model_timing(self, start_time: datetime64) -> None:
+        """Set of model timing by setting up initial last update time."""
+
+        if hasattr(self, "last_update"):
+            log_and_raise(
+                "Model timing was already set up, it should not be setup again!",
+                ImproperFunctionCall,
+            )
+        else:
+            self.last_update = start_time
+
     def should_update(self, current_time: datetime64) -> bool:
         """Determines whether a model should be updated for a specific time step."""
 
-        if (
-            not self.last_update
-            or current_time > self.last_update + self.update_interval
-        ):
+        if current_time > self.last_update + self.update_interval:
             self.last_update = current_time
             return True
         return False
--- a/virtual_rainforest/core/model.py
+++ b/virtual_rainforest/core/model.py
@@ -31,8 +31,8 @@ methods, child classes have to overwrite the methods with new functions, otherwi
 inheritance fails. This ensures that a consistent api (set of functions) is used across
 models, while also ensuring that functions don't default to an inappropriate generic
 behaviour due to a function not being defined for a particular model. At the moment, we
-expect every model to have a setup, spinup, solve and cleanup process, though this might
-change in the future.
+expect every model to have a setup, spinup, update and cleanup process, though this
+might change in the future.
 
 We also define an abstract class method to perform model initialisation. This method
 (:func:`~virtual_rainforest.core.model.BaseModel.from_config`) is a factory method which
@@ -67,14 +67,10 @@ from typing import Any, Type
 
 from numpy import datetime64, timedelta64
 
-from virtual_rainforest.core.logger import LOGGER, log_and_raise
+from virtual_rainforest.core.logger import LOGGER
 
 MODEL_REGISTRY: dict[str, Type[BaseModel]] = {}
 """A registry for different models."""
-
-
-class ImproperFunctionCall(Exception):
-    """Custom exception class for cases where functions should not have been called."""
 
 
 class InitialisationError(Exception):
@@ -85,7 +81,7 @@ class BaseModel(ABC):
     """A superclass for all `vr` models.
 
     Describes the common functions and attributes that all `vr` models should have. This
-    includes functions to setup, spin up and solve the specific model, as well as a
+    includes functions to setup, spin up and update the specific model, as well as a
     function to cleanup redundant model data. At this level these functions are not
     define and are mere placeholders to be overwritten (where appropriate) by the
     inheriting classes.
@@ -101,18 +97,21 @@ class BaseModel(ABC):
 
     name = "base"
 
-    def __init__(self, update_interval: timedelta64, **kwargs: Any):
+    def __init__(
+        self, update_interval: timedelta64, start_time: datetime64, **kwargs: Any
+    ):
         self.update_interval = update_interval
+        self.next_update = start_time + update_interval
         # Save variables names to be used by the __repr__
-        self._repr = ["update_interval"]
+        self._repr = ["update_interval", "next_update"]
 
     @abstractmethod
     def spinup(self) -> None:
         """Function to spin up the model."""
 
     @abstractmethod
-    def solve(self) -> None:
-        """Function to solve the model."""
+    def update(self) -> None:
+        """Function to update the model."""
 
     @abstractmethod
     def cleanup(self) -> None:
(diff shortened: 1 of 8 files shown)


virtual_rainforest/main.py
```diff
@@ -108,20 +230,38 @@ def vr_run(
...
+    current_time = setup_timing_loop(start_time, end_time, update_interval)
+    while current_time < end_time:
+
+        current_time += update_interval
+
+        to_refresh, fixed = get_models_to_update(current_time, models_cfd)
+
+        # TODO - Solve models to steady state
+
+        models_cfd = to_refresh + fixed
```
davidorme, line 262: I'm not convinced by this - there is a lot of model passing around for what need to be some frequent checks. And I'm also not convinced by the separate   time 'grain' for the models and update_interval.

* I think update_interval should small (let's say 1 minute). There is a cost to repeating this loop at a fast frequency but all it should be doing really is keeping track of changing time.
* The whole `get_models_to_update` is _expensive_. I think we should track just a set of `next_update`times. Something like this:

```python
# Get the list of date times of the next update.
update_due = {mod.name: mod.next_update for mod in models_cfd}

# Loop until the end
while current_time < end_time:
    # Get the names of models that have expired due dates
    update_needed = [nm for nm, upd in update_due.items() if upd < current_time ]

    # If there are expired model, run their update() method and update their update due dates
    if update_needed:
        for mod_nm in update_needed:
            models_cfg[mod_nm].update()
            update_due[mod_nm] = models_cfg[mod_nm].next_update()

```

davidorme: There's a huge issue in here with models running at different rates and how to synchronize them. Ultimately that would be great to actually do but we need to kick this massively into touch for now and enforce synchronised model updates.

virtual_rainforest/main.py
```diff
@@ -81,6 +85,124 @@ def configure_models(
...
+        )
+    except (pint.errors.DimensionalityError, pint.errors.UndefinedUnitError):
+        log_and_raise(
+            "Units for core.timing.main_time_step are not valid time units: %s"
+            % config["core"]["timing"]["main_time_step"],
+            InitialisationError,
+        )
+    else:
+        # Round raw time interval to nearest minute
+        update_interval = timedelta64(int(raw_interval.magnitude), "m")
```
davidorme, line 119: This worries me - I don't think we're going to be troubled by it immediately, but why do we need to round? I think this gets a lot easier if we reconsider the grain of the timing loop (see comment below)

virtual_rainforest/core/model.py
```diff
@@ -135,10 +137,21 @@ def __init_subclass__(cls, model_name: str):
...
+        """Set of model timing by setting up initial last update time."""
+
+        if hasattr(self, "last_update"):
+            log_and_raise(
+                "Model timing was already set up, it should not be setup again!",
+                ImproperFunctionCall,
+            )
+        else:
+            self.last_update = start_time
+
```
davidorme, line 150: Just make `start_time` part of the `__init__` arguments? No need to modularise this into a method and then have to guard against improper use.

virtual_rainforest/core/model.py
```diff
@@ -135,10 +137,21 @@ def __init_subclass__(cls, model_name: str):
...
     def should_update(self, current_time: datetime64) -> bool:
         """Determines whether a model should be updated for a specific time step."""
 
-        if current_time > self.last_update + self.update_interval:
+        if current_time >= self.last_update + self.update_interval:
             self.last_update = current_time
             return True
         return False
```
davidorme, line 157: I think rather than having a method which has to calculate for each time step, we can have `Model.update()` (I think I prefer that to `solve()` 😄 ) do stuff and then simply have: `self.next_update += self.update_interval`. See the sketch below for how this might work in the timing loop. That seems _much_ cheaper.

virtual_rainforest/main.py
davidorme: Oh - comments in the review show up in posting order, not position in file 😖

jacobcook1995: Yeah this is a problem with using `timedelta64`, as far as I can tell you can't supply it with a float and then units. So, e.g. `timedelta64(0.5, "D")` results in a `TypeError`. I've been tackling this by converting units to minutes and then rounding them to minimise precision loss

virtual_rainforest/main.py
```diff
@@ -80,6 +84,124 @@ def configure_models(
...
+    if fast_models:
+        LOGGER.warning(
+            "The following models have shorter time steps than the main model: %s"
+            % fast_models
+        )
+
+
+def setup_timing_loop(
+    start_time: datetime64, end_time: datetime64, update_interval: timedelta64
+) -> datetime64:
```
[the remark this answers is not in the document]
jacobcook1995: Yeah probably makes sense as a single function, particularly if the warning can go

virtual_rainforest/core/model.py
jacobcook1995: Yeah this is probably sensible

virtual_rainforest/main.py
davidorme: OK - so can only have integer values (but numpy can combine them sanely):

```python
 np.timedelta64(1, 'm') + np.timedelta64(12, 's')
numpy.timedelta64(72,'s')
```

I may be being paranoid here, but we could easily go finer:

```python
In [83]: np.datetime64('2012-06-12 01:00:00') + np.timedelta64(np.timedelta64(32,'m'))
Out[83]: numpy.datetime64('2012-06-12T01:32:00')

In [86]: np.datetime64('2012-06-12 01:00:00') + np.timedelta64(np.timedelta64(32,'m'), 'us')
Out[86]: numpy.datetime64('2012-06-12T01:32:00.000000')
```

Maybe microsecond is going too far 😄 . We can't go finer as the range of representable dates is [too short](https://numpy.org/doc/stable/reference/arrays.datetime.html#datetime-units) 😁

virtual_rainforest/main.py
```diff
@@ -80,6 +84,124 @@ def configure_models(
...
+        start_time: Time the model starts at
+        end_time: Time the model should end at
+        update_interval: Time step of the main model loop
+    """
+
+    # Print number of steps needed to cover the interval
+    n = (end_time - start_time) / update_interval
+    # Convert this to a percentage of time span not covered
+    _, d = divmod(n, 1)
+    p = 100 * d / n
```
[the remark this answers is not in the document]
jacobcook1995: Hadn't come across PEP8 naming stuff before (probably should have though :grimacing:). We could add that option, though it does seem like it mainly deals with underscores and functions naming and wouldn't have caught the problem above. Are single letter names frowned upon in python? (Aside from "l", "I" and "O")

virtual_rainforest/main.py
jacobcook1995: Hahahaha, did consider switching to second precision but decided against. Finest scale I can imagine someone realistically wanting is 20 or 30 mins, which does need minute level precision, but sub minute precision is almost certainly never going to be possible

jacobcook1995: Or possible with meaningful output produced within a reasonable simulation time

virtual_rainforest/main.py
davidorme: Fair enough, but I suspect that cost will be trivial compared to the update() costs! I've just tried running the loop (including list comprehension due date check) using 1 hour intervals and a century took about 10 seconds, so minutes would probably be a bit much (guessing at 10 minutes).

virtual_rainforest/main.py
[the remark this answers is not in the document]
davidorme: Yes - I'm thinking of `pylint`, which has opinions of its own about names:

https://stackoverflow.com/questions/21833872

virtual_rainforest/main.py
davidorme: In [169]: print(end - start)
0:10:31.752243

virtual_rainforest/main.py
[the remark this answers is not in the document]
jacobcook1995: The above code is going anyway, so nothing to change there. Do you think I should add the `pep8-naming` extension?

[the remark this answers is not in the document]
dalonsoa: It depends on how picky you are - and `pylint` is really picky! Things like `l` and `I` are obviously banned, but forbidding `n`, `d` or `t` will drive nuts people in certain fields where those are perfectly valid names for magnitudes and variables. The bottom line is: does the context make it clear what the variables you are defining mean? If it does, then single letter names are fine. If it does not, you need to use something more explicit.

virtual_rainforest/main.py
dalonsoa: About to start my review, but one thing I want to highlight is that, keeping aside if sub-minute resolution is needed or not - which is for you to decide -, the more steps you do, the higher is going to be the loss of precision due to the numerical accuracy of floating point numbers. Not sure if this might become a problem, but I just flag this here so you can consider it.

virtual_rainforest/core/model.py
dalonsoa: I agree. Creating an attribute on the fly, when it is something you can provide directly in `__init__`, does not make much sense.

virtual_rainforest/main.py
dalonsoa: By the way, `int()` truncates, not rounds float numbers. So `int(0.7) = 0`. I guess this might not be what you want. Maybe `int(round(0.7)) = 1` will be best.

virtual_rainforest/main.py
dalonsoa: I like very much David's alternative. It should definitely be faster than copying models around. It can also be simplified slightly:

```python 
# Get the list of date times of the next update.
update_due = {mod.name: mod.next_update for mod in models_cfd}

# Loop until the end
while current_time < end_time:
    # Get the names of models that have expired due dates
    update_needed = [nm for nm, upd in update_due.items() if upd <= current_time ]

    # If there are expired model, run their update() method and update their update due dates
    for mod_nm in update_needed:
        models_cfg[mod_nm].update()
        update_due[mod_nm] = models_cfg[mod_nm].next_update()
```

dalonsoa (changes requested):
A couple of minor comments. David's done most of the hard work.

virtual_rainforest/main.py
davidorme: Oooh, yes. Entirely redundant `if` statement 😞

virtual_rainforest/main.py
```diff
@@ -109,18 +192,39 @@ def vr_run(
         "All models successfully configured, now attempting to initialise them."
     )
 
-    # This is just a step to pass flake8 checks (DELETE LATER)
-    print(models_cfd)
+    # Define the basic model time grain (set to 10 minutes for now)
+    update_interval = timedelta64(10, "m")
```
dalonsoa, line 196: Is this something you want to hardcode rather than reading it from the configuration?

dalonsoa (approved):
Looks great! I've just have a very minor question.

virtual_rainforest/main.py
jacobcook1995: Yeah my worry was with users setting an update interval that wasn't small enough to capture the timing of models well, i.e. setting it to be 3 weeks when one model updates every three weeks while another updates monthly would generate undesirable behaviour.

Can definitely see the value in this being configurable though, I'll probably add it to the config schema with the default set to 10 minutes, and the description for the user warning them to set it significantly lower than the model time step

davidorme (approved):
LGTM - the only ting I'd have picked up is the hardcoded step interval and that's fixed now!

PR #127, opened 2022-12-05
Command line documentation

### LINKED ISSUE #108: Basic command line documentation
v0.1 will introduce a command line entry point to run the virtual rainforest model. This will need to be documented somewhere.

We need to decide on the best way to present this, and then go ahead and document it

jacobcook1995:
# Description

Added a page to the docs detailing how the `vr_run` command line tool should be used. Also added a bash script `update_command_line_usage.sh` which extracts information from the --help option of the command line tool, and writes it to a text file which is then included in the docs. This avoids the need to manually update the docs every time the command line usage information is changed. At the moment this bash script has to be run manually to update the files, but in future we could do something cleverer with it.

This pull request also fixes a fair few typos and formatting problems I noticed in the docs. So, it might be good to build docs for this branch for the writing workshop on Tuesday.

Fixes #108 

## Type of change

## Key checklist

## Further checks


docs/source/command_line_tools/vr_run.md
```diff
@@ -0,0 +1,25 @@
...
+this consists of validated a set of schema files, using their contents to initialise a
+set of models and then setting up the main simulation loop. It goes no further than
+that, and no actual simulation steps are performed. This will change in the future.
+
+### Using `vr_run`
+
+The usage instructions for `vr_run` are given below:
+
+```{include} command_line_usage/vr_run.txt
+```
```
alexdewar, line 19: I don't think this is needed. Users can just run `vr_run --help` themselves and see the output. You can tell them to run this for more info if you like, but I don't think you need to go to the extent of actually including the output in the documentation.

While we're on the topic, what happens if you run `vr_run` without arguments? Or with bad arguments? In those cases it would also be worth pointing users towards `vr_run --help`.

docs/source/command_line_tools/vr_run.md
```diff
@@ -0,0 +1,25 @@
...
+
+Essentially, you can run the Virtual Rainforest model by calling:
+
+```bash
+vr_run path/to/config/file.toml path/to/second/config/file.toml
+```
```
alexdewar, line 25: This bit is all the documentation you really need. Users need to have `vr_run` installed anyway before they can do anything.

docs/source/development/design/core.md
```diff
@@ -144,12 +144,12 @@ values directly to avoid having to create NetCDF files with trivial variables.
 
 ```toml
 [core.data.precipitation]
-file = /path/to/netcdf.nc
-var = prec
+file = "/path/to/netcdf.nc"
+var = "prec"
```
alexdewar, line 148: Is it not valid toml if you don't have the quotes or something?

docs/source/refs.bib
```diff
@@ -1,214 +1,183 @@
 
 @article{Wang:2017go,
```
alexdewar, line 2: What tool did you use for reformatting this btw?

docs/source/virtual_rainforest/soil/soil_details.md
```diff
@@ -50,7 +50,7 @@ Another mechanism by which carbon is protected from microbial activity is minera
 association, whereby mineral surfaces take up organic matter by adsorption, conferring
 chemical protection. It is generally microbial biomass and LMWC that gets absorbed in
 this way. Aggregates can form from MAOM, which they then release upon breakdown. When
-desorption occurs carbon is released from mineral associated in the form of LMWC.
+desorption occurs carbon is released from mineral association in the form of LMWC.
```
alexdewar, line 53: Is "mineral assocation" a domain-specific term?

poetry.lock
```diff
@@ -16,7 +16,7 @@ python-versions = "*"
```
alexdewar, line 16: Is there any particular reason you updated the dependencies on this branch? It's fine but it doesn't really fit with the theme of documentation tidy-ups.

In general, I wouldn't explicitly run `poetry update` on a branch unless you really need to. If you're just bumping dependencies for the sake of it, let dependabot bump them instead.

docs/source/command_line_tools/command_line_usage/update_command_line_usage.sh
```diff
@@ -0,0 +1,16 @@
+# This script echoes the command line usage for script tools to files in this directory
```
alexdewar, line 1: I don't think you should include this (see below) but in future, I'd add a shebang to scripts and make them executable.

alexdewar (changes requested):
I've made some suggestions but otherwise all looks good.

docs/source/virtual_rainforest/soil/soil_details.md
jacobcook1995: Yeah, I think it is. In soil science it refers to carbon bound to soil minerals which is therefore not accessible for use by soil microbes.

Just for clarity I added a bunch of small doc fixes while I was working on the command line docs (not sure if that is best practice?), so this bit forms part of the soil module specific docs, rather than the command line docs.

docs/source/refs.bib
jacobcook1995: I didn't use a specific tool to reformat, just whatever the built in linter for VSCode is. I also manually deleted a lot of the categories that we would never want to display (e.g. abstracts) and the stuff that doesn't relate to the reference itself (e.g. paths to local storage). Is there a more sensible way I should have done this?

poetry.lock
jacobcook1995: I added a dependancy (`markdown_include`) because I wanted to try something out and committed the change. Then down the line I decided against using it and removed it. Because I had committed the change I couldn't figure out how to return the poetry lock to its original state. Is there a better way to handle situations like this?

docs/source/development/design/core.md
jacobcook1995: Yeah any string in `toml` is supposed to be in quotes, and `sphinx` won't format this block as `toml` for the docs when this rule is broken. Looking at it again though I do think that `file` should be a literal string (') rather than a basic string (")

docs/source/virtual_rainforest/soil/soil_details.md
alexdewar: Ok cool. That's what I thought, but I just wanted to check that it made sense to someone!

I think it's fine to mix in various fixes on the same PR, although generally you want a theme -- and "misc doc fixes" is fine as a theme. I would usually do them as separate commits though for revertability.

poetry.lock
alexdewar: You basically have two options:

1. `git revert [commit hash]` to undo the change and make a new commit
2. Rewrite the history in some way to drop that commit altogether (e.g. interactive rebase: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

They have slightly different use cases, but I think either would be fine here. `git revert` is generally less of a faff.

davidorme: I'd definitely do `git revert` here. Rewriting the history just to avoid a commit here is pain for little obvious gain 😄

docs/source/development/design/core.md
alexdewar: Sounds good. Just checking!

docs/source/refs.bib
alexdewar: Not really. I just wondered what you'd change and whether you had a useful tool I hadn't heard of.

I've found JabRef is really handy for working with bibtex files: jabref.org

It does have some built-in tools for fixing broken tags etc.

docs/source/command_line_tools/vr_run.md
jacobcook1995: Good point about running `vr_run` without arguments (or with bad arguments) at the moment that just raises a critical error without telling the user where the problem can be resolved. Which now I think about isn't very useful for anyone who doesn't already know how the package works

poetry.lock
alexdewar: Yeah you're probably right :smile:

docs/source/command_line_tools/vr_run.md
alexdewar: Ok cool. I think it's totally fine to leave an error message in there btw, but I'd just print another line saying "For more information please run vr_run --help" (or something)

jacobcook1995:
Okay so started randomly getting tests failing on `macosx` for python `3.9` and `3.10`, due to `types-jsonschema`. This hasn't just happened on this branch but seems to be a problem with every branch simultaneously (or at the very least also impacts #131).

Not sure why this happened but I managed to fix it by pinning `python 3.9` and `3.10` to their newest versions (i.e. `3.9.16` and `3.10.16`). This fixes the problem but `test (windows-latest, 3.9.16)` fails as `3.9.16` apparently is yet available for `x64` architecture. I feel like this is an acceptable test to fail though

docs/source/command_line_tools/vr_run.md
jacobcook1995: One of the benefits of including the command output here is that it gives us somewhere in the docs to check that the command line output is formatted properly, which is hard to work out from the docstring alone. With normal functions their docstrings are displayed in the api, but as far as I'm aware it doesn't make sense to add in command line entry points there?

davidorme: I actually disagree relatively strongly. It does no real _harm_ here and it does provide a prospective user with the ability to read and learn about the interface without _having_ to install it. If I was teaching a student to use this tool, having a page where they could review the main interfaces would be a useful step. This is partly a _tutorial_ page, so hand-holding is completely in its place here.

alexdewar (approved):
Looks good!

One thing I would say though is that I personally wouldn't merge it with a test failing. I'd suggest removing the test from your GitHub actions and opening an issue saying that e.g. pytest is currently failing for Windows (or whatever the problem is). You can then later try fixing it on another branch when the time is right and revert the change.

davidorme:
Looks like there is an exclude option there?

```toml
  exclude:
    - os: windows-latest
      python-version: 3.9.16 # Broken
```

docs/source/command_line_tools/vr_run.md
alexdewar: Maybe. If you want to explain what the different options mean here, then I think that's fine. I mostly just object to having a script to dump the program output into your documentation for the purposes of checking that it's formatted correctly!

Up to you though.

jacobcook1995:
Hmmmm, it will not let me merge as "Required statuses must pass before merging". The required statuses seem to be "test (macos-latest, 3.9)" and "test (ubuntu-latest, 3.9)" which seem to have being hanging around for a while without running. Any idea how to force these to run?

alexdewar:
That's odd. It looks like GitHub is running all the tests you specified :thinking:. Maybe it's complaining because there are a couple of tests that you're now no longer running.

I'd be inclined to override the requirements and merge it anyway. It might well just start working. If it breaks things you can open an issue and fix it elsewhere.

jacobcook1995:
As far as I can tell I do not

davidorme:
OK -  I will merge now.

PR #130, opened 2022-12-07
first draft radiation.py

vgro:
# Description

This is a first (very conceptual) draft of a radiation sub module. The file contains the main processes I want to consider and dummy functions with doc strings (sometimes a bit of code). This is very much a starting point for discussion.

I would appreciate feedback on the overall structure, if I should break up functions into more smaller (better testable) units, if it makes sense to add results as attributes of the class or should they be returned as variables?
What would be efficient ways of running these functions over a numpy array (for each of our grid cells individually, as they are not linked this could happen in parallel)?

## Type of change

## Key checklist

## Further checks

- [NA] Tests added that prove fix is effective or that feature works


virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+# the following structural components are not implemented yet
+TODO core.constants
+TODO include vertical structure (canopy layers)
+TODO include data structure (->numpy array)
+TODO include time dimension
+TODO logging, raise errors
+"""
+
+# import numpy as np
+# from core.constants import kc, kd, kfFEC, bolz # this doesn't exist yet
```
davidorme, line 31: I think this might operate as `from core.constants import CONSTANTS` and then use `CONSTANTS.k` or `CONSTANTS['k']`in code?

Or `from core.constants import CONSTANTS as C` and then `C.k` if you wanted more brevity 😄

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        ppfd: float, photosynth. photon flux density, top of canopy [mol m-2]
+        rn_profile: float, net radiation profile, below canopy [W m-2]
+    """
+
+    def __init__(self) -> None:
+        """Initializes point-based radiation method."""
+        pass
+
+    def calc_sw_down(self, n: int, y: int, lat: float) -> float:
+        """Calculate shortwave downward radiation.
```
davidorme, line 52: Do we need to worry about spatial variation in this? So mapping different points within the simulation grid? Probably not to begin with, but this could potentially be written to handle an array of `lat` values?

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        # check if shortwave down is provided, else calculate
+        if sw_in:
+            sw_down = sw_in
+        else:
+            sw_down = self.calc_sw_down(n, y, lat)
+
+        # atmospheric filtering (clouds, aerosols)
+        tau_o = kc + kd * sf
+        tau = tau_o * (1.0 + (2.67e-5) * elev)
```
davidorme, line 110: Should the `2.67e-5` be part of the constants?

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        """Calculates daily net radiation profile within the canopy.
+
+        The output is a vector of net radiation, the first entry is the top
+        of the canopy, the last entry is the net radiation that reaches the surface.
+
+        Args:
+            n_level: number of canopy layers
+            canopy_abs: absorption by canopy (from plant module) [W m-2]
+            emis: vector with canopy and soil emmissivity
+            tc: vector of canopy temperatures
```
davidorme, line 134: `n_level` is not in the function signature. Should this handle multiple cells or not (or not _now_). We should be using `numpy` types (or possibly `xarray` types) for vectors

davidorme (changes requested):
I think the structure looks good - I think this probably should operate at a whole simulation level and so the class `__init__` should work for the whole set of cell ids? The individual methods might then work on a single cell (and hence the full `radiation_balance` method might use a loop) or it might be possible to vectorise the methods across all cells.

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        sw_down_f = sw_down * tau
+
+        # topography and albedo effects on shortwave radiation
+        sw_net_toc = (1 - sw_albedo) * sw_down_f * topo_factor
+        self.sw_net_toc = sw_net_toc
+
+        # PPFD
+        # TODO integrate over duration of the day?
+        ppfd = (1.0e-6) * kfFEC * (1.0 - vis_albedo) * sw_down_f
+        self.ppfd = ppfd
```
dalonsoa, line 120:
```suggestion
        self.ppfd = (1.0e-6) * kfFEC * (1.0 - vis_albedo) * sw_down_f
```

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        self,
+        sw_in: float,
+        n: int,
+        y: int,
+        lat: float,  # this is optional for the case of sw_in == FALSE
+        elev: float,
+        sf: float,
+        topo_factor: float,
+        sw_albedo: float,
+        vis_albedo: float,
```
dalonsoa, line 79: Some of these variables could have a more descriptive name (`shortwave_in`, `sunshine_hours`, `day`, `year`, etc.) . It does not save much space or time to shorten them and will make the code more explicit.

virtual_rainforest/abiotic/radiation.py
dalonsoa: Agree. If the inputs are numpy arrays (or xarray.DataArrays) the type annotations should reflect that. The same in `radiation_balance` below. So, something like:

```python
...
    def calc_canopy_radiation(
         self, canopy_abs: float, tc: NDArray[float], ts: float, emis: NDArray[float]
     ) -> None:
```

Also, probably you can get `n_level` (should it be `n_layers`?) from the length of `tc`, right?

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        # Calculate net radiation cross-over hour angle (hn), degrees
+        # Calculate daytime net radiation (rn_d), J/m^2
+        # Calculate nighttime net radiation (rnn_d), J/m^2
+
+        # calculate vertical canopy shortwave radiation profile
+        sw_profile = self.sw_net_toc - canopy_abs
+
+        # calculate vertical canopy net radiation profile
+        rn_profile = sw_profile - lw_profile
+        self.rn_profile = rn_profile
```
dalonsoa, line 157:
```suggestion
        self.rn_profile = sw_profile - lw_profile
```

dalonsoa (changes requested):
A good starting point. Just a few suggestions of things to improve.

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+
+        Args:
+            n: day of the year
+            y: year
+            lat: latitude
+
+        Ref:
+            Davis et al. (2017): Simple process-led algorithms for simulating habitats
+            (SPLASH v.1.0): robust indices of radiation, evapotranspiration and plant-
+            available moisture, Geosci. Model Dev., 10, 689-708
```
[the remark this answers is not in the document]
vgro: The latter. As I mentioned above, we might move away from this functionality for now but it is a good point, I assume if I used the package I wouldn't include so much detail as reference.

virtual_rainforest/abiotic/radiation.py
vgro: There are multiple dimensions, one is the cell_id and one is the vertical structure in each cell (n_layers) which can be derived from the length of tc, very good point, as it will change with the vegetation dynamics.
So far I have only thought about the types of the 'smallest unit' without taking space or time into account, I guess most inputs will be arrays, I will change that in the next iteration.

alexdewar (changes requested):
Looking good! I've made some minor suggestions throughout.

In terms of bigger-picture stuff, I would also say:

- I'd add another subfolder, so that this file is in `models/abiotic/radiation.py` to make it clearer when other models are added
- It seems like many of your functions have some of the same arguments. Assuming that the argument won't change between calls, that's usually a good indication that you should make it an argument to `__init__` and store it there so it can be reused in the other member functions (e.g. `self.lat = lat`)
- I'd raise `NotImplementedError`s everywhere where functions are unfinished to make it more obvious where bits are missing

Re your question about making the code numpy friendly, at the moment the main thing you need to change is arguments to various functions: where you want them to be vectors they should be of type `NDArray[float]` (or similar). When it comes to implementing the algorithm, you often don't have to make substantial changes to code for numpy arrays, e.g. you can use the `+` operator to do element-wise addition of two vectors, just as you can do with scalars. For common functions like `sin`, numpy provides its own which work just like the ones in the standard library but also work with vectors.

Parallelisation is something that I think should be left for a later stage for a couple of reasons:

1. Parallel code always ends up being more complicated (even if it's not by much)
2. It's also harder to test because the order in which things are done can vary between runs

It's also possible that your module might not actually need to deal with parallelisation at all directly, if it's something done by the core.

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+TODO include time dimension
+TODO logging, raise errors
+"""
+
+# import numpy as np
+# from core.constants import kc, kd, kfFEC, bolz # this doesn't exist yet
+kc = 1
+kd = 1
+kfFEC = 1
+bolz = 1
```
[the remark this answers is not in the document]
vgro: Thanks, that is is a good suggestions. Some of these constants will be so universal that they are part of scipy, others aren't. For now I leave the (real :-) )values here, once we start core.constants.py we should make a collective decision on that.

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        topofcanopy_radiation: float, shortwave radiation, top of canopy [W m-2]
+        ppfd: float, photosynth. photon flux density, top of canopy [mol m-2]
+        rn_profile: float, net radiation profile, below canopy [W m-2]
+    """
+
+    def __init__(self) -> None:
+        """Initializes point-based radiation method."""
+        pass
+
+    def calc_sw_down(self, n: int, y: int, lat: float) -> float:
```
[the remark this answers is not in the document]
vgro: Thanks for this comment, I'll adjust the variables names to make more sense

virtual_rainforest/abiotic/radiation.py
vgro: I will adjust as soon as we have the file in place :-)

virtual_rainforest/abiotic/radiation.py
vgro: This is not a universal constant, the equation is based on the regression of Beer’s radiation extinction function at elevations below 3000 m, so I think it could stay like thi,s maybe best to add the reference?

virtual_rainforest/abiotic/radiation.py
vgro: I have a problem with NDArray[float] , it says: 'Could not specialize type "NDArray[ScalarType@NDArray]"
  Type "float" cannot be assigned to type "generic" "float" is incompatible with "generic"'. Suggestions?

vgro: 'emis' contains only two constant values, I added it to the list of constants at the top

virtual_rainforest/abiotic/radiation.py
davidorme: Yes - but then if someone publishes a new version, you don't want to have to change the hard-coding and people might want to try different versions. If this is is coded as part of a `constants` library then it can be changed flexibly.

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+        sf: float,
+        topo_factor: float,
+        sw_albedo: float,
+        vis_albedo: float,
+        canopy_abs: float,
+        tc: float,
+        ts: float,
+        emis: list,
+    ) -> None:
+        """Calculate radiation balance."""
```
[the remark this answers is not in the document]
vgro: There are a few functions that don't return variables but only attributes of the class, would i still indicate that the return value is a float?

virtual_rainforest/abiotic/radiation.py
davidorme: There are likely better ways to do this, but `pyrealm` has basically no hard-coded parameters in it at all and has sets of lookup tables to store things like this:
https://github.com/davidorme/pyrealm/blob/develop/pyrealm/param_classes.py

alexdewar: @vgro Even if you don't make this a global constant, you could make it a local one. I'd personally make it caps to show it's a constant and then just define it at the start of the function, maybe with a docstring saying what it is:

```py
BEER_REGRESSION = 2.67e-5
"""Parameter which does something that Alex doesn't understand."""
```

virtual_rainforest/abiotic/radiation.py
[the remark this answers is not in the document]
alexdewar: Yep. If the attribute you're returning is a `float` then your function will also be `float`.

virtual_rainforest/abiotic/radiation.py
davidorme: Some bigger picture considerations:

1. Over in `pyrealm` I want to be able to have parallel instances of a PModel that might differ in these 'constant' parameters. That is the _main_ reason they are bundled up in classes: each PModel instance can have a separate isolated instance of the parameters. If they were declared as global constants, you could still change them but it would get extremely error prone. I don't think we want that here - but it is a big part of why that example is structured as it is.

2. But we do have an existing similar system: this  _could_ be moved into `config`? Part of the point of that system is to fill in defaults when users don't want to tweak them, so it could be declared there.

virtual_rainforest/abiotic/radiation.py
davidorme: Or `NDArray[floating]`?

vgro: Perfect, that seems to work. That's also what I concluded from my internet search but wasn't sure how that was compatible with what we have set up.

virtual_rainforest/abiotic/radiation.py
vgro: @davidorme I agree that we might want to have different values for certain constants but I don't think this is one of them.

virtual_rainforest/abiotic/radiation.py
davidorme: The parallel issue I've run into is when a value could be a scalar or an array: `Union[float, np.ndarray]` . That causes `mypy` irritation down the line because you constantly have to trap where you can use e.g. `var[idx]` . I think another problem here is that 0-dimension np.arrays then get treated as scalars: `x = np.array(6.0)` ?

Is the best way to handle this to aggressively cast scalar types to one-dimensional np.arrays?

virtual_rainforest/abiotic/radiation.py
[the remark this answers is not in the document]
vgro: And how would I indicate this if there are multiple outputs? As a list?

virtual_rainforest/abiotic/radiation.py
davidorme: I've certainly taken an extreme view over in `pyrealm` 😄 You could set it up to run on Mars or a parallel universe...

virtual_rainforest/abiotic/radiation.py
davidorme: Actually - maybe I'm looking for `npt.ArrayLike`, which should handle this?

virtual_rainforest/abiotic/radiation.py
alexdewar: > @alexdewar good pont, I'll do that. is it a general convention to have constants in caps?

Yep, though not everyone does it. I like that it makes them stand out when you're skimming the code.

virtual_rainforest/abiotic/radiation.py
alexdewar: > The parallel issue I've run into is when a value could be a scalar or an array: `Union[float, np.ndarray]` . That causes `mypy` irritation down the line because you constantly have to trap where you can use e.g. `var[idx]` . I think another problem here is that 0-dimension np.arrays then get treated as scalars: `x = np.array(6.0)` ?
> 
> Is the best way to handle this to aggressively cast scalar types to one-dimensional np.arrays?

I think so, yes. Either that or just don't handle the scalar case and make the user convert it themselves. In general I imagine you won't be passing around scalar values a whole lot anyway, so I think that would be fine. Don't make your life harder than it needs to be.

I think `ArrayLike` is for other types that are like numpy arrays in certain respects, mainly in that you can convert them to a numpy array, so you could pass a tuple or a list into your function instead. I don't know though.

vgro: Is this a n issue if I want to run a function for all cell_ids but not necessarily in parallel?

davidorme: @vgro Do you mean you have a function that can be applied using a vectorised calculation across all `cell_id`s? That isn't technically 'in parallel' as it is a single calculation but just applied to array inputs. Or that you don't want to do that and need to carry out calculations on a cell-by-cell basis?

vgro: Could you rephrase your questions please, I'm not sure if I understand the terminology here. I want to do a single calculation but applied to array inputs. What type of processes would you run in parallel?

davidorme: By parallel, I meant code where you run each cell in a different process (like if you run a bunch of jobs on different cores on the HPC). If you were looping over cells, you could either do each in series using a `for` loop or farm out different cells to different processors and speed it up. As @alexdewar noted elsewhere here, that is something to think about later.

I've used 'vectorised' to mean applying the same calculation across all of the elements within arrays of inputs, which is something `numpy` is brilliant at, and is what you're describing here.

davidorme: That's useful to know, @dalonsoa . In my case, I was trying to support situations where someone might be working with a single site as opposed to calculating the same things across lots of locations. It would be better there to stipulate that users need to provide single sites using np.arrays, even if 0-dimensional.

davidorme: @vgro A good example here would be calculating pressure from elevation (assuming no _weather_ :smile:). If you had a `numpy` array of cell elevations from the data store, you could just do something like:

```python
data['air_pressure'] = 101325 *  (1 - 0.0000225577 * data['elevation']) ** 5.25588    
```

And that is all the cells done in one step

virtual_rainforest/abiotic/radiation.py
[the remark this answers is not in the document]
alexdewar: It depends what you mean by multiple outputs.

You can return multiple values as a tuple, e.g.:

```py
return 3.14, False
```

In that case the return value is a tuple, or specifically a `Tuple[float, bool]`.

If you're returning a numpy array of floats, again it's `NDArray[np.float32]` and if it's e.g. a list of strings you can describe it as a `list[str]` etc.

mypy's usually quite helpful with this. If you accidentally pick the wrong return type for your function then it should tell you what the right one is in the error message.

virtual_rainforest/abiotic/radiation.py
vgro: > I've certainly taken an extreme view over in `pyrealm` 😄 You could set it up to run on Mars or a parallel universe...

I agree it is good to have these sort of parameters in one place. I just had a closer look into the pyrealm code and it makes a lot of sense to me to structure it that way. I'll keep it as it is for now and we can merge it once we have the location for it :-)

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    def __init__(self) -> None:
+        """Initializes point-based radiation method."""
+        pass
+
+    def calc_sw_down(self, n: int, y: int, lat: float) -> float:
+        """Calculate shortwave downward radiation.
+
+        This function calculated incoming shortwave radiation based on location and time
```
vgro, line 54: I set the return type to NDArray but got a 'missing return' warning from mypy, I guess because I dont actually return anything other than adding/updating an attribute of the class. Do I need a different approach here?

virtual_rainforest/abiotic/radiation.py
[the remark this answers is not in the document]
vgro: When I set the return type to NDArray for a function that only adds to self, I got a 'missing return statement' warning from mypy

virtual_rainforest/abiotic/radiation.py
```diff
@@ -0,0 +1,231 @@
...
+                canopy layers [W m-2]
+            longwave_soil: NDArray[np.float32], longwave radiation from soil [W m-2]
+            self.longwave_out: NDArray[np.float32], total longwave radiation [W m-2]
+        """
+        raise NotImplementedError("Implementation of this feature is missing")
+
+        # longwave emission canopy
+        # TODO define the longwave_canopy array to match grid+vertical structure
+        # which is not uniform (different number of layers in each grid cell)
+        longwave_canopy = np.full_like(canopy_temperature, 0)
```
vgro, line 141: I already know that this needs changing, need to dive into core docs, so please skip

davidorme: I think this is going to be _much_ easier if we assert a configurable maximum number of canopy layers. That would give us a `canopy_temperature` array that would have the shape (n_cells, n_canopy_layers) and we could then do:

```python
longwave_canopy = (CANOPY_EMISSIVITY * BOLZMAN_CONSTANT * canopy_temperature ** 4)
```

There are three issues:

 - the memory use is higher
 - we need to keep track of which of those layers are "above" the canopy
 - we either have to abort the simulation if we exceed that number of layers (which might be good if the number of layers is not sane) or alternatively have some kind of mechanism for extending that maximum and adding layers to the canopy as needed.

The alternative here is dealing with ragged arrays, which neither `numpy` nor `xarray` handle elegantly. There are some really neat tricks for speeding up calculations on ragged arrays but it seems best to use memory rather than needing that complexity. @alexdewar @dalonsoa ?

davidorme: The canopy model is then basically going to populate an array `canopy_layer_heights` with shape (n_cells, n_canopy_layers, n_timesteps) and for any one timestep you might have in an example with two cells and five canopy layers:

```python
In [22]: np.array([[16.1, 19.6, 23.1, np.nan, np.nan],[10.1, 15.9, 20.1, 23.5, np.nan]])
Out[22]: 
array([[16.1, 19.6, 23.1,  nan,  nan],
       [10.1, 15.9, 20.1, 23.5,  nan]])
```

alexdewar:
> > * I'd add another subfolder, so that this file is in `models/abiotic/radiation.py` to make it clearer when other models are added
> 
> @alexdewar - so that basically we have namespace where (at the moment anyway) everything is `vr.core.xxx` or `vr.models.abiotic.yyy`, `vr.models.tree_eating_alien_invasion.zzz`? I can see that `core` might also get subdivisions at some point but that seems less pressing.

Yeah, that's what I was thinking.

virtual_rainforest/abiotic/radiation.py
vgro: Thanks David, I'll consider that as we discussed in person today

tests/test_radiation.py
```diff
@@ -0,0 +1,98 @@
...
+
+def test_calc_ppfd():
+    """Simple check for correct numbers, better test to be decided."""
+    from virtual_rainforest.models.abiotic import radiation
+
+    dummy = radiation.Radiation(100)
+    tau = dummy.calc_ppfd(29376000, 1.0)
+    assert tau == pytest.approx(0.752, 0.1)
+    assert dummy.ppfd == pytest.approx(43.713, 0.1)
+
```
davidorme, line 44: A couple of options here on structuring tests. The first is using `pytest.fixtures` - these are re-usable bits of code you can share between tests. So you could do simplify all the tests to do this:

```python
@pytest.fixture
def radiation_fixture():
    """Create a reusable radiation fixture"""
    from virtual_rainforest.models.abiotic.radiation import Radiation

    return Radiation(100)

def test_calc_ppfd(radiation_fixture):
    """Simple check for correct numbers, better test to be decided."""
    tau = radiation_fixture.calc_ppfd(29376000, 1.0)
    assert tau == pytest.approx(0.752, 0.1)
    assert radiation_fixture.ppfd == pytest.approx(43.713, 0.1)
```

The other thing is that you're using scalars as your inputs to methods that support arrays and it is worth testing that array inputs work as expected. You could simply replace the test values, but I would be tempted to use parameterisation:

```python
@pytest.mark.parameterise(
    argnames="input,exp_tau,exp_ppfd",
    argvalues=[
        (29376000, 0.752, 43.713),
        (
            np.array([29376000, 29376000]),
            np.array([0.752, 0.752]),
            np.array([43.713, 43.713]),
        ),
    ],
)
def test_calc_ppfd(radiation_fixture, input, exp_tau, exp_ppfd):
    """Simple check for correct numbers, better test to be decided."""
    tau = radiation_fixture.calc_ppfd(input, 1.0)
    assert tau == pytest.approx(exp_tau, 0.1)
    assert radiation_fixture.ppfd == pytest.approx(exp_ppfd, 0.1)
```

vgro: Thanks for this suggestion, I included something like this for all the tests :-)

davidorme:
@vgro Could you update `feature/abiotic_dummy` to be in line with `develop`? I think from @alexdewar's suggestion, that should make the changes on this branch easier to see. I think that is just:

```sh
git checkout feature/abiotic_dummy
git merge develop
git push
```

At that point, a lot of the changes in this PR should drop away, because they are things this branch already has from `develop` but `feature/abiotic_dummy` does not - it is a bit stale.

alexdewar:
> 

Yes please. You may need to check out `develop` and do a `git pull` first to make sure it's up to date before you merge it into `feature/abiotic_dummy`.

vgro:
> @vgro Could you update `feature/abiotic_dummy` to be in line with `develop`? I think from @alexdewar's suggestion, that should make the changes on this branch easier to see. I think that is just:
> 
> ```shell
> git checkout feature/abiotic_dummy
> git merge develop
> git push
> ```
> 
> At that point, a lot of the changes in this PR should drop away, because they are things this branch already has from `develop` but `feature/abiotic_dummy` does not - it is a bit stale.

I've done that but I;m not sure if it worked as expected, there are still all the files that were changed in different context...

vgro:
1. Is there a way to highlight the files I would like to have reviewed?
2. I will update some minor suggestions that were made for the energy_balance and change this PR to draft in the meantime, so no rush to review if you haven't started.

vgro:
Is there a way of highlighting what files need to be checked?

alexdewar:
@vgro I don't think so, unfortunately.

It looks like you've merged the branches as I suggested, but GitHub is still insisting that other files have been changed. It's not a biggie -- I can see which files are yours -- it's just something to bear in mind for future PRs. When you start work on a new feature, you should make sure you're basing it on the correct branch.

davidorme:
@alexdewar @vgro I think what might have happened here is that `feature/radiation_dummy` _is_ a branch of of `feature/abiotic_dummy` but that at some point `feature/abiotic_dummy` has had the new stuff in `develop` merged into it, bypassing `feature/radiation_dummy`. 

If we are going to use PRs into feature branchs to build up systems, then I guess you could get updates into the branches but you'd have to merge `develop` into the main feature branch and then merge the main feature branch into the subfeature? Might be useful to consistently use `feature` and `subfeature` here?

Does that sound right, @alexdewar?

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,226 @@
...
+                the surface (=forest floor)[J m-2]
+        """
+
+        tau = self.calc_ppfd(shortwave_in, sunshine_hours, albedo_vis)
+        self.calc_topofcanopy_radiation(
+            tau, shortwave_in, sunshine_hours, albedo_shortwave
+        )
+        self.calc_longwave_radiation(canopy_temperature, surface_temperature)
+        self.calc_netradiation_surface(canopy_absorption)
+        del tau
```
davidorme, line 226: I'm not sure if the `del` is needed here? I don't think it is wrong, but I think `tau` will just be garbage collected when the method returns.

tests/test_radiation.py
```diff
@@ -0,0 +1,209 @@
...
+    assert TRANSMISSIVITY_COEFFICIENT == 0.50
+    assert FLUX_TO_ENERGY == 2.04
+    assert BOLZMAN_CONSTANT == 5.67e-8
+    assert SOIL_EMISSIVITY == 0.95
+    assert CANOPY_EMISSIVITY == 0.95
+    assert BEER_REGRESSION == 2.67e-5
+    assert ALBEDO_VIS == np.array(0.03, dtype=np.float32)
+    assert ALBEDO_SHORTWAVE == np.array(0.17, dtype=np.float32)
+    assert CELSIUS_TO_KELVIN == 273.15
+    assert SECOND_TO_DAY == 86400
```
davidorme, line 34: I'm not sure this is helpful. We would have to update it if the values changed and then you need to change a value in two places (module and here). Essentially, this is primarily a test that you typed the same number correctly twice.

davidorme (changes requested):
This looks good. I have focussed on `radiation.py` and `test_radiation.py`  as I think those are the core files of this PR.

@alexdewar and @dalonsoa: I have made some claims about:
* defining all attributes in __init__
* use of Returns
* use of `del`

I think I'm right but please do review my review!

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,226 @@
...
+            sunshine_hours: NDArray[np.float32], fraction of sunshine hours, between 0
+                (100% cloud cover) and 1 (cloud free sky)
+            albedo_shortwave: NDArray[np.float32], shortwave albedo, default = 0.17
+
+        Returns:
+            self.topofcanopy_radiation: NDArray[np.float32], top of canopy downward
+                shortwave radiation, [J m-2]
+        """
+        self.topofcanopy_radiation = (1.0 - albedo_shortwave) * tau * shortwave_in
+        """Daily top of canopy downward shortwave radiation, [J m-2]"""
```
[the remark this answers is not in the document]
davidorme: And for other attributes in other functions.

tests/test_radiation.py
davidorme: And why are two of the constants defined as `np.array`?

vgro: The idea was to test that all constants were imported correctly.
Albedo is an array here because it is the default if there is no array for albedo provided as an input (albedo will be a function of land cover in the future).

davidorme: That makes sense, but I think the thing to test here is that the required constants _can_ be imported rather than their values. With the current test, it is possible that a constant might be mistyped in the code, but equally could be mistyped here. It's really a test of duplicated typing rather than the code. And - because the import is outside of the test - if one is missing, the test file itself will fail with an AttributeError. 

Does this do what you want? It checks that each constant is provided.

```python

@pytest.mark.parametrize(
    argnames=["cname"],
    argvalues=[
        ("ALBEDO_SHORTWAVE",),
        ("ALBEDO_VIS",),
        ("BEER_REGRESSION",),
        ("BOLZMAN_CONSTANT",),
        ("CANOPY_EMISSIVITY",),
        ("CELSIUS_TO_KELVIN",),
        ("CLOUDY_TRANSMISSIVITY",),
        ("FLUX_TO_ENERGY",),
        ("SECOND_TO_DAY",),
        ("SOIL_EMISSIVITY",),
        ("TRANSMISSIVITY_COEFFICIENT",),
    ],
)
def test_import_constants2(cname):
    """Test constants can be imported."""

    # Get the module that should contain the constants
    import virtual_rainforest.models.abiotic.radiation as rad
   
    # Check that the constant can be retrieved from the module without
    # raising an AttributeError.
    with does_not_raise():
        assert getattr(rad, cname)
```

davidorme: I think defaults for an argument should probably be in the function signature rather than the constants? That is a grey area though - my `pyrealm` package has a whole bunch of 'constants' tucked away in a general settings container which are really just experimentally derived values. I think the distinction is that I'd expect someone to use a constant set of those values across multiple models (unless they are explicitly testing sensitivity). But something like albedo might very well change across multiple models of  study site?

vgro: It should be quite straight forward to use different albedo values, it does not need to be the default. I could just stick the default value in the function instead.

vgro:
Close this branch after restructuring and cleaning, new PR #155

PR #133, opened 2022-12-13
Update to axis validation in data system

davidorme:
# Description

This PR updates the handling of axis validation when adding data arrays to a `core.data.Data` instance. The main changes are:

* A data array no longer _has_ to match a validator for every registered axis. For example, elevation data would be expected to match the `spatial` axis but not a `temporal` axis.
* However, it _is_ an error to use any of the dimension names associated with a registered axis and not match a validator. For example, you can't arbitrarily use `x` as a dimension name - that is used by the `spatial` axis validators.

The PR updates the tests and adds two new test params to check the new behaviour.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/core/axes.py
```diff
@@ -164,6 +177,10 @@ def get_validator(axis: str, data: Data, darray: DataArray) -> Optional[Callable
     #   empty set.
     for (ld_dims, ld_coords, ld_grid_type), ld_fnm in AXIS_VALIDATORS[axis].items():
```
dalonsoa, line 178: What does `ld` stands for here? Maybe give a more descriptive name - or explain somewhere what that is?

virtual_rainforest/core/axes.py
```diff
@@ -140,12 +140,21 @@ def get_validator(axis: str, data: Data, darray: DataArray) -> Optional[Callable
...
-    provided signatures. It returns either the appropriate function or None.
+    provided signatures.
+
+    If the input data array uses dimension names that have been registered to validate a
+    particular axis, then a matching validator must be found and an exception is raised
+    if no signature is matched. If the dimension names on the data array do not match
+    any of the registered dimensions, then the function returns None.
 
     Args:
         axis: The core axis to get a validator for
```
dalonsoa, line 151: I think I might be missing something here: what is the difference between an `axis` and a `dimension`? How could an `axis` be related with multiple dimensions?

virtual_rainforest/core/axes.py
```diff
@@ -140,12 +140,21 @@ def get_validator(axis: str, data: Data, darray: DataArray) -> Optional[Callable
...
         data: A Data object providing parameter information
```
dalonsoa, line 152: Given that you're only using `grid_type` I would pass only that as argument and not the whole `data` object. You say in the doctoring above that in the future, some validators might use other information contained in the `data` object, but given that `data` is not passed to the validators themselves, only used in `get_validator`, there's no point on passing the whole `data` to this function. That will also save you from importing `Data` here, removing the potential issue of circular imports.

dalonsoa (changes requested):
I've made a few comments, and also raised some questions, but otherwise it looks good.

virtual_rainforest/core/axes.py
davidorme: Yes - good point. There is a special case here with the `spatial` axis, which coerces `x` and `y` dimensions onto the `cell_id` spatial axis. In general, there will be only one dimension per axis!

virtual_rainforest/core/axes.py
davidorme: Actually, `data` is passed on to the validators and `data.grid` is used within the spatial validators. At the moment though, using `Data` is a complicated way to pass in a `Grid` object. 

I think we'll ultimately need a `Config` class. Then essentially all information needed to configure axis validators (and anything else) will be accessed by passing that around? For the moment, just using the `Grid` rather than the `Data` seems more sensible.

virtual_rainforest/core/axes.py
davidorme: Hah - it was an abbreviation for `loaders` before I renamed lots of things. They should be `validator_dims` or some abbreviation of that.

jacobcook1995 (approved):
LGTM

virtual_rainforest/core/axes.py
dalonsoa: I don't have the big picture in my head, but I'm not sure that creating another class is the way to go. You already have the configuration dictionary named `config`, so having a `Config` class will make things confusing. 

About `Data`, it only has `grid` and `data` attributes, so I'm not sure what else a validator might use. If you're referring to the `attrs` of the `data` Dataset, I'd pass those as `kwargs` and let each validator pick whatever they need. The same applies to other dimensions or coordinates already contained in the `data` Dataset. Ideally, you want to include as arguments to functions the information actually used in the function, and not big, complex objects that contain the information used. 

So, for the validators you have now, I'd rather use:

```python
def some_spacial_validator(data: DataArray, grid: Grid, **kwargs) -> DataArray:
    ...
```

And if you ever get a validator that require something else, like `param_1` contained in `data.attrs`, in addition to `grid`:

```python
def another_validator(data: DataArray, grid: Grid, param_1: float, **kwargs) -> DataArray:
    ...
```

And if you have validators that have nothing to do with the grid, but just some existing coordinate in the `data` dataset, for example `time`, then:

```python
def a_time_validator(data: DataArray, time: DataArray, **kwargs) -> DataArray:
    ...
```

All of them being called as:

```python
axis_validator_func(value, grid = self.grid, **dict(self.data.data_vars), **dict(self.data.coords), **self.data.attrs)
```

If we follow the variable names used in `data.py:170`.

As cumbersome as this call might look, it is cheap because all in there is passed as reference, it needs to be written only once and, more importantly, it let each validator to be explicit on what it actually needs to run, much better and clear to document.

virtual_rainforest/core/axes.py
dalonsoa: OK, so `loader` is the same as `validator`? I think I've seen used loader in other places, too, so it might be worth re-checking the naming.

virtual_rainforest/core/axes.py
dalonsoa: Quite confusing, but OK. I'm not sure I have a better solution to this at the moment.

virtual_rainforest/core/axes.py
davidorme: I was thinking longer term about _replacing_ the `config` dict with a class. At the moment the config process uses a chain of functions which  pass information back and forth between variables - most of that is about updating the state of the config dict and the schema that should be applied to it. The code would be simpler and cleaner if we had a `UserDict` style class which had methods updating attributes. That class would also retain more information about the config process - files used etc.

But - that's for a different branch 😄

dalonsoa (commented):
I've been going through this code over and over again, and while I'm totally convinced about the value of the validators, I'm not convinced at all on the implementation. It looks as if you're trying to make something generally applicable to more complex situations but, at the same time, you're tailoring it to tackle the case of `spatial` axis, which is a special case, in your own words. As a result, the whole process is pretty complex and inflexible. 

Although it might look counterintuitive, I'd suggest to setup the validators as classes and run ALL of them on the input array, letting the validator itself decide if it can validate the array.

This would be the skeleton of what I've in mind:

### axes.py

```python
AXIS_VALIDATORS: List[Callable] = []

#An abstract class
class Validator(ABC):
	@classmethod
	def __init_subclass__(cls):
		AXIS_VALIDATORS.append(cls)

	@classmethod
	def process(cls, value: NDArray, data: Data):
		val = cls()
		if val.can_validate(value, data):
			return val.validate(value, ...)
		return value
		
	@abstractmethod
	def can_validate(self, value: NDArray, data: Data) -> bool:
		#Logic to check if this validator can validate the value array
		#As complex or simple as you want.
		
	@abstractmethod
	def validate(self, value: NDArray, ...) -> NDArray:
		#The actual validation
```

### data.py

Replace `for` loop in 159 by

```python
from functools import reduce
...

value = reduce(lambda previous, fun: fun.process(previous, self), AXIS_VALIDATORS, value)
```

I understand if you do not want to make such a massive change like this, now, but I think it will be good to consider the long term sustainability and readability of your code.

davidorme:
> I'm totally convinced about the value of the validators

Well that's a good starting point :smile:

If I understand correctly, essentially you think that the structure of the DataArray 'signature' is a hostage to fortune, where encapsulating that check inside a Class essentially frees the validators up to use bespoke 'signatures'. I can definitely see the advantages - I don't think it is a massive change.

The 'downside' is that the encapsulation involves more processing (calling the various methods etc.) rather than a quicker pattern matching exercise, but I can absolutely see that the pattern matching is brittle.

davidorme:
One thing I can't immediately see how to fit into that framework is the grouping of validators by specific axis. So, for example, checking that an input array that uses `cell_id`, `x` or `y` as dimension or coordinate names _must_ get validated by at least one of the spatial validators.

My immediate solution is to retain that grouping by axis name in `AXIS_VALIDATORS` and use class attributes to set the axis name and dimension names?

```python
class SpatialCellID(Validator):
    axis_name = 'spatial'
    dim_names = ('cell_id',)
```

This way, `__init_subclass__` can add each new validators to a list of validators in under the axis name key in AXIS_VALIDATORS. Then in `data.py` you would still need to loop over the axis names, something like:

```python
for this_axis, validators in AXIS_VALIDATORS:
    # Get the dimension names associated with this axis
    axis_dimensions = set([v.dim_names for v in validators])

    # If the dataarray includes any of those dimension names, one of the validators for that 
    # axis must be able to validate the array, otherwise we can skip validation on this axis
    if axis_dimensions.intersect(darray.dims):
        # There should be one and only validator that can validate for this axis.
        validator_found = [v.can_validate(darray, data, grid) for v in validators]

        if not any(validator_found):
             raise RuntimeError(f"Data dimensions match {this_axis} axis but no validator can validate")

        if sum(validator_found) > 1:
             raise RuntimeError(f"Validators on {this_axis} axis not mutually exclusive")

        # Get the appropriate Validator class and then use it to update the data array
        this_validator = validators[validator_found.index(True)]
        darray = this_validator().validate(darray, data, grid)
```

That's basically pseudocode - too late in the day - but I can't see that you can do this easily without retaining this nesting within axis names. That doesn't take anything away from the advantages of the Validator ABC approach.

dalonsoa:
I think we are seeing the logic in a different order - or I am less worried about not executing code that is not needed; I honestly don't think a loss of performance here is a big deal.

To incorporate part of that functionality in the approach I'm suggesting I would modify `Validator` as follows:


```python
class Validator(ABC):
	
	axis_name = ""

	@classmethod
	def __init_subclass__(cls):
		if cls.axis_name == "":
			raise ValueError("Axis name cannot be an empty string.")
		AXIS_VALIDATORS.append(cls)

	@classmethod
	def process(cls, value: NDArray, data: Data):
		val = cls()
		if val.can_validate(value, data):
			if existing := value.attrs.get(f"{val.axis_name}_validator", None):
				raise RuntimeError(f"Axis {val.axis_name} already validated by {existing}")

			value = val.validate(value, ...)
			value.attrs[f"{val.axis_name}_validator"] = cls.__name__

		return value
		
	...
```

The check about the dimensions goes inside `can_validate`, together with any other checks on coordinates, grid or whatever. Only if the validator can validate the data we check if it has already been validated, and raise and error accordingly.

Granted that this does not account for the case when there are validators for that axis but none can actually validate the data  - is that a problem? - , and that it *might* involved more processing, but is this really a performance barrier?

davidorme:
Yeah - fair point about performance as this should be a pretty low use activity: it should basically only happen when initial data is loaded or when models create an internal variable.

The case where a DataArray uses dimension names that imply a particular axis but which don't then match a validator is important to catch: a user is being careless with dimension names and the dataset structure may get corrupted as a result. I can't see a way of handling that without some kind of grouping structure in the validators.

Using the `DataArray.attrs` to record axis validation in `process()` is neat - and something I was planning to do anyway, as the next (and hopefully last!) PR onto this branch is creating an `xarray.Accessor` to provide a `.has_axis('spatial')` method for DataArrays. The use case for that is to allow models to verify that any data loaded has been validated onto the required axes.

dalonsoa:
Ok, let's try the following option - and I am aware that things are becoming a bit more complicated that I initially intended:


```python
class Validator(ABC):
	
	axis_name = ""
	dims = ()

	@classmethod
	def __init_subclass__(cls):
		if cls.axis_name == "":
			raise ValueError("Axis name cannot be an empty string.")
		if cls.dims == ():
			raise ValueError("A validator needs to affect at least 1 dimension.")
		AXIS_VALIDATORS.append(cls)

	@classmethod
	def process(cls, value: DataArray, data: Data):
		val = cls()
		key = f"{val.axis_name}_validator"
		if not val.check_dims(value.dims):
			return value
			
		value.attrs[key] = value.attrs.get(key, "")

		if val.can_validate(value, data):
			if existing := value.attrs.get(key, ""):
				raise RuntimeError(f"Axis {val.axis_name} already validated by {existing}")

			value = val.validate(value, ...)
			value.attrs[key] = cls.__name__

		return value
		
	def check_dims(self, dims: Sequence) -> bool:
		return all([d in dims for d in self.dims])
		
	...

def assert_validation(attrs: Dict):
	missing = [k.strip("_validator") for k, v in attrs.items() if k.endswith("_validator") and v == ""]
	if any(missing):
		raise RuntimeError(f"Data dimensions match {missing} axis but no validator can validate")
```

And in `data.py`, once all validation has taken place, you run:

```python
assert_validation(value.attrs)
```

This has the potential advantage that the user gets information about all the failing validations at once, but I'm not sure how common it will be to have an array with multiple dimensions that need validating. I still thing it is a cleaner alternative than your version but, as I mention, it is a bit less simple than I was hoping for, so feel free to ignore it.

davidorme:
I think on balance that keeping the Validator class simple and having a small amount of extra code when applying the validators to handle the groups of validators per axis is easier to understand. It means we can't just use that `reduce` to apply validation, but I have to admit I find a `for` loop more comprehensible than `reduce` :smile: So, I will commit changes to adopt the ABC approach for the Validators but keep the axis grouping logic external to that class. For now anyway.

davidorme:
Or, perhaps better, wrap that group axis logic in a function:

```python
def check_axis(validators: list[Validator], value: DataArray, ....):
```

dalonsoa:
That was going to be my suggestion if you follow that route. Create a function that does all the validation stuff in axis.py and import that function in data.py, so it is the only validation step in __setitem__

davidorme:
@dalonsoa Many thanks for those reviews - the recent pushes implement that compromise solution of using the ABC for the validators and a wrapper `validate_dataarray` to handle validation over the groups of validators by axis. I've deliberately left out the use of the `DataArray.attrs` to record successful validation on an axis - I have small additional PR to create an accessor for that.

virtual_rainforest/core/axes.py
```diff
@@ -1,323 +1,384 @@
...
+        When new subclasses are created this method automatically extends the
+        :var:`~virtual_rainforest.core.axes.AXIS_VALIDATORS` registry. The subclass is
+        added to the list of AxisValidators that apply to the subclass core axis.
+        """
 
-    Args:
-        axis: The core axis that the validator applies to
-        signature: A tuple of the validator signature.
-    """
+        if cls.core_axis == "":
```
dalonsoa, line 84: Shouldn't you also check that the `dim_names` are not an empty set?

virtual_rainforest/core/axes.py
```diff
@@ -1,323 +1,384 @@
...
-@register_axis_validator("spatial", (("cell_id",), ("cell_id",), ("any",)))
-def spld_cellid_coord_any(data: Data, darray: DataArray) -> DataArray:
-    """Spatial loader for cell id coordinates onto any grid.
+            # Get the appropriate Validator class and then use it to update the data
+            # array
+            this_validator = validators[validator_found.index(True)]
+            try:
+                value = this_validator().validate(value, grid, **kwargs)
+            except Exception as excep:
+                log_and_raise(str(excep), excep.__class__)
```
dalonsoa, line 209: With this approach, you're running `can_validate` twice and have a too verbose code. If you follow this route, then I'd do the following:


```python
validator_found = [v() for v in validators if v().can_validate(value, grid)]

if len(validator_found) == 0:
    log_and_raise(
        f"DataArray uses '{axis}' axis dimension names but does "
        f"not match a validator: {','.join(matching_dims)}",
        ValueError,
    )

elif len(validator_found) > 1:
    log_and_raise(
        f"Validators on '{axis}' axis not mutually exclusive", RuntimeError
    )

 # Get the appropriate Validator class and then use it to update the data
 # array
 try:
	 value = validators[0].run_validator(value, grid, **kwargs)
 except Exception as excep:
	 log_and_raise(str(excep), excep.__class__)
```

And remove the `validate` class method because it is not necessary.

dalonsoa (approved):
Looks much cleaner! I approve it, yet, I've made a comment to further simplify the code, because in mixing both approaches there's some redundant code you can get rid off.

virtual_rainforest/core/axes.py
davidorme: Ha - I did wonder about that 😄 I thought that having that all in one method might be a useful tool at some point, but agree we should keep it simple. I will remove that and merge.

virtual_rainforest/core/axes.py
dalonsoa: That's fine, let's not overthink tests! Parametric testing is useful when parameters are simple stuff, otherwise it is not worth the effort and, indeed, masks what you are actually testing.

virtual_rainforest/core/axes.py
dalonsoa: Mind that I've also updated how the `validator_found` list is constructed and the conditional statements, making thins, I think, cleaner.

PR #134, opened 2022-12-15
Minimal carbon pool model

### LINKED ISSUE #99: Demo soil model
Need a basic model for v0.1 to run so that we can demonstrate that the overall simulation setup and run process works as we want it to.

Planning to start with the soil module. Here, we will define a simple set of differential equations describing outcomes of Rock Paper Scissors. These can then be solved through time and across the grid.

Considered defining the basic model in terms of a few carbon pools and their equations. Decided against as I thought it was better to put in information is definitely wrong (i.e. soil doesn't play rock, paper, scissors), rather than incomplete/poorly parameterised versions of the final model.

jacobcook1995:
# Description

I've added a basic carbon pool to the soil module. It only includes 2 pools (the full model will probably include 5 pools), and a function that handles transfers between these two pools. This is all bundled together as part of a `SoilCarbon` class.

I was originally intending to implement an alternate equation form and a method to switch between the two, but decided that was a bit ambitious for a single pull request. Structurally there's a couple of things that I'm not completely happy with and would want to clear up before I moved onto something like that.

1. The constant dictionaries defined at the top of the script define a set of fitting parameters for the various equations, these are taken from fits to data carried out in the literature. I'm not sure if this is a good way to store this kind of information, but it didn't seem sensible either to define these fitting parameters as independant constants either. I'm also unsure whether treating these as constants is a sensible approach either as I could see us wanting to do sensitivity analysis with them down the line.
2. I wrote the code with no sanity checking for variables except when the class is initialised. My reasoning for this is that too common sanity checking would slow down execution, and that we probably don't want to simulation to halt entirely because a negative value has crept into one differential equation in one grid cell. Is this broadly sensible?

No hurry on reviewing this, I mainly wanted to get the pull request submitted before I finished for Christmas. 

Fixes #99

## Type of change

## Key checklist

## Further checks


virtual_rainforest/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+            1 + self.lmwc * binding_affinity
+        )
+
+        # Find scalar factors that multiple rates
+        temp_scalar = scalar_temperature(soil_temp)
+        moist_scalar = scalar_moisture(soil_moisture)
+
+        flux = temp_scalar * moist_scalar * self.lmwc * (equib_maom - self.maom) / Q_max
+
+        return -flux, flux
```
alexdewar, line 158: I'm not sure why you're returning both `flux` and `-flux`. Surely just returning `flux` is enough? The user can work out the other one.

If you do need both and they're always both the same length, then *maybe* this would be better expressed as a 2D numpy array.

virtual_rainforest/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def scalar_temperature(soil_temp: NDArray[np.float32]) -> NDArray[np.float32]:
```
alexdewar, line 161: It's a good rule of thumb to start your function names with a verb as they are (hopefully) doing something. So in this case, maybe it could be `convert_temperature_to_scalar` or something. Yes it can end up being overly verbose and I don't think it should be a *hard* rule, but I think it's worth considering here for this and other function names.

virtual_rainforest/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+
+    # This expression is drawn from Abramoff et al. (2018)
+    scaler = 1 / (
+        1
+        + MOISTURE_SCALAR["coefficient"]
+        * e ** (-MOISTURE_SCALAR["exponent"] * soil_moisture)
+    )
+
+    return scaler
```
alexdewar, line 203: Two things:

1. I'd just return the expression directly without the temporary variable
2. Use `np.exp()` to calculate the exponent rather than `e ** x` (I know these things look the same but I *think* a separate `exp()` function is usually provided by programming languages because it's either faster or more accurate to do it that way)

```suggestion
    return 1 / (
        1
        + MOISTURE_SCALAR["coefficient"]
        * np.exp(-MOISTURE_SCALAR["exponent"] * soil_moisture)
    )
```

alexdewar (changes requested):
This looks really good and it's also a good example of how a PR should be done (i.e. you provided a good explanation, and your code has tests and lots of docstrings). Well done!

I've made some small suggestions throughout. Some slightly "bigger picture suggestions I had were:

1. It'd be helpful to say what units each of your variables are throughout (in docstrings etc.) where they represent physical values (e.g. I'd assume all your temperatures are in °C but would like to know for sure, I don't know what the unit of soil moisture is at all etc.)
2. While it's great that you have tests for each of your functions, you're oftentimes only testing that it works for a single set of parameters. This is *much* better than having nothing at all, but I think you should go one step further and test multiple sets or, ideally, ranges of values (i.e. by using `@pytest.mark.parametrize`). I was working on something similar myself today: https://github.com/ImperialCollegeLondon/FINESSE/blob/main/tests/test_tc4820.py

virtual_rainforest/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
+"""The `soil.carbon` module.
```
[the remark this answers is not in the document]
jacobcook1995: Yeah we did agree that that nesting would be better. I didn't change it because the nesting has been changed in a different pull request and I was concerned about merge conflicts, but is it more sensible to just change it wherever appropriate and let git handle the merging?

alexdewar:
I realise I forgot to answer your questions:

1. Seems sensible to me!
2. I would just add the sanity checks. In the grand scheme of things it shouldn't hurt performance much and it could potentially save you a lot of debugging pain further down the line.

davidorme:
On question 2, I think this overlaps with #135 - I've added a sketch of how this might work across modules to that discussion.

virtual_rainforest/models/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+        binding_affinity = 10.0 ** (
+            BINDING_WITH_PH["slope"] * pH + BINDING_WITH_PH["intercept"]
+        )
+
+        # This expression is also drawn from Mayes et al. (2012)
+        # Original paper also depends on Fe concentration, but we are ignoring this for
+        # now
+        Q_max = bulk_density * 10 ** (
+            MAX_SORPTION_WITH_CLAY["slope"] * np.log10(percent_clay)
+            + MAX_SORPTION_WITH_CLAY["intercept"]
```
davidorme, line 144: You could make use of matrix multiplication here to speed this up? I don't know if makes a massive difference with `a + bx` , so maybe not here but this could be:

```python
msc = np.array([2.328, 0.483])

np.sum(
    msc * np.column_stack([np.ones_like(percent_clay), np.log10(percent_clay)]), axis=1
)
```

virtual_rainforest/models/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+       soil_temp: soil temperature for each soil grid cell
+    """
+
+    # This expression is drawn from Abramoff et al. (2018)
+    numerator = TEMP_SCALAR["t_2"] + (TEMP_SCALAR["t_3"] / np.pi) * np.arctan(
+        np.pi * (soil_temp - TEMP_SCALAR["t_1"])
+    )
+    denominator = TEMP_SCALAR["t_2"] + (TEMP_SCALAR["t_3"] / np.pi) * np.arctan(
+        np.pi * TEMP_SCALAR["t_4"] * (TEMP_SCALAR["ref_temp"] - TEMP_SCALAR["t_1"])
+    )
```
davidorme, line 179: There are repeated dictionary lookups here. I don't know if this is a big deal (and I probably do something similar in `pyrealm`!) but you could do:

```
TEMP_SCALAR = np.array([15.4, 11.75,  29.7,  0.031,  30.0])
```

And then:

```python
t_1, t_2, t_3, t_4, ref_temp = TEMP_SCALAR

# This expression is drawn from Abramoff et al. (2018)
numerator = t_2 + (t_3 / np.pi) * np.arctan(np.pi * (soil_temp - t_1))
denominator = t_2 + (t_3 / np.pi) * np.arctan(np.pi * t_4 * (ref_temp - t_1))
```

virtual_rainforest/models/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # Check that arrays are of equal size and shape
+        if maom.shape != lmwc.shape:
+            log_and_raise(
+                "Dimension mismatch for initial carbon pools!",
+                InitialisationError,
+            )
+
+        # Check that negative initial values are not given
+        if any(i < 0 for i in maom) or any(i < 0 for i in lmwc):
```
davidorme, line 65: I think this is better and likely faster as `if np.any(maom < 0) or np.any(lmwc < 0)`

virtual_rainforest/models/soil/carbon.py
```diff
@@ -0,0 +1,203 @@
...
+            percent_clay: Percentage clay for each soil grid cell
+
+        Returns:
+            lmwc_to_maom: The net flux from LMWC to MAOM
+        """
+
+        # This expression is drawn from (Mayes et al. (2012))
+        binding_affinity = 10.0 ** (
+            BINDING_WITH_PH["slope"] * pH + BINDING_WITH_PH["intercept"]
+        )
```
davidorme, line 137: Hah - was going to suggest `np.power` here, but turns out be no better:

```python
In [18]: import timeit
In [19]: timeit.timeit("10.0 ** np.arange(10)", setup="import numpy as np")
Out[19]: 1.8152062078006566
In [23]: timeit.timeit("np.power(10.0, np.arange(10))", setup="import numpy as np")
Out[23]: 1.8450454170815647
```

And should you accidentally set that up with an integer? 
```
In [20]: timeit.timeit("np.power(10, np.arange(10))", setup="import numpy as np")
Out[20]: 3.010193958878517
```

davidorme (changes requested):
Looks good - some suggestions on the tests and some possible tweaks on the implementation, but nothing serious. I think the tests are worth updating, so requesting those changes.

jacobcook1995:
@alexdewar I definitely agree that there is a value in testing a wider range of values for the functions. I hit a problem in doing so though, because I wasn't sure the best way to generate expected values in this case.

I generated the current expected return values of the functions using a calculator to ensure that I wasn't just assuming that the current implementation of the function is correct (which in a couple of cases it wasn't). Testing a broader set of parameter values means I would have to generate a lot more expected values, and if I do that by calculator would be pretty time consuming.

I was wondering whether I am safe to assume that the current implementation is fine once it has been tested for a few values, or if I should always be externally verifying values? If the later I was wondering if there is a smarter way to do this than manual calculation?

davidorme:
> I was wondering whether I am safe to assume that the current implementation is fine once it has been tested for a few values, or if I should always be externally verifying values? If the later I was wondering if there is a smarter way to do this than manual calculation?

I had this discussion with @vgro before Christmas. With the plants module, some of the things I'm doing are re-implementations and so I can generate a test set of data points with independent implementations (or at least robust to implementation differences - they might both be wrong for some other reason). I think there is still value in using a wider range of test values generated from the same implementation - useful for detecting breaking changes in output values and probably also detecting surprising edge and corner cases?

virtual_rainforest/models/soil/carbon.py
alexdewar: I think it'll probably make things slower tbh, but either way there's not likely to be much in it time-wise and I'd be inclined to leave it as is for clarity's sake.

alexdewar:
> @alexdewar I definitely agree that there is a value in testing a wider range of values for the functions. I hit a problem in doing so though, because I wasn't sure the best way to generate expected values in this case.
> 
> I generated the current expected return values of the functions using a calculator to ensure that I wasn't just assuming that the current implementation of the function is correct (which in a couple of cases it wasn't). Testing a broader set of parameter values means I would have to generate a lot more expected values, and if I do that by calculator would be pretty time consuming.
> 
> I was wondering whether I am safe to assume that the current implementation is fine once it has been tested for a few values, or if I should always be externally verifying values? If the later I was wondering if there is a smarter way to do this than manual calculation?

Tagging @dalonsoa because I think he'll be about to give better feedback about this (he's back on Monday).

I think the right thing to do is to test the smaller individual functions with ranges of values. Then you can just test that your higher-order functions are just calling those functions with the correct arguments etc.

As I said, I think what you've done is def a good start though. While it's harder to think about how to do this when it comes to computing values, the other thing you can check for is whether your methods raise exceptions or not depending on the input values, which should be easier. So you could do that for starters.

alexdewar:
> > I was wondering whether I am safe to assume that the current implementation is fine once it has been tested for a few values, or if I should always be externally verifying values? If the later I was wondering if there is a smarter way to do this than manual calculation?
> 
> I had this discussion with @vgro before Christmas. With the plants module, some of the things I'm doing are re-implementations and so I can generate a test set of data points with independent implementations (or at least robust to implementation differences - they might both be wrong for some other reason). I think there is still value in using a wider range of test values generated from the same implementation - useful for detecting breaking changes in output values and probably also detecting surprising edge and corner cases?

Yeah, this isn't a bad idea provided you don't spend too long on reimplenting stuff. In the link I sent, I did some reimplementing of functionality.

virtual_rainforest/models/soil/carbon.py
davidorme: I almost didn't post it 😄 I'd just been reviewing some code where there are several consecutive sums of powers and there it was definitely easier than writing each term!

https://github.com/davidorme/pyrealm/blob/master/pyrealm/pmodel.py#L73-L85

virtual_rainforest/models/soil/carbon.py
```diff
@@ -0,0 +1,201 @@
...
+        if np.any(maom < 0) or np.any(lmwc < 0):
+            log_and_raise(
+                "Initial carbon pools contain at least one negative value!",
+                InitialisationError,
+            )
+
+        self.maom = maom
+        """Mineral associated organic matter pool"""
+        self.lmwc = lmwc
+        """Low molecular weight carbon pool"""
```
dalonsoa, line 72: These strings describing the attributes should go in the Class docstring (to document the attributes) and in the `__init__` docstring to document what those parameters mean.

davidorme: Shouldn't there usually only be one of those two things? It seems to be common practice to either document the class in the class docstring or in the class.__init__ docstring, but not to have both?

@alexdewar raised the question of where to put attribute docstrings [here](https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/140#discussion_r1064568241)  - we just need to agree on a common practice.

dalonsoa:
It's a tricky thing to test functions over a range of possible input values that implement specific formulas because in those cases you are interested in checking that the specific numerical output is the correct one, and not just properties of the output (which is what `hypothesis` would do). 

Alex's suggestion on mocking the low level functions when testing the higher level ones is a good one, as well as breaking down functions in as small logical bits as possible - eg. implementing one formula per function - but otherwise I think that manually picking a few sensible and edge cases is the only possible way to go here.

virtual_rainforest/models/soil/carbon.py
davidorme: Although I note the Google pystyle still has a stub doc for `__init__`:
https://google.github.io/styleguide/pyguide.html#384-classes

dalonsoa: @alexdewar suggestion is correct and so is mine according to the [Google Style Docstring guide for python](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). So, as you say, it is a question of agreeing on the common practice. I prefer my option, obviously :P

dalonsoa: Just check that the option you choose renders as you would like in Read The Docs!

jacobcook1995: Have checked that these render okay in Read The Docs. Personally I prefer the style with docstrings after each attribute as I feel it makes it easier to notice when they are missing. What do others think @davidorme @vgro @TaranRallings @robewers01 ?

jacobcook1995:
I think I've addressed most of your comments.

In terms of sanity checks, I've added them in a couple of cases where there are very clear bounds that wouldn't necessarily be obvious when be debugging. Other cases I've left to be handled by the bounds module down the line, as we probably want to constrain e.g. soil bulk density more tightly than just being strictly positive, but it wouldn't be sensible to bake a particular choice into the function

Anyway let me know your thoughts

virtual_rainforest/models/soil/carbon.py
davidorme: Shall we discuss this in person next week. We can put together a few examples of different structures and see what we feel about the rendering and code style.

In the past, I have generally put `__init__` is documented in the class docstring. It then appears naturally in docs right where the class is introduced. I think I prefer documenting the instance attributes using docstrings in  `__init__` (or above `__init__` for class variables). It is clearer to see when one is missing and the list of attributes in the compiled docs is built automatically.

I'm not convinced about that model - having a class overview in the class docstring and then `__init__` specific content with `__init__` also works.

```python
class MyClass:
    """This is my class.

    There are many like it, but this one is mine. Lots of text.

    Args:
        myarg: A value for my class
    """"

    unknown: int = 0
    """Need to explicitly say this is a class attribute - it gets lumped in."""

    def __init__(self, myarg: int) -> None:
        """Create a new MyClass instance."""

        self.myarg: int = myarg
        """The myarg value used to create the instance."""
        self.twice_myarg: int = self.myarg * 2
        """Twice my arg"""
```

docs/source/api/soil.md
```diff
@@ -14,15 +14,24 @@ kernelspec:
...
-## The `virtual_rainforest.soil.model` module
+## The `virtual_rainforest.models.soil.model` module
 
 ```{eval-rst}
-.. automodule:: virtual_rainforest.soil.model
+.. automodule:: virtual_rainforest.models.soil.model
+    :autosummary:
+    :members:
+    :special-members: __init__
+```
```
davidorme, line 29: Depending on how large these are likely to be, could use subpages within a folder, so:
```
soil_overview.md
soil/model.md
soil/carbon.md
```

virtual_rainforest/models/soil/carbon.py
jacobcook1995: Yeah probably a sensible thing to decide in person. I guess I can leave them as they are for now, and then if we decide to get rid of them I can remove them during the session next week?

docs/source/api/soil.md
jacobcook1995: Yeah that is probably a good idea `soil/carbon.md` in particular is likely to get fairly large, so splitting the content makes sense

virtual_rainforest/models/soil/carbon.py
```diff
@@ -156,6 +146,73 @@ def mineral_association(
...
+
+    Q_max = bulk_density * 10 ** (
+        MAX_SORPTION_WITH_CLAY["slope"] * np.log10(percent_clay)
+        + MAX_SORPTION_WITH_CLAY["intercept"]
+    )
+    return Q_max
+
+
+def calculate_equilibrium_maom(
+    pH: NDArray[np.float32],
```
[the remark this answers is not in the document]
davidorme: I did look to see if there was a switch somewhere, but the `numpy` maintainers have been here before 😄 https://github.com/numpy/numpy/issues/6860

virtual_rainforest/models/soil/carbon.py
```diff
@@ -190,9 +247,14 @@ def convert_moisture_to_scalar(
     needed.
 
     Args:
-        soil_moisture: soil moisture for each soil grid cell
+        soil_moisture: relative water content for each soil grid cell (unitless)
     """
 
+    if any(soil_moisture > 1.0) or any(soil_moisture < 0.0):
+        log_and_raise(
```
davidorme, line 254: I think these have to be `np.any` to handle the array inputs.  Otherwise you get:

```python
In [8]: x = np.arange(24).reshape(4,6)

In [9]: any(x > 4)
ValueError                                Traceback (most recent call last)
Input In [9], in <cell line: 1>()
----> 1 any(x > 4)

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

davidorme (changes requested):
Looks good but a few things to look at.

virtual_rainforest/models/soil/carbon.py
```diff
@@ -1,4 +1,4 @@
-"""The `soil.carbon` module.
+"""The `models.soil.carbon` module.
 
 This module simulates the radiation soil carbon cycle for the Virtual Rainforest. At the
```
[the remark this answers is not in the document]
jacobcook1995: Ahh yes, that should not be there

tests/test_soil_carbon.py
```diff
@@ -61,69 +61,233 @@ def test_soil_carbon_class(caplog, maom, lmwc, raises, expected_log_entries):
...
-        pH, bulk_density, soil_moisture, soil_temp, percent_clay
+    soil_carbon = SoilCarbonPools(
+        maom=np.array(maom, dtype=np.float32), lmwc=np.array(lmwc, dtype=np.float32)
+    )
+
+    # Mock required values
+    mock_t_scalar = mocker.patch(
+        "virtual_rainforest.models.soil.carbon.convert_temperature_to_scalar"
+    )
+    mock_t_scalar.return_value = np.array(temp_scalar, dtype=np.float32)
```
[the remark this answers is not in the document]
davidorme: Well, yes it does isolate the problem but mocking absolutely everything but the function you are testing:

* is a lot of extra typing,
*  makes the test flow much less clear,
*  and is really difficult to debug if there is a typo in your mock value because the source of that value is obscure.

Sure, if a low level function breaks, you _will_ get a cascade of dependent failures, but that dependency is real and I think relying on units that are tested elsewhere is part of the test framework. And is _much_ easier to write.

virtual_rainforest/models/soil/carbon.py
```diff
@@ -192,9 +247,14 @@ def convert_moisture_to_scalar(
...
 
     Args:
-        soil_moisture: soil moisture for each soil grid cell
+        soil_moisture: relative water content for each soil grid cell (unitless)
     """
 
+    if np.any(soil_moisture > 1.0) or np.any(soil_moisture < 0.0):
+        log_and_raise(
+            "Relative water content cannot go below zero or above one!", ValueError
+        )
```
alexdewar, line 256: Note you will need to update this if #158 is merged before this.

alexdewar (approved):
I've made some very minor comments, but otherwise I think this is good to go. Good job!

virtual_rainforest/models/soil/carbon.py
```diff
@@ -156,6 +146,73 @@ def mineral_association(
...
+
+    Args:
+        pH: pH values for each soil grid cell
+
+    Returns:
+        binding_coefficient: Langmuir binding coefficients for mineral association of
+            labile carbon (m^3 kg^-1)
+    """
+
+    return 10.0 ** (BINDING_WITH_PH["slope"] * pH + BINDING_WITH_PH["intercept"])
```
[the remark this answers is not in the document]
jacobcook1995: Ahh right is that the case for everywhere that I'm using dictionaries of constants? Or is this specific to `BINDING_WITH_PH`?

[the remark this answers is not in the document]
jacobcook1995: And the class would be defined like this

```python
@dataclass
class BindingWithPH:
    """From linear regression (Mayes et al. (2012))."""
    intercept: float = -0.186
    slope: float = -0.216
```

and accessed like this?

```python
return 10.0 ** (BindingWithPH.slope * pH + BindingWithPH.intercept)
```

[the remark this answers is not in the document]
davidorme: If it is just the `object.thing` notation then there are other options for having `dot` style dictionaries (`dotmap` for example). But @alexdewar did you mean that `BindingWithPH` would also provide `BindingWithPH.calc_binding`?

``` python
@dataclass
class BindingWithPH:
    """From linear regression (Mayes et al. (2012))."""
    intercept: float = -0.186
    slope: float = -0.216
    
    def calc_binding(ph: NDArray) -> NDArray
        return 10.0 ** (self.slope * pH + self.intercept)
```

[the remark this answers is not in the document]
davidorme: More generally, I think there's an argument for the `constants` module providing something that can be easily serialised to JSON or YAML so that the `constants` can become part of the configuration if needed. `core.constants` might provide a default set, but people might want to play around with the settings.

I've used `dataclasses` for this kind of role here: https://github.com/davidorme/pyrealm/blob/master/pyrealm/param_classes.py

I will accept that the size of the dataclasses may not be ideal 😬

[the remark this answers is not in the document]
alexdewar: @jacobcook1995 Yes, that's essentially what I had in mind, but I'm open to suggestions if something like `dotmap` (which I hadn't heard of till now) is better.

The reason I suggested a dataclass was so you can get better type linting (i.e. `mypy` will be able to see what fields your `BindingWithPH` class has and what types they are). You could also just have them as separate variables.

jacobcook1995:
Made the decision to switch to data classes rather than dot map as I could work out how to add parameter specific docstrings more easily with data classes. Happy for this to change down the line when we agree upon a consistent approach between modules for constants units etc

davidorme (approved):
LGTM. Obvious caveats about how we are going to do constants and possibly having some kind of shared utility for bounds checking and if so where.

davidorme (changes requested):
Wait. I think there's something off about using dataclasses like that.

davidorme:
So I may be not understanding how dataclasses are used, but I've always seen them used to create an instance of the dataclass. So given:

```python
@dataclass
class BindingWithPH:
    """From linear regression (Mayes et al. (2012))."""
    intercept: float = -0.186
    slope: float = -0.216
```

Then a function might work like this:

```python
def calculate_binding_coefficient(pH: NDArray[np.float32], coef: BindingWithPH = BindingWithPH())
```

And then the code can use `coef.slope`, etc. You could then run that calculation with the default values or something else:

```python
res = calculate_binding_coefficient(pH_array, BindingWithPH(slope=-0.23))
```

Here, we're using the attributes more like class attributes. The dataclass does *expose* those attributes but they seem to fundamentally not do what I expect. So for example:

```python
In [40]: @dataclass
    ...: class BindingWithPH:
    ...:     """From linear regression (Mayes et al. (2012))."""
    ...:     intercept: float = -0.186
    ...:     slope: float = -0.216
    ...: 

# create a normal instance

In [41]: inst  = BindingWithPH()
In [42]: inst
Out[42]: BindingWithPH(intercept=-0.186, slope=-0.216)

# and with a different slope
In [43]: inst  = BindingWithPH(slope=3)
In [44]: inst
Out[44]: BindingWithPH(intercept=-0.186, slope=3)

# The value _can_ be accessed directly from the class, not an instance
# as if it is a class attribute

In [45]: BindingWithPH.slope
Out[45]: -0.216

# And it can be changed too

In [46]: BindingWithPH.slope = 3
In [47]: BindingWithPH.slope
Out[47]: 3

# BUT that doesn't interact with the instance __init__, which keeps
# returning the original defaults.

In [48]: inst  = BindingWithPH()
In [49]: inst
Out[49]: BindingWithPH(intercept=-0.186, slope=-0.216)
In [50]: BindingWithPH.__init__
Out[50]: <function __main__.BindingWithPH.__init__(self, intercept: float = -0.186, slope: float = -0.216) -> None>
```

I'm not sure what the advantage of `dataclass` over a `namedtuple` is here - we're not using instances of it. 

I'm also wary of how users change these values. I'm not sure, but at the moment I think you _could_ do it by changing the "class attribute" value as above, but you then have absolutely no control over what other functions might be accessing that altered value. If an instance is created, with defaults or altered, then the user knows what version they are passing.

This all seems really familiar for some reason!

alexdewar:
Yeah, I guess a `namedtuple` would be another way to go. Alternatively you could just have two separate constants defined at the module level.

I hear your concerns about the values being mutable, but that is kind of what you get with Python. Module-level constants are mutable too.

davidorme:
I think `dataclasses` are probably the way to go - see the discussion here #162 - and I agree about the mutable being Python for you. I have used frozen `dataclasses` over in `pyrealm`, which is probably just unnecessary paranoia. Well, actually, not paranoia because someone will do it, but stopping people changing key constants half way through an analysis should be a job for common sense, not code.

jacobcook1995:
So the suggestion here @davidorme would be to keep the classes as they are but move them to a `models/soil/constants.py` script so that all model data classes for the module are stored in a single location? And then these would be imported in and supplied as a default value for functions that use them?

davidorme:
Yes - I think so. That's very close to what you have already and seems like a sensible way to organise things anyway. Something else might come of #162 but no alarm bells yet!

davidorme (approved):
LGTM!

PR #136, opened 2022-12-19
Added xarray dataset accessor to record axis validation

davidorme:
# Description

This PR extends `xarray.DataArray` to add a `core_axes` property that is used to record the axis validation applied to that data array. It uses  the `xarray` Accessor framework.

The `core_axis` property is basically a dictionary of which AxisValidator was applied to each core axis (or `None` if no validator was applied. The property is also made callable as a simple shorthand boolean test of whether a named axis has been validated. For example, a model `__init__` might confirm:

```python
if not elevation.core_axis('spatial'):
    raise ValueError("Elevation data not validated on spatial axis")
```

I worry a little that making the `dict` callable is a bit too clever. The main reason is to keep a short API - it would otherwise have to be something like `elevation.core_axis.has_axis('spatial')`, which admittedly isn't exactly verbose.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -200,13 +200,13 @@ class Data:
         """
 
         if var_name not in self.data:
-            raise KeyError(f"Unknown variable name: {var_name}")
-
-        if axis_name not in AXIS_VALIDATORS:
-            raise KeyError(f"Unknown core axis name: {var_name}")
+            raise ValueError(f"Unknown variable name: {var_name}")
 
         if var_name not in self._variable_validation:
             raise RuntimeError(f"Missing variable validation data: {var_name}")
+
+        if axis_name not in AXIS_VALIDATORS:
+            raise ValueError(f"Unknown core axis name: {axis_name}")
 
         return False if self._variable_validation[var_name][axis_name] is None else True
 
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -197,6 +197,13 @@ class Data:
         Args:
             var_name: The name of a variable
             axis_name: The core axis name
+
+        Returns:
+            A boolean indicating if the variable was validated on the named axis.
+
+        Raises:
+            ValueError: Unknown variable or core axis name
+            RuntimeError: Incomplete variable validation data in Data instance.
         """
 
         if var_name not in self.data:
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -300,7 +300,7 @@ class Data:
                 # processed
                 try:
                     self.load_from_file(
-                        file=each_var["file"],
+                        file=Path(each_var["file"]),
                         file_var_name=each_var["file_var_name"],
                         data_var_name=each_var.get("data_var_name"),
                     )
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -84,7 +84,7 @@ from typing import Any, Optional
 
 from xarray import DataArray, Dataset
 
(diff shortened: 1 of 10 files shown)


davidorme:
I have in the back of my mind that we might provide a `check_variables` function that might take a dict of variable names and the axes they should be validated against as a utility function for model startups: 

```python
check_variables({"elevation": ('spatial',), "temperature": ('spatial', 'temporal')})
```

or even a decorator?

```python
@require_axes({"elevation": ('spatial',), "temperature": ('spatial', 'temporal')})
```
But I think worth seeing what the common use patterns are first!

dalonsoa (commented):
As it is, it is OK, but I'm not sure I see the need of it - even from the functionality point of view. What's wrong with:
```python
value.attrs["core_axis"][axis_name] = ...
```
It is reusing an existing interface, one that will be known by `xarray` users, one adequate to store the information you want, it is simple... I feel you are over-engineering something that is pretty simple, in nature.

davidorme:
That's true, but `attrib` is an interface that forms part of the NetCDF data structure. So, if we export objects from the `Data` instance using the `xarray` writer functions, the user will get data files with noisy attributes. This approach isolates these internal structures from the data structures. But that's quite a niche argument.

dalonsoa:
I still don't see the need, but it is you call. I'd suggest you gather the opinion from other members of the team.

davidorme:
I was just playing around to check if I _really_ object to noisy attributes in output NetCDF files or if I just like playing with the Accessor framework 😄. 

It turns out that you although you can create `dict` attributes in `DataArray.attrs`, you _cannot_ then export that to NetCDF files. We'd need to explicitly scrub the `core_axes` key before writing data. 

```python
In [42]: ds = xr.DataArray([1,2,3,4], name='test')
In [43]: ds.attrs['core_axes'] = {'spatial': "Spat_CellId_Coord_Any", "temporal": "Temporal_Date", "soil_layers": None}
In [44]: ds.to_dataset().to_netcdf('Attrs_test.nc')
TypeError                                 Traceback (most recent call last)
...
TypeError: Invalid value for attr 'core_axes': {'spatial': 'Spat_CellId_Coord_Any', 'temporal': 'Temporal_Date', 'soil_layers': None}. For serialization to netCDF files, its value must be of one of the following types: str, Number, ndarray, number, list, tuple
```

Changing to having a `Dataset` as the central data store and being able to use its export methods is a real advantage, so having to wrap those to get rid of `core_axes` attributes seems like a hack. I'm sure there are ways we could work around it (two lists: `core_axis_name` and `core_axis_validator`, although we'd then need a placeholder for `None`, which is also not a valid attribute type).

dalonsoa:
Ok, that is a **strong** reason for the accessor you created. So all good 👍

dalonsoa:
That's annoying! Given that is no longer linked to the DataArray of interest, I would make all of these methods/attributes of the `Data` object, so no using accessors anymore.

davidorme:
Sorry @dalonsoa - yet another iteration...

virtual_rainforest/core/data.py
```diff
@@ -185,6 +188,28 @@ def __contains__(self, key: str) -> bool:
...
+    def on_core_axis(self, var_name: str, axis_name: str) -> bool:
+        """Check core axis validation.
+
+        This function checks if a given variable loaded into a Data instance has been
+        validated on one of the core axes.
+
+        Args:
+            var_name: The name of a variable
+            axis_name: The core axis name
+        """
```
dalonsoa, line 200: Need `Return` and `Raises` explanations in docstring.

dalonsoa (approved):
Looks good! Just a couple of comments on doc strings. 

Sorry it took me so long for so little input...

PR #140, opened 2023-01-09
Feature/energy balance dummy

vgro:
# Description

This is the first draft of the energy_balance sub-module. So far, this includes the basic structure, initialization, and three small functions to derive basic variables for further calculations. 
Could you please give me feedback on the init, basically should all of this go in the json schemer?
No tests yet.

## Type of change

## Key checklist

## Further checks


davidorme (changes requested):
I think this looks really good. For the config side of things, I've made some comments about how the flow of the `EnergyBalance` class works. At the moment, it looks like something that is created for each time step, but you could have a flow like this:

* Abiotic module is initialised
    * `EnergyBalance` is configured with simulation wide constants (n layers etc.)
* During simulation:
    * `EnergyBalance.set_canopy(args)` is run to configure the current canopies across the grid cells and the absorbed radiation.
    * `EnergyBalance.calculate_energy_balance(args)` is run to calculate the energy balances.
* Go to next time step.

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
+"""The `abiotic.energy_balance` module.
+
+This module calculates the energy balance for the Virtual Rainforest.
+The sequence of processes is based on Maclean et al, 2021: Microclimc: A mechanistic
+model of above, below and within-canopy microclimate. Ecological Modelling
+Volume 451, 109567. https://doi.org/10.1016/j.ecolmodel.2021.109567.
+
+# --- this is the sequence of processes in microclimc for one time step ---
```
jacobcook1995, line 8: Is this taken from the package you are extracting the equations from? If it is I think it might be worth contextualising this bit more in the docstring

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+        "psi_e": 0.1,  # Matric potential (J / m^3)
+    },
+}
+
+
+class EnergyBalance:
+    """EnergyBalance method.
+
+    Attributes:
+        air_temperature: NDArray[np.float32],
```
jacobcook1995, line 125: Don't know if giving type information here is necessary, as it's already covered by the init function signature

jacobcook1995: Also think that description of what the attributes are would be useful here

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+            np.array(MAX_LEAF_AIR_CONDUCTIVITY),
+            self.canopy_layers,
+        )
+
+        # set initial heights
+        self.canopy_node_heights = [
+            (x + 0.5) / self.canopy_layers * self.canopy_height
+            for x in range(0, self.canopy_layers)
+        ]
+        self.soil_node_depths = [
```
jacobcook1995, line 236: Think it might be sensible to make the more complex interpolations separate functions as that would make testing and potential future changes the interpolation method easier. Probably not necessary for the ones that are just straightforward linear interpolations though

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def calc_molar_density_air(
+        self,
+        temperature: NDArray[np.float32] = np.array(20.0, dtype=float),
+        atmospheric_pressure: NDArray[np.float32] = np.array(101.3, dtype=float),
+    ) -> NDArray[np.float32]:
+        """Calculate temperature-dependent molar density of air.
+
+        Args:
+            temperature: NDArray[np.float32], temperature [C]
```
jacobcook1995, line 265: Here again I'm not sure you really need to include the types as they are already in the function signature

jacobcook1995 (changes requested):
Looks generally sensible to me! In terms of comments the majority of them are basically nitpicks about docstrings. On a slightly higher level I also wonder if some of the functionality of the class `__init__` could be split out into separate functions? It might be better practice to keep everything in one function for the initialisation, but feel that once you start doing non-linear interpolations that will make the `__init__` a very complex and hard to test function. So could see the value in splitting this out

jacobcook1995:
I think the things you have in the `__init__` are pretty sensible, I don't see that any of it could be handled by the `json_schema` as we are really only using that the validate that the relevant input data exists and is of the right type (and if it doesn't exist to provide default values). So, the actual conversion of input values and data to a usable model setup has to happen elsewhere.

If the interpolations that `__init__` performed are specific to the `EnergyBalance` class then I think this is a sensible place for them to happen. If they are relevant to the whole abiotic module I think they probably should take place when the `AbioticModel` class is instantiated, but I think setting that up now would overcomplicate the pull request. If it needs to be changed the line I think it would be pretty straightforward to do and just involve cutting and pasting code

virtual_rainforest/models/abiotic/energy_balance.py
davidorme: Probably worth doing for all of them.

virtual_rainforest/models/abiotic/energy_balance.py
davidorme: Yup. `sphinx` should pick up the types from the signatures, so they get inserted in the API docs without needing to repeat them here.

davidorme:
> I don't see that any of it could be handled by the `json_schema`

I'm not sure about that - the number of soil layers should be constant. Although on reflection, some of this is going to be better provided in data. So `soil_depth` will be spatially variable and a pain to put in TOML, so we might need `data` in the `__init__` arguments and require `soil_depth` in there.

virtual_rainforest/models/abiotic/energy_balance.py
alexdewar: I'd personally *not* include the attribute names here, though it isn't exactly wrong and I haven't checked what style you guys are using elsewhere.

Instead, just add a docstring after the attributes when you create them in `__init__()`, e.g.:

```py
self.my_attr = 5.0
"""A very important parameter between 0.0 and 10.0"""
```

Again, this way your IDE and documentation generator can pick this up.

jacobcook1995:
In terms of the soil layers the `__init__` here doesn't do any of the checking that the `json schema` would perform, i.e. that the number of layers is a positive integer. It might be worth moving some of the soil depth stuff, either to `Data` or to `AbioticModel`, but I don't think it can be moved into the configuration

davidorme:
This is probably something for that `core` discussion but IIUC we want a constant number of soil layers across the simulation. So configuration would require the users to set `n_soil_layers` to be a constant integer value and then `Data` can check that anything that claims to have soil layer data has that number of layers - it would be a `soil_layers` core axis.

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+    """
+
+    def __init__(
+        self,
+        data: dict[str, float],
+        soil_type: NDArray[np.string_],
+        soil_depth: NDArray[np.float32] = np.array(2.0, dtype=float),
+        mean_annual_temperature: NDArray[np.float32] = np.array(20, dtype=float),
+        canopy_layers: NDArray[np.int32] = np.array(3, dtype=int),
+        soil_layers: NDArray[np.int32] = np.array(2, dtype=int),
```
alexdewar, line 159: Are `canopy_layers` and `soil_layers` supposed to be arrays? You seem to be using them as scalars below and I don't think the code will work if actual arrays are passed in.

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+            self.temperature_above_canopy,
+            self.canopy_layers + self.soil_layers,
+        )
+
+        self.air_temperature = temperature_interpolation[
+            (int(self.soil_layers)) : int((self.canopy_layers + self.soil_layers))
+        ]
+        self.soil_temperature = temperature_interpolation[0 : int(self.soil_layers)][
+            ::-1
+        ]
```
alexdewar, line 195: What's the `[::-1]` on the end for? Is it because you want to cast to a `np.array` or something?

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+            np.array(MIN_LEAF_AIR_CONDUCTIVITY),
+            np.array(MAX_LEAF_AIR_CONDUCTIVITY),
+            self.canopy_layers,
+        )
+
+        # set initial heights
+        self.canopy_node_heights = [
+            (x + 0.5) / self.canopy_layers * self.canopy_height
+            for x in range(0, self.canopy_layers)
+        ]
```
alexdewar, line 235: Two things here:

1. Just checking: Is the order of operations here definitely correct? You don't mean `(x + 0.5) / (self.canopy_layers * self.canopy_height)`?
2. This could be written more simply using numpy:

```py
self.canopy_node_heights = (np.arange(self.canopy_layers) + 0.5) / self.canopy_layers * self.canopy_height
```

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+            self.soil_depth
+            / (self.soil_layers**SOIL_DIVISION_FACTOR)
+            * (x**SOIL_DIVISION_FACTOR)
+            for x in range(1, self.soil_layers + 1)
+        ]
+        self.height_of_above_canopy = self.canopy_height
+        self.canopy_wind_speed = [
+            (x / self.canopy_layers) * self.wind_speed_10m
+            for x in range(1, self.canopy_layers + 1)
+        ]
```
alexdewar, line 246: Could use `numpy` here. See comment above

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+
+        # set initial fluxes
+        self.sensible_heat_flux = np.array(0.0)
+        self.latent_heat_flux = np.repeat(0.0, self.canopy_layers)
```
alexdewar, line 250:
```suggestion
        self.latent_heat_flux = np.zeros(self.canopy_layers)
```

virtual_rainforest/models/abiotic/energy_balance.py
alexdewar: Does `soil_type` need to be an array either?

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+        self.ground_heat_flux = np.array(0.0)
+        self.diabatic_correction_factor = np.array(0.0)
+
+        # get soil parameters
+        self.soil_parameters = soil_parameters[str(soil_type)]
```
alexdewar, line 255: Again, I think this will break if `soil_type` is an array

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def calc_molar_density_air(
+        self,
+        temperature: NDArray[np.float32] = np.array(20.0, dtype=float),
+        atmospheric_pressure: NDArray[np.float32] = np.array(101.3, dtype=float),
```
alexdewar, line 260: If you use `dtype=float`, I think that's the same as `np.float64`, which is not what you've put in the type hint. I'd just use `np.float32` everywhere.

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+)
+AIR_CONDUCTIVITY = 50.0  # typical for decidious forest with wind above canopy 2 m/s
+MIN_LEAF_AIR_CONDUCTIVITY = (
+    0.13  # typical for decidious forest with wind above canopy 2 m/s
+)
+MAX_LEAF_AIR_CONDUCTIVITY = (
+    0.19  # typical for decidious forest with wind above canopy 2 m/s
+)
+KARMANS_CONSTANT = 0.4  # constant to calculate mixing length
+CELCIUS_TO_KELVIN = 273.15  # calculate absolute temperature in Kelvin
```
alexdewar, line 71: Spelling: should be "Celsius"

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Reference:
+            Maclean et al, 2021: Microclimc: A mechanistic model of above, below and
+            within-canopy microclimate. Ecological Modelling Volume 451, 109567.
+            https://doi.org/10.1016/j.ecolmodel.2021.109567.
+
+        TODO identify remaining constants
+        """
+        latent_heat_of_vaporisation = -42.575 * temperature + 44994
+        return latent_heat_of_vaporisation
```
alexdewar, line 321:
```suggestion
        return -42.575 * temperature + 44994
```

davidorme:
> It's unusual (but not unheard of) to have PRs submitted against branches other than the main one.

@alexdewar This follows on from @dalonsoa's comments on #126 . One way to build up newer more complex functionality is by using PRs onto a feature branch. That allows us to review code of something larger in reasonably sized PRs - once it is assembled and well integrated, it is then much easier to move into `develop`.

alexdewar:
Ah, ok. As long as it's by design.

I'd suggest merging `develop` into `abiotic_dummy` to make the diff cleaner though.

davidorme:
Ah - that's a great hint. I'd been wondering how to avoid that issue and hadn't realised that if you just merge both sides of the PR onto the same point on develop then you're back to the PR having _only_ the changes unique to the PR. Clever!

virtual_rainforest/models/abiotic/energy_balance.py
```diff
@@ -0,0 +1,321 @@
...
+        self.soil_layers = soil_layers
+        self.canopy_height = canopy_height
+        self.soil_depth = soil_depth
+
+        self.air_temperature_2m = data["air_temperature_2m"]
+        self.relative_humidity_2m = data["relative_humidity_2m"]
+        self.atmospheric_pressure_2m = data["atmospheric_pressure_2m"]
+        self.wind_speed_10m = data["wind_speed_10m"]
+        self.mean_annual_temperature = mean_annual_temperature
+        self.soil_moisture_reference = data["soil_moisture_reference"]
```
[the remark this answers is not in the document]
vgro: good point, I can access the data directly. My thought was that I want the data for the current time step, so I would need to access that somehow. That means I first need to separate ini from the timesteps like you and others suggested.

virtual_rainforest/models/abiotic/energy_balance.py
vgro: good idea. I moved the interpolation function out and made it extendable for other interpolation methods.

virtual_rainforest/models/abiotic/energy_balance.py
vgro: Ok, I'll remove the types from the doc strings. I think it resulted from an earlier comment that suggested copying the args to the docstring. I'm happy to keep it shorter, I didn't realize that this is picked up already.

virtual_rainforest/models/abiotic/energy_balance.py
vgro: I'll put a more detailed description of the processes in the docstring or a separate documentation

virtual_rainforest/models/abiotic/energy_balance.py
alexdewar: As I think I mentioned before, `self.soil_layers` should probably just be of type `int`. There's no point its being a `np.array` if it's a scalar.

virtual_rainforest/models/abiotic/energy_balance.py
vgro: The initial idea was that canopy layers could be different across cells, but as it will be the same everywhere, yes, it should be a scalar. Soil type differs between grid cells

vgro:
major structural changes needed and many changes in develop, will repoen with smaller requests and link to discussion here

PR #146, opened 2023-01-23
Unpinning CI python versions as no longer needed

jacobcook1995:
# Description

Github actions python versions for 3.9 and 3.10 now default to 3.9.16 and 3.10.9, respectively. This means there's no longer any reason to keep the versions pinned, so I have unpinned them. (They were originally pinned to fix a weird CI edge case where a specific test would fail on macOS for python versions <3.9.16 and <3.10.9)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         # Pinned to fix problem with macosx and older python versions (change when fixed)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         # Pinned to fix problem with macosx and older python versions (change when fixed)
```
davidorme, line 22: Remove comment.

davidorme (approved):
Sounds good! Might as well remove out of date comment too

PR #152, opened 2023-01-25
How to set up new model documentation

### LINKED ISSUE #148: Document how to setup a new model
At the moment the API documentation explains what the `core.model` class is, and gives an example model (in `models.soil.model`), but doesn't explain how to actually setup a new model. This is problematic as there are a few key details buried in `__init__.py` files that need to be right for the model to be correctly picked up by the core framework.

A "How to guide" should be started that describes how a developer would add a new model to the `virtual_rainforest` package. I think with time this will probably grow to be something more user focused, but at the moment this will be very much aimed at the package developers.

Don't think this fits any of the other Diátaxis categories (Tutorials, Explanation or Reference).

jacobcook1995:
# Description

This is my first attempt at a how to doc for setting up a new model, let me know if it makes sense and more importantly whether it actually covers all the necessary steps.

Fixes #148 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -174,3 +174,48 @@ def spinup(self) -> None:
 def cleanup(self) -> None:
     """Placeholder function for freshwater model cleanup."""
 ```
+
+## Including a configuration schema
+
+A detailed description of the configuration system can be found
+[here](../virtual_rainforest/core/config.md). The key thing to note is that a `JSON`
+schema file should be saved within your model folder. This file should have a name of
+the format "{MODEL_NAME}_schema.json". In order for this schema to be generally
+accessible, it needs to be registered in the model's `__init__.py` (i.e. the
+`__init__.py` in the model folder). This means that when the model is imported, it's
+schema is automatically added to the schema registry.
+
+```python
+from virtual_rainforest.core.config import register_schema
+
+@register_schema("freshwater")
+def schema() -> dict:
+    """Defines the schema that the freshwater module configuration should conform to."""
+
+    schema_file = Path(__file__).parent.resolve() / "freshwater_schema.json"
+
+    with schema_file.open() as f:
+        config_schema = json.load(f)
+
+    return config_schema
+```
+
+## Ensuring that schema and models are always added to the registry
+
+At the moment,a configuration schema only get added to the schema registry when the
+model it belongs to is imported, and a `Model` class only gets added to the registry
+when the class itself is imported. This is a problem because the script that runs the
+main Virtual Rainforest simulation does not import these things directly. To circumvent
+this these imports needed to be placed in the top level `__init__.py` file (the one in
+the same folder as `main.py`). This won't pass the `pre-commit` checks unless `flake8`
+checks are turned off for the relevant lines. It's only strictly necessary to import the
+`Model` class, as this implicitly entails importing the specific model as a whole.
+However, for the sake of clarity we currently include both imports.
+
+```python
+# Import all module schema here to ensure that they are added to the registry
+from virtual_rainforest.models.freshwater import schema  # noqa
+
+# Import models here so that they also end up in the registry
+from virtual_rainforest.models.freshwater.model import FreshWaterModel  # noqa
+```
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -1,19 +1,3 @@
----
-jupytext:
-  cell_metadata_filter: -all
-  formats: md:myst
-  main_language: python
-  text_representation:
-    extension: .md
-    format_name: myst
-    format_version: 0.13
-    jupytext_version: 1.13.8
-kernelspec:
-  display_name: vr_python3
-  language: python
-  name: vr_python3
----
-
 # How to add a new model to the Virtual Rainforest
 
 The Virtual Rainforest is designed to be modular, meaning that the set of models to be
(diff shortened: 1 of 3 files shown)


docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,221 @@
...
+mkdir virtual_rainforest/models/freshwater
+```
+
+Within this folder a `python` script defining the model should be created.
+
+```bash
+touch virtual_rainforest/models/freshwater/model.py
+```
+
+This script must import the [`BaseModel` class](../api/core/model.md) as new class must
```
jacobcook1995, line 28: Just to mention I am aware that this link is broken, it links to a page that exists in the `feature/data_system` branch but not in `develop`, so once that is merged the link won't be broken any longer.

docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,221 @@
...
+inherit from this abstract base class.
+
+```python
+from virtual_rainforest.core.model import BaseModel
+```
+
+The new model class is created using a class method, this means that a model name must
+be provided when class inheritance occurs, so that the model can automatically be added
+to the registry under that name. This name does not strictly have to match the
+`self.name` attribute, but it generally should.
```
davidorme, line 38: Maybe the registry should extract `cls.name` rather than having duplicate locations - there is no reason why these shouldn't be identical?

jacobcook1995: Yeah relatively easy fix. I've added the change to the pull request

davidorme: Honestly, I'd do it the other way around. So rather than synchronising two locations:

```python
    @classmethod
    def __init_subclass__(cls):
        """Method that adds new model classes to the model registry."""

        if getattr(cls, 'name') is None:
             LOGGER.critical("Model is missing the model name class attribute")
        if not isinstance(cls.name, str)
             LOGGER.critical("Model name is not a string")

        # Add the grid type to the registry
        if cls.name in MODEL_REGISTRY:
            LOGGER.warning(
                "Model with name %s already exists and is being replaced", cls.name 
            )
        MODEL_REGISTRY[cls.name ] = cls
```

davidorme: I think there's a way of decorating a class attribute so that is part of the ABC and hence cannot be left empty. Still might have bogus content though.

davidorme: And maybe use `f"The 'name' class attribute is not set in model class {cls.__name__}"`

jacobcook1995: I kind of prefer supplying a model name though, think it makes it clearer that a something other than normal class inheritance is being used.

jacobcook1995: I'm only keeping the `self.name` in `BaseModel` so that the base model unit tests don't break, could just remove that name entirely, and rewrite the unit tests a bit

docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,216 @@
+# How to add a new model to the Virtual Rainforest
+
+The Virtual Rainforest is designed to be modular, meaning that the set of models to be
+used in a particular run is configurable at the start of the simulation. We are starting
+out by defining a core set of models (`abiotic`, `animals`, `plants` and `soil`), which
+will generally all be included for the vast majority of simulations. In future, it might
```
vgro, line 6:
```suggestion
will generally all be included for the vast majority of simulations. In the future, it might
```

docs/source/development/defining_new_models.md
davidorme: The problem though, is that they should be identical, and I think it is _much_ stranger to have a class attribute written in the code that is then replaced with another value.

docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,216 @@
...
+be desirable to include models for other aspects of rainforests (e.g. `freshwater`), or
```
vgro, line 7:
```suggestion
be desirable to include models for other aspects of rainforests (e.g. `freshwater`, `disturbance`), or
```

vgro: I think it might be worth mentioning disturbance as this will be one of the major applications

docs/source/development/defining_new_models.md
davidorme: I think `cls.name` is important. When the model is created and running, that is the only place where the model name is stored and hence can be used as `self.name` in logger calls etc. And that name is _fundamentally_ an attribute of the class.

docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,216 @@
...
+This script must import the [`BaseModel` class](../api/core/model.md) as new class must
+inherit from this abstract base class.
+
+```python
+from virtual_rainforest.core.model import BaseModel
+```
+
+The new model class is created using a class method, this means that a model name must
+be provided when class inheritance occurs, so that the model can automatically be added
+to the registry under that name. N.B. The `self.name` attribute for the class is
```
vgro, line 37: What does N.B. mean?

docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,216 @@
...
+@classmethod
+def from_config(cls, config: dict[str, Any]) -> FreshWaterModel:
+    """Factory function to initialise the freshwater model from configuration.
+
+    This function unpacks the relevant information from the configuration file, and
+    then uses it to initialise the model. If any information from the config is
+    invalid rather than returning an initialised model instance an error is raised.
+
+    Args:
+        config: The complete (and validated) virtual rainforest configuration.
```
vgro, line 101: General comment/question: do we capitalize 'Virtual Rainforest' or 'virtual rainforest'

docs/source/development/defining_new_models.md
```diff
@@ -0,0 +1,216 @@
...
+
+    with schema_file.open() as f:
+        config_schema = json.load(f)
+
+    return config_schema
+```
+
+## Ensuring that schema and models are always added to the registry
+
+At the moment,a configuration schema only get added to the schema registry when the
```
vgro, line 200:
```suggestion
At the moment, a configuration schema only gets added to the schema registry when the
```

docs/source/development/defining_new_models.md
davidorme: Also it isn't great to have a class attribute that doesn't appear in the code with a docstring. So, it should be there in the subclass definitions.

vgro (commented):
This is a very clear step by step description, very easy to follow. I'll try it out and let you know if something doesn't work or is missing. I made a few minor comments.

vgro:
Could you please describe all the imports that are necessary at the start, for example the numpy stuff and the required logger component? I found it in the sol example but not here in the documentation. Maybe a complete example file would be good.

docs/source/development/defining_new_models.md
davidorme: I think "Virtual Rainforest" in text? @robewers01

docs/source/development/defining_new_models.md
jacobcook1995: I meant it to be nota bene, but realise that you are supposed to either abbreviate that as "n.b." or "NB". Looking at it again though, I think it's a bad way to start a sentence so I will switch it to "It should be noted" or similar

docs/source/development/defining_new_models.md
jacobcook1995: I've been capitalising it generally (though I didn't here 😅), but agree that should probably settle on an agreed style

docs/source/development/defining_new_models.md
jacobcook1995: I think I might be misunderstanding something here. Are `cls.name` and `self.name` different things?

jacobcook1995: I was imagining that `cls.name` was `self.name` for the class in question

davidorme: They do refer to the same thing, but in the context of the `class` method, which uses `cls` as an argument to indicate that it is referring to the properties of the class itself, rather than `self.name` referring to the properties of an _instance_ of the class. The point of class attributes specifically is that property value is set across all possible instances of the class.

davidorme: But in defining a subclass, which has a class attribute `name` (maybe we should using `model_name`), I think that attribute should be set and docstring in the model code, rather than set via an argument to `__init_subclass`. It won't appear in the class attributes for example. Given that needs to happen, it makes sense to use it to register the class?

jacobcook1995: So the issue with the current way I'm doing it is that only class instances produced using `__init_subclass__` would end with a name populated?

jacobcook1995: If someone called `SoilModel(**kwargs)`, it would create an instance without a name?

jacobcook1995: Or at the moment I think it would be called "base", but was planning on deleting that

davidorme: The use of `class ModelName(BaseModel)` _guarantees_ that `BaseModel.__init_subclass__` will be called, so `cls.name` would be populated on any BaseModel subclass. **But** that is an obscure way to populate a class attribute and those are typically included and documented as:

```python
class ModelName(BaseModel):

     model_name = "ModelName"
     """The model name for use in registering the model and logging"
``

Comment (davidorme) [reply docs/source/development/defining_new_models.md]: No - the name is _only_ set in the creation of the class itself. So your `soil.model` object _creates_ the `SoilModel` object, and at that point the subclass itself (not a class instance) is registered using `__init_subclass__`.

If someone creates an _instance_ using `SoilModel(**kwargs)`, then that creates an instance, with the `name` class attribute populated when the package was loaded. The difference here is really where it is clearer to set that attribute, and I think the direct option above is better for clarity and having attributes appear in rendered docstrings.

Comment (jacobcook1995) [reply docs/source/development/defining_new_models.md]: Happy to switch it around if you think that makes sense! My only hesitation is that it feels like a fairly big functional change to the code (which would require new tests, etc). Wonder if that is sensible to be adding to a pull request about docs, or whether I should make a new issue + PR for it?

Comment (davidorme) [reply docs/source/development/defining_new_models.md]: @dalonsoa @alexdewar - Thoughts? It's not a large change but does need code change in the BaseModel and tests. Seems like a small PR to me. And also - am I talking nonsense?

File: docs/source/development/defining_new_models.md [hunk 8]
```diff
@@ -0,0 +1,233 @@
...
+
+This script must import a number of things to be able to set up a new `Model` class
+correctly. It's worth noting that `log_and_raise` will be removed soon.
+
+```python
+# One of the member functions of the Model class returns a class instance. mypy doesn't
+# know how to handle this unless annotations are imported from __future__
+from __future__ import annotations
+
+# Any needed for type hints of the config dictionary as the values are of various types
```
davidorme, line 36: I believe you can simply quote the return name to avoid the `mypy` error too. We should do this systematically though.

jacobcook1995: Didn't realise that worked! Prefer this way, partly because its a little neater, partly because I like importing from `__future__`

TaranRallings (approved):

vgro (approved):
Documentation looks good to me, the technical details I cannot judge at the moment.

docs/source/development/defining_new_models.md
alexdewar: @davidorme Sounds sensible to me.

I would do it more simply though (I didn't know `__init_subclass__` was a thing till now). I'd just have a check in `BaseModel`'s `__init__` along the lines of:

```py
if not hasattr(self, "name"):
    raise ValueError("Models must have a name attribute")
````

alexdewar (commented):
Sorry, added myself as a reviewer by mistake. Looks good though :laughing:

davidorme:
Oddly can't reply direct to that comment @alexdewar. Don't know why!

We need `__init_subclass__` (it was new to me too!) to register the model when the package loads anyway. Does it make makes sense to keep the tests here to make it fall over before before someone actually tries to make an instance of a nameless model, hence calling `super().__init__()`

jacobcook1995:
I've moved where the `model_name`'s get set to be defined in each model and then pulled from there to be used as a registry name. I put the checks in the `__init_subclass__` rather than the `__init__` so that invalid models cannot be added to the registry.

I also had to bump the version of `isort` as the `pre-commit` check had started failing (on Github) with the old version. I updated the minimum value in `pyproject.toml` to match this + generated a new poetry lock

davidorme:
I've just PR'd that patch on `isort` on `develop` to merge in #122 . Shouldn't be an issue.

PR #153, opened 2023-01-26
First draft of supplementary classes for dummy_animal_module

TaranRallings:
# Description

This contains the supplementary classes for the dummy animal module, accompanying tests, and the animals_schema.json. These are classes for plants, soils, and carcasses used for testing functions of the animal class. The plant class contains simple functions for growth (increasing energy storage) and death.

These supplementary classes are being uploaded first because the animals class contains references to them and I needed to break up the work I'd done into smaller chunks for reviewing.

Fixes # (issue)

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Key checklist

- [X] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`

## Further checks

- [X] Code is commented, particularly in hard-to-understand areas
- [X] Tests added that prove fix is effective or that feature works


alexdewar:
@TaranRallings I've added Diego as a reviewer too, in case he wants to add his two cents.

I'll take a look now

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+
+class CarcassPool:
+    """This is a class of carcass pools."""
+
+    def __init__(self, energy: float, position: int) -> None:
+        """The constructor for Carcass class."""
+        self.energy = energy
+        """The amount of energy in the carcass pool [j]."""
+        self.position = position
+        """The grid position of carcass pool pool [0-8]."""
```
davidorme, line 93: Is it `[0-8]` because your toy example is using a 3x3 grid? I wouldn't tie down the input values here even in the toy.

More generally, you're going to want a `CarcassPool` for each grid cell, so I would lean towards having a list of them `carcasses = [<CarcassPool>, <CarcassPool>, ..., <CarcassPool>]`. If those are in `cell_id` order then you can get the pool for cell_id `n` by getting `carcasses[n]`. That removes the need to encode `position` in the `CarcassPool` object. 

You could do it anyway, for information, but I wouldn't use it as a mechanism for finding the right one for a cell.

davidorme: Compare:

```python
for cell_index in cell_ids:
    this_carcass = carcasses[cell_index]
```

```python
for cell_index in cell_ids:
    this_carcass = [cp for cp in carcasses if cp.position == cell_index][0] 
```

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, name: str, mass: float, position: int) -> None:
+        """The constructor for Plant class."""
+        self.name = name
+        """The functional-type name of the plant cohort."""
+        self.mass = mass
+        """The mass of the plant cohort [g]."""
+        self.energy = mass * 100
+        """The amount of energy in the plant cohort [j] [toy]."""
+        self.alive = "alive"
+        """Whether the cohort is alive [True] or dead [False]."""
```
alexdewar, line 41: I'd make `self.alive` a boolean, as you suggest it is in the docstring. I'd also rename it to `is_alive` just to be extra explicit.

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+            None: Toy implementation of death is only a function of the
+                    current aliveness state.
+
+        Returns:
+            Modified value of self.alive.
+            An alert (str) informing you the cohort has died.
+        """
+        self.alive = "dead"
+        return f"""A {self.name} cohort died"""
+
```
davidorme, line 72: I get that this is debugging a toy example, but you don't really want functions returning things that aren't going to be used. A better way to do this is to use the logger:

At the top of the file:
```python
```

And then you can do:
```python
  def plant_death(self) -> None:
        """The function to change the self.alive state from True to False.
        Parameters:
            None: Toy implementation of death is only a function of the
                    current aliveness state.
        Returns:
            Modified value of self.alive.
            An alert (str) informing you the cohort has died.
        """
        self.alive = "dead"
        LOGGER.debug(f"A {self.name} cohort died")
```

No return statement is needed (it is implicitly going to `return None`)

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+        self.mass = mass
+        """The mass of the plant cohort [g]."""
+        self.energy = mass * 100
+        """The amount of energy in the plant cohort [j] [toy]."""
+        self.alive = "alive"
+        """Whether the cohort is alive [True] or dead [False]."""
+        self.energy_max = mass * 100
+        """The maximum amount of energy that the cohort can have [j] [toy]."""
+        self.position = position
+        """The grid location of the cohort [0-8]."""
```
davidorme, line 45: See the note about the `CarcassPool`. Here you are possibly going to want multiple plants per cell so `all_the_plants` could be a list of length `n_cells` with entries that are a list of the available stems.

tests/test_dummy_animal_module.py
```diff
@@ -0,0 +1,61 @@
...
+import virtual_rainforest.animals.dummy_animal_module as am
+
+
+class TestPlant:
+    """Test Plant class."""
+
+    def test_plant_growth0(self):
+        """Testing plant_growth at 100% energy."""
+        p = am.Plant("tree", 10.0, 1)
+        print("testing plant growth : plant")
```
davidorme, line 20: I wouldn't put `print` statements in tests or use LOGGER in them. When they fail `pytest` prints information on why they failed, but otherwise they should generally be silent.

tests/test_dummy_animal_module.py
```diff
@@ -0,0 +1,61 @@
...
+        p.plant_growth()
+        assert p.energy == 1000
+
+    def test_plant_growth1(self):
+        """Testing plant_growth at 100% energy."""
+        p = am.Plant("tree", 10.0, 1)
+        print("testing plant growth : plant")
+        p.plant_growth()
+        assert p.energy == 1000
```
davidorme, line 29: duplicated?

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def plant_growth(self) -> None:
```
alexdewar, line 47: A good rule of thumb is to start your function names with a verb, because they should be doing something. Given that it's a member of the `Plant` class, the `plant_` prefix is redundant too. I'd go for something like `grow()` or `increase_energy()`.

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+
+        Parameters:
+            None: Toy implementation of death is only a function of the
+                    current aliveness state.
+
+        Returns:
+            Modified value of self.alive.
+            An alert (str) informing you the cohort has died.
+        """
+        self.alive = "dead"
```
alexdewar, line 70: As above, I'd use booleans for this (i.e. set its value to `False`).

davidorme (changes requested):
Looks good to me. Some minor stuff to change.

Another thing I wondered: do you really need realistic plants that die or are indeed finite as a placeholder. There's nothing wrong with the placeholder you have, but you could just have:

```python
class Plants():
 
     def __init__(self):
         self.energy: float = np.inf

     def herbivory(self, energy_consumed: float):
          self.energy -= energy_consumed
```

You could have one instance of that per cell. Obviously the energetics are not remotely sane but you have a simpler placeholder to keep everything from starving?

virtual_rainforest/animals/dummy_animal_module.py
alexdewar, line 93: Unless you're planning to add more complicated logic to these `__init__`s, I'd change these to dataclasses: https://docs.python.org/3/library/dataclasses.html

It's just a slightly terser way of expressing simple classes like these.

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+            None: Toy implementation of growth is only a function of the
+                  current energy state.
+
+        Returns:
+            Modified value of self.energy
+        """
+        self.energy += 1 * self.energy * (1 - self.energy / self.energy_max)
+
+    def plant_death(self) -> str:
+        """The function to change the self.alive state from True to False.
```
alexdewar, line 60: Assume that the user doesn't know anything about the internals of your class. Say what the function does in "real" terms, e.g. "Kill the entire plant cohort instantly".

tests/test_dummy_animal_module.py
```diff
@@ -0,0 +1,61 @@
...
+        p.energy = 0
+        p.plant_growth()
+        assert p.energy == 0
+
+    def test_plant_death(self):
+        """Testing plant_death."""
+        p = am.Plant("tree", 10.0, 1)
+        print("testing plant growth : plant")
+        p.plant_death()
+        assert p.alive == "dead"
```
alexdewar, line 52: I'd also check that the plants are alive before calling `plant_death()`.

tests/test_dummy_animal_module.py
```diff
@@ -0,0 +1,61 @@
...
+        p.plant_growth()
+        assert p.energy == 750
+
+    def test_plant_growth3(self):
+        """Testing plant_growth at 0% energy."""
+        p = am.Plant("tree", 10.0, 1)
+        print("testing plant growth : plant")
+        p.energy = 0
+        p.plant_growth()
+        assert p.energy == 0
```
alexdewar, line 45: Instead of the separate `test_plant_growth*` functions, I'd use parameterised testing, as e.g. @jacobcook1995 has done here: https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/develop/tests/test_main.py

That way you can test lots of different input/output combinations succinctly.

alexdewar (changes requested):
Hi @TaranRallings.

Looking good! I've made various small suggestions, but I think the general structure seems sensible. And it's good that you submitted it with placeholders in for now. That way we can hash out the structure before other elements are added.

As a general comment, as you include dummy models for soil and plant bits, I would add the soil and plant leads as reviewers for this PR. (I'm afraid I've forgotten who they are -- sorry! -- and I'm guessing there's no "carcass lead" :wink: ?) No one *has* to review a PR -- and if they do, they can just review the bits relevant to their expertise -- but by adding someone as a reviewer you're giving them a chance to weigh in if they want to. I guess in general you guys will want to have e.g. the plant lead as a reviewer on all PRs that touch on plant stuff. You can also have code owners, which is something you guys might want to look at longer term: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners

It would also be worth asking the plant and soil leads whether it would work better if your dummy models should live in "their" submodules along with their code or not.

virtual_rainforest/animals/dummy_animal_module.py
alexdewar, line 93: Typo:

```suggestion
        """The grid position of the carcass pool [0-8]."""
```

Unless you're planning to add more complicated logic to these `__init__`s, I'd change these to dataclasses: https://docs.python.org/3/library/dataclasses.html

It's just a slightly terser way of expressing simple classes like these.

tests/test_dummy_animal_module.py
davidorme: The other think that is useful (but only marginally shorter here) is to use a `pytest.feature` for the repeated input:

```python
@pytest.feature
def plant_instance():
    from virtual_rainforest.animals.dummy_animal_module import Plant
    return am.Plant("tree", 10.0, 1)
```

And then tests can share that plant:

```python
def test_plant_growth3(plant_instance):
        """Testing plant_growth at 0% energy."""
        print("testing plant growth : plant")
        plant_instance.energy = 0
        plant_instance.plant_growth()
        assert plant_instance.energy == 0
```

davidorme: Also note that style of importing the `vr` functionality inside the test itself. That makes the ability to import the bit you want part of the test and an import fail only breaks the test that uses that functionality rather than the whole file.

davidorme:
Also note that we're moving to a structure of having a `models` folder, so this should be in `virtual_rainforest/models/animals`

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+# - only iteroparity (want: semelparity)
+# - no development
+
+# notes to self
+# assume each grid = 1 km2
+# assume each tick = 1 day (28800s)
+# damuth ~ 4.23*mass**(-3/4) indiv / km2
+# waste_energy pool likely unnecessary
+#   better to excrete directly to external pools
+# only elephants disperse atm
```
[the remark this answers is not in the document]
dalonsoa: My thought, too. This way, you also document your todo list and inform potential developers of where they can chip in.

tests/test_dummy_animal_module.py
dalonsoa: Agree. Tests should not print or log things directly (they might do so indirectly when testing something, but not include those as part of the code of the test)

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+
+class Plant:
+    """This is a dummy class of plant cohorts for testing the animal module."""
+
+    def __init__(self, name: str, mass: float, position: int) -> None:
+        """The constructor for Plant class."""
+        self.name = name
+        """The functional-type name of the plant cohort."""
+        self.mass = mass
+        """The mass of the plant cohort [g]."""
```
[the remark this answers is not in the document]
dalonsoa: Probably it will be a good idea to use a more robust approach for units, using Pint as done in the `soil` model: https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/1df932dec85044aa562757ee8a01d1e560e95f63/virtual_rainforest/soil/model.py#L90

dalonsoa (changes requested):
Great job! Placeholders and dummy functions are a great starting point to put your thoughts in order and figure out what you need to do next. 

I've left a couple of comments, but mostly I've just given 👍 to @alexdewar and @davidorme comments, which were already very thorough.

virtual_rainforest/animals/dummy_animal_module.py
TaranRallings: I was using 0-8 for a toy 3x3 grid. I intended that [0-8] comment as a note-to-self but I need to get out of the habit of thinking about comments that way. 

I have a list of CarcassPool in the, yet unpulled, AnimalModel class but I haven't been using the ordering for anything yet. I will try to build towards using that ordered structure for tracking.

 I would like to keep a position attribute on everything for now so I can keep a close eye on how dispersal and positional generation are working (or not working). It's less of an issue for CarcassPool but I anticipate things getting hairy for AnimalCohorts and I'd like to keep the tracking consistent. Does that make sense?

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: I think you're right, at this point Plant could get the same treatment as CarcassPool. I can build out a plant growth system when I need to test starvation patterns over time. Now that we're doing module links earlier, it might not even be necessary. I'll ditch the extras on Plants.

tests/test_dummy_animal_module.py
TaranRallings: Good catch!

tests/test_dummy_animal_module.py
TaranRallings: Interesting, thanks.

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: This is really interesting, thanks for mentioning it. I'm so used to thinking about comments and to-do's as comments to myself that I didn't even consider it.

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: On further consideration, I'm going to keep some of the extra plant stuff because I want the practice building in a bunch of these suggestions. I'll try to trim it down though.

virtual_rainforest/animals/dummy_animal_module.py
davidorme: Yup. From plants, it makes absolute sense to have a single `PlantCommunity` object assigned to each cell (until we do the `Triffid` extension at least). But for animal cohorts, it is probably much easier to have a single cohort with a spatial map of where the individuals are, rather than constantly having to divide that cohort onto the cell-specific impacts. So then `AnimalCohort` definitely needs a structure like:

```
AnimalCohort.spatial = {
    cell_id = [2,3,4,5,6],
    proportion = [0.1, 0.1, 0.6, 0.1, 0.1]
}
```
that records the cells being used by each cohort and the relative impacts in terms of energy extraction (waves hands vaguely) to allow core and peripheral range use.

Or something.

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+# assume each grid = 1 km2
+# assume each tick = 1 day (28800s)
+# damuth ~ 4.23*mass**(-3/4) indiv / km2
+# waste_energy pool likely unnecessary
+#   better to excrete directly to external pools
+# only elephants disperse atm
+
+
+class Plant:
+    """This is a dummy class of plant cohorts for testing the animal module."""
```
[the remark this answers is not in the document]
davidorme: _Probably_ what we will have is `PlantFunctionalType`, `PlantDemography` (PFT specific) and `PlantCommunity` (a list of `PlantDemography` objects for a cell). We will record the size structured abundance of an individual PFT within a cell in `PlantDemography`, but probably won't have a separate class for the size-based cohorts.

So I suspect, you're going to want to think about using `PlantCommunity`.

virtual_rainforest/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,93 @@
...
+        Args:
+            None: Toy implementation of growth is only a function of the
+                  current energy state.
+
+        Returns:
+            Modified value of self.energy
+        """
+        self.energy += 1 * self.energy * (1 - self.energy / self.energy_max)
+
+    def plant_death(self) -> str:
```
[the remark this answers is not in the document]
davidorme: I think `self.die()` is clearer about what is ceasing to be 😄

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: Fixed the boolean point in the function and the test.

virtual_rainforest/animals/dummy_animal_module.py
TaranRallings: I've replaced the printed message with this logger report.

tests/test_dummy_animal_module.py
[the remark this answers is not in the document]
alexdewar: I assumed you had (hope I didn't sound snarky!). I just wanted to double-check that you weren't testing in some roundabout way instead, e.g. by importing these functions into a script and running it, which would explain the `print` statements.

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
alexdewar: It's important to remember that code is more or less always a work in progress and it's fine to acknowledge this in the documentation too :smile:

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
alexdewar: How about `self.kick_the_bucket()`? :wink:

I guess it depends on whether you're thinking in terms of what the code is "doing" to the object or what the object is "doing" to itself. Both `kill()` and `die()` sound a bit off to me tbh :laughing:, but I still think it's better to have a verb in there somewhere!

tests/test_dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: The print statements got in there because I only kinda knew how to set up this sort of testing and was following online tutorials that used them! You didn't sound snarky at all.

virtual_rainforest/animals/dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: I've always worked in grams so its just force of habit. I'm happy to slip over to SI. 

We had a chat about Pint and unit tracking this morning. I think the battle plan is for me to leave it at the moment and then for us to sort out an overall strategy for using Pint in the project and then implement simultaneously. Does that sound reasonable?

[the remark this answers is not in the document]
alexdewar: Sounds reasonable to me!

virtual_rainforest/animals/dummy_animal_module.py
TaranRallings: That makes sense to me. I'll get a basic spatial distribution structure  in there for animal cohorts before I pull it.

tests/test_dummy_animal_module.py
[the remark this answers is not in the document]
TaranRallings: With respect to putting dummy classes in other people's modules, now that we're shifting to module connections I'll bring it up as a early step. Thanks for the suggestion.

virtual_rainforest/animals/dummy_animal_module.py
TaranRallings: Those dataclasses are quite concise, thanks!

tests/test_dummy_animal_module.py
```diff
@@ -3,59 +3,53 @@
...
+    )
+    def test_grow(self, plant_instance, initial, final):
+        """Testing grow at 100%, 50%, and 0% maximum energy."""
+        plant_instance.energy = initial
+        plant_instance.grow()
+        assert plant_instance.energy == final
+
+    def test_die(self, plant_instance):
+        """Testing die."""
+        if plant_instance.is_alive:
```
alexdewar, line 31: Not sure why you have the if statement here... `plant_instance` should be newly created in which case it will be alive. You probably want to `assert plant_instance.is_alive` to check this.

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,77 @@
...
+- only endothermy (want: ectothermy)
+- only iteroparity (want: semelparity)
+- no development
+
+Notes to self:
+- assume each grid = 1 km2
+- assume each tick = 1 day (28800s)
+- damuth ~ 4.23*mass**(-3/4) indiv / km2
+- waste_energy pool likely unnecessary, better to excrete directly to external pools
+"""
```
alexdewar, line 22: Much cleaner!

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,77 @@
...
+        """The constructor for Plant class."""
+        self.mass = mass
+        """The mass of the plant cohort [kg]."""
+        self.energy_density: float = 1000.0
+        """The energy (J) in a kg of plant. [small toy J/kg value for convenience.]"""
+        self.energy_max: float = self.mass * self.energy_density
+        """The maximum amount of energy that the cohort can have [J] [toy]."""
+        self.energy = self.energy_max
+        """The amount of energy in the plant cohort [J] [toy]."""
+        self.is_alive: bool = True
```
alexdewar, line 42: You don't need the `bool` because `mypy` can infer the type from the fact that you're assigning `True` to it. It's fine as is though.

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,77 @@
...
+        """Whether the cohort is alive [True] or dead [False]."""
+        self.position = position
+        """The grid location of the cohort [0-8]."""
```
alexdewar, line 45: Was there a discussion about this being a value between 0 to 8? Is that still the case? If so, you could consider enforcing it with a check.

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,77 @@
...
+        """The function to logistically modify cohort energy to the energy_max value."""
+        self.energy += 1 * self.energy * (1 - self.energy / self.energy_max)
+
+    def die(self) -> None:
+        """The function to kill a plant cohort."""
+        if self.is_alive:
+            self.is_alive = False
+            LOGGER.debug("A Plant Community has died")
+        elif not self.is_alive:
+            LOGGER.debug("A Plant Community which is dead cannot die.")
```
alexdewar, line 57: If you *really* want to forbid this, you could raise an exception if the plant community is already dead. Otherwise I'd say it should probably be a `warning`, rather than a `debug` statement.

alexdewar (changes requested):
I've made a few very minor comments, but otherwise it's all looking good. Good work!

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@dataclass
+class SoilPool:
```
jacobcook1995, line 61: Would it be possible to name this something else, soil pool is a quite commonly used term within the `soil` module, e.g. for different forms of organic matter, inorganic nutrients, etc. Would be good to keep the naming a bit separate here? Could maybe be `PalatableSoil` or something similar?

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -0,0 +1,77 @@
...
+    """This is a dummy class of soil pools for testing the animal module."""
+
+    energy: float
+    """The amount of energy in the soil pool [J]."""
+    position: int
+    """The grid position of the soil pool."""
+
+
+@dataclass
+class CarcassPool:
```
jacobcook1995, line 71: Think it would be good to move this to a separate module. Basically would help to have a shared carcasses module for me to work with when I get round to working on carcass decay stuff. Obviously wouldn't be much in it for the time being but might as well make the separation now

jacobcook1995: Not really bothered about moving the `SoilPool` class to a separate module as I think the changes to that will be much more significant

jacobcook1995 (changes requested):
Looks good to me generally, just a couple of comments about class names and submodules

dalonsoa (approved):
Looks good. I've just added a minor comment and added 👍 to those comments by other that I agree with.

tests/test_dummy_animal_module.py
TaranRallings: In hindsight, it seems obvious that's what you meant!

virtual_rainforest/models/animals/dummy_animal_module.py
TaranRallings: There was a bit of a discussion and I believe the position was that it's better to not hardcode any grid sizes, even in toy implementation. I'll take the [0-8] note out, which was really just a note to myself anyway.

virtual_rainforest/models/animals/dummy_animal_module.py
TaranRallings: Absolutely!

alexdewar (approved):
Looking good. I think this is ready to merge, once the CI is passing.

jacobcook1995 (approved):
LGTM!

davidorme (approved):
LGTM. 🎉

Issue #129, opened 2022-12-07
core.constants.py


vgro:
Different modules will need access to a core set of constants such as density of air, Stephan Bolzman constant,...

Proposed solution: centrally stored dictionary core.constants.py

davidorme:
I think that makes sense as a mini module. Another possibility for the contents - which I used over in `pyrealm` - is to use  `dataclasses.dataclass`. There are a couple of possible advantages:

* You can freeze the contents, so no-one can accidentally edit them.
* It gives you a slightly neater access method (`constants.density_of_air` not `constants['density_of_air']`)

dalonsoa:
The second point is easily covered by by putting all the constants in a module and using it as `from vr.core import constants`. Granted that it does not prevent importing specific constants nor to edit them, but it seems like the most straight forward approach. 

Either way (data class or module) will work fine. My main concern always with universal constants are units. What units they will be defined in? How is that communicated to the users? I tried [a not-so-popular approach](https://github.com/qpv-research-group/solcore6/blob/develop/solcore/constants.py) in a project some time ago using `pint`, but yet it might work here.

davidorme:
Completely agree about care needed on units. There are various python solutions for units such as `pint` and `cfunit` but it does add a layer of complexity - I've also seen pushback against using coded units. We certainly should state them in docs/comments and perhaps we park whether to actually programatically set them for now?

davidorme:
Closing - see #364

PR #155, opened 2023-01-27
Feature/radiation dummy 2

vgro:
# Description

This is a new clean PR for the radiation submodule including all changes that were requested up to last week in PR #130. 

## Type of change

## Key checklist

## Further checks


tests/test_radiation.py
```diff
@@ -0,0 +1,212 @@
...
+    """Simple check for correct numbers, better test to be decided."""
+
+    radiation_fixture.topofcanopy_radiation = toc_radiation
+    radiation_fixture.longwave_canopy = lw_canopy
+    radiation_fixture.longwave_soil = lw_soil
+    radiation_fixture.calc_netradiation_surface(rad_absorbed)
+    assert radiation_fixture.netradiation_surface == pytest.approx(exp_netrad)
+
+
+@pytest.mark.parametrize(
```
jacobcook1995, line 138: I was told on one of my pull requests to add a wider range of values to unit tests, so think you probably should here. As what your `parameterise` is testing here is basically that the functions work for a single float and an array of (the same) float.

jacobcook1995: Obviously depends a bit on how pinned down the parameterisation is, not work sinking loads of time into developing extensive tests for functions that don't have remotely sensible parameters

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
...
+            canopy_absorption: absorption by canopy [J m-2]
+        """
+
+        tau = self.calc_ppfd(shortwave_in, sunshine_hours, albedo_vis)
+        self.calc_topofcanopy_radiation(
+            tau, shortwave_in, sunshine_hours, albedo_shortwave
+        )
+        self.calc_longwave_radiation(canopy_temperature, surface_temperature)
+        self.calc_netradiation_surface(canopy_absorption)
+        del tau
```
jacobcook1995, line 215: Think this was mentioned on a different pull request, but explicit deletion of data is not often necessary/useful in python (this is in contrast to some other languages)

jacobcook1995 (commented):
Generally looks good to me, just had a couple of comments

tests/test_radiation.py
davidorme: It's useful to have a few values covering maybe minimum, median and maximum sensible values? Honestly with float functions you are very unlikely to have one test value stay unchanged when an error is introduced - although it could be at some kind of stable point or limit* - but having a few is a bit safer from the point of view of having test values that act as a canary when the changing the code changes predictions unexpectedly.

* I'm just basically selecting maths words at random here.

virtual_rainforest/models/abiotic/radiation.py
davidorme: They will be garbage collected when the function exits (see #150)

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
+"""The `abiotic.radiation` module.
+
+The radiation module calculates the radiation balance of the Virtual Rainforest.
+The top of canopy net shortwave radiation at a given location depends on
+    1. extra-terrestrial radiation (affected by the earth's orbit, time of year and
+        day, and location on the earth),
+    2. terrestrial radiation (affected by atmospheric composition and clouds),
+    3. topography (elevation, slope and aspect),
+    4. surface albedo (vegetation type and fraction of vegetation/bare soil), and
+    5. emitted longwave radiation.
```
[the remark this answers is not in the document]
davidorme: Also RST is very picky about indentation - the d of day needs to align with the e of extra or this gets interpreted as a definition list and the first line gets rendered in bold...

```suggestion
The top of canopy net shortwave radiation at a given location depends on

1. extra-terrestrial radiation (affected by the earth's orbit, time of year and
   day, and location on the earth),
2. terrestrial radiation (affected by atmospheric composition and clouds),
3. topography (elevation, slope and aspect),
4. surface albedo (vegetation type and fraction of vegetation/bare soil), and
5. emitted longwave radiation.
```

davidorme (changes requested):
This looks really good. I have suggested some structural changes but the core of this is all fine.

One thing I would add - I think when you're adding a new module python file with docstrings, it also makes sense to include the `sphinx` API stub and an index entry. Those are small and mean that the API doc building can be checked along with reviewing the docstrings. I added those locally to build the API docs so let me know if you want me to commit those files to the branch.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
+"""The `abiotic.radiation` module.
+
```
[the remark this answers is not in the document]
jacobcook1995: Yeah I wrote my "how to doc" using that structure, which obviously conflicts with the current state of `develop` but I think is the right direction to be going in

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
...
+        self.netradiation_surface = (
+            self.topofcanopy_radiation
+            - canopy_absorption  # np.sum(canopy_absorption, axis=1)  # what axis?
+            - self.longwave_soil
+            - self.longwave_canopy  # np.sum(self.longwave_canopy, axis=1)
+        )
+
+    def radiation_balance(
+        self,
+        elevation: NDArray[np.float32],
```
davidorme, line 177: This is passed as an argument but is not used as `calc_ppfd` reads it directly from `self`.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
...
+        """Daily top of canopy downward shortwave radiation, [J m-2]"""
+        self.longwave_canopy = NDArray[np.float32]
+        """Daily longwave radiation from n individual canopy layers, [J m-2]"""
+        self.longwave_soil = NDArray[np.float32]
+        """Daily longwave radiation from soil, [J m-2]"""
+        self.netradiation_surface = NDArray[np.float32]
+        """Daily net shortwave radiation at the surface (= forest floor), [J m-2]"""
+
+    def calc_ppfd(
+        self,
```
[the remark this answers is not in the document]
davidorme: Actually, I think this is important. `Radiation` as a class is currently only intended ever to be used as:

```
r = Radiation(elev)
r.radiation_balance(lots_of_args)
```

If you try and use the other methods, it all works _provided_ you run them in the right order and pass `tau` correctly. If `Radiation` is basically a recipe for applying a set of radiation calculations in the right order, then making those `calc_` functions external and only providing `Radiation.radiation_balance` ensures that the class can't be used incorrectly.

I don't know if there is stuff to be added to `Radiation` but it could be that all of the `radiation_balance` code can be moved into `__init__` and then you can just use:

```
r = Radiation(elev, lots_of_args)
```

That doesn't make sense if there are other things to add in though.

davidorme:
@vgro I have pushed those `sphinx` files.

alexdewar:
@vgro Don't hesitate to add me and/or @dalonsoa as reviewers to this new PR if think it'd be useful. We still have plenty of allocated time for you guys!

tests/test_radiation.py
vgro: Good point, it makes sense to include a wider range of values including some non-valid ones to check that it fails. Would it be worth using an advance method for that or could I just extend the parameterize functionality I currently use?

virtual_rainforest/models/abiotic/radiation.py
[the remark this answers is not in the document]
vgro: Yes, I think that is what we agreed on.

tests/test_radiation.py
davidorme: I would probably just extend the parameterisation. This [existing test](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/develop/tests/test_data.py#L305) provides a possible template for adding failing values in - you can add an expected Exception (or expected lack of exception) to the parameter sets. 

I really like the `pytest.param` and then `id=param_set_name` syntax used in that example - the name acts both as a short explanation of what those inputs test and also makes it easier to identify failing tests.

virtual_rainforest/models/abiotic/radiation.py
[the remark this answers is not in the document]
davidorme: Good - `develop` doesn't have that yet but I also added this is a comment on @TaranRallings PR #153 so I don't think we'll get any issues!

virtual_rainforest/models/abiotic/radiation.py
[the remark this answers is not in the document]
vgro: Thanks @davidorme for this suggestion. I agree that this is an important point, something I was trying to get to some time ago when we had the meetings on overall model structure, but I couldn't articulate it very well without the examples to show. If the concept of a class is not appropriate for the microclimate (this might not only affect radiation but other components, too), I'm happy to change it. What do you think @alexdewar  and @dalonsoa ?

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
...
+"""factor to convert temperature in Celsius to absolute temperature in Kelvin"""
+SECOND_TO_DAY = 86400
+"""factor to convert between days and seconds."""
+
+
+class Radiation:
+    """Radiation balance."""
+
+    def __init__(self, elevation: NDArray[np.float32]) -> None:
+        """Initializes point-based radiation method."""
```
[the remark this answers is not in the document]
vgro: Good point. I set `ALBEDO_VIS` and `ALBEDO_SHORTWAVE` to default in the `calc_` function for now and removed them from the constants, let's see what comes out of the discussion on where to put the `clac_` functions, then a solution might become more obvious :-)

virtual_rainforest/models/abiotic/radiation.py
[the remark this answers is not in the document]
davidorme: Just to add that I think a class is right here. It acts as a convenient container for multiple variables calculated from the radiation inputs and you want to keep those packaged together.

It could be that `Radiation` makes most sense as a [`dataclass`](https://docs.python.org/3/library/dataclasses.html). This is a kind of stripped down syntax to create classes that at their simplest just hold data. That would package up all of the `lots_of_args` into one block and then the `__post_init__` method can be used to calculate the rest. That really would be very little different from what you have now, but it is just a cleaner syntax.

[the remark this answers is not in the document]
davidorme: This is a mangled example because I'm not using the right inputs etc... but:

```python
@dataclass
class Radiation():
    elev: NDArray
    other_init_args: NDArray
    ppfd: NDArray = field(init=False)

    def __post_init__(self):
        self.ppfd = calc_ppfd(self.other_init_args)
```

And then you get `Radiation(elev, other_init_args)` as the class signature and `__post_init__` is basically your `radiation_balance` method.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,215 @@
...
+"""Albedo of shortwave radiation"""
+CELSIUS_TO_KELVIN = 273.15
+"""factor to convert temperature in Celsius to absolute temperature in Kelvin"""
+SECOND_TO_DAY = 86400
+"""factor to convert between days and seconds."""
+
+
+class Radiation:
+    """Radiation balance."""
+
```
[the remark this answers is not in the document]
vgro: I agree, I followed the 'I dump everything here for now' approach until we decided on where we put which part of the documentation ;-)

vgro:
> @vgro Don't hesitate to add me and/or @dalonsoa as reviewers to this new PR if think it'd be useful. We still have plenty of allocated time for you guys!

I would appreciate your feedback on the structure of the module, see discussion above

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,208 @@
...
+    ) -> None:
+        """Calculate top of canopy shortwave radiation [J m-2].
+
+        Args:
+            tau: atmospheric transmissivity, unitless
+            shortwave_in: daily downward shortwave radiation [J m-2]
+            sunshine_fraction: fraction of sunshine hours, between 0 (100% cloud cover)
+                and 1 (cloud free sky)
+            albedo_shortwave: shortwave albedo, default = 0.17
+        """
```
alexdewar, line 125: I'd add a sanity-check that `sunshine_fraction` is between 0 and 1.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,208 @@
...
+# from core.constants import CONSTANTS as C
+# this doesn't exist yet; optional scipy
+CLOUDY_TRANSMISSIVITY = 0.25
+"""Cloudy transmittivity :cite:p:`Linacre1968`"""
+TRANSMISSIVITY_COEFFICIENT = 0.50
+"""Angular coefficient of transmittivity :cite:p:`Linacre1968`"""
+FLUX_TO_ENERGY = 2.04
+"""From flux to energy conversion, umol/J :cite:p:`Meek1984`"""
+BOLZMAN_CONSTANT = 5.67e-8
+"""Stephan Bolzman constant W m-2 K-4"""
```
alexdewar, line 48: Spelling. Also, I'd refer to this as Stefan's constant to avoid confusion with the Boltzmann constant.

```suggestion
STEFANS_CONSTANT = 5.67e-8
"""Stefan-Boltzmann constant W m-2 K-4"""
```

tests/test_radiation.py
```diff
@@ -0,0 +1,207 @@
...
+def test_import_constants(cname):
+    """Test constants can be imported."""
+
+    # Get the module that should contain the constants
+    import virtual_rainforest.models.abiotic.radiation as rad
+
+    # Check that the constant can be retrieved from the module without
+    # raising an AttributeError.
+    with does_not_raise():
+        assert getattr(rad, cname)
```
alexdewar, line 32: As I think I said on the last version of this PR, I don't think you need to test that the constants are present. Just test whether your functions work.

virtual_rainforest/models/abiotic/radiation.py
alexdewar: There are also a bunch of other places where similar sanity-checks would be useful.

tests/test_radiation.py
```diff
@@ -0,0 +1,212 @@
...
+        sunshine_hours=sun_hours,
+        albedo_vis=ALBEDO_VIS,
+        albedo_shortwave=ALBEDO_SHORTWAVE,
+        canopy_temperature=temp_canopy,
+        surface_temperature=temp_soil,
+        canopy_absorption=rad_absorbed,
+    )
+
+    assert radiation_fixture.ppfd == pytest.approx(exp_ppfd, 0.1)
+    assert radiation_fixture.topofcanopy_radiation == pytest.approx(exp_toc_rad, 0.1)
```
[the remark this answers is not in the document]
alexdewar: If we're expecting these values to be arrays at some point in the future, then using `np.allclose` makes sense.

alexdewar (changes requested):
Getting there! I've made a few suggestions.

virtual_rainforest/models/abiotic/radiation.py
dalonsoa: The correct should be is `STEFAN_BOLTZMAN_CONSTANT`, or `SIGMA`. Just `BOLTZMAN` or `STEFAN` would be misleading. https://en.wikipedia.org/wiki/Stefan–Boltzmann_constant

virtual_rainforest/models/abiotic/radiation.py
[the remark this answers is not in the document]
dalonsoa: I agree with @alexdewar and @davidorme comments. Making things more functional by extracting some of these methods as functions will simplify the testing and maintenance of the code, as well as the chances of using it wrong. I will add using the full `calculate_ppfd`. You're not saving much space and it makes the name of the function readable.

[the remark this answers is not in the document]
vgro: Thank you both, I'll change the structure to have more pure functions.
About the values to be stored, most of the attributes are inputs to other modules (for example PPFD is a key input for the plants, net_radiation is the main input to the energy balance) and I would ideally like to store them in our central data object. This is where the top down and bottom up clashes a little bit (in my brain at least), I didn't fully understand the discussion we had around the Modelclass/loading the data object/inheritence/overwriting/timing...When I first drafted this I only thought about the functions without the VR context, now I will need to fit this in the overall structure.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,208 @@
...
+
+        Reference: :cite:t:`Davis2017`
+        """
+        # calculate transmissivity (tau), unitless
+        tau_o = CLOUDY_TRANSMISSIVITY + TRANSMISSIVITY_COEFFICIENT * sunshine_fraction
+        tau = tau_o * (1.0 + BEER_REGRESSION * self.elevation)
+
+        # Calculate daily photosynth. photon flux density (ppfd_d), mol/m^2
+        self.ppfd = (1.0e-6) * FLUX_TO_ENERGY * (1.0 - albedo_vis) * tau * shortwave_in
+        return tau
```
dalonsoa, line 108: The pure version of the functional programming paradigm will tell you that the function outputs should depend exclusively on its inputs. Based on that, and something I've used in the past and has came handy for testing, **even constants should be passed as inputs**:

```python
def calc_something(a, b, sigma = STEFAN_BOLTZMAN_CONSTANT):
    ...
```

I think this is overkilling for really mathematical constants (eg. pi), but it might be useful for constants that might not be constants all the time (eg. is it `CLOUDY_TRANSMISSIVITY = 0.25` always? Or is it something that depends on the specific model you are implementing?

This might linked with something mentioned above about the albedo.

.github/workflows/ci.yml
```diff
@@ -19,11 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        # Pinned to fix problem with macosx and older python versions (change when fixed)
-        python-version: [ "3.9.16" , "3.10.9", "3.11" ]
-        exclude:
-           - os: windows-latest
-             python-version: 3.9.16 # Not yet available
+        python-version: [ "3.9" , "3.10", "3.11" ]
```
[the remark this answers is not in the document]
vgro: I'm very sorry about this, I was hoping to avoid this on the clean branch :-(

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,208 @@
...
+            * longwave radiation from soil [J m-2]
+            * net shortwave radiation at the surface (=forest floor)[J m-2]
+
+        Args:
+            elevation: elevation [m]
+            shortwave_in: daily downward shortwave radiation[J m-2]
+            sunshine_fraction: fraction of sunshine hours, between 0
+                (100% cloud cover) and 1 (cloud free sky)
+            albedo_vis: visible light albedo, default = 0.03
+            albedo_shortwave: shortwave albedo, default = 0.17
```
dalonsoa, line 197: If you indicated in the docstring a default value, then that default value should be included in the signature. For that to work, you will need to add these two to the end of the list of input arguments.

dalonsoa (changes requested):
I've made a couple of comments, but mostly agreed with what @davidorme and @alexdewar have already said. 

Well done - this is taking shape!

.github/workflows/ci.yml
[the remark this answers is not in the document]
alexdewar: That's ok! No need to be sorry. It's pretty obvious which bits are yours -- just trying to figure out why it's happening for future PRs :smile:

virtual_rainforest/models/abiotic/radiation.py
alexdewar: I'll defer to the physicist here :wink:. Mind there are two Ns in Boltzmann though!

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,208 @@
...
+        canopy_absorption: NDArray[np.float32],
+    ) -> None:
+        """Calculates daily net radiation at the forest floor [J m-2].
+
+        Args:
+            canopy_absorption: absorption by canopy layers, [J m-2]
+        """
+        self.netradiation_surface = (
+            self.topofcanopy_radiation
+            - canopy_absorption  # np.sum(canopy_absorption, axis=1)  # what axis?
```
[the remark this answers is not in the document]
vgro: This is still open because I wasn't sure how we are going to structure the vertical dimension and how I access it. The `canopy_absorption` should be an array of length `cell_id` and the other dimension is canopy layers, for example 3. I want to sum over all canopy layers here

virtual_rainforest/models/abiotic/radiation.py
davidorme: There's a parallel here to my `pyrealm` package which has a _ton_ of these kind of empirical 'constants'.

There, I made the decision that if people would typically be using a consistent set of those values across an analysis, then packaging them into a (user editable) `dataclass` of constants that could be used consistently was easier than having every last constant as a named input. So functions have a `params` input that accepts a `PModelParams` object. 

As well as keeping keeping the function/method signatures simple, this also isolates a particular set of values in a (frozen) dataclass so it is always clear which values are being used. You can't - for example - accidentally update `CONSTANTS.STEFAN_BOLTZMAN_CONSTANT` half way through. You can of course use two different parameter sets inconsistently but that's on you 😄 

This idea though has not been exposed to review...

alexdewar: Honestly, I slightly prefer @dalonsoa's approach of the two. I'm not sure when you'll want to define a whole set of constants, other than the one place where you define them with their "real" values, so it makes more sense to me to pass in a single constant value to a pure function (e.g. for testing purposes). I might be missing something.

Either way though, it seems like we are broadly in agreement and the details don't matter so much.

vgro: Thank you all for the suggestions, although I must admit I got a bit confused by the end of the day ;-) I see the point in adding constants as arguments to the functions, and to set defaults if they are not true constants, but then I no longer have the CONSTANTS in capitals, which makes it less clear when looking at the code.
I suggest to go back to an earlier suggestion and make more use of scipy (for example for Stefan Boltzmann constant) and other (more variable) inputs can be set to default but changes in one of the config files in the future?

virtual_rainforest/models/abiotic/radiation.py
vgro: Thanks for pointing this out, I think it goes back to the problem of constants discussed earlier. I hope this can be solved with the restructuring of the module

virtual_rainforest/models/abiotic/radiation.py
davidorme: Let's not rush to do this. @vgro - can you leave those constants as they are (using global values) for the moment.

davidorme: We can always revisit this and move to the purer version but we have a functioning approach for making these initial drafts and we are going to want a coherent approach across all modules. The reason I'm suggesting waiting and addressing this in a later PR (maybe across modules) is that  this project is going to have an awful lot of parameters:

* Some of those are going to be 'real' constants which _might_ be in `scipy` but might not, so having a single source that brings them all together is going to be cleaner. Notably, `scipy` only provides 'true' constants and doesn't cover Earth specific constants like the adiabatic lapse rate etc.

* Some functions are going to use physical equations from the literature that have complex empirical parameterisation. Those are not things that a user is likely to want to tinker with: 'Hmm, I wonder what happens to elephants if the temperature / water density relationship changes?'.

  I don't think these should be hard coded into functions because they might be revised but if you take a purely functional approach and have them all as arguments, but then the function signatures can get very long. That isn't inherently a problem, but it does make the API feel a lot more complex. Essentially there are "args that are real data variables" and "lots of args you will never touch".

  Those are some of the reasons I took the parameter set approach in `pyrealm` - the most obvious application there is the set of `calc_` functions [here](https://github.com/davidorme/pyrealm/blob/master/pyrealm/pmodel.py#L128).

* And finally there are `vr` empirical 'not-constants-at-all', which are going to be empirical parameters. Those are likely to be values that should be specified in the config and are expected to be changed frequently between simulations. I think *those* should be pure function arguments!

I'm definitely not saying that the approach in `pyrealm` is better (or even good or well implemented) but it is a solution to something like the unusually complex parameterisation we will have. I just think it is worth leaving things as they are until we've got a feel for how we want to consider this holistically.

alexdewar: > Let's not rush to do this. @vgro - can you leave those constants as they are (using global values) for the moment.

Agreed. Maybe we have ventured a bit off topic. Constants aside, the main thing I think is turning some of the member variables into input arguments, as we were initially talking about.

davidorme: Just tagging in #129 - and we should maybe convert that issue to a discussion.

We now have a couple of whole system things where we want to synchronise approaches: constants and also using `pint` to add units to everything. Both of those should be adopted universally so it seems better to continue with the individual module developments and then have a joint branch/PR to adopt those overarching strategies once we've firmed up what we want to do.

tests/test_radiation.py
davidorme: The steps need separating:

```python
@pytest.mark.parametrize(
    argnames="elev, sun_frac, exp_tau, exp_err",
    argvalues=[
        pytest.param(
            np.array([100, 1000]),
            np.array([0.5, 1.0]),
            np.array([0.501335, 0.770025]),
            does_not_raise(),
            id="standard_array_should_get",
        ),
        pytest.param(
            np.array([100, 1000]),
            np.array([1.1, 1.0]),
            None,
            pytest.raises(ValueError),
            id="value_error_sf",
        ),
    ],
)
def test_calculate_atmospheric_transmissivity(elev, sun_frac, exp_tau, exp_err):
    """Test atmospheric tansmissivity across elevation and sunshine fractions."""

    # Run the code that is expected to raise the exception, testing the assertion that it is raised
    with exp_err:
        result = radiation.calculate_atmospheric_transmissivity(elev, sun_frac)
        # If the code does not raise, we can now proceed to testing the result
        assert np.allclose(result, exp_tau)
```

davidorme: Note that the `assert np.allclose(result, exp_tau)` has to happen within the `with` block, because otherwise the `assert` will be tested even if the code raises an exception.

vgro: hmm it still doesn't like the `np.allclose` for `None`

davidorme: Oh - thought that would work. HOw about this:

```python
  # Run the code that is expected to raise the exception, testing the assertion that it is raised
  with exp_err as err:
      result = radiation.calculate_atmospheric_transmissivity(elev, sun_frac)

if not err:
      assert np.allclose(result, exp_tau)
```

vgro: Perfect, this works, thank you! Now I just need to think of ways in which this could go wrong ;-)

vgro:
I considered your suggestions  on the main code and updated tests. The first test does not work because it does not raise an error, I couldn't figure out how to overcome this, suggestions are welcome.
The handing of constants and the data object will need to be updated once we've decided how we want to go about this.

tests/test_radiation.py
```diff
@@ -0,0 +1,150 @@
...
+def test_calculate_atmospheric_transmissivity(elev, sun_frac, exp_tau, exp_err):
+    """Test atmospheric tansmissivity across elevation and sunshine fractions."""
+
+    with exp_err as err:
+        result = radiation.calculate_atmospheric_transmissivity(elev, sun_frac)
+
+        # if no error is raised
+    if not err:
+        assert np.allclose(result, exp_tau)
+
```
vgro, line 39: Failed: DID NOT RAISE <class 'virtual_rainforest.core.model.InitialisationError'>

tests/test_radiation.py
```diff
@@ -0,0 +1,173 @@
...
+    ],
+)
+def test_radiation_class(elev, exp_err, exp_message):
+    """Test elevation above zero."""
+
+    with exp_err:
+        radiation.data["elevation"] = elev
+        result = radiation.Radiation()
+
+        assert result.elevation.all() == elev.all()
```
vgro, line 30: Failed: DID NOT RAISE <class 'virtual_rainforest.core.model.InitialisationError'>

alexdewar: So you've told `pytest` that this block should raise an `InitialisationError` (for some parameters) and it's complaining that it wasn't (either no error was raised or a different type was).

I have a suspicion as to why this is, but have you checked that the error is raised by doing something like this outside of `pytest`?

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+    # check that elevation is above sea level
+    if np.any(data["elevation"] < 0):
+        to_raise = InitialisationError(
+            "Initial elevation contains at least one negative value!"
+        )
+        LOGGER.error(to_raise)
+        raise to_raise
+
+    elevation: NDArray[np.float32] = data["elevation"]
+    """Elevation above sea level, [m]"""
```
alexdewar, line 110: I *think* this might be the reason your tests are failing. I think this check will only be run once, when the class is defined, not every time you create a `Radiation` object. So I think all of this block needs to go into `__post_init__`, otherwise if `data["elevation"]` is later changed then a) the `elevation` field won't be updated and b) the check won't be run.

I may be wrong about this, but I think this is worth checking.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+            [300, 200, 100],
+            [300, 200, 100],
+        ],
+        dtype=np.float32,
+    ),
+}
+
+
+@dataclass
+class Radiation:
```
alexdewar, line 80: Given that all but one of these fields is initialised in `__post_init__` (and that one seems to be a source of test failures), why not just make this a regular class and do the necessary checks/calculations in `__init__`? I don't see what we get by making this a `dataclass`, except weirder-looking code.

@davidorme, I think this was your idea, right?

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+        beer_regression: parameter in equation for atmospheric transmissivity based on
+            regression of Beer's radiation extinction function :cite:p:`Allen1996`,
+            default set in config
+
+    Returns:
+        atmospheric transmissivity, unitless
+    """
+
+    # check sunshine fraction between 0 and 1
+    if 0 > np.any(sunshine_fraction) > 1:
```
alexdewar, line 177: There are two problems here:

1. Doesn't this need to be between 0 and 1 *inclusive*? Currently you can't have 0% or 100% sunshine
2. The syntax is wrong. This will check whether any of `sunshine_fraction` is "truthy" first, *then* it will check whether this true/false value is between 0 and 1

I think this will work:

```suggestion
    if np.any(0 => sunshine_fraction => 1):
```

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Reference: :cite:t:`Davis2017`
+    """
+
+    # check if sunshine fraction is in data object, if not set to 1
+    if sunshine_fraction.size > 1:
+        sunshine_fraction = data["sunshine_fraction"]
+    else:
+        sunshine_fraction = np.ones(len(elevation), dtype=np.float32)
+        LOGGER.info("The sunshine fraction is set to default = 1")
```
alexdewar, line 217: This `if`-statement means that *whatever* value for `sunshine_fraction` is passed into the function, it will be discarded and replaced either by ones or `data["sunshine_fraction"]`. I don't think this is what you want.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+) -> NDArray[np.float32]:
+    """Calculate top of canopy photosynthetic photon flux density, [mol m-2].
+
+    Args:
+        shortwave_in: downward shortwave radiation, [J m-2]
+        elevation: elevation above sea level, [m]
+        sunshine_fraction: fraction of sunshine hours, between 0 (100% cloud cover)
+            and 1 (cloud free sky), default = 1
+        albedo_vis: visible light albedo, default = 0.03
+        flux_to_energy: flux to energy conversion factor, [umol J-1],default from config
```
alexdewar, line 204: Out of interest is umol/J the usual unit for this? If so that's ok, but if there's not a standard way of representing it, I'd just use SI units.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+    Returns:
+        top of canopy radiation shortwave radiation, [J m-2]
+    """
+
+    # check if sunshine fraction is in data object
+    if sunshine_fraction.size > 1:
+        sunshine_fraction = data["sunshine_fraction"]
+    else:
+        sunshine_fraction = np.ones(len(elevation), dtype=np.float32)
+        LOGGER.info("The sunshine fraction is set to default = 1")
```
alexdewar, line 247: This function has the same problem as `calculate_ppfd`.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+        * STEFAN_BOLTZMANN_CONSTANT
+        * (CELSIUS_TO_KELVIN + surface_temperature) ** 4
+    )
+
+    # return array of longwave radiation for all canopy layers and surface
+    return np.append(
+        longwave_canopy.transpose(),
+        longwave_soil.reshape([1, len(surface_temperature)]),
+        axis=0,
+    ).transpose()
```
alexdewar, line 295: What do these operations do to the shape of the returned data? Do you just want to concatenate some vectors together?

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+        canopy_absorption: shortwave radiation absorbed by canopy layers, [J m-2]
+        longwave_radiation: longwave radiation from canopy layers and soil, [J m-2]
+
+    Returns:
+        net shortwave radiation at the surface ( = forest floor), [J m-2]
+    """
+    return (
+        topofcanopy_radiation
+        - np.sum(canopy_absorption, axis=1)  # sum over all canopy layers
+        - np.sum(longwave_radiation, axis=1)  # sum over all canopy layers and topsoil
```
alexdewar, line 316: It's not clear in my head what shapes these various arrays can be. Are they all either vectors or scalars? If so, you shouldn't need the `axis=1` argument, which only makes sense if there are two or more dimensions.

It would be helpful if the docstring mentioned how many dimensions the arguments can/must have.

alexdewar (changes requested):
I've made a few suggestions. It's getting there!

virtual_rainforest/models/abiotic/radiation.py
davidorme: I think that is right - in fact, that code might not run at all? I don't know how `dataclasses` handle that kind of extra code that isn't in a special method?

virtual_rainforest/models/abiotic/radiation.py
davidorme: Hey! We all said it might work: https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/155#discussion_r1094218746

I see what you mean about the 'weirder-looking' code though.

@vgro - I think we're getting into the realms of "equal functionality / flip flopping advice / implementation ping pong" here.

I think this is a good opportunity for pair programming because then you can have those "what the _hell_ are you talking about" discussions in real time, rather than working in a vacuum via GitHub commits.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,317 @@
...
+    ),
+    "surface_temperature": np.array([20, 20], dtype=np.float32),
+    "canopy_absorption": np.array(
+        [
+            [300, 200, 100],
+            [300, 200, 100],
+        ],
+        dtype=np.float32,
+    ),
+}
```
davidorme, line 76: Now that the `data` branch is in, we can start using the actual data system 😄 The code below creates an actual `Data` object with the same contents:

```python
from xarray import DataArray


grid = Grid(cell_nx=2, cell_ny=1)
data = Data(grid)

data['elevation'] = DataArray([100, 100], dims=['cell_id'])
data['sunshine_fraction'] = DataArray([0.5, 1.0], dims=['cell_id'])
data['canopy_temperature'] = DataArray([
            [25, 22, 20],
            [25, 22, 20],
        ],
        dims = ['cell_id', 'canopy_layer']
    )
data['surface_temperature'] = DataArray([20, 20], dims=['cell_id'])
data['canopy_absorption'] = DataArray([
            [300, 200, 100],
            [300, 200, 100],
        ],
        dims = ['cell_id', 'canopy_layer']
    )

```

virtual_rainforest/models/abiotic/radiation.py
vgro: I agree, there are some very basic structural problems that I find hard to solve which leads to just more different ways of doing it wrong, a direct conversation would be very useful.

virtual_rainforest/models/abiotic/radiation.py
vgro: I tested this outside pytest and it turns our that it is not possible to concatenate arrays just as they are but the shape needs to be provided, that was also concluded by others. I'm sure there is a better way to ensure that the dimensions match but in the end, thia will also depend on the shape of the data object entries, so it is a bit hard to set this up properly...

davidorme (commented):
@alexdewar has picked up some functional issues and I can't see anything else apart from how we structure that `Radiation` class. I think the overall structure is now good.

I've added a comment showing how to create a dummy dataset using the actual Data system. At some point we might want to convert your constants into a dataclass, but the discussion on how to do this across the project is still open at #162 so this is fine for now.

virtual_rainforest/models/abiotic/radiation.py
alexdewar: I'd forgotten that :laughing:. Sorry, I wasn't trying to accuse you!

I did like the general idea of removing intermediate values from the class, but I'm not sure a `dataclass` is the way to go here (despite what I may have said before :wink:). As @davidorme says, there are a bunch of ways you could rewrite this, but I think it should be relatively easy to convert this into a regular class, with all the `__post_init__` stuff in `__init__` instead, which will just make it a bit clearer. Does anyone else have a better idea?

virtual_rainforest/models/abiotic/radiation.py
alexdewar: Fair enough! As I said, I'd just try sticking this into `__post_init__` for now and seeing if that fixes your tests.

vgro:
Hello @alexdewar @dalonsoa @jacobcook1995 @TaranRallings ,

@davidorme and I had a 2.5 h long pair-programming session this morning to discuss the issues that have come up here. As a result, we made major changes to the radiation class and discovered a whole range of new issues that we would like to discuss with the whole team (please ignore the tests for now).
Here the main points:

**bigger picture (proof of concept):**
- we moved all `constants` in a `dataclass` which is given to the `Radiation` class as an input and provides defaults for the non-class functions
- we integrated a `data` object, which seems to work but brings up a new issue (as already mentioned elsewhere): our data comes as DataArrray but the radiation.py is based on numpy arrays/functionalities. We need to discuss this.
- we did some low key testing with the `data` object and the `constants` which worked ok, however, I moved other parts which means that the tests need to be updated

**class specific:**
- we turned the `Radiation` class back into a 'normal' class with a 'normal' `__init__` method (previously dataclass)
- for now we keep all functions in __init__, but we see the opportunity to move some of it the the `AbioticModel` as more of a pre-run (in the future)
- updated docstrings
- incorporated other smaller suggestions

davidorme:
This uses the approach mentioned in the discussion item on constants (#162). Basically, the `radiation` module has a `RadiationConstants` dataclass. The class `__init__` includes that as an input, so those 'constants' can be modified.

```python
def __init__(
    self, data: Data, const: RadiationConstants = RadiationConstants()
) -> None:
```

There is then a choice about how to pass constants down to methods/functions used by classes. We've got this function:

```python
def calculate_atmospheric_transmissivity(
    elevation: NDArray[np.float32],
    sunshine_fraction: NDArray[np.float32] = np.array(1.0, dtype=np.float32),
    cloudy_transmissivity: float = RadiationConstants.CLOUDY_TRANSMISSIVITY,
    transmissivity_coefficient: float = RadiationConstants.TRANSMISSIVITY_COEFFICIENT,
    beer_regression: float = RadiationConstants.BEER_REGRESSION,
) -> NDArray[np.float32]:
```

That signature is extremely clear but means that within the `Radiation` class, the `data` and `const` need to be unpacked:

```python
tau = calculate_atmospheric_transmissivity(
    data["elevation"],
    sunshine_fraction,
    const.cloudy_transmissivity,
    const.transmissivity_coefficient,
    const.beer_regression,
)
```
 
You _could_ collapse those function definitions a bit and just pass the `RadiationConstants` object around and let the functions handle getting the named constants. 

```python
def calculate_atmospheric_transmissivity(
    elevation: NDArray[np.float32],
    sunshine_fraction: NDArray[np.float32] = np.array(1.0, dtype=np.float32),
    const: RadiationConstants = RadiationConstants()
) -> NDArray[np.float32]:
```

And hence:

```python
tau = calculate_atmospheric_transmissivity(
    data["elevation"],
    sunshine_fraction,
    const
)
```

We've kept the explicit constant arguments for the moment and we think that is clearer. If you did it the other way, you probably should then document _which_ constants are used in a particular function in the docstring, so you're more or less back at explicit arguments.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -89,10 +94,17 @@ class Radiation:
...
+    * cloudy_transmissivity: Cloudy transmittivity
+    * transmissivity_coefficient: Angular coefficient of transmittivity
+    * flux_to_energy: float From flux to energy conversion, umol J-1
+    * stefan_boltzmann_constant: Stefan-Boltzmann constant = 5.67e-8 W m-2 K-4
+    * soil_emissivity: Soil emissivity
+    * canopy_emissivity: Canopy emissivity
+    * beer_regression: Parameter in equation for atmospheric transmissivity based on
+        regression of Beer's radiation extinction function
+    * celsius_to_kelvin: Factor to convert temperature in Celsius to absolute
+        temperature in Kelvin= 273.15
```
davidorme, line 107: The API documentation for the `RadiationConstants` dataclass is going to show all of this information, so I'd keep it simple here and just say:

```rst
The ``const`` argument takes an instance of class
:class:`~virtual_rainforest.models.abiotic.radiation.RadiationConstants`, which provides
a user modifiable set of required constants.
```

That link then allows people to see the details and avoids breaking DRY.

virtual_rainforest/models/abiotic/radiation.py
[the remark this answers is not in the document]
vgro: will discuss below

virtual_rainforest/models/abiotic/radiation.py
vgro: we decided to change the class back to a 'normal' class and this resolved this issue :-)

virtual_rainforest/models/abiotic/radiation.py
vgro: yes, that's true. We changed this to check if the input is provided and if not, set to default. This does not check if any values are missing but this should be picked up by earlier checks?

virtual_rainforest/models/abiotic/radiation.py
vgro: According to the reference, for daily values, the unit is moles per megajoules, but the hourly values were expressed in micromoles per joule to maintain consistency with energy units of global solar radiation energy-based photosynthetically active radiation (PARE) at hourly and daily scales. I guess I could introduce a scaling factor if I wanted t express the constant in mol J-1.

virtual_rainforest/models/abiotic/radiation.py
vgro: This sparked a discussion around dimension names, the use of xarray, and handling 3D data (time dimension not included). What I initially wanted was to combine the values for the topsoil [1,cell_id] with the canopy values [canopy_layers,cell_id]. To avoid mixup with the vertical axis, we could 1) have all vertical elements in one array (soil +canopy) and select chunks of it if they need to be treated differently, or 2) keep them separate and have two vertical axis

vgro:
I changed all np.arrays to DataArray, updated tests will follow soon

dalonsoa:
OK, so I don't see any problem with your changes. All of them make sense and have been somewhat discussed elsewhere. Using a `dataclass` for constants and be explicit in the function calls, unpacking the constants, I think is better and, to be honest, it does not seem too much extra work.

Not sure why DataArray vs numpy array is a problem. Is there anything that cannot be done with DataArrays and you need numpy arrays? In that case, just use the `data` attribute of the DataArray, which is the underlying numpy array.

virtual_rainforest/models/abiotic/radiation.py
davidorme: We might want to turn this into a helper function because I think it will get widely used. 😄 Once we get the data input bounds checking setup (#135) that could include a basic function we could import to do this.

virtual_rainforest/models/abiotic/radiation.py
vgro: outdated

vgro:
I will create a new PR to merge directly in develop, no need for intermediate feature branch

Issue #160, opened 2023-02-02
Approach to logging and error handling


davidorme:
@alexdewar raised this point in a comment on #158 and I have converted it to an issue to keep it live when I merge that PR.

This is def a cleaner approach than the `log_and_raise` business that came before and this PR could totally be merged as is. I do have some broader suggestions about how error handling is being done in VR though, which could be dealt with at another point.

Firstly, you're using the `critical` log level everywhere when it should probably be `error`. Critical errors are generally (I think) *unrecoverable* errors, but if you're re-raising an exception then it's not really unrecoverable by definition, because the caller could catch it and continue execution of the program. You could have places where you use the critical log level, but these really should be higher-level errors (e.g. "failed to load data files") at a higher level in the call stack, rather than lower-level stuff (e.g. "data file X is missing required property Y"), and it should really be the last thing you do before you terminate the program. It's fine to just call `sys.exit(1)` if your error really is unrecoverable.

Secondly, while it's good that you've removed the `log_and_raise` function, I think the general approach of "log an exception then (re-)raise it" isn't the best way to handle some of the errors you're dealing with. Let's take the `validate_array` function, for example. There are three places where you raise errors in this function, logging the errors before doing so. `validate_array` is currently only called in `Data.__setitem__`. To my mind, if you *are* going to catch and re-raise errors for this (though I'm not sure that it's necessary), it would make more sense to do it there rather than inside `validate_array`, e.g.:

```py
try:
    # Validate and store the data array
    value, valid_dict = validate_dataarray(value=value, grid=self.grid)
    self.data[key] = value
    self.variable_validation[key] = valid_dict
except Exception as ex:
    LOGGER.error(f"Data validation failed: {str(ex)}")
    raise
```

Note that this also tells gives the user more context as to what the errors are actually about. That said, in this case, personally I would probably raise a new exception here (maybe `from ex`), rather than logging the error and then the caller (or the caller's caller) can decide what to do with it.

An alternative approach to explicitly logging and reraising errors everywhere would be to write a custom handler for uncaught exceptions, by overriding [`sys.excepthook`](https://docs.python.org/3/library/sys.html#sys.excepthook). That way you'd only have to write this boilerplate error-handling code once, rather than every time any error happens.

[...]

PR #165, opened 2023-02-07
Feature/abiotic model

vgro:
# Description

Basic `AbioticModel` with `__init__`, `abiotic_schema.json`, and `model.py`. I followed the documentation and the soil model as an example. So far, this does only set the number of soil and canopy layers. The timestep is also taken from the soil model and might need to be adjusted.

Tests are added to `test/test_models.py` for now, see #166 for further discussion.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/__init__.py
```diff
@@ -2,6 +2,8 @@
 
 # Import all module schema here to ensure that they are added to the registry
 from virtual_rainforest.core import schema  # noqa
+from virtual_rainforest.models.abiotic import schema  # noqa
+from virtual_rainforest.models.abiotic.model import AbioticModel  # noqa
 from virtual_rainforest.models.plants import schema  # noqa
```
vgro, line 7: I tried to add the `AbioticModel` to the list of models below but it was sorted back in here, I think it might do it alphabetically?

jacobcook1995: Ahh that make sense, not completely sure why `isort` let me get away with that in the past

jacobcook1995 (commented):
LGTM, but in fairness it derives from something I originally wrote, so I might not be the fairest judge. I had a few minor comments, but nothing major

tests/test_models.py
```diff
@@ -187,3 +198,143 @@ def test_generate_soil_model(
...
+
+        # In cases where it passes then checks that the object has the right properties
+        assert set(["setup", "spinup", "update", "cleanup"]).issubset(dir(model))
+        assert model.model_name == "abiotic"
+        assert (
+            repr(model)
+            == f"AbioticModel(update_interval = 1 weeks, next_update = 2022-11-08, "
+            f"soil_layers = {int(soil_layers)}, "
+            f"canopy_layers = {int(canopy_layers)})"
+        )
```
alexdewar, line 282: Do you actually want to check that the `repr()` for your model looks like this or do you just want to check that it has certain properties with certain values? If the latter, I'd write this out more explicitly, even if it means being more verbose.

virtual_rainforest/models/abiotic/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from pathlib import Path
+
+from virtual_rainforest.core.config import register_schema
+
+
+@register_schema("abiotic")
+def schema() -> dict:
+    """Defines the schema that the abiotic module configuration should conform to."""
+
+    schema_file = Path(__file__).parent.resolve() / "abiotic_schema.json"
```
alexdewar, line 11: This is fine for now, but note that it isn't *strictly* how you should load files embedded in a Python package. See issue #167.

It will work for now though, so don't worry about it yet

virtual_rainforest/models/abiotic/model.py
```diff
@@ -0,0 +1,146 @@
...
+        canopy_layers: int,
+        **kwargs: Any,
+    ):
+
+        if soil_layers < 1:
+            to_raise = InitialisationError(
+                "There has to be at least one soil layer in the abiotic model!"
+            )
+            LOGGER.critical(to_raise)
+            raise to_raise
```
alexdewar, line 50: Again, this is fine for now as it's how things are being done elsewhere, but FYI I don't think you guys should be using the `critical` log level everywhere as it's really only for fatal errors (see #160)

virtual_rainforest/models/abiotic/model.py
```diff
@@ -0,0 +1,146 @@
...
+
+        elif soil_layers != int(soil_layers):
```
alexdewar, line 52: *Really* minor style point here, but I'd personally use `if` here rather than `elif`, as the function will already have exited if the first `if` statement was true (via the `raise to_raise`). It's fine as is though

alexdewar (approved):
I've added a few comments, but I think this is good as is, so I've approved it.

virtual_rainforest/models/abiotic/model.py
vgro: Thanks for pointing this out, we can update this once we've reached a conclusion on this :-)

virtual_rainforest/models/abiotic/__init__.py
vgro: ok, I'll keep an eye on the issue and update when we have resolved it

virtual_rainforest/models/abiotic/model.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+            return cls(update_interval, start_time, soil_layers, canopy_layers)
+        else:
+            raise InitialisationError()
+
+    # THIS IS BASICALLY JUST A PLACEHOLDER TO DEMONSTRATE HOW THE FUNCTION OVERWRITING
+    # SHOULD WORK
+    def setup(self) -> None:
+        """Function to set up the abiotic model."""
+        for layer in range(0, self.soil_layers):
```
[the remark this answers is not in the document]
alexdewar: I'd personally remove it if it doesn't do anything, but up to you.

tests/test_models.py
vgro: I guess both, I updated the test

docs/source/api/abiotic.md
```diff
@@ -0,0 +1,24 @@
...
+---
+
+# API reference for `abiotic` modules
+
+The {mod}`~virtual_rainforest.models.abiotic` module is one of the component models of
+the Virtual Rainforest. It is comprised of several submodules that calculate the
+radiation balance, the energy balance, the water balance and the atmospheric CO2
+balance.
+
+Each of the soil sub-modules has its own API reference page.
```
jacobcook1995, line 24: abiotic

virtual_rainforest/models/abiotic/model.py
jacobcook1995: Yeah this usage mirrors `models/soil/model.py`, I definitely agree it should be changed (in both locations + the docs). But probably more sensible to do it everywhere down the line

jacobcook1995 (approved):
LGTM! Only noticed a very minor typo

docs/source/api/abiotic.md
vgro: upsi :-)

PR #169, opened 2023-02-08
Remove unnecessary critical messages in model.py scripts

### LINKED ISSUE #113: Use logger error codes better
**Describe the bug**

At the moment, we are using the INFO, ERROR and CRITICAL codes from `logger` but we need to be clearer about the audience for different messages and target them better.

The key one here at the moment is that we are using INFO for all sorts of internal setup things (like the various function registries). Those are not messages that a normal user will want to see - they make much more sense as DEBUG messages - and we can add a `debug` or `log_level` item to the config (defaulting to INFO) that allows them to be exposed. Developers can of course programatically change the logging level.

We also have WARNING - which I don't think is used yet. This should also be for user focussed messages. We _can_ create custom levels, but I don't see any need at present.

jacobcook1995:
# Description

Critical logging messages were being raised by `Model` `__init__`'s when the errors they were emitting were being handled by higher level functions in `main.py`. This is obviously not correct, as it led to multiple `CRITICAL` messages being logged. I thought it was worth fixing this immediately as the documentation informing people how to setup a model was basically instructing people to make this mistake.

This does not fix #113 or #160, which are much broader discussion. It does however remove an egregious example of the problem.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -89,7 +89,7 @@ def __init__(
         to_raise = InitialisationError(
                 "There has to be at least one pond in the freshwater model!"
             )
-        LOGGER.critical(to_raise)
+        LOGGER.error(to_raise)
         raise to_raise
         
     # Provide general attributes to the __init__ of the base class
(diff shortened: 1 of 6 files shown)


vgro (approved):
LGTM

davidorme (approved):
LGTM

Issue #168, opened 2023-02-08
Restructure the `tests` folder.


davidorme:
The `tests` folder has a very flat structure, which is going to get confusing fast. @dalonsoa has pointed out:

> In complex projects, I normally keep for the tests the same folder/file structure than in the actual code. In that sense, I would restructure completely the location of the tests to follow that. I'm sorry I did not spot this back at the beginning, when starting to write the tests.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/virtual_rainforest/issues/166#issuecomment-1422476713_

jacobcook1995:
So would the intended file structures be e.g. `tests/test_core/test_config.py` and `tests/test_models/test_soil/test_carbon.py`?

davidorme:
Probably just `tests/core/test_config.py` and `tests/models/soil/test_carbon.py`?

vgro:
I would go with @davidorme 's version, but I think both would work

jacobcook1995:
I was just wondering if `pytest` would pick up intermediate folders without the `test_*` structure? If it does then @davidorme's version definitely works better

vgro:
good point, this I don't know

dalonsoa:
Yes, the `test_` is only needed on the files, not on the intermediate folders. `pytest` will pick all of those files as containing tests, no matter the folder structure.

TaranRallings:
Would we be keeping the model.py tests in the central `test_models.py` file or decentralize those tests into something like `tests/models/animals/test_model.py`?

jacobcook1995:
The latter. That was actually the exact file that sparked this issue.

PR #171, opened 2023-02-09
Feature/dummy animal - AnimalModel 

TaranRallings:
# Description

This adds a `model.py` file containing an AnimalModel class and an accompanying `test_model.py` file testing that class.
The testing file has been placed in `tests/models/animals/` in line with #168 .
Both files follow the template from SoilModel.

Fixes # (issue)

## Type of change

## Key checklist

- [x ] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [x ] All tests pass: `$ poetry run pytest`

## Further checks

- [x ] Code is commented, particularly in hard-to-understand areas
- [x ] Tests added that prove fix is effective or that feature works


virtual_rainforest/models/animals/model.py
```diff
@@ -0,0 +1,118 @@
...
+class AnimalModel(BaseModel):
+    """A class describing the animal model.
+
+    Describes the specific functions and attributes that the animal module should
+    possess. Currently it is incomplete and mostly just a copy of the template set out
+    in AnimalModel.
+
+    Args:
+        update_interval: Time to wait between updates of the model state.
+
```
jacobcook1995, line 40: Probably should get rid of this long gap after the last Argument

virtual_rainforest/models/animals/model.py
```diff
@@ -0,0 +1,118 @@
...
+        if valid_input:
+            LOGGER.info(
+                "Information required to initialise the animal model successfully "
+                "extracted."
+            )
+            return cls(update_interval, start_time)
+        else:
+            raise InitialisationError()
+
+    # THIS IS BASICALLY JUST A PLACEHOLDER TO DEMONSTRATE HOW THE FUNCTION OVERWRITING
```
jacobcook1995, line 101: This comment can probably go

virtual_rainforest/models/animals/model.py
```diff
@@ -0,0 +1,118 @@
+"""The :mod:`~virtual_rainforest.animal.model` module creates a
```
jacobcook1995, line 1: A lot of the links in this doc string are broken, i.e. they are `~virtual_rainforest.animals` when they should be `~virtual_rainforest.models.animals`

jacobcook1995 (approved):
LGTM! Had a few minor comments.

There was also the thing that David raised about not having `model.py` but that probably makes as something to change down the line all at once

dalonsoa (approved):
Good start! I just have the same comments already made, so nothing extra from my side.

davidorme (approved):
Also LGTM. No major comments. There are some changes coming, but we can update when they get merged.

virtual_rainforest/models/animals/model.py
alexdewar: I'm slightly surprised that flake8-docstring doesn't pick up on this tbh.

virtual_rainforest/models/animals/model.py
```diff
@@ -0,0 +1,118 @@
...
+
+        Raises:
+            InitialisationError: If configuration data can't be properly converted
+        """
+
+        # Assume input is valid until we learn otherwise
+        valid_input = True
+        try:
+            raw_interval = pint.Quantity(config["animal"]["model_time_step"]).to(
+                "minutes"
```
alexdewar, line 75: I'm not sure about the choice of minutes as a unit here. Unless there's a good reason not to, I'd go with SI units everywhere to minimise confusion.

If you do want to go with a non-obvious unit, I'd incorporate this into the field name (i.e. `"model_time_step_minutes"`) to make it less ambiguous.

virtual_rainforest/models/animals/model.py
```diff
@@ -0,0 +1,118 @@
...
+        except (
+            ValueError,
+            pint.errors.DimensionalityError,
+            pint.errors.UndefinedUnitError,
+        ) as e:
+            valid_input = False
+            LOGGER.error(
+                "Configuration types appear not to have been properly validated. This "
+                "problem prevents initialisation of the animal model. The first"
+                "instance of this problem is as follows: %s" % str(e)
```
alexdewar, line 89: I'd make this an `f`-string

alexdewar (approved):
I've made some minor suggestions, but otherwise it looks good.

virtual_rainforest/models/animals/model.py
davidorme: There's a specific recommendation not to for logging - something to do with not evaluating the string unless that logger call is being emitted.

alexdewar: Interesting.... TIL

jacobcook1995:
By the way the naming scheme for `model.py` files has now been changed in `develop` (see #174). So the `model.py` file should be renamed to `animals_model.py` or `animal_model.py` to be consistent with `develop`

virtual_rainforest/models/animals/model.py
alexdewar: That makes sense, if you're already using minutes elsewhere.

I'd open an issue for it and then fix it everywhere in one PR.

virtual_rainforest/models/animals/model.py
davidorme: It's not obvious from the `logging` docs - they do mention optimisation but not this explicitly. There's a timing example right at the bottom of this link:
https://coralogix.com/blog/python-logging-best-practices-tips/

virtual_rainforest/models/animals/model.py
davidorme: It does do some odd things - there was an issue with it allowing incomplete `Args` when there were other sections? `Raises` in particular?

virtual_rainforest/models/animals/model.py
davidorme: More generally, the `int(round(raw_interval.magnitude))` patterns worries me: we shouldn't be silently changing settings. If someone had put in `1.5 minutes` then we shouldn't be chopping off those 30 seconds. If we don't want fractional time units, then we should block that in the config setup.

There is a problem here with using `pint` - it gives nice human readable units in the config, but we need to make sure that whatever comes in via `pint` gives an integer value on whatever our baseline temporal resolution is in `np.datetime64`. If we go for a fine enough baseline resolution then rounding is probably ok, but honestly I think we want to avoid rounding and it would be better to explicitly check that the value is a whole number of seconds.

alexdewar: Agreed. You also don't need to cast to `int` if you're using `round` anyway.

virtual_rainforest/models/animals/model.py
TaranRallings: Right! Thanks

jacobcook1995 (approved):
LGTM!

PR #173, opened 2023-02-10
Feature/model autodiscovery

### LINKED ISSUE #167: Use `importlib` to load files inside the package
I've noticed this in the places where the various JSON schemas are being loaded, but this may be a problem elsewhere in the code. For example, see here: https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/7f445d9e6fd729c7f398576ba51c30b5efda0e55/virtual_rainforest/models/plants/__init__.py#L11

The code looks something like this:

```py
schema_file = Path(__file__).parent.resolve() / "plants_schema.json"
```

Intuitively, this *should* be correct, but unfortunately it doesn't work with all the weird and wonderful ways Python code can be packaged, so technically you should always use the `importlib.resources` module: https://docs.python.org/3/library/importlib.resources.html

### LINKED ISSUE #172: Autodiscovery system for models

It isn't ideal that we need to remember to add models and schema to `virtual_rainforest.__init__.py`. We have to do that to get the objects imported and to populate the model and schema registries.

We can instead put this in the base `__init_.py`:

```python
import importlib.metadata
from importlib import import_module
import pkgutil

# Autodiscover models in the models module
import virtual_rainforest.models as vfm

for module_info in pkgutil.iter_modules(vfm.__path__):
    import_module(f"virtual_rainforest.models.{module_info.name}")

__version__ = importlib.metadata.version("virtual_rainforest")
```

Then we need to consistently:
 * register model schema in `virtual_rainforest/models/modelname/__init__.py` (which we already do)
 * in that model `__init__.py` file, also import the `BaseClass` subclass:

```python
from virtual_rainforest.models.modelname.modelname_model import ModelName   # noqa: F401
```

The `import_module` function then imports the members defined in `__init__.py` and, hey presto, everything is registered. 

What this **doesn't** work for is when a user has a custom model outside of the `virtual_rainforest` package structure. At the Python command line , you could simply import `vr` and then `from mymodule import EarthquakeModel`, but that doesn't currently fit into the command line configuration: you can't simply include an `earthquake` model config and expect it to work. So we probably need some kind of configuration option for custom models (rather than forcing people to put their code into the pip installed package location).

But that's a way off in the distance.

davidorme:
# Description

This PR adds model autodiscovery to the `virtual_rainforest/__init__.py` file. 

Any submodule within `virtual_rainforests/models` will automatically get its schema and BaseModel subclass registered, **as long as** the model `__init__.py` contains the schema registration and imports its BaseModel. The PR also updates the existing `soil` and `abiotic` modules to implement that requirement.

The other advantage here is that the import path for the main models is shorter. That seems like a sensible thing for the core object in each model:

```python
from virtual_rainforest.models.soil import SoilModel
```

rather than:

```python
from virtual_rainforest.models.soil.model import SoilModel
```

Fixes #172 

ETA: Fixes #167

## Type of change

## Key checklist

## Further checks


davidorme:
@dalonsoa @alexdewar

@jacobcook1995 and @vgro and I discussed this recently so I haven't added them as reviewers, but it would be great to get a check for any issues. @TaranRallings is ill.

jacobcook1995:
If possible could you update the documentation in `source/development/defining_new_models.md` to reflect the change?

davidorme:
That occurred to me too. I was going to lazily change it when I update the `data` part of the new model definition. But that is lazy.

docs/source/development/defining_new_models.md
```diff
@@ -217,23 +255,3 @@ def schema() -> dict:
 
     return config_schema
 ```
```
alexdewar, line 257: Is this code the same for all models, except for the name of the schema file?

If so, you could remove some of the boilerplate and just do something like this in all the `__init__.py` files:

```py
from virtual_rainforest.core.config import default_schema_setup

schema = default_schema_setup()
```

You'll probably need to examine the stack for this.

virtual_rainforest/__init__.py
```diff
@@ -1,13 +1,18 @@
...
+# Import the core schema to register it
+from virtual_rainforest.core import schema  # noqa: F401
+
+# Autodiscover models in the models module to add their schema and BaseModel subclass
+# All modules within virtual_rainforest.model are expected to:
+# - import their BaseModel subclass to the module root, for example:
+#   from virtual_rainforest.models.soil.model import SoilModel  # noqa: F401
+# - define a schema member, decorated with @register_schema
+for module_info in pkgutil.iter_modules(vfm.__path__):
+    import_module(f"virtual_rainforest.models.{module_info.name}")
```
alexdewar, line 16: ...alternatively, you could check here whether the module has an attribute called `schema` and if not, add a default one.

alexdewar (approved):
Looks ok to me. I've made a suggestion.

docs/source/development/defining_new_models.md
davidorme: It is the same for all models. That's a good thought.

virtual_rainforest/__init__.py
davidorme: One thing that crossed my mind was having a `config.add_module` command which basically wraps this import and schema registration. That could then be used to add custom modules _outside_ of the package tree. If you're at the Python command line, you can always import them, but if you wanted to use them in a config, we'd need a `custom_model` section that could use this too. But, that is probably one for the future.

davidorme:
@alexdewar @jacobcook1995 

I have updated this PR to implement Alex's suggestion of having a function to setup the schema. I've basically moved all of the reading and validation out of model `__init__.py` and into `config`. I have separated the `register_schema` decorator into `get_schema` function, which handles most of the heavy lifting and error cases, and a `register_schema` function, which just does the registering and wraps `get_schema`. There's a mild update to how we extend the validator.

I've updated the user docs and docstrings and then patched all the existing `__init__.py` files. This also fixes #167.

docs/source/development/defining_new_models.md
```diff
@@ -193,47 +193,63 @@ def cleanup(self) -> None:
...
-checks are turned off for the relevant lines. It's only strictly necessary to import the
-`Model` class, as this implicitly entails importing the specific model as a whole.
-However, for the sake of clarity we currently include both imports.
+1. The `__init__.py` file needs to register the JSONSchema file for the module. The
+  {meth}`~virtual_rainforest.core.config.register_schema` function takes a module name
+  and the path to the schema file and then, after checking the file can be loaded and is
+  valid, adds the schema to the schema registry
+  {data}`~virtual_rainforest.core.config.SCHEMA_REGISTRY`.
+1. It also needs to import the main BaseModel subclass. This is not going to be used
+   within the file, so needs `#noqa: 401` to surpress a `flake8` error. This step is
```
jacobcook1995, line 232: suppress

docs/source/development/defining_new_models.md
```diff
@@ -193,47 +193,63 @@ def cleanup(self) -> None:
...
+   primarily to ensure that the subclass is always loaded from the same location but
+   also means that the model class can now be imported more directly from the module
+   root: ``virtual_rainforest.models.freshwater.FreshWaterModel``.
+
+The `__init__.py` file will then look something like this:
 
 ```python
```
jacobcook1995, line 239: Is it worth adding in a docstring here, as the `__init__.py` now requires a docstring here to document the module correctly (granted this isn't necessarily linked to this pull request)

docs/source/development/defining_new_models.md
```diff
@@ -193,47 +193,63 @@ def cleanup(self) -> None:
...
-`Model` class, as this implicitly entails importing the specific model as a whole.
-However, for the sake of clarity we currently include both imports.
+1. The `__init__.py` file needs to register the JSONSchema file for the module. The
+  {meth}`~virtual_rainforest.core.config.register_schema` function takes a module name
+  and the path to the schema file and then, after checking the file can be loaded and is
+  valid, adds the schema to the schema registry
+  {data}`~virtual_rainforest.core.config.SCHEMA_REGISTRY`.
+1. It also needs to import the main BaseModel subclass. This is not going to be used
+   within the file, so needs `#noqa: 401` to surpress a `flake8` error. This step is
+   primarily to ensure that the subclass is always loaded from the same location but
```
jacobcook1995, line 233: Not 100% sure I grasp the relevance of "loaded from the same location" isn't it more about ensuring that it's always loaded (so that `pytest` etc are guaranteed to find it)?

davidorme: It was opaque - I've given a bit more detail

virtual_rainforest/core/config.py
```diff
@@ -100,95 +88,122 @@ def log_all_validation_errors(
...
+    Args:
+        module_name: The name to register the schema under
+        schema_file_path: The file path to the JSON Schema file
+
+    Raises:
+        ValueError: the module name has no registered schema path
+        FileNotFoundError: the schema path does not exist
+        JSONDecodeError: the file at the schema path is not valid JSON
+        SchemaError: the file contents are not valid JSON Schema
+        ValueError: the JSON Schema is missing required keys
```
jacobcook1995, line 172: Should these two Value errors be grouped, or are they better left separate?

jacobcook1995 (approved):
LGTM, made some minor comments but nothing major

virtual_rainforest/core/config.py
davidorme: Actually, I think one of them is redundant! Good spot.

PR #174, opened 2023-02-10
Feature/split testing

jacobcook1995:
# Description

The test files have now been split such that their structure mirrors the structure of the package code. This PR also changes the name of the model specific `model.py` to `{MODEL_NAME}_model.py` to make things a bit clearer.

Fixes #166 
Fixed #168

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

## Key checklist

- [x] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [x] All tests pass: `$ poetry run pytest`

## Further checks

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #176, opened 2023-02-13
Make a `Data` object part of the `BaseModel.__init__` 

### LINKED ISSUE #97: Unpacking data from data system to initialise simulation run
Once the data system is setup we will need a way of reading from it and using the data stored in it to initialise a simulation run.

Somewhat sketchy on what this would entail as the system for reading in and storing data is not yet established. However the data is stored it would need to be unpacked and used to setup the models. Probably in a specific `initialise_models_step`

Definitely need to use the data for initialisation eventually, so this seems unavoidable

davidorme:
# Description

This PR:

1. Makes a `Data` object a core part of model creation, by adding it to the `__init__` method for the `BaseModel` ABC. All subclasses now have a way to connect through to a shared `Data`instance.
2. Adds the `required_init_vars` property to `BaseModel` and adds running the new `check_required_init_vars` to the `BaseModel.__init__` method. Model subclasses can now define what variables (and on what axes) they need in Data to successfully `__init__` via that property. (Discussed #97)
3. Makes the two class properties (`model_name` and `required_init_vars`) part of the abstract BaseModel API, rather than just class attributes.
4. Updates `__init_subclass__`
5. Updates `test_model.py` to add new tests for.
6. Updates module docstring.
7. Updates `main.py`, `test_main.py` and existing models to match new signatures.

Fixes #97

Some notes:

* I wanted to use a parameterised test for `__init_subclass_` because there are multiple new subclass failure modes. That is awkward because the new class has to be defined within the test. I've used `exec()` to run the different subclass attempts within the model. That feels a bit clumsy but I can't see a better way and I don't think it is actively harmful.
* This PR has revealed that we need a new `Grid.from_config` function but I'll leave that for another PR.
* The `test_main.py` tests are using the `SoilModel` as a demo - that is a constantly moving target as that model develops so we should probably just define a new test BaseModel subclass in here for use in testing `main.py`.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -24,6 +24,16 @@ from virtual_rainforest.models.soil.model import SoilModel
 from .conftest import log_check
 
 
+@pytest.fixture
+def data_instance():
+    """Creates an emmpty data instance."""
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    grid = Grid()
+    return Data(grid)
+
+
 @pytest.mark.parametrize(
     "model_list,no_models,raises,expected_log_entries",
     [
@@ -155,13 +165,15 @@ def test_select_models(caplog, model_list, no_models, raises, expected_log_entri
         ),
     ],
 )
-def test_configure_models(caplog, config, output, raises, expected_log_entries):
+def test_configure_models(
+    caplog, data_instance, config, output, raises, expected_log_entries
+):
     """Test the function that configures the models."""
 
     with raises:
         model_list = select_models(["soil"])
 
-        models = configure_models(config, model_list)
+        models = configure_models(config, data_instance, model_list)
 
         if output is None:
             assert models == [None]
@@ -175,7 +187,7 @@ def test_vr_run_miss_model(mocker, caplog):
     """Test the main `vr_run` function handles missing models correctly."""
 
     mock_conf = mocker.patch("virtual_rainforest.main.validate_config")
-    mock_conf.return_value = {"core": {"modules": ["topsoil"]}}
+    mock_conf.return_value = {"core": {"modules": ["topsoil"], "data": []}}
 
     with pytest.raises(InitialisationError):
         vr_run("tests/fixtures/all_config.toml", Path("./delete_me.toml"))
@@ -183,6 +195,7 @@ def test_vr_run_miss_model(mocker, caplog):
         Path("./delete_me.toml").unlink()
 
     expected_log_entries = (
+        (INFO, "Loading data from configuration"),
         (INFO, "Attempting to configure the following models: ['topsoil']"),
         (
             CRITICAL,
@@ -205,6 +218,7 @@ def test_vr_run_bad_model(mocker, caplog):
                 "start_date": "2020-01-01",
                 "end_date": "2120-01-01",
             },
+            "data": [],
         },
         "soil": {
             "model_time_step": "0.5 martian days",
@@ -217,6 +231,7 @@ def test_vr_run_bad_model(mocker, caplog):
         Path("./delete_me.toml").unlink()
 
     expected_log_entries = (
+        (INFO, "Loading data from configuration"),
         (INFO, "Attempting to configure the following models: ['soil']"),
         (
             INFO,
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -27,7 +27,7 @@ from .conftest import log_check
 @pytest.mark.parametrize(
     "model_list,no_models,raises,expected_log_entries",
     [
-        (
+        pytest.param(
             ["soil"],  # valid input
             1,
             does_not_raise(),
@@ -37,8 +37,9 @@ from .conftest import log_check
                     "Attempting to configure the following models: ['soil']",
                 ),
             ),
+            id="valid input",
         ),
-        (
+        pytest.param(
             ["soil", "core"],
             1,
             does_not_raise(),
@@ -48,8 +49,9 @@ from .conftest import log_check
                     "Attempting to configure the following models: ['soil']",
                 ),
             ),
+            id="ignores core",
         ),
-        (
+        pytest.param(
             ["soil", "freshwater"],  # Model that hasn't been defined
             0,
             pytest.raises(InitialisationError),
@@ -65,6 +67,7 @@ from .conftest import log_check
                     " the registry: ['freshwater']",
                 ),
             ),
+            id="undefined model",
         ),
     ],
 )
@@ -82,7 +85,7 @@ def test_select_models(caplog, model_list, no_models, raises, expected_log_entri
 @pytest.mark.parametrize(
     "config,output,raises,expected_log_entries",
     [
-        (
+        pytest.param(
             {  # valid config
                 "soil": {"no_layers": 1, "model_time_step": "7 days"},
                 "core": {"timing": {"start_time": "2020-01-01"}},
@@ -98,8 +101,9 @@ def test_select_models(caplog, model_list, no_models, raises, expected_log_entri
                     "extracted.",
                 ),
             ),
+            id="valid config",
         ),
-        (
+        pytest.param(
             {  # invalid soil config tag
                 "soil": {"no_layers": -1, "model_time_step": "7 days"},
                 "core": {"timing": {"start_time": "2020-01-01"}},
@@ -123,8 +127,9 @@ def test_select_models(caplog, model_list, no_models, raises, expected_log_entri
                     "simulation.",
                 ),
             ),
+            id="invalid soil config tag",
         ),
-        (
+        pytest.param(
             {  # model_time_step missing units
                 "soil": {"no_layers": 1, "model_time_step": "7"},
                 "core": {"timing": {}},
@@ -146,6 +151,7 @@ def test_select_models(caplog, model_list, no_models, raises, expected_log_entri
                     "simulation.",
                 ),
             ),
+            id="model_time_step missing units",
         ),
     ],
 )
@@ -235,7 +241,7 @@ def test_vr_run_bad_model(mocker, caplog):
 @pytest.mark.parametrize(
     "config,output,raises,expected_log_entries",
     [
-        (
+        pytest.param(
             {
                 "core": {
                     "timing": {
@@ -259,8 +265,9 @@ def test_vr_run_bad_model(mocker, caplog):
                     "requested 15778800 minutes",
                 ),
             ),
+            id="timing correct",
         ),
-        (
+        pytest.param(
             {
                 "core": {
                     "timing": {
@@ -279,8 +286,9 @@ def test_vr_run_bad_model(mocker, caplog):
                     "larger than the run length (1 minutes)",
                 ),
             ),
+            id="run length < update interval",
         ),
-        (
+        pytest.param(
             {
                 "core": {
                     "timing": {
@@ -299,8 +307,9 @@ def test_vr_run_bad_model(mocker, caplog):
                     " days",
                 ),
             ),
+            id="invalid run length units",
         ),
-        (
+        pytest.param(
             {
                 "core": {
                     "timing": {
@@ -319,6 +328,7 @@ def test_vr_run_bad_model(mocker, caplog):
                     " long minutes",
                 ),
             ),
+            id="invalid update_interval units",
         ),
     ],
 )
@@ -337,8 +347,8 @@ def test_extract_timing_details(caplog, config, output, raises, expected_log_ent
 @pytest.mark.parametrize(
     "update_interval,expected_log_entries",
     [
-        (timedelta64(2, "W"), ()),
-        (
+        pytest.param(timedelta64(2, "W"), (), id="valid"),
+        pytest.param(
             timedelta64(5, "W"),
             (
                 (
@@ -347,10 +357,11 @@ def test_extract_timing_details(caplog, config, output, raises, expected_log_ent
                     "['soil']",
                 ),
             ),
+            id="fast model",
         ),
     ],
 )
-def test_check_for_fast_models(caplog, mocker, update_interval, expected_log_entries):
+def test_check_for_fast_models(caplog, update_interval, expected_log_entries):
     """Test that function to warn user about short module time steps works."""
 
     # Create SoilModel instance and then populate the update_interval
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -10,7 +10,7 @@ from logging import CRITICAL, ERROR, INFO, WARNING
 import pytest
 from numpy import datetime64, timedelta64
 
-from virtual_rainforest.core.model import BaseModel, InitialisationError
+from virtual_rainforest.core.model import InitialisationError
 from virtual_rainforest.models.abiotic.model import AbioticModel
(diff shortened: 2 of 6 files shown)


davidorme:
Just fixing the conflicts arising from the recently merged #174

virtual_rainforest/core/model.py
```diff
@@ -1,61 +1,69 @@
...
-unique. Although registration is automatic, it only happens when class definition
-happens, which requires the module which defines the child class to be imported
-somewhere. At the moment, we import all child models in the top level ``__init__.py`` to
-ensure that they are added to the registry.
-
-An example of ``Model`` inheritance from
-:class:`~virtual_rainforest.core.model.BaseModel` can been seen in the :mod:`soil.model`
-module.
+"""The :mod:`~virtual_rainforest.core.model` module defines the high level API for the
+different models within the Virtual Rainforest. The module creates the
```
vgro, line 2: for consistency, should we call this `core_model`?

jacobcook1995: I think it's probably better to call it something like `model_api`. The reason I didn't update the name here to match the other model naming scheme is that it isn't really a model that would be used in the same way as other models, which giving it a consistent name could implicitly suggest that it was

vgro (approved):
As far as I can tell, this all makes sense :-)

tests/test_models.py
```diff
@@ -5,339 +5,295 @@
...
-    mocker.patch.object(BaseModel, "__abstractmethods__", new_callable=set)
-
-    # Initialise model
-    model = BaseModel(timedelta64(1, "W"), datetime64("2022-11-01"))
+@pytest.fixture(scope="module")
+def data_instance():
+    """Creates a simple data instance for use in testing."""
+    from xarray import DataArray
 
-    # In cases where it passes then checks that the object has the right properties
```
jacobcook1995, line 29: Is this section being removed because the functionality is tested when inheriting classes are tested for this?

jacobcook1995 (approved):
This all seems sensible to me. Added a few queries, but nothing significant enough to prevent merging if others are happy for it to be merged

virtual_rainforest/core/model.py
davidorme: I think giving it a less ambiguous name would be good. I'd lean towards `base_model` - that seems to me clearer that it is the base for real models (and the object is called BaseModel!). `base_abc_model` would be even clearer but seems a bit over the top.

Perhaps make this a separate PR?

vgro: Side note: once we have an agreement on this, could you please update the API links to the updated model names for soil and abiotic? We missed that when updating the model names.

virtual_rainforest/core/model.py
```diff
@@ -118,34 +163,58 @@ def cleanup(self) -> None:
...
+        if isinstance(cls.model_name, property):
+            excep = NotImplementedError(
+                f"Property model_name is not implemented in {cls.__name__}"
+            )
+            LOGGER.error(excep)
+            raise excep
+
+        if not isinstance(cls.model_name, str):
+            excep = TypeError(f"Property model_name in {cls.__name__} is not a string")
+            LOGGER.error(excep)
```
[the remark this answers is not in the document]
davidorme: Good point. Same for the other ones here too - they will all occur when importing `virtual_rainforest`.

virtual_rainforest/core/model.py
jacobcook1995: Yeah I would be happy with `base_model.py`

tests/test_models.py
davidorme: Yup - I'm running this testing on an example minimal real class rather than making it possible to create an instance of BaseModel itself. The `test_check_failure_on_missing_methods` tackles this specifically. It would also catch undefined `model_name` and `required_init_vars` but those are checked in more detail in `__init_subclass__` so cannot  be missing when an instance is created.

jacobcook1995:
Actually while I remember, once the changes to setting up models are set in stone, would you be able to update `docs/development/defining_new_models.md` to reflect the changes?

davidorme:
@jacobcook1995 I'll do that in a mini PR, once #176 and #173 are in `develop`, otherwise I'll be documenting across branches 😖

jacobcook1995:
Yeah that makes sense

virtual_rainforest/core/base_model.py
```diff
@@ -119,34 +163,58 @@ def cleanup(self) -> None:
...
-        if not hasattr(cls, "model_name"):
-            excep = ValueError("Models must have a model_name attribute!")
+        excep: Exception
+
+        # Check that model_name exists and is a string - if it is not implemented in the
+        # subclass then the object will be of type property
+        if isinstance(cls.model_name, property):
+            excep = NotImplementedError(
+                f"Property model_name is not implemented in {cls.__name__}"
+            )
```
dalonsoa, line 190: Given that `model_name` is an `abstractmethod`, shouldn't be just impossible to create a subclass without overwriting it? That's the whole point of `abstractmethods`, right?

virtual_rainforest/core/base_model.py
```diff
@@ -119,34 +163,58 @@ def cleanup(self) -> None:
...
+        if not isinstance(cls.model_name, str):
+            excep = TypeError(f"Property model_name in {cls.__name__} is not a string")
+            LOGGER.critical(excep)
+            raise excep
+
+        # Check that required_init_vars is set - TODO - not testing structure here
+        if isinstance(cls.required_init_vars, property):
+            excep = NotImplementedError(
+                f"Property required_init_vars is not implemented in {cls.__name__}"
+            )
```
dalonsoa, line 203: Ans the same here.

tests/test_main.py
```diff
@@ -24,10 +24,20 @@
...
 
 
+@pytest.fixture
+def data_instance():
+    """Creates an emmpty data instance."""
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    grid = Grid()
+    return Data(grid)
```
dalonsoa, line 34: Given that this fixture is identical to the one in `test_soil_model`, I would add it to a `conftest.py` file in the tests root directory.

dalonsoa (approved):
All looks good. I think it is the right move.

I just made a couple of comments. It might be I don't understand how abstract classes work, in particular in reference to properties.

virtual_rainforest/core/base_model.py
davidorme: I got hung up on this too! The distinction is (hat tip, @jacobcook1995) that the automatic testing for missing `abstractmethod` objects only happens when you try and create an instance of the class. When `__init_subclass__` runs, it is just dealing with the `cls` object. In this function, we want to check the name is valid, but do need to gatekeep first for undefined `model_name`. 

Arguably, we could leave out the `required_init_vars` - we don't currently test the contents - and then it would get flagged at instantiation if missing. I think testing the contents is probably valuable though.

dalonsoa: Interesting. Indeed, I didn't know abstract classes worked that way. Yes, I agree it is useful to check during the definition of the subclasses.

davidorme: Is this kind of structure using `assert` ok for checking the contents of `required_init_vars`. I've not really used `assert` like this before but it seems a compact way to raise exceptions without having to nest a bunch of `if` statements to catch specific failure modes:

```python
        # Check the structure
        try:
            assert isinstance(cls.required_init_vars, tuple)
            for vname, axes in cls.required_init_vars:
                assert isinstance(vname, str)
                assert isinstance(axes, tuple)
                assert all([isinstance(a, str) for a in axes])
        except (AssertionError, ValueError):
            raise ValueError(
                f"Property required_init_vars has the wrong structure in {cls.__name__}"
            )
```

dalonsoa: There is nothing conceptually wrong with using `assert` I think, but `assert` checks are disabled when running python with the `-O` flag. From `python -h`:

```bash
-O     : remove assert and __debug__-dependent statements;
```

I'm not sure if you intend to run the code in optimised mode, but just in case I would not use `assert` to control the flow of the code.

davidorme: Ah. Damn. Yes - we don't want error checking to disappear in optimised mode. More long winded approach it is then.

tests/test_main.py
```diff
@@ -24,10 +24,20 @@
 from .conftest import log_check
 
 
+@pytest.fixture
+def data_instance():
+    """Creates an emmpty data instance."""
```
alexdewar, line 29: Spelling

virtual_rainforest/models/abiotic/abiotic_model.py
```diff
@@ -66,22 +69,23 @@ def __init__(
...
-    def from_config(cls, config: dict[str, Any]) -> AbioticModel:
+    def from_config(cls, data: Data, config: dict[str, Any]) -> AbioticModel:
         """Factory function to initialise the abiotic model from configuration.
 
         This function unpacks the relevant information from the configuration file, and
         then uses it to initialise the model. If any information from the config is
         invalid rather than returning an initialised model instance None is returned.
 
         Args:
+            data: A :class:`~virtual_rainforest.core.data.Data` instance.
```
alexdewar, line 88: Minor nit: the user will already be able to see `data`'s type, so there's no point in repeating it. How about describing it as "loaded user data" or something?

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -42,9 +43,11 @@ class SoilModel(BaseModel):
 
     model_name = "soil"
     """The model name for use in registering the model and logging."""
+    required_init_vars = []
```
alexdewar, line 46: How about using a type annotation here so that `mypy` knows what *kind* of list it will be?

alexdewar (approved):
I've made some minor comments, but it all looks good to me.

virtual_rainforest/models/soil/soil_model.py
davidorme: Does it not _have_ to follow the type of the abstract property in BaseModel? `mypy` certainly throws an error when it doesn't.

Having said that - it seems that we _have_ to manually add docstrings for the subclasses or these properties don't get documented. I think we probably could do something clever and make them inheritable (probably in `__init_subclass__` but they probably should be subclass specific.

davidorme:
OK - I have now updated this to check `required_init_vars` has the right structure and uses known axes when the `__init_subclass` runs. That means that unknown core axes from bad model setup don't get flagged when users are trying to use the model class - the whole thing falls over when `vr` is imported. That seems correct for a fundamental programming error.

I've also updated the docstrings for existing models and the API stubs in source to better document the new data argument and the subclass attributes.

davidorme:
@jacobcook1995 I've updated the new models document to match - can you have a look?

tests/models/abiotic/test_abiotic_model.py
```diff
@@ -11,6 +11,16 @@
...
+@pytest.fixture
+def data_instance():
+    """Simple data instance to use in model init."""
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    grid = Grid()
+    return Data(grid)
+
+
```
dalonsoa, line 23:
```suggestion
```
As this is now in `conftest.py`, it can be removed from here.

virtual_rainforest/core/base_model.py
```diff
@@ -119,34 +177,96 @@ def cleanup(self) -> None:
...
     @classmethod
     @abstractmethod
-    def from_config(cls, config: dict[str, Any]) -> Any:
+    def from_config(cls, data: Data, config: dict[str, Any]) -> BaseModel:
         """Factory function to unpack config and initialise a model instance."""
 
     @classmethod
     def __init_subclass__(cls) -> None:
-        """Method that adds new model classes to the model registry.
+        """Initialise subclasses deriving from BaseModel.
```
dalonsoa, line 185: The functionality of this method is totally ok, but I would split it into smaller functions for checking the individual attributes, leaving this as a top level method that just call those internally. That will make understanding what's going here on and testing easier.

virtual_rainforest/core/base_model.py
```diff
@@ -160,7 +280,57 @@ def __repr__(self) -> str:
 
     def __str__(self) -> str:
         """Inform user what the model type is."""
-        if hasattr(self, "model_name"):
-            return f"A {self.model_name} model instance"
-        else:
-            return "A base model instance"
+        return f"A {self.model_name} model instance"
+
+    def check_required_init_vars(self) -> None:
```
dalonsoa, line 285: Just to be sure I get this: the checks in `init_subclass` are done when setting up the class - i.e. when importing `vr` and these ones are checked when running the simulation, at the config stage, right? 

I would rename this method because it seems a bit confusing checking the `required_init_vars` in two places considering that in one case you are checking the class and in another the contents of data.

dalonsoa (changes requested):
I've made a couple of suggestions, maybe to improve clarity of the code. The functionality itself is totally fine.

tests/core/test_base_model.py
```diff
@@ -1,95 +1,350 @@
-"""Test module for base_model.py.
+"""Test module for model.py (and associated functionality).
 
-This module tests the functionality of base_model.py, which defines the basic model API
-that specific models (e.g. `soil_model.py`) utilise.
+This module tests the functionality of model.py, as well as other bits of code that
```
jacobcook1995, line 3: It doesn't actually test other bits of code anymore as this has been moved to separate scripts

jacobcook1995 (approved):
The doc changes are good! Have added a few minor comments

virtual_rainforest/core/base_model.py
davidorme: That's correct on the functionality - and I agree that the naming is ambiguous.

docs/source/development/defining_new_models.md
```diff
@@ -93,23 +167,116 @@ def __init__(
...
+    "title": "Root Schema",
+    "required": [
+        "update_interval",
+        "no_of_ponds"
+    ],
+    "properties": {
+        "update_interval": {
+            "type": "string",
+            "default": "",
+            "title": "The update_interval Schema",
```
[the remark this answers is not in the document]
davidorme: Those fields are straight out of the JSONSchema app link shown above. I might add a note to say that we don't need all of them!

virtual_rainforest/core/base_model.py
davidorme: This was _much_ more irritating to do than anticipated. `mypy` basically complained about _everything_ once the components of the type checking (which is essentially the whole of the function) were isolated in separate functions. And at the end, we now have `# type: ignore`s because `mypy` in `__init_subclass` is no longer aware that the properties have been checked and have the correct structure.

Sometimes I _hate_ `mypy`.

davidorme: I mean, I agree that the end point is cleaner, but.... gaaaaaah!

alexdewar (approved):
LGTM!

dalonsoa (approved):
Well, `mypy` can be annoying, but it is easy to silence it and, in this case, it is justified and clear, so ignoring it is not dangerous.

PR #179, opened 2023-02-16
Feature/radiation dummy 2

vgro:
After extensive discussions in #155 and #130, and the number of files that changed every time we used the intermediate feature branch, we decided to now merge the radiation component directly in develop.

This version includes the `radiation.py` and corresponding tests (with increased range of values to test and test for errors) as well as the api components.

What is still missing is the link to the `config`to include the number of `canopy_layers` in the setup and there is no time dimension included. A number of additional sanity checks could also be added, but I left it for now as comments because this might come with a general checker. I think it would be easier to implement these changes or further structural changes as smaller pull requests rather than leaving these big PRs open for much longer.

P.S. the link to the `energy_balnace` raises a warning, please ignore.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,340 @@
...
+    * elevation: elevation above sea level, [m]
+    * shortwave_in: downward shortwave radiation, [J m-2]
+    * canopy_temperature: canopy temperature of individual layers, [C]
+    * surface_temperature: surface soil temperature, [C]
+    * canopy_absorption: absorption by canopy, [J m-2]
+
+    The `data` object can also optionally provide these variables, but will default to
+    the values given below:
+    * albedo_vis: visible light albedo, default = 0.03
+    * albedo_shortwave: shortwave albedo, default = 0.17
```
alexdewar, line 97: Why were these defaults chosen?

I'd be inclined to make these values constants in the module (e.g. `DEFAULT_ALBEDO_VIS` etc.) and not refer to the values here explicitly, because inevitably at some point someone will change them and forget to update the docstring.

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,340 @@
...
+        # Set the default values if variables not provided in data
+        sunshine_fraction: Union[DataArray, float]
+        if "sunshine_fraction" not in data:
+            sunshine_fraction = 1
+        else:
+            sunshine_fraction = data["sunshine_fraction"]
+
+        albedo_vis: Union[DataArray, float]
+        if "albedo_vis" not in data:
+            albedo_vis = 0.03
```
alexdewar, line 131: See above

virtual_rainforest/models/abiotic/radiation.py
```diff
@@ -0,0 +1,340 @@
...
+        topofcanopy_radiation: top of canopy radiation shortwave radiation, [J m-2]
+        canopy_absorption: shortwave radiation absorbed by canopy layers, [J m-2]
+        longwave_canopy: longwave radiation from canopy layers, [J m-2]
+        longwave_soil: longwave radiation from soil layers, [J m-2]
+
+    Returns:
+        net shortwave radiation at the surface ( = forest floor), [J m-2]
+    """
+
+    # TODO raise error if canopy_layers don't have correct dimensions, not here?
```
alexdewar, line 334: If `canopy_layers` doesn't have the correct dimensions, won't you get an error anyway?

alexdewar (approved):
I've made some minor suggestions, but none of them are terribly important, so feel free to ignore them.

I'm not totally convinced that `Radiation` needs to be a class as it stands: it doesn't have any methods other than `__init__` atm and usually you want to *do* something with objects. As I think I said before, I wouldn't bother saving `ppfd` etc. if it's cheap to calculate -- you could add a member function to calculate it as needed instead. But as I said, not too important.

docs/source/index.md
```diff
@@ -76,8 +76,7 @@ team.
 .. toctree::
   :maxdepth: 4
   :caption: API reference
-  
-  Main <api/main.md>
+
```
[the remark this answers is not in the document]
vgro: yes it was, well spotted :-)

virtual_rainforest/models/abiotic/radiation.py
vgro: probably yes. I was just thinking of what could go wrong...

alexdewar: You could add a test to check that the error *is* raised if a bad `canopy_layers` is passed in....

vgro:
> I've made some minor suggestions, but none of them are terribly important, so feel free to ignore them.
> 
> I'm not totally convinced that `Radiation` needs to be a class as it stands: it doesn't have any methods other than `__init__` atm and usually you want to _do_ something with objects. As I think I said before, I wouldn't bother saving `ppfd` etc. if it's cheap to calculate -- you could add a member function to calculate it as needed instead. But as I said, not too important.

I agree, this is something that became obvious as the code developed. I will leave it as it is for now but will consider your suggestions when we integrate this in the `AbioticModel`; I commented that a lot of this might go in the initialization step

vgro:
Merge first version of radiation module in develop. Might need restructuring to integrate in `AbioticModel`, but this will be a separate PR.

PR #181, opened 2023-02-16
Added docs build check to the CI

jacobcook1995:
# Description

This pull request adds a task to the CI to check that the docs build correctly. This has been added as a separate job and runs using `python 3.9` on `ubuntu`. The documentation build is run using `-W` which converts all warnings to errors, this ensures that the build process fails for broken links in the documentation.

This pull request is being pushed with docs build errors not fixed as I couldn't find a simple fix for them. This problem relates to the `GRID_STRUCTURE_SIG` object in `core/grid.py`. `mypy` is quite happy with how this object is used, but `sphinx` build produces a number of warnings about being unable to find `GRID_STRUCTURE_SIG.__repr__`, `GRID_STRUCTURE_SIG.count`, or `GRID_STRUCTURE_SIG.index`. Interestingly these errors disappear if you build the docs using `python 3.11` (though that introduces another error, which may be easier to fix).

Let me know if you have any ideas how this could be fixed. Worse case scenario the specific problems could be added to `nitpick_ignore` to suppress the warnings.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,9 @@ jobs:
 
       - name: Install dependencies
         run: poetry install
+
+      - name: Set ipython kernel
+        run: poetry run python -m ipykernel install --user --name=vr_python3
       
       - name: Build docs using sphinx
         run: poetry run sphinx-build -W --keep-going docs/source build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,4 +63,4 @@ jobs:
         run: poetry install
       
       - name: Build docs using sphinx
-          run: "poetry sphinx-build -W --keep-going . build"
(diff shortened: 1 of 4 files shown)


davidorme:
The lines causing the issue are here:

https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/c3117195d2f99eaadad33136a7b35447b4d438d5/virtual_rainforest/core/grid.py#L32-L33

The intention there is to create a common signature type that can be used across functions, but it may be the wrong way of doing things. The errors suggest to me that `sphinx` is trying to create links for this object as if it is a `tuple` rather than a type.

davidorme (approved):
LGTM - although I have no solution for the remaining signature issues.

jacobcook1995:
Yep that works! Bizarre, thanks a lot!

alexdewar (approved):

PR #183, opened 2023-02-22
Feature/wind dummy

vgro:
This is the first version of a wind submodule. In contrast to the other submodules of the abiotic model, it is relatively simple - it does not have a class structure but only a series of functions that are combined to calculate above and within canopy wind profiles.

At the moment, it is not 100% clear how the vertical axis is handled, i.e. how we combine above canopy heights and within canopy heights to fit on one axis with everything pointing in the right directions.

I have also not considered all possible ways in which the input could be wrong/sanity checks, but this can follow here or will be handled with the bounds checking.

The bib references are like to break once we updated the way we load references.

P.S. I also had to fix a few broken links related to radiation and energy balance, you can skip over this.

## Type of change

## Key checklist

## Further checks


tests/models/abiotic/test_wind.py
```diff
@@ -0,0 +1,227 @@
...
+            dims=["cell_id", "canopy_layers"],
+        ),
+    )
+
+
+def test_calculate_wind_profile(dummy_data):
+    """Test wind profile above and within canopy.
+
+    PROBLEM: all wind profiles go to zero with this set of data, need to find out why.
+    """
```
vgro, line 198: This test needs more digging, there is something going wrong in the calculations. I set the assert to zero only to get to the end of errors, this is not what it is supposed to look like. Will update asap.

docs/source/conf.py
```diff
@@ -78,6 +78,7 @@
     ("py:class", "numpy.int64"),
     # TODO - Delete this once Vivienne has merged this feature into develop
     ("py:class", "virtual_rainforest.models.abiotic.Energy_balance"),
```
jacobcook1995, line 80: Could probably delete this line as you changed the links in your docstrings to `virtual_rainforest.models.abiotic.energy_balance.Energy_balance`, so this one isn't required any more

virtual_rainforest/models/abiotic/wind.py
```diff
@@ -0,0 +1,551 @@
+"""The `abiotic.wind` module.
```
jacobcook1995, line 1: With the new documentation setup you shouldn't put a heading for module docstrings as `sphinx` adds a heading itself. The `# noqa: D205, D415` flags at the bottom of the docstring stop `flake8` from complaining about this

virtual_rainforest/models/abiotic/wind.py
```diff
@@ -0,0 +1,551 @@
...
+        wind_profile_above: wind profile above canopy, [m s-1]
+        wind_attenuation_coefficient: wind attenuation coefficient, dimensionless
+        canopy_height: canopy height, [m]
+
+    Returns:
+        wind profile below canopy, [m s-1]
+
+    # TODO: find correct fill value
+    """
+    topofcanopy_wind_speed = wind_profile_above[:, -1]
```
jacobcook1995, line 229: Feel this should be `top_of_canopy_wind_speed` or `wind_speed_canopy_top` if you need it shorter

tests/models/abiotic/test_wind.py
```diff
@@ -0,0 +1,227 @@
+"""Test module for wind.py."""
+
+# from contextlib import nullcontext as does_not_raise
```
jacobcook1995, line 3: You could probably delete this rather than commenting it out

jacobcook1995 (approved):
I just had a few minor comments

TaranRallings (approved):
LGTM!

virtual_rainforest/models/abiotic/wind.py
```diff
@@ -0,0 +1,553 @@
...
+    * friction velocity
+    * 10m wind speed
+    * 2m temperature
+    * atmospheric pressure
+    * canopy height
+    * leaf area index
+    * sensible heat flux at the top of the canopy
+
+    The ``const`` argument takes an instance of class
+    :class:`~virtual_rainforest.models.abiotic.radiation.RadiationConstants`, which
```
dalonsoa, line 91:
```suggestion
    :class:`~virtual_rainforest.models.abiotic.wind.WindConstants`, which
```

virtual_rainforest/models/abiotic/wind.py
```diff
@@ -0,0 +1,553 @@
...
+                    -np.sqrt(
+                        zero_plane_scaling_parameter * plant_area_index_displacement
+                    )
+                )
+            )
+            / np.sqrt(zero_plane_scaling_parameter * plant_area_index_displacement)
+        )
+        * canopy_height,
+        dims="cell_id",
+    )
```
dalonsoa, line 282: The maths here are obscured because of `black` breaking down the lines in a pretty insane way. I would suggest to calculate the two `sqrt` separately, assigning them to a (short-named) variable, and then use those in the calculation. Hopefully, the layout will be more readable.

dalonsoa (changes requested):
Looks good! I've just made a couple of comments/corrections. My only doubt at the moment is how this integrates with the workflow of the application, given that it is not a class like other models.

dalonsoa (approved):
Looks good!

PR #184, opened 2023-02-22
Switched to using a consistent key format for citations

jacobcook1995:
# Description

This pull request changes the bibliography file (`refs.bib`) to use the `BibTeX` output of our shared Zotero reference library (found [here](www.zotero.org/virtual_rainforest/)). The links have been updated to match this new system of tags. 

In future, whenever items are added to the bibliography they should be added to the shared Zotero library and the `refs.bib` file should be redownloaded and used to replace the previous `refs.bib` file.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,10 +13,14 @@ documentation root, use os.path.abspath to make it absolute, like shown here.
 
 import os
 import sys
+from dataclasses import dataclass, field
 
 # Import Matplotlib to avoid this message in notebooks:
 # "Matplotlib is building the font cache; this may take a moment."
 import matplotlib.pyplot  # noqa: F401
+import sphinxcontrib.bibtex.plugin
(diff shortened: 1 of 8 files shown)


davidorme (approved):
LGTM. Not entirely sure that I like Zotero's default keys ('name_firstwordoftitle_year') but that is just personal preference!

jacobcook1995:
Yeah I can't find any setting that allows us to configure the format of the keys that `Zotero` spits out, so I guess we are stuck with them

vgro:
> Yeah I can't find any setting that allows us to configure the format of the keys that `Zotero` spits out, so I guess we are stuck with them

It is unusual but I think it might become handy in case we have several papers published in the same year by one author, the a b format always causes a bit of a mess. This way we can avoid this (assuming they start their papers with different words ;-) )

vgro (approved):
Looks good!

PR #186, opened 2023-02-28
abiotic_tools introduced

vgro:
I found a number of functions and universal constants that will be used by multiple submodules of the `abiotic_model`.
I introduced a new submodule called `abiotic_tools` which contains them all and can be accessed easily by other submodules. 

The `AbioticConstants` might eventually go to a central point or can be accessed by others, for now it just makes the `abiotic` part of the model easier to manage.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/abiotic/test_abiotic_tools.py
+++ b/tests/models/abiotic/test_abiotic_tools.py
@@ -1,0 +1,54 @@
+"""Test abiotic_tools.py."""
+
+import pytest
+import xarray as xr
+from xarray import DataArray
+
+
+@pytest.fixture
+def dummy_data():
+    """Creates a dummy data object for use in wind tests.
+
+    One grid cell has no vegetation, two grid cells represent a range of values.
+    """
+
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    # Setup the data object with two cells.
+    grid = Grid(cell_nx=3, cell_ny=1)
+    data = Data(grid)
+
+    # Add the required data.
+    data["temperature_2m"] = DataArray([30, 20, 30], dims=["cell_id"])
+    data["atmospheric_pressure"] = DataArray([101, 102, 103], dims=["cell_id"])
+
+    return data
+
+
+def test_calc_molar_density_air(dummy_data):
+    """Test calculate molar desity of air."""
+
+    from virtual_rainforest.models.abiotic import abiotic_tools
+
+    result = abiotic_tools.calc_molar_density_air(
+        temperature=dummy_data.data["temperature_2m"],
+        atmospheric_pressure=dummy_data.data["atmospheric_pressure"],
+    )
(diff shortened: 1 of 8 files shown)


tests/models/abiotic/test_abiotic_tools.py
```diff
@@ -0,0 +1,54 @@
+"""Test abiotic_tools.py."""
+
```
vgro, line 2: The tests are cut and paste from the `wind` submodule, so they already went through PR process.

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -0,0 +1,70 @@
+"""The ``models.abiotic.models`` module contains a set of general functions and
```
jacobcook1995, line 1: "models.abiotic.abiotic_tools"?

jacobcook1995 (approved):
This seems like a sensible idea, I noticed a typo, but otherwise looks good to me!

virtual_rainforest/models/abiotic/abiotic_tools.py
vgro: Thanks, I noticed a few of them and made a small update

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -0,0 +1,70 @@
+"""The ``models.abiotic.abiotic_tools`` module contains a set of general functions and
+universal constants that are shared across submodels in the abiotic model.
```
davidorme, line 2: Is this deliberately not a link? It might well be, because it would be a link to itself, basically, but just checking.

davidorme (approved):
The only thing I'd query here is that `abiotic_tools` is a bit vague. You basically have `functions` and `constants`. I could definitely see us having a standard structure within models of having `model.constants` and `model.functions`. 

I don't think that should stop this PR - the aim of centralising those objects is fine - but we might revisit how these are arranged. For example, one thing might be that all constants classes typically go in a file (Wind and Radiation going with the general ones). There's also a case for them staying with their classes though.

You could then put

virtual_rainforest/models/abiotic/abiotic_tools.py
vgro: Yes, I followed the syntax from other modules here. I could add a link to the `abiotic_model`, but I think that can wait until I had a think about what else to write in the docstring that is a bit more descriptive. For know it's just to continue with other things more easily.

We should definitely have a general discussion about constants and where they should go, at the moment each abiotic submodule has a data class at the top but this is not meant to stay there, it could well go in an `abiotic.constants` file that constains all data classes.

vgro:
Further discussion on uniform structure of constants needed, see #129 and https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/186#pullrequestreview-1317041139

PR #187, opened 2023-02-28
Change SoilModel to use the data object

jacobcook1995:
# Description

This pull request alters the `SoilModel` so that it uses the `data` object to initialise itself and updates the data object as it goes. As part of this, the soil carbon pools are no longer attributes of the `SoilCarbonPools` class, instead these soil carbon pools are only stored as part of the data object. This is mainly for the sake of code clarity, as changes to the attributes (initialised from `data`) would change the `data` object for everyone. This still happens now but the code is explicit about it happening.
 
This PR also sets the logger level used by `pytest` has been to `DEBUG` globally, in order to ensure that debug messages are always seen by `pytest`. Previously, they were seen if `pytest` was run on the entire repo, but not if it was run on individual testing files. This option can be overridden in individual test functions or files, but if it proves to be problematic we can set it to another level.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,8 +18,7 @@ def log_check(caplog: pytest.LogCaptureFixture, expected_log: tuple[tuple]) -> N
 
     Arguments:
         caplog: An instance of the caplog fixture
-        expected_log: An iterable of 2-tuples containing the
-            log level and message.
+        expected_log: An iterable of 2-tuples containing the log level and message.
     """
 
     assert len(expected_log) == len(caplog.records)
@@ -112,7 +111,10 @@ def dummy_carbon_data():
     grid = Grid(cell_nx=4, cell_ny=1)
     data = Data(grid)
 
-    # Add the required data.
+    # The required data is now added. This includes the two carbon pools: mineral
+    # associated organic matter and low molecular weight carbon. It also includes
+    # various factors of the physical environment: pH, bulk density, soil moisture, soil
+    # temperature, percentage clay in soil.
     data["mineral_associated_om"] = DataArray([2.5, 1.7, 4.5, 0.5], dims=["cell_id"])
     """Mineral associated organic matter pool (kg C m^-3)"""
     data["low_molecular_weight_c"] = DataArray(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -98,6 +98,29 @@ def data_instance():
 
 
 @pytest.fixture
+def dummy_carbon_data():
+    """Creates a dummy carbon data object for use in tests."""
+
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    # Setup the data object with four cells.
+    grid = Grid(cell_nx=4, cell_ny=1)
+    data = Data(grid)
+
+    # Add the required data.
+    data["maom"] = DataArray([2.5, 1.7, 4.5, 0.5], dims=["cell_id"])
+    data["lmwc"] = DataArray([0.05, 0.02, 0.1, 0.005], dims=["cell_id"])
+    data["pH"] = DataArray([3.0, 7.5, 9.0, 5.7], dims=["cell_id"])
+    data["bulk_density"] = DataArray([1350.0, 1800.0, 1000.0, 1500.0], dims=["cell_id"])
+    data["soil_moisture"] = DataArray([0.5, 0.7, 0.6, 0.2], dims=["cell_id"])
+    data["soil_temperature"] = DataArray([35.0, 37.5, 40.0, 25.0], dims=["cell_id"])
+    data["percent_clay"] = DataArray([80.0, 30.0, 10.0, 90.0], dims=["cell_id"])
+
+    return data
+
+
+@pytest.fixture
 def new_axis_validators():
     """Create new axis validators to test methods and registration."""
     from virtual_rainforest.core.axes import AxisValidator
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -113,9 +113,11 @@ def dummy_carbon_data():
     data = Data(grid)
 
     # Add the required data.
-    data["maom"] = DataArray([2.5, 1.7, 4.5, 0.5], dims=["cell_id"])
+    data["mineral_associated_om"] = DataArray([2.5, 1.7, 4.5, 0.5], dims=["cell_id"])
     """Mineral associated organic matter pool (kg C m^-3)"""
-    data["lmwc"] = DataArray([0.05, 0.02, 0.1, 0.005], dims=["cell_id"])
+    data["low_molecular_weight_c"] = DataArray(
+        [0.05, 0.02, 0.1, 0.005], dims=["cell_id"]
+    )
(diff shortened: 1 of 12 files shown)


jacobcook1995:
Oh yeah it's probably worth mentioning that the way that the `data` object gets updated is a bit of a placeholder. At the moment the value is simply overwritten, down the line we either want to output data at each step (i.e. before it gets overwritten), or want to make the `data` object into a time series.

tests/conftest.py
```diff
@@ -97,6 +101,31 @@ def data_instance():
...
+
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    # Setup the data object with four cells.
+    grid = Grid(cell_nx=4, cell_ny=1)
+    data = Data(grid)
+
+    # Add the required data.
+    data["maom"] = DataArray([2.5, 1.7, 4.5, 0.5], dims=["cell_id"])
```
vgro, line 116: general comment on variable names, could you make them a bit more descriptive?

tests/conftest.py
```diff
@@ -1,12 +1,16 @@
...
 import pytest
 from xarray import DataArray
 
 # An import of LOGGER is required for INFO logging events to be visible to tests
 # This can be removed as soon as a script that imports logger is imported
-import virtual_rainforest.core.logger  # noqa
+from virtual_rainforest.core.logger import LOGGER
+
+# Class uses DEBUG
+LOGGER.setLevel(DEBUG)
```
alexdewar, line 13: I'm not sure about this... I really don't think you need to check that log statements at the debug level are emitted. The debug level is aimed at developers only so no one should be relying on its output. If you really want everyone to see it, use the info level and then it makes more sense to test it.

tests/conftest.py
alexdewar: The comment could be more descriptive too

tests/conftest.py
alexdewar: Tbf, I've just realised that this was the existing behaviour and that you've just moved things around. I don't think you guys want to check for debug logging though. Maybe open an issue/discussion?

virtual_rainforest/models/soil/carbon.py
```diff
@@ -29,71 +30,88 @@ class SoilCarbonPools:
...
+            data: The data object to be used in the model.
             pH: pH values for each soil grid cell
             bulk_density: bulk density values for each soil grid cell (kg m^-3)
             soil_moisture: relative water content for each soil grid cell (unitless)
             soil_temp: soil temperature for each soil grid cell (degrees C)
             percent_clay: Percentage clay for each soil grid cell
             dt: time step (days)
+
+        Returns:
+            A vector containing net changes to each pool. Order [lmwc, moam].
```
alexdewar, line 87:
```suggestion
            A vector containing net changes to each pool. Order [lmwc, maom].
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -57,35 +66,21 @@ def __init__(
...
         """A Data instance providing access to the shared simulation data."""
         self.update_interval
         """The time interval between model updates."""
         self.next_update
         """The simulation time at which the model should next run the update method"""
+        self.carbon = SoilCarbonPools(data)
+        """Set of soil carbon pools used by the model"""
+
+        # Save variables names to be used by the __repr__
+        # self._repr.append("")
```
alexdewar, line 83: What's happening here?

tests/core/data/all_config.toml
```diff
@@ -22,5 +22,4 @@ pft_name = "broadleaf"
 maxh = 50.0
 
 [soil]
-no_layers = 1
```
alexdewar, line 25: What's happened to `no_layers`?

alexdewar (approved):
It looks good from a cursory glance, but I'm not sure whether you said you were still having the problem of clobbering existing values in `Data`?  If so, I guess you'll need to solve that before merging.

virtual_rainforest/models/soil/soil_model.py
jacobcook1995: I used to append things to the `_repr`, so that they would show up when `__repr__` is called. I'm not doing this for any variable any more, but I left it in commented out to remind myself. It probably should just go

tests/core/data/all_config.toml
jacobcook1995: No longer planing for the `soil` module to use multiple soil layers, was basically told that this would be complete and utter nightmare to parameterise, so switching to a simpler model where there's a single homogenous layer where all the processes of interest happens. The `abiotic` model is planning to keep discrete soil layers though for calculation of energy and water transfers through soil (which I *think* are a bit easier to parameterise).

jacobcook1995:
In terms of the `data` update question, I've now set it up so that a single function updates the `data` object (`SoilCarbonPools.update_soil_carbon_pools`), and when it does so it updates the `data` object explicitly (rather than changing an attribute that was populated from the `data` object).

Down the line I might move this function into the `soil_model` module, so that soil model updates to the data object only happen within a single function, but that feels like too big a restructure to make midway through the review process of a PR.

tests/conftest.py
jacobcook1995: I've changed how both `maom` and `lmwc` are named in the `data` object (to `mineral_associated_om` and `low_molecular_weight_c`). I haven't changed them within my module as there they are contained in functions with docstrings that explain the acronyms. I guess the naming of variables within the `data` object is the most important place to be explicit as in there the variables are presented severed from the context they are generally used/defined/documented in.

tests/core/data/all_config.toml
alexdewar: Ok cool. I was just curious!

vgro (approved):
LGTM :-)

virtual_rainforest/models/soil/carbon.py
```diff
@@ -116,20 +137,28 @@ def mineral_association(
...
         temp_scalar = convert_temperature_to_scalar(soil_temp)
         moist_scalar = convert_moisture_to_scalar(soil_moisture)
 
-        return temp_scalar * moist_scalar * self.lmwc * (equib_maom - self.maom) / Q_max
+        return (
+            temp_scalar
+            * moist_scalar
+            * data["low_molecular_weight_c"]
+            * (equib_maom - data["mineral_associated_om"])
+            / Q_max
```
davidorme, line 153: It seems a little odd to have each variable passed in as a dataarray , _except_ for `maom` and `lwmc`, which are wrapped up within `data` still? Is there any reason not to pass those arrays in?

jacobcook1995: This is basically an artefact of `maom` and `lmwc` previously being attributes of `SoilCarbonPools`. When they stopped being attributes I changed how they were accessed without making them consistent with what had gone before.

However, I am tempted to leave them as is, as the way this data is accessed has already been changed on my integration branch. So, it might be more productive to have the DataArray vs NDArray discussion when the pull request for that is in?

PR #188, opened 2023-03-01
Add 'writable' DataArray accessor

dalonsoa:
# Description

I was thinking on adding this in a discussion point but it takes the same effort to make it in a draft PR and, if we all agree, developed it further into a proper PR. If not, we just close the PR.

Based on the conversations today, it seems advisable to protect the Data object from accidental writing given that all the DataArrays it contains are always passed by reference and modifying them somewhere will affect the calculations everywhere else.

This PRs makes **all** DataArrays added to the Data object read-only and implements a DataArray accessor that doubles as context manager and makes the chosen DataArray (any DataArray, not just those in the Data object) temporarily writable. It requires explicitly setting the context manager for the particular array to modify a requirement in order to modify it, reducing the chances - although not eliminating completely - of things being modified where they should not.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/accessors.py
+++ b/virtual_rainforest/core/accessors.py
@@ -1,0 +1,38 @@
+import xarray as xr
+
+
+@xr.register_dataarray_accessor("writable")
+class WritableArray:
+    def __init__(self, xarray_obj):
+        self._obj = xarray_obj
+        self._flag = xarray_obj.data.flags["WRITEABLE"]
+
+    def __enter__(self):
+        self._obj.data.setflags(write=True)
+        return self._obj
+
+    def __exit__(self, exc_type, exc_value, exc_tb):
+        self._obj.data.setflags(write=self._flag)
+
+
+if __name__ == "__main__":
+    import numpy as np
+
+    T = xr.DataArray(np.ones(3))
+    T.data.setflags(write=False)
+
+    try:
+        T[1] = 42
+    except:
+        print("T is not writable!")
+
+    with T.writable:
+        T[1] = 42
+        print("Now it is writable.")
+
+    try:
+        T[0] = 42
+    except:
+        print("T is not writable... again!")
+    
+    print(T)
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -209,6 +209,7 @@ class Data:
         # Validate and store the data array
         value, valid_dict = validate_dataarray(value=value, grid=self.grid)
         self.data[key] = value
+        self.data[key].data.setflags(write=False)
         self.variable_validation[key] = valid_dict
 
     def __getitem__(self, key: str) -> DataArray:


alexdewar:
Looks useful! Were you also thinking of adding a decorator that can be used to do the same thing?

dalonsoa:
You would need as input the DataArray you want to make writeable, so I'm not sure how this could be used in decorator form.

Besides, what would be the use case of having it in decorator form?

davidorme (commented):
I think this is neat and seems like clearly the best code solution for this issue.  I'm a bit on the fence about whether we adopt it. 

* Pros: We absolutely should not be altering data unless we are sure we are doing it at the right time. And I can also see that it could be easy to update an array, forgetting that it is - by reference - part of the shared data and this stops you doing that.

* Cons: we are running extra processing to stop us programming forgetfully. It's a useful sanity check, but - once the models are running correctly - I can imagine us wanting to turn this off. We could make the context manager and `write` flag setting respect a `SAFEMODE` setting, but that isn't going to stop the context manager having to run those methods. 

Also - it needs comments 😄  Can I check what I think is going on:

* `xarray` has built-in flags that we can use to make a variable read only. When we access a data array (either just using `data['var']` or explicitly extracting it using`var = data['var']`) that flag is preserved.

* `xarray` then has a system for extending what you can do with Datasets and DataArrays using 'accessors'. These basically sandbox custom behaviour rather than people hacking the core functions.

* The `WriteableArray`  defines an accessor for our data arrays that provides a context manager. When you use it:
    * The `with array.writable:` line initialises the context manager: it runs `__init__` and stores the data array and the current `write` flag value.  
    * It then calls the `__enter__` method - which sets the `write` flag to True.
    * Next, any code within the context manager block - such as setting the values - gets run.
    * Last, after the code and before moving to the code after the context manager, it runs the `__exit__` method, which restores the original value of the `write` flag.
    
So:

```python
with T.writable:   # calls WritableArray.__init__ and then WritableArray.__enter__
    T[1] = 0  # gets run, with write now set to True

# calls WritableArray.__exit__ and restores `write` to original state,
# Other stuff happens
```

dalonsoa:
Quick answers:

1. There's no comments, docstrings or tests, and I bypassed pre-commit because this is meant to be a discussion. That's what's also a Draft PR. If you give me the green light, I'll put all those things in place.
2. [Flags are a `numpy` feature](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.flags.html), not a `xarray` feature, that's why they are applied to the `data` attribute of the `DataArray`, which is the underlying numpy array.
3. For this reason, it obviously does not prevent a whole replacement of the DataArray - that brings a whole new numpy array to the table. If you want to prevent that, to some extend, put measures in place in the `__set__` method of the `Data` object to prevent overwriting an existing DataArray.
5.  This measure prevents you from making changes to entries when you should not (as it happened to @jacobcook1995 and as it was a concern for @vgro), but the real benefit is precisely for the future developers who will not have the broad vision and awareness that you all have now and can easily make those mistakes. So, if anything, I see this as something more useful in the future than now. Therefore, for me a safemode mode does not make sense - it defeats the purpose of setting this in the first place.

davidorme:
Aha - I hadn't fully got that it was setting the attribute on the `ndarray` via `data`.  I also wasn't expecting it to prevent (3) - just noting that this is a limitation.

On point 4 - yes, I agree that this is a future-proofing feature that helps developers get their models right. And that is really valuable, particularly because data being updated at the wrong time is a particularly evil kind of bug to track down: the predictions are wrong, but no errors are occurring. I think that has a huge amount of value. My concern is that it's a bit like training wheels on a bike: great to make sure you don't fall off to begin with, but eventually they start to get in the way. This PR would insist that all models use training wheels all the time.

I may be over-estimating the computational cost, but writing data is going to be a common activity. If a model has internal fast time scales, then it might be updating a lot and that could build up to be an appreciable cost.  In an ideal world, we'd like to be able to remove the context manager wrapping for trusted models, and that is why I mentioned a safe mode. 

I've been trying to think of ways that we could allow us to take those wheels off for trusted models, but I can't see a good approach to do it at any level.

* There doesn't seem to be much point in having a safe mode inside of the context manager. Once the update code is wrapped inside a context manager block then those special methods (`__init__`, `__enter__`, `__exit__`) have to run and there isn't a great deal to shortcut. 
* The only way to avoid running the context manager is to take the code outside of the manager, but  we definitely don't want something like this all over the place:
  
  ```python
  if SAFEMODE:
      with T.writable:
          T[1] = 0
  else:
      T[1] = 0
  ```

* We _could_ have a set of  'privileged' module names, but there is no easy way that the numpy array can know which module is writing to it. I think that the module identity would have to be passed at the level of individual attempts to edit:

  ```python
  # Trusted module
  T[1] = (0, 'trusted_module')
  
  # Untrusted module
  with T.writable:
      T[1] = 0
  ```
  This is genuinely closer to what I could see us wanting. Write the model using the context manager and then swap out the context managers once the model is trusted, but it seems clumsy and a really unintuitive use of '='.

*  Setting the `writable` flag as a global option simply doesn't work:
  
    ```python
    if SAFEMODE:
        self.data[key].data.setflags(write=False)
    ```
    When `SAFEMODE = False` then modules can use a mix of direct writing and context manager wrapped writing,  which does allow wheels to be taken off. But when `SAFEMODE = True`, all the trusted modules fail.

alexdewar:
> You would need as input the DataArray you want to make writeable, so I'm not sure how this could be used in decorator form.
> 
> Besides, what would be the use case of having it in decorator form?

Ahh, good point. I hadn't thought of that. I only mentioned it because I thought it came up in our discussion. But using a `with` block is probably clearer in any case.

davidorme:
Of _course_ it's an error to write data inefficiently! This PR wouldn't necessarily stop that, it would just be written inefficiently within a context manager. 

All I'm saying is that this PR introduces an unavoidable layer in our data API whose sole purpose is to make people stop and think that they are doing the right thing. And I think that's valuable, but I'm concerned that this layer can't easily be removed in production code.

alexdewar:
Right, but what I'm saying is how frequently are you realistically going to be using this context manager? For example, if you only needed to use it once a second, the code of setting and unsetting a flag is so negligible then it's not worth worrying about. And if it's 10,000 times a second, then you're doing something odd and should stop.

vgro (commented):
Apologies for the late reply, I was out of office for a while.
I think the general idea you suggest is very suitable for the virtual rainforest. I guess when it comes to the technical implementation, you are more experienced to judge what is the most efficient and transparent method to chose. About the performance, I agree that we should see how much it actually affects to flow, I don't think we will need to constantly read and write the data object.

dalonsoa:
So... any further comment on this? Shall I make it a proper contribution or do we ditch the idea?

I'm in no rush, but it's better to avoid having open PR without actioning them.

PR #189, opened 2023-03-02
Added utility function to extract model specific timing details

### LINKED ISSUE #180: Use second rather than minute units
Currently some steps of the model timing loop convert time units to minutes as a small unit to minimise rounding loss when converting from `datetime64`. The problem is that we should be using SI units where at all possible for consistency and minutes are not SI units

Replace all uses of minutes in the model time loops with seconds. These are currently found in `models/abiotic/abiotic_model.py`, `models/soil/soil_model.py` and `main.py`. They may also have to be fixed in `models/plants/plants_model.py` and `models/animals/animal_model.py` once they are added to `develop`.

jacobcook1995:
# Description

This pull request adds a utility function to extract the timing details based on @davidorme's suggestion. This utility function then gets used by each model class, instead of each class having nearly identical boiler plate code to extract the timing.

This pull request also addresses @alexdewar's comment about time units used by the Model classes not being SI by switching from using minutes to using seconds.

Fixes #180 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -85,6 +85,9 @@ from virtual_rainforest.core.logger import LOGGER
 # InitialisationError is a custom exception, for case where a `Model` class cannot be
 # properly initialised based on the data contained in the configuration
 from virtual_rainforest.core.base_model import BaseModel, InitialisationError
+
+# A utility function to unpack the model specific timing details from the config
+from virtual_rainforest.core.utility_functions import extract_model_time_details
 ```
 
 ### Defining the new class and class attributes
@@ -283,7 +286,7 @@ The job of the `from_config` method is to take that dictionary, along with the s
 configuration into the arguments required by the `__init__` method.
 
 The method then uses those parsed arguments to actually call the `__init__` method and
-return an initialised instance of the model using the setttings. The `from_config`
+return an initialised instance of the model using the settings. The `from_config`
 method should raise an `InitialisationError` if the configuration fails.
 
 As an example:
@@ -311,10 +314,9 @@ def from_config(cls, config: dict[str, Any]) -> FreshWaterModel:
         raw_interval = pint.Quantity(config["freshwater"]["model_time_step"]).to(
             "minutes"
         )
-        # Round raw time interval to nearest minute
-        update_interval = timedelta64(int(round(raw_interval.magnitude)), "m")
-        start_time = datetime64(config["core"]["timing"]["start_time"])
-        no_of_pools = config["freshwater"]["no_of_pools"]
+        start_date, update_interval = extract_model_time_details(
+                config, cls.model_name
+            )
     # Catch cases where Values or dimensions are wrong
     except (
         ValueError,
--- a/virtual_rainforest/core/utility_functions.py
+++ b/virtual_rainforest/core/utility_functions.py
@@ -1,0 +1,6 @@
+"""The ``core.utility_functions`` module contains functions that are used across the
+Virtual Rainforest, but which don't have a natural home in a specific module. At the
+moment, this module only contains a single function, but it will probably expand in
+future. Adding functions here can be a good way to reduce the amount boiler plate code
+generated for tasks that are repeated across modules.
+"""  # noqa: D205, D415
--- a/virtual_rainforest/core/utility_functions.py
+++ b/virtual_rainforest/core/utility_functions.py
@@ -29,9 +29,9 @@ def extract_model_time_details(
         pint.errors.UndefinedUnitError: If the unit is not known to pint.
     """
 
-    raw_interval = pint.Quantity(config[model_name]["model_time_step"]).to("minutes")
+    raw_interval = pint.Quantity(config[model_name]["model_time_step"]).to("seconds")
     # Round raw time interval to nearest minute
-    update_interval = timedelta64(int(round(raw_interval.magnitude)), "m")
+    update_interval = timedelta64(int(round(raw_interval.magnitude)), "s")
 
     start_date = datetime64(config["core"]["timing"]["start_date"])
 
--- a/virtual_rainforest/core/utility_functions.py
+++ b/virtual_rainforest/core/utility_functions.py
@@ -4,3 +4,37 @@ moment, this module only contains a single function, but it will probably expand
 future. Adding functions here can be a good way to reduce the amount boiler plate code
 generated for tasks that are repeated across modules.
 """  # noqa: D205, D415
+
+from typing import Any
+
+import pint
+from numpy import datetime64, timedelta64
+
+
+# TODO - WRITE TEST FOR THIS
+# TODO - IMPROVE MODEL SETUP DOCUMENTATION TO REFLECT THIS
+def extract_model_time_details(
+    config: dict[str, Any], model_name: str
+) -> tuple[datetime64, timedelta64]:
+    """Function to extract the timing details required to setup a specific model.
+
+    Args:
+        config: The configuration for the Virtual Rainforest simulation.
+        model_name: The name of the specific model of interest.
+
+    Returns:
+        A tuple containing the start date and the update interval for the model
+
+    Raises:
+        pint.errors.DimensionalityError: If the desired conversion between units isn't
+        possible, e.g. kg to s.
+        pint.errors.UndefinedUnitError: If the unit is not known to pint.
+    """
+
+    raw_interval = pint.Quantity(config[model_name]["model_time_step"]).to("minutes")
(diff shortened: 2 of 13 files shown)


TaranRallings (approved):
LGTM!

virtual_rainforest/core/utility_functions.py
```diff
@@ -0,0 +1,38 @@
...
+        pint.errors.UndefinedUnitError: If the unit is not known to pint.
+    """
+
+    raw_interval = pint.Quantity(config[model_name]["model_time_step"]).to("seconds")
+    # Round raw time interval to nearest minute
+    update_interval = timedelta64(int(round(raw_interval.magnitude)), "s")
+
+    start_date = datetime64(config["core"]["timing"]["start_date"])
+
+    return start_date, update_interval
```
davidorme, line 38: I think it might be worth moving the logging in here too? At the moment, the `from_config` methods are building up one long monolithic `try` for config errors. That works but as and when `utils` functions grow, it will probably be easier to keep some logical sections: core `util` checking, model specific checking. That also makes logging messages a bit less generic? If we did this:

```suggestion
def extract_model_time_details(
    config: dict[str, Any], model_name: str
) -> tuple[datetime64, timedelta64]:
    """Function to extract the timing details required to setup a specific model.

    Args:
        config: The configuration for the Virtual Rainforest simulation.
        model_name: The name of the specific model of interest.

    Returns:
        A tuple containing the start date and the update interval for the model

    Raises:
        pint.errors.DimensionalityError: If the desired conversion between units isn't
            possible, e.g. kg to s.
        pint.errors.UndefinedUnitError: If the unit is not known to pint.
    """

    try:
        raw_interval = pint.Quantity(config[model_name]["model_time_step"]).to("seconds")
        # Round raw time interval to nearest minute
        update_interval = timedelta64(int(round(raw_interval.magnitude)), "s")

        start_date = datetime64(config["core"]["timing"]["start_date"])

    except (pint.errors.DimensionalityError, pint.errors.UndefinedUnitError) as excep:
        LOGGER.error("Model timing error:" + excep)
        raise excep
```

Then `from_config` could all start:

```
valid_input = True
try:
    start_time, update_interval = extract_model_time_details(
        config, cls.model_name
    )
except Exception:
     valid_input = False
```

And then move on to model specific stuff? There's no way to avoid that `try` block, because of the `valid_input` sentinel. Well, unless we made `extract_model_time_details` define it:

```
valid_input,  start_time, update_interval = extract_model_time_details(
        config, cls.model_name
    )
```

Then `extract_model_time_details` would return `True, datetime64, timedelta64` or `False, None, None`. But although shorter, that seems confusing.

virtual_rainforest/core/utility_functions.py
```diff
@@ -0,0 +1,38 @@
+"""The ``core.utility_functions`` module contains functions that are used across the
+Virtual Rainforest, but which don't have a natural home in a specific module. At the
```
davidorme, line 2: I've a mild preference for `virtual_rainforest.core.utils` but only really for brevity.

docs/source/development/defining_new_models.md
```diff
@@ -311,10 +314,9 @@ def from_config(cls, config: dict[str, Any]) -> FreshWaterModel:
...
             "minutes"
         )
-        # Round raw time interval to nearest minute
-        update_interval = timedelta64(int(round(raw_interval.magnitude)), "m")
-        start_time = datetime64(config["core"]["timing"]["start_time"])
-        no_of_pools = config["freshwater"]["no_of_pools"]
+        start_date, update_interval = extract_model_time_details(
+                config, cls.model_name
+            )
     # Catch cases where Values or dimensions are wrong
```
davidorme, line 320: Was `no_of_pools` meant to go? It is an example of a model specific config check? If we do switch to a pattern of separating the timing checks off in their own `try` (see other comments), then having an example of then checking model specific bits would be good.

davidorme (changes requested):
No real problems, some minor comments about how to structure checking within `from_config` that might impact this PR.

virtual_rainforest/core/utility_functions.py
jacobcook1995: Yeah I was wondering if logging would be better inside the function, agree that what you've proposed makes sense as it minimises code complexity in the individual model classes, which was kind of the point of this pull request

virtual_rainforest/core/utility_functions.py
jacobcook1995: Yeah if that is a fairly standard shortening I'm happy to use it

docs/source/development/defining_new_models.md
jacobcook1995: Nope, was not meant to go!

virtual_rainforest/core/utils.py
```diff
@@ -0,0 +1,46 @@
...
+            possible, e.g. kg to s.
+        pint.errors.UndefinedUnitError: If the unit is not known to pint.
+    """
+
+    try:
+        raw_interval = pint.Quantity(config[model_name]["model_time_step"]).to(
+            "seconds"
+        )
+        # Round raw time interval to nearest minute
+        update_interval = timedelta64(int(round(raw_interval.magnitude)), "s")
```
alexdewar, line 39: `round()` already casts to `int`:

```suggestion
        update_interval = timedelta64(round(raw_interval.magnitude), "s")
```

virtual_rainforest/core/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+        start_date = datetime64(config["core"]["timing"]["start_date"])
+    except (pint.errors.DimensionalityError, pint.errors.UndefinedUnitError) as excep:
+        LOGGER.error("Model timing error: %s" % str(excep))
+        raise excep
```
alexdewar, line 44: These errors basically signify that there's something wrong with the configuration, right? If so, I'd be inclined to use a custom `ConfigurationError` class (or something) that you can share with other parts of the core code.

virtual_rainforest/models/abiotic/abiotic_model.py
```diff
@@ -113,25 +113,14 @@ def from_config(cls, data: Data, config: dict[str, Any]) -> AbioticModel:
...
-            update_interval = timedelta64(int(round(raw_interval.magnitude)), "m")
-            start_time = datetime64(config["core"]["timing"]["start_time"])
-            soil_layers = config["abiotic"]["soil_layers"]
-            canopy_layers = config["abiotic"]["canopy_layers"]
-        except (
-            ValueError,
-            pint.errors.DimensionalityError,
-            pint.errors.UndefinedUnitError,
-        ) as e:
+        except Exception:
```
alexdewar, line 119: I don't like this blanket catching of all exceptions here, as it seems a little bug prone. If someone introduces an bug into `extract_model_time_details` then the error message will be a bit confusing.

I'd just use a custom exception type, as described above.

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -74,23 +74,11 @@ def from_config(cls, data: Data, config: dict[str, Any]) -> AnimalModel:
...
             )
-            # Round raw time interval to nearest minute
-            update_interval = timedelta64(int(round(raw_interval.magnitude)), "m")
-            start_time = datetime64(config["core"]["timing"]["start_time"])
-        except (
-            ValueError,
-            pint.errors.DimensionalityError,
-            pint.errors.UndefinedUnitError,
-        ) as e:
+        except Exception:
```
alexdewar, line 80: Ditto above.

alexdewar (changes requested):
I think in general this is a sensible approach, but I'm slightly suspicious of the copied boilerplate code you now have in each of your models. Is there a better way you could abstract this?

virtual_rainforest/core/utils.py
```diff
@@ -26,21 +27,19 @@ def extract_model_time_details(
...
         )
         # Round raw time interval to nearest minute
-        update_interval = timedelta64(int(round(raw_interval.magnitude)), "s")
+        update_interval = timedelta64(round(raw_interval.magnitude), "s")
 
         start_date = datetime64(config["core"]["timing"]["start_date"])
     except (pint.errors.DimensionalityError, pint.errors.UndefinedUnitError) as excep:
         LOGGER.error("Model timing error: %s" % str(excep))
-        raise excep
+        raise InitialisationError()
```
alexdewar, line 43: I'd keep this info as an inner exception:

```suggestion
        raise InitialisationError() from excep
```

alexdewar (approved):
Looks fine to me. Made one small suggestion.

PR #191, opened 2023-03-08
Numerical integration of the soil module

jacobcook1995:
# Description

This pull request adds functionality to integrate the soil model (or at least the simple dummy model that currently exists). This involves a function to set up and run the integration process, and a function that combines the existing functions into a form suitable for integration using `scipy`'s `solve_ivp` function.

One thing that definitely needs thought is where DataArrays get used and where plain `numpy` arrays are used. I basically just adopted the structure that worked that was the smallest change from what I had previously. This might be an area where it's worth being more systematic as I am guessing this has performance implications (at least if data is being copied/altered).

This pull request adds `dotmap` as a dependency. I added it because I needed a container for mocking integration output for one of the tests. I considered using a `dataclass` rather than a `DotMap` for but it seemed like it would end up being significantly more longwinded. I am happy to try and implement this if adding `dotmap` as a dependancy is a problem.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+            raise IntegrationError()
+
+
+def construct_full_soil_model(
+    t: float, pools: np.ndarray, data: Data, no_cells: int
+) -> np.ndarray:
+    """Function that constructs the full soil model in a solve_ivp friendly form.
+
+    Args:
+        pools: And area containing all soil pools in a single vector
```
dalonsoa, line 250: Argument `t` is missing in the docstring and, actually, it is not used at all in the body of the auction. Is this an error or it is really not used for anything?

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+        pools[:no_cells],
+        pools[no_cells:],
+        data["pH"],
+        data["bulk_density"],
+        data["soil_moisture"],
+        data["soil_temperature"],
+        data["percent_clay"],
+    )
+
+    return pool_rate_of_changes
```
dalonsoa, line 268:
```suggestion
    return calculate_soil_carbon_updates(
        pools[:no_cells],
        pools[no_cells:],
        data["pH"],
        data["bulk_density"],
        data["soil_moisture"],
        data["soil_temperature"],
        data["percent_clay"],
    )
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+        # Convert from second to day units
+        t_span = (0.0, update_time / (60.0 * 60.0 * 24.0))
+
+        # Construct vector of initial values y0
+        y0 = np.concatenate(
+            [
+                self.data["low_molecular_weight_c"].to_numpy(),
+                self.data["mineral_associated_om"].to_numpy(),
+            ]
+        )
```
dalonsoa, line 215: In general, you will have more than 2 pools, right? Could you define a naming convention for the pools such that they can be automatically found and concatenated, eg, all of them start with `c_pool_` or something along those lines? If you do that, you could use:

```suggestion
        y0 = np.concatenate(
            [
                self.data[name].to_numpy() 
                for name in self.data 
                if name.startswith("c_pool_")
            ]
        )
```

Naturally, `construct_full_soil_model` and functions called within will need to be adapted to use this convention. 

I understand this is a somewhat dummy/reduced soil model, but if can be designed such that it can be easily expanded with the real stuff when ready, that will save time down the line.

virtual_rainforest/models/soil/carbon.py
```diff
@@ -22,136 +20,100 @@
 # to track down a reliable parameterisation for this section.
 
 
-class SoilCarbonPools:
-    """Class containing the full set of soil carbon pools.
+def calculate_soil_carbon_updates(
+    low_molecular_weight_c: np.ndarray,
+    mineral_associated_om: np.ndarray,
```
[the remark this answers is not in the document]
dalonsoa: Ok, now I've read the full PR I understand why it is the way it is. In that case, I would use numpy arrays in all cases as you don't seem to be exploiting any of the `DataArray` functionality anyway.

dalonsoa (changes requested):
This is a good starting point, and you have definitely got the idea of wrapping the functionality to be used by the solver. I've just made a suggestion to make the code more generalisable down the line.

tests/models/soil/test_soil_model.py
```diff
@@ -198,17 +270,147 @@ def test_generate_soil_model(
...
+    assert np.allclose(dummy_carbon_data["mineral_associated_om"], end_maom)
+    assert np.allclose(dummy_carbon_data["low_molecular_weight_c"], end_lmwc)
+
+
+@pytest.mark.parametrize(
+    argnames=["output", "raises", "expected_log"],
+    argvalues=[
+        pytest.param(
+            DotMap(
+                success=True,
```
davidorme, line 338: I don't see that we need `DotMap` here. You can just have the `argnames` as:

```python
argnames=["success", "y", "msg", "raises", "expected_log"],
```

And then have `y = None` when it fails and `msg  = None` when it succeeds.

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+    """
+
+    pool_rate_of_changes = calculate_soil_carbon_updates(
+        pools[:no_cells],
+        pools[no_cells:],
+        data["pH"],
+        data["bulk_density"],
+        data["soil_moisture"],
+        data["soil_temperature"],
+        data["percent_clay"],
```
davidorme, line 265: In addition to @dalonsoa's change, I've a vague preferences for using named keywords, particularly with long arguments? Here, the data keys makes it pretty obvious, but being explicit about what the different inputs are seems more defensive?

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
         """Placeholder function to spin up the soil model."""
 
     def update(self) -> None:
-        """Placeholder function to update the soil model."""
+        """Function to update the soil model."""
```
alexdewar, line 150: This can be terser:

```suggestion
        """Update the soil model."""
```

I'd also say *how* you're updating the soil model.

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
         # Update carbon pools (attributes and data object)
         # n.b. this also updates the data object automatically
-        self.carbon.update_soil_carbon_pools(self.data, carbon_pool_updates)
+        self.replace_soil_pools(updated_carbon_pools)
 
         # Finally increment timing
         self.next_update += self.update_interval
 
     def cleanup(self) -> None:
         """Placeholder function for soil model cleanup."""
```
alexdewar, line 163: (I know this isn't part of this PR, but I've only just noticed it.)

Do all models have a `cleanup()` method? What's it needed for? You don't normally have to worry about manually cleaning up resources with Python.

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+
+        Args:
+            new_pools: Array of new pool values to insert into the data object
+        """
+
+        self.data["low_molecular_weight_c"] = new_pools["new_lmwc"]
+        self.data["mineral_associated_om"] = new_pools["new_maom"]
+
+    def integrate_soil_model(self) -> Dataset:
+        """Function to integrate the soil model.
```
alexdewar, line 183: We already know it's a soil model, so I think you can just call this `integrate()`. And, again, you can drop the "Function to..." bit.

```suggestion
    def integrate(self) -> Dataset:
        """Integrate the soil model.
```

davidorme (changes requested):
Looking good - and I think this is the only way to go. Some minor comments. I don't think we need `DotMap` but it isn't the most expensive thing in the world.

Oh _wait_. I've just got why it is being used - because you're mocking, you need to pass something like the real output of `solve_ivp` forward to `integrate_soil_model`. I have to admit that for a simple integration test, I'd just run `solve_ivp` rather than mocking it. If we do mock it, then we can import the actual expected return class, which is basically just `scipy`s own implementation of a dict with dot attribute access.

``` python
from scipy.integrate._ivp.ivp import OdeResult
# or avoiding the private special subclass
from scipy.optimize import OptimizeResult
```

and then we just define the expected `output` as:

```
OptimizeResult(
                success=True,
                y=np.array(
                    [
                        [5.000e-02, 3.210e-02],
                        [2.000e-02, 1.921e-02],
                        [4.500e00, 4.509e00],
                        [5.000e-01, 5.000e-01],
                        [3.210e-02, 5.000e-02],
                        [1.921e-02, 2.000e-02],
                        [4.509e00, 4.500e00],
                        [5.000e-01, 5.000e-01],
                    ]))
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+            IntegrationError: When the integration cannot be successfully completed.
+        """
+
+        # Find number of grid cells integration is being performed over
+        no_cells = self.data.grid.n_cells
+
+        # Extract update interval
+        update_time = self.update_interval.astype("timedelta64[s]").astype(float)
+        # Convert from second to day units
+        t_span = (0.0, update_time / (60.0 * 60.0 * 24.0))
```
alexdewar, line 207: I *think* you can do this:

```suggestion
        t_span = (0.0, np.timedelta64(update_time, "D"))
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,122 @@ def spinup(self) -> None:
...
+                    new_lmwc=DataArray(output.y[:no_cells, -1], dims="cell_id"),
+                    new_maom=DataArray(output.y[no_cells:, -1], dims="cell_id"),
+                )
+            )
+        else:
+            LOGGER.error(
+                "Integration of soil module failed with following message: %s"
+                % str(output.message)
+            )
+            raise IntegrationError()
```
alexdewar, line 241: Seeing as both of these blocks exit the function scope, you don't need the else part. I'd also write the error condition first for clarity:

```suggestion
        # Check if integration failed
        if not output.success:
            LOGGER.error(
                "Integration of soil module failed with following message: %s"
                % str(output.message)
            )
            raise IntegrationError()

        return Dataset(
            data_vars=dict(
                new_lmwc=DataArray(output.y[:no_cells, -1], dims="cell_id"),
                new_maom=DataArray(output.y[no_cells:, -1], dims="cell_id"),
            )
        )
```

alexdewar (changes requested):
I've made some minor suggestions in places.

virtual_rainforest/models/soil/soil_model.py
davidorme: Not that this gives the floor of the number of days. Which might be appropriate, but equally it might be an error for this not to be an exact number of days. Although leap seconds...

virtual_rainforest/core/base_model.py
```diff
@@ -87,6 +87,10 @@ class InitialisationError(Exception):
     """Custom exception class for model initialisation failures."""
 
 
+class IntegrationError(Exception):
```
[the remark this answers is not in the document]
jacobcook1995: The reason this is defined here is linked to an annoying bug/feature of `sphinx`. If you define a custom exception in a particular script then it can't be used in docstrings of functions/classes in the same script without `sphinx` producing a "reference target not found" warning.

This warning can be fixed by importing the custom exception, i.e. by moving it to be defined in another script. It's a bit annoying to have to do this in general though, as often the natural place to define a custom exception it will be a place where it's raised by one of the functions.

One possible solution is to place all custom exceptions in a `core.exceptions` module, or for each module to have a `exceptions` module of its own. If either of those things seem like sensible things to do?

virtual_rainforest/models/soil/soil_model.py
jacobcook1995: The docs for `solve_ivp` won't accept `timedelta64` as an input.

I can get rid of that unpleasant hard coded sum by just calculating `update_time` as
```python
update_time = self.update_interval / np.timedelta64(1, "D")
```
which returns a float

davidorme: I've run into the same issue with `scipy.interpolate.interp1d`: the interpolation axis inputs need converting to floatlike things first. That division trick is very clean, although I still think timespans that aren't multiples of a day are probably trouble 😄

jacobcook1995: I don't think time steps that aren't multiple of days are a problem, e.g.

```python
from numpy import timedelta64

timedelta64(5, "h") / timedelta64(1, "D")
```

results in 0.20833333333333334.

Leap seconds are still a potential problem, but I presume that's a were are a few seconds wrong problem rather than something that causes `timedelta64` to catch fire and die?

davidorme: Oh sure,  the division works fine, but I'm just unsure whether models should be using, say,  an update spacing of 7.6734 days. 😄 I don't know if we want to trap that kind of thing.

jacobcook1995: I'm guessing if we want to trap it we would want to trap it higher in the simulation flow, trapping it here would involve trapping it for every model individually

virtual_rainforest/models/soil/soil_model.py
dalonsoa: Yes, the order will be as set in the `Data` object. But you can easily keep track of that order but looking at the `Data` object at any time, in particular, when reconstructing the output.

virtual_rainforest/core/base_model.py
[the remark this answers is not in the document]
dalonsoa: That's annoying... Maybe something to add to the next meeting on the architecture of the tool. Seems like an Sphinx bug (or at least, odd behaviour), so maybe we could raise the issue upstream.

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,165 @@ def spinup(self) -> None:
...
+                for name in self.data.data.keys()
+                if str(name).startswith("soil_c_pool_")
+            ]
+        )
+
+        # Find and store order of pools
+        pool_order = []
+        for name in self.data.data.keys():
+            if str(name).startswith("soil_c_pool_"):
+                pool_order.append(str(name))
```
dalonsoa, line 220:
```suggestion
        pool_order = [
                 str(name)
                 for name in self.data.data.keys()
                 if str(name).startswith("soil_c_pool_")
             ]
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,165 @@ def spinup(self) -> None:
...
+
+    # Construct index slices
+    slices = make_slices(no_cells, len(pool_order))
+
+    # Construct dictionary of numpy arrays (using a for loop)
+    soil_pools = {}
+    slice_no = 0
+    for pool in pool_order:
+        soil_pools[str(pool)] = pools[slices[slice_no]]
+        slice_no += 1
```
dalonsoa, line 283:
```suggestion
    soil_pools = {str(pool): pools[slc] for slc, pool in zip(slices , pool_order)}
```

dalonsoa (changes requested):
I've made a few suggestions - formatting will be wrong - to make the code more concise. I'm not an enthusiast of using `locals`, as you do in line 72, but I cannot think on a better, scalable method at the moment.

jacobcook1995:
I do agree that using `locals()` is a bit weird, but I couldn't come up with an alternative (apart from `vars()` which is either worse or much the same)

virtual_rainforest/models/soil/carbon.py
```diff
@@ -22,143 +20,114 @@
...
-        )
+    # Determine net changes to the pools
+    delta_pools_ordered["soil_c_pool_lmwc"] = -lmwc_to_maom
+    delta_pools_ordered["soil_c_pool_maom"] = lmwc_to_maom
+
+    # Create output array based on first pool
+    delta_pools = np.array([])
+    # Then append every subsequent pool
+    for pool in delta_pools_ordered.keys():
+        delta_pools = np.concatenate([delta_pools, delta_pools_ordered[pool]])
```
dalonsoa, line 73:
```suggestion
    delta_pools = np.concatenate(list(delta_pools_ordered.values()))
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -47,8 +54,8 @@ class SoilModel(BaseModel):
     model_name = "soil"
     """An internal name used to register the model and schema"""
     required_init_vars = (
```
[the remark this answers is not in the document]
jacobcook1995: In my case these don't have to be present in the `data` object to initialise the class as my class doesn't have any attributes. I've just put all the things that need to be in `data` for the functions used by the `SoilModel` to work in here. This does cause initialisation to fail when the `data` object technically contains everything needed to initialise the class, which might not be ideal behaviour.

I don't have any data for the soil variables as of yet, so they are not at present loaded, but they should be loaded along when the data object is created.

On the latter 2 questions, I haven't really given them much thought, and as far as I know there isn't any functionality in place to handle them as of yet.

tests/models/soil/test_soil_model.py
```diff
@@ -6,7 +6,7 @@
 
 import numpy as np
 import pytest
-from dotmap import DotMap  # type: ignore
+from scipy.optimize import OptimizeResult  # type: ignore
```
alexdewar, line 9: Why are you ignoring the type here?

tests/models/soil/test_soil_model.py
```diff
@@ -368,25 +363,80 @@ def test_replace_soil_pools(dummy_carbon_data, soil_model_fixture):
...
 
     # Check that integrator is called once (and once only)
-    mock_integrate.assert_called_once()
+    if mock_output:
+        mock_integrate.assert_called_once()
 
     log_check(caplog, expected_log)
 
 
+def test_order_independance(dummy_carbon_data, soil_model_fixture):
```
alexdewar, line 389: This is fine as is, but it could be part of a parametrized test, with an `order` argument that determines what order the data is in.

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -201,68 +200,109 @@ def integrate_soil_model(self) -> Dataset:
...
+        no_pools: Number of soil pools being integrated
+
+    Returns:
+        A list of containing the correct number of correctly spaced slices
+    """
+
+    # Construct index slices
+    slices = [slice(n * no_cells, (n + 1) * no_cells) for n in range(no_pools)]
+
+    return slices
```
alexdewar, line 308:
```suggestion
    return [slice(n * no_cells, (n + 1) * no_cells) for n in range(no_pools)]
```

alexdewar (changes requested):
I've got a couple of small queries/suggestions, but otherwise LGTM!

alexdewar (approved):
Sorry, I meant to approve before...

tests/models/soil/test_soil_model.py
jacobcook1995: As far as I can tell `scipy.optimize` is not typed even though other `scipy` modules are typed. I could have missed them but also couldn't track down type stubs

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -132,29 +147,162 @@ def spinup(self) -> None:
...
+        the new values for all soil module components can be calculated on a single
+        state, which is then updated (using this function) when all values have been
+        calculated.
+
+        Args:
+            new_pools: Array of new pool values to insert into the data object
+        """
+
+        self.data["soil_c_pool_lmwc"] = new_pools["soil_c_pool_lmwc"]
+        self.data["soil_c_pool_maom"] = new_pools["soil_c_pool_maom"]
```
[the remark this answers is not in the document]
davidorme: @jacobcook1995 Let me know if you want a chat about adding that time dimension! But I agree that it is fine to park that for now.

davidorme (approved):
LGTM

PR #192, opened 2023-03-08
Feature/animal module - Basic AnimalCohort functions

TaranRallings:
# Description

This pull request adds basic functionality to the AnimalCohort class: eat, excrete, birth, aging, and death. The functions themselves are quite simple but some of them do connect the AnimalCohorts to my toy implementations of other pools: plants, soil, carcasses. Included in this is a function, forage(), that combines two connected processes, eat and excrete, together into one function call. Let me know if there is a better way to approach this. 

I have also included some more realistic parameter values in constants.py. These are almost all values specific to terrestrial mammals but they at least get us in the ball park. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/models/animals/constants.py
+++ b/virtual_rainforest/models/animals/constants.py
@@ -12,7 +12,7 @@ class DamuthsLaw:
 
     exponent: float = -0.75
     """"""
-    coefficienct: float = 4.23
+    coefficient: float = 4.23
     """"""
 
 
@@ -22,7 +22,7 @@ class MetabolicRate:
 
     exponent: float = 0.75
     """"""
-    coefficienct: float = 10**-5
+    coefficient: float = 10**-5
     """"""
 
 
@@ -32,5 +32,5 @@ class StoredEnergy:
 
     exponent: float = 0.75
     """"""
-    coefficienct: float = 10
+    coefficient: float = 10
     """"""
--- a/virtual_rainforest/models/animals/constants.py
+++ b/virtual_rainforest/models/animals/constants.py
@@ -7,30 +7,20 @@ from dataclasses import dataclass
 
 
 @dataclass
-class DamuthsLaw:
-    """From (Damuth 1987)."""
+class Term:
+    """Dataclass for handling the terms of scaling equations."""
 
-    exponent: float = -0.75
-    """"""
-    coefficient: float = 4.23
-    """"""
-
-
-@dataclass
-class MetabolicRate:
-    """Toy values."""
-
-    exponent: float = 0.75
-    """"""
-    coefficient: float = 10**-5
-    """"""
+    exponent: float
+    """Handles the exponent on body-mass for scaling equations."""
+    coefficient: float
+    """Handles the coefficient of scaling equations"""
 
 
-@dataclass
-class StoredEnergy:
-    """Toy values."""
+"""From (Damuth 1987)."""
+DamuthsLaw = Term(-0.75, 4.23)
 
-    exponent: float = 0.75
-    """"""
-    coefficient: float = 10
-    """"""
+"""Toy values."""
+MetabolicRate = Term(0.75, 10**-5)
+
+"""Toy values."""
+StoredEnergy = Term(0.75, 10)
--- a/virtual_rainforest/models/animals/constants.py
+++ b/virtual_rainforest/models/animals/constants.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass
 
 
 @dataclass
-class Term:
+class ScalingTerms:
     """Dataclass for handling the terms of scaling equations."""
 
     exponent: float
@@ -16,11 +16,36 @@ class Term:
     """Handles the coefficient of scaling equations"""
 
 
-"""From (Damuth 1987)."""
-DamuthsLaw = Term(-0.75, 4.23)
+"""Mammalian herbivore population density, observed allometry (Damuth 1987). [kg]"""
+DamuthsLaw = ScalingTerms(-0.75, 4.23)
 
-"""Toy values."""
-MetabolicRate = Term(0.75, 10**-5)
+"""Temporary basal metabolic rate values for mammals (citation from Rallings). [g]"""
+MetabolicRate = ScalingTerms(0.75, 0.047)
 
-"""Toy values."""
-StoredEnergy = Term(0.75, 10)
+"""Scaling of mammalian herbivore fat mass (citation from Rallings). [g]"""
+FatMass = ScalingTerms(1.19, 0.02)
+
+"""Scaling of mammalian herbivore muscle mass (citation from Rallings).[g]"""
+MuscleMass = ScalingTerms(1.0, 0.38)
+
+"""Mammalian maximum intake rate (kg/day) from (Shipley 1994)."""
+IntakeRate = ScalingTerms(0.71, (0.63 * 480) * (1 / 1000))
+""" Converts original g/min rate to kg/day where 'day' is an 8hr foraging window."""
+
+
+@dataclass
+class Term:
+    """Dataclass for handling fixed parameter values."""
+
+    value: float
+    """Contains the float value of the parameter"""
+
+
+"""The energy of a unit mass of mammal meat (check citation from Rallings). [J/g]"""
+MeatEnergy = Term(7000.0)
+
+"""Toy value expressing how much of a consumed food is absorbed as energy. [unitless]"""
+ConversionEfficiency = Term(0.1)
+
+"""Temporary realistic plant food value: Alfalfa ¬ 18,200,000 J/kg DM."""
+PlantEnergyDensity = Term(18200000.0)
--- a/virtual_rainforest/models/animals/constants.py
+++ b/virtual_rainforest/models/animals/constants.py
@@ -1,0 +1,36 @@
+"""The `models.animals.constants` module contains a set of dataclasses containing
+constants" (fitting relationships taken from the literature) required by the broader
+:mod:`~virtual_rainforest.models.animals` module
+"""  # noqa: D205, D415
+
+from dataclasses import dataclass
+
+
+@dataclass
+class DamuthsLaw:
+    """From (Damuth 1987)."""
+
+    exponent: float = -0.75
+    """"""
+    coefficienct: float = 4.23
+    """"""
+
+
+@dataclass
+class MetabolicRate:
+    """Toy values."""
+
+    exponent: float = 0.75
+    """"""
+    coefficienct: float = 10**-5
+    """"""
+
+
+@dataclass
+class StoredEnergy:
+    """Toy values."""
+
+    exponent: float = 0.75
+    """"""
+    coefficienct: float = 10
+    """"""
--- a/virtual_rainforest/models/animals/dummy_animal_module.py
+++ b/virtual_rainforest/models/animals/dummy_animal_module.py
@@ -179,3 +179,12 @@ class AnimalCohort:
         """
         consumed_energy = self.eat(food)
         self.excrete(soil, consumed_energy)
+
+    def aging(self, dt: timedelta64) -> None:
+        """The function to modify cohort age as time passes.
+
+        Args:
+            dt: The number of days that should be added to cohort age.
+
+        """
+        self.age += float(dt / timedelta64(1, "D"))
--- a/virtual_rainforest/models/animals/dummy_animal_module.py
+++ b/virtual_rainforest/models/animals/dummy_animal_module.py
@@ -22,6 +22,8 @@ Notes to self:
 from dataclasses import dataclass
 from math import ceil
 
+from numpy import timedelta64
+
 from virtual_rainforest.core.logger import LOGGER
 from virtual_rainforest.models.animals.constants import (
     DamuthsLaw,
@@ -97,11 +99,17 @@ class AnimalCohort:
         self.stored_energy: float = (
             StoredEnergy.coefficient * self.mass**StoredEnergy.exponent
         )
-        """The current indiv energetic reserve [J]."""
+        """The current individual energetic reserve [J]."""
 
-    def metabolize(self) -> None:
-        """The function to reduce stored_energy through basal metabolism."""
-        energy_burned = 28800 * self.metabolic_rate
+    def metabolize(self, dt: timedelta64) -> None:
+        """The function to reduce stored_energy through basal metabolism.
+
+        Args:
+            dt: Number of days over which the metabolic costs should be calculated.
+
+        """
+
+        energy_burned = float(dt / timedelta64(1, "s")) * self.metabolic_rate
         """Number of seconds in a day * J/s metabolic rate, consider daily rate."""
         if self.stored_energy >= energy_burned:
             self.stored_energy -= energy_burned
--- a/virtual_rainforest/models/animals/dummy_animal_module.py
+++ b/virtual_rainforest/models/animals/dummy_animal_module.py
@@ -218,3 +218,18 @@ class AnimalCohort:
             LOGGER.warning("An animal cohort has died")
         elif not self.is_alive:
             LOGGER.warning("An animal cohort which is dead cannot die.")
+
+    def birth(self) -> "AnimalCohort":
+        """The function to produce a new AnimalCohort through reproduction.
+
+        Currently, the birth function returns an identical cohort of adults with age
+        0. In the future, the offspring will be modified to have appropriate juvenile
+        traits based on parental type. This will also include supplemental functionality
+        to attached the birthed cohort to relevant cohort tracking lists in AnimalModel
+        and Grid.
+
+        Returns:
+            An AnimalCohort instance having appropriate offspring traits for the
+            location and functional type of the parent cohort.
+        """
+        return AnimalCohort(self.name, self.mass, 0, self.position)
--- a/virtual_rainforest/models/animals/dummy_animal_module.py
+++ b/virtual_rainforest/models/animals/dummy_animal_module.py
@@ -85,17 +85,17 @@ class AnimalCohort:
         self.position = position
         """The grid position of the animal cohort."""
         self.individuals: int = ceil(
-            DamuthsLaw.coefficienct * self.mass ** (DamuthsLaw.exponent)
+            DamuthsLaw.coefficient * self.mass ** (DamuthsLaw.exponent)
         )
         """The number of individuals in the cohort."""
         self.is_alive: bool = True
         """Whether the cohort is alive [True] or dead [False]."""
         self.metabolic_rate: float = (
-            MetabolicRate.coefficienct * self.mass**MetabolicRate.exponent
+            MetabolicRate.coefficient * self.mass**MetabolicRate.exponent
         )
         """The rate at which energy is expended to [J/s]."""
         self.stored_energy: float = (
-            StoredEnergy.coefficienct * self.mass**StoredEnergy.exponent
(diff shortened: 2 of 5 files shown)


virtual_rainforest/models/animals/constants.py
```diff
@@ -16,11 +16,36 @@ class Term:
...
+
+"""Scaling of mammalian herbivore muscle mass (citation from Rallings).[g]"""
+MuscleMass = ScalingTerms(1.0, 0.38)
+
+"""Mammalian maximum intake rate (kg/day) from (Shipley 1994)."""
+IntakeRate = ScalingTerms(0.71, (0.63 * 480) * (1 / 1000))
+""" Converts original g/min rate to kg/day where 'day' is an 8hr foraging window."""
+
+
+@dataclass
```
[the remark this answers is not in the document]
TaranRallings: If there is a specific advantage, I don't know what it is. My motivation here has more to do with keeping these temporary structures streamlined and in as few different data types as possible. Do you think it's better to just have it as a standard variable?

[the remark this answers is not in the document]
jacobcook1995: Personally I would just save them as standard variables, but there might be some advantage that I am missing, so I would wait to see what others think

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -93,12 +99,20 @@ def __init__(self, name: str, mass: float, age: float, position: int) -> None:
         self.is_alive: bool = True
         """Whether the cohort is alive [True] or dead [False]."""
         self.metabolic_rate: float = (
-            MetabolicRate.coefficient * self.mass**MetabolicRate.exponent
+            MetabolicRate.coefficient * (self.mass * 1000) ** MetabolicRate.exponent
```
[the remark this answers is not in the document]
jacobcook1995: I think that depends whether you think you will use the function in question elsewhere or just within/upon the AnimalCohort class. If it's the later then making it a member function makes sense to keep things neat, but if you want to use it elsewhere making it independent saves you from having to rewrite an identical function elsewhere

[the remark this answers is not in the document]
jacobcook1995: It's an easy thing to change down the line if needed though, like literally copy + paste, change a few arguments

virtual_rainforest/models/animals/constants.py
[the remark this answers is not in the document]
jacobcook1995: Ahhh that makes sense!

jacobcook1995 (changes requested):
Looks good generally, but you probably should write tests for the `scaling_functions.py` module

TaranRallings:
> Looks good generally, but you probably should write tests for the `scaling_functions.py` module

Haha, somehow I completely overlooked that!

virtual_rainforest/models/animals/constants.py
```diff
@@ -7,7 +7,7 @@
 
 
 @dataclass
-class Term:
+class ScalingTerms:
```
dalonsoa, line 10: If you're using this classes this way, and they are meant to be immutable, I would use [`NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple) rather than a `dataclass`.

virtual_rainforest/models/animals/constants.py
```diff
@@ -16,11 +16,35 @@ class Term:
...
+
+"""Scaling of mammalian herbivore muscle mass (citation from Rallings).[g]"""
+MuscleMassTerms = ScalingTerms(1.0, 0.38)
+
+"""Mammalian maximum intake rate (g/min) from (Shipley 1994)."""
+IntakeRateTerms = ScalingTerms(0.71, 0.63)
+
+
+@dataclass
+class Term:
```
dalonsoa, line 36: I think I miss the point of this. If it is just one number, why not just setting it directly:

```python
ConversionEfficiency: float = 0.1
...
```

Also, the convention in python is using UPPERCASE for constants. I'm surprised `flake8` does not complain about you using CamelCase.

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -0,0 +1,124 @@
...
+    DamuthsLawTerms,
+    FatMassTerms,
+    IntakeRateTerms,
+    MeatEnergy,
+    MetabolicRateTerms,
+    MuscleMassTerms,
+)
+
+
+def DamuthsLaw(mass: float) -> int:
```
dalonsoa, line 18: Functions should be lower_case

```suggestion
def damuths_law(mass: float) -> int:
```

And the same with all the other functions in this file.

dalonsoa (changes requested):
Functional-wise, this looks good to me, but the style of the code need to be changed to follow the usual standards.

As I mention, I'm surprised that `flake8` or at least your code editor is not flagging this as an issue.

virtual_rainforest/models/animals/constants.py
TaranRallings: I have things set up as classes because they will end up being quite a bit more complex in the not-so-distant future. Both the scaling relationships and the single constants will end up taking information about the animal functional type, environmental conditions, and/or food types to sort out the specific final values. I imagined this as making the most sense as a class but I'm open to alternatives.

dalonsoa: Ah, ok. In that case, it makes sense as it is now. Maybe add a `ToDo` to the docstring of the class explaining this.

jacobcook1995 (approved):
LGTM!

dalonsoa (approved):
Looks good!

alexdewar (approved):
I've made a few really small suggestions, but it's looking good. Happy to approve it and you can fix it up as you like.

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -110,8 +114,113 @@ def metabolize(self, dt: timedelta64) -> None:
         """
 
         energy_burned = float(dt / timedelta64(1, "s")) * self.metabolic_rate
-        """Number of seconds in a day * J/s metabolic rate, consider daily rate."""
+        "Number of seconds in a day * J/s metabolic rate, consider daily rate."
```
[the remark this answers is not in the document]
davidorme: Should simply be a comment then?

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -110,8 +114,113 @@ def metabolize(self, dt: timedelta64) -> None:
...
+        also be used to perform supplementary actions like removing the cohort from
+        lists of active cohorts in AnimalModel or Grid instances.
+
+        """
+
+        if self.is_alive:
+            self.is_alive = False
+            LOGGER.warning("An animal cohort has died")
+        elif not self.is_alive:
+            LOGGER.warning("An animal cohort which is dead cannot die.")
```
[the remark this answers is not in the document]
davidorme: And are you trying to sneak the [Cthulhu mythos](https://en.wikipedia.org/wiki/That_Is_Not_Dead#:~:text=%22That%20is%20not%20dead%20which%20can%20eternal%20lie%2C%20/%20And%20with%20strange%20aeons%20even%20death%20may%20die.) into our logs?

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -110,8 +114,113 @@ def metabolize(self, dt: timedelta64) -> None:
...
+
+    def birth(self) -> "AnimalCohort":
```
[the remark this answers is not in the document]
davidorme: I did not know this option! Apologies, @TaranRallings!

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -0,0 +1,127 @@
...
+    FAT_MASS_TERMS,
+    INTAKE_RATE_TERMS,
+    MEAT_ENERGY,
+    METABOLIC_RATE_TERMS,
+    MUSCLE_MASS_TERMS,
+)
+
+
+def damuths_law(mass: float) -> int:
+    """The function set initial population densities .
```
davidorme, line 19: I think the coefficients should go in the signature - something @dalonsoa encouraged. We have some slight differences in how this is done

* `soils` passes in a `coef` argument containing the whole constants dataclass as in [here](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/cec0f9c69f15f914ba886122d562f2495aac9a45/virtual_rainforest/models/soil/carbon.py#L157).

It would be something like:


* `abiotic` is more explicit in the low level functions about naming each constant as in [here](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/cec0f9c69f15f914ba886122d562f2495aac9a45/virtual_rainforest/models/abiotic/radiation.py#L199)

 ```suggestion
def damuths_law(
    mass: float,
    coef: float = DAMUTHS_LAW_TERMS.coefficient, 
    exp: float = DAMUTHS_LAW_TERMS.exponent
) -> int:
    """The function set initial population densities .
```

**Review by davidorme — CHANGES_REQUESTED** [review]
Looks good - some changes requested for consistency across modules.

In one of the comments, I mentioned handling arrays. I think that's a general question really about whether some or all of the animal groups can be handled using arrays rather than individual calculations? I guess as soon as you move towards an array, then that is inherently moving towards a discretised set of size classes and integral projection model territory?

File: virtual_rainforest/models/animals/constants.py [hunk 10]
```diff
@@ -7,20 +7,56 @@
...
+    the final required parameter value.
+
+    """
+
+    value: float
+    """Contains the float value of the parameter"""
+
+
+"""The energy of a unit mass of mammal meat (check citation from Rallings). [J/g]"""
+MEAT_ENERGY = Term(7000.0)
```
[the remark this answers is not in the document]
TaranRallings: I would buy it!

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -0,0 +1,127 @@
...
+    """The function to set the amount of muscle mass on individual in an AnimalCohort.
+
+        Currently, this scaling relationship is only accurate for terrestrial mammals.
+        This will later be updated for additional functional types.
+
+    Args:
+        mass: The body-mass [kg] of an AnimalCohort.
+
+    Returns:
+        The mass [g] of muscle on an individual of the animal cohort.
```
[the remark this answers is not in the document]
TaranRallings: I left it as-is because the grams are just a transitional step and are only used to generate energetic values in joules. Right now, I'm just trying to get basic forms in there. I'll make a note to streamline some of these.

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -0,0 +1,127 @@
...
+        multi-grid occupation system for large animals is implemented.
+
+    Args:
+        mass: The body-mass [kg] of an AnimalCohort.
+
+    Returns:
+        The population density of that AnimalCohort [individuals/km2].
+
+    """
+    return ceil(DAMUTHS_LAW_TERMS.coefficient * mass ** (DAMUTHS_LAW_TERMS.exponent))
```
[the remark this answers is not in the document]
TaranRallings: @davidorme , the problem with my constants is that each one will end up needing to be or to have a sorting function that takes in mass, functional group traits, and environmental factors to determine which constants need to be employed. I was imagining this being done within a constant class like `DAMUTHS_LAW_TERMS` and so I set them up in a more complicated way. 

If doing this within the constant classes is too incongruent, I could create an intermediary layer of sorting functions to pull from that unstructured pool of constants-as-data-stores.

tests/models/animals/test_scaling_functions.py
```diff
@@ -0,0 +1,72 @@
...
+)
+def test_damuths_law(mass, population_density):
+    """Testing damuth's law for various body-masses."""
+    testing_pop = damuths_law(mass)
+    assert testing_pop == population_density
+
+
+@pytest.mark.parametrize(
+    "mass, met_rate",
+    [(0.0, 0.0), (1.0, 8.357913227182937), (1000.0, 1486.2705002791383)],
```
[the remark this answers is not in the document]
TaranRallings: That's a handy pytest feature, I'll build that in.

Wrt arrays - I am not currently set up for arrays. I think arrays may end up being a good solution to some later-on problems but, as you suggested, that choice would likely push us in some specific directions that I don't have a clear understanding of yet. Once I've got the minimum viable product up and running it'll be much easier to weigh out a choice like that.

virtual_rainforest/models/animals/constants.py
```diff
@@ -1,62 +1,69 @@
...
+    "bird": (0.7, 0.50),
+    # Toy Values
+}
+
+
+ENERGY_DENSITY: dict[str, float] = {
+    "meat": 7000.0,
+    # The energy of a unit mass of mammal meat (check citation from Rallings). [J/g]
+    "plant": 18200000.0
+    # Temporary realistic plant food value: Alfalfa ¬ 18,200,000 J/kg DM.
```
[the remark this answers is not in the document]
TaranRallings: It is in constants because it is the first of a variety of different plant energy densities for different plant parts/types. I can put it in it's own category of constants if that helps with the units issue but I'd rather not put 4-5 different energy density constants inside the PlantCommunity. I know it's all temporary but I'll soon be needing to check that herbivores are foraging successfully over multiple different plant resources.

davidorme:
Oh - absolutely, that makes a lot of sense. The issue at the moment is that that generator functionality is being built into the actual scaling functions, by passing in the `diet` etc arguments. That doesn't scale well, so my requested changes are centred on having those scaling functions start off with a clean FunctionalGroup object interface. We can then separately work on expanding that placeholder for the  `FunctionalGroup` object and creating a `generate_functional_group` factory function. Those files I pushed are then a sketch for how those objects are likely to behave.

If I understand correctly, what you then want is a function:

```python
temp = generate_functional_group(
    thermal="ecothermic",
    repro="semelparous", 
    diurnal="nocturnal", 
    level="canopy-dwelling", 
    diet ="insectivore",  
    abm=300.0
) -> Taxon:
```

The internals of that are likely to be... spiky, shall we say... but it will be a valuable tool. In particular, I can imagine a user wanting to define auto-generating FunctionalGroups in a `csv` like:

```csv
name,thermal,diurnal,...
temp,ecothermic,nocturnal,...
```

You'd still want to be able to directly specify values in a more complex file, but that could be a great alternate option.

.vscode/settings.json
```diff
@@ -18,4 +18,9 @@
     "workbench.editorAssociations": {
         "*.ipynb": "jupyter-notebook"
     },
+    "cSpell.words": [
+        "allometric",
+        "coef",
+        "Testasaurus"
+    ],
```
davidorme, line 25: Was this intentionally committed?

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -0,0 +1,54 @@
...
+
+    Needs:
+        - auto-generation system
+        - ability to manually populate specific constants
+
+    """
+
+    def __init__(self, name: str, taxa: str, diet: str) -> None:
+        """The constructor for the FunctionalGroup class."""
+        self.name = name
```
davidorme, line 37: Ultimately, I think this would be better having the actual terms as the signature (as in that `dataclass` example I posted).  That way you can create a `FunctionalGroup` instance by specifying all the terms. But... then your current `__init__` method can become a factory method:

```python
class FunctionalGroup:
    ...
    @classmethod
    def from_keywords(cls, name, diet, taxa) -> FunctionalGroup:
          # Contents of current __init__ into a constructor
          return cls(
               name=name, 
               diet=diet, 
               taxa=taxa,
               metabolic_rate_terms = ENDOTHERM_METABOLIC_RATE_TERMS[taxa],
               ...
               )
```

That sets up the two ways of populating them (totally bespoke values versus presets). So I think that is better, but not necessarily for this PR.

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -35,22 +21,18 @@ def damuths_law(
...
 
     Returns:
         The population density of that AnimalCohort [individuals/km2].
 
     """
-    exp = terms[taxa][diet][0]
-    coef = terms[taxa][diet][1]
+    exp = terms[0]
+    coef = terms[1]
     return ceil(coef * mass**exp)
```
davidorme: There's no real need to extract them to actual variables?

```suggestion
    return ceil(terms[1] * mass**terms[0])
```

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -9,22 +9,8 @@
...
-
-
-def damuths_law(
-    mass: float,
-    taxa: str,
-    diet: str,
-    terms: dict = DAMUTHS_LAW_TERMS,
-) -> int:
+
+def damuths_law(mass: float, terms: tuple) -> int:
```
davidorme: This has only just occurred to me, but most of the functions in here are allometric laws. I can see the value in documenting each function, but the name for each law could be described in the docs and then apply a generic allometric function within the animal cohort calculations. This would reduce the number of functions and testing quite a lot.

As a sketch, you could do this. Note something else that just occurred to me: you can use `*argvalue` in a python function call to expand an iterable of values into positional arguments, which would get rid of all of the `terms[0]` and `terms[1]`.

```python
from math import ceil

def allometric_law(mass, coef, exp):

    return coef * mass ** exp

mass = 0.23
terms = (0.75, 0.047)

allometric_law(100, *terms)  # unpacks to allometric_law(100, 0.75, 0.047)
ceil(allometric_law(100, *terms))
```

Again I don't think this is necessary for this PR, but I think it is probably cleaner.

davidorme (changes requested):
LGTM for right now, except for that last step we discussed of moving the `FunctionalGroup` instance creation out of `AnimalCohorts` and passing it actual instances of `FunctionalGroup`.

The other suggestions are things that work as is but where I think I see a cleaner way to do it. Those could happen now, but if not then maybe worth adding them as issues?

virtual_rainforest/models/animals/scaling_functions.py
davidorme: I think I got `exp` and `coef` the wrong way around in that sketch 😄

.vscode/settings.json
TaranRallings: Haha no it was not, that's pretty funny though. I was tired of seeing them in my local space, clicked for what I thought was visual studio to learn the words and stop highlighting them as spelling errors. Best way to undo that?

davidorme: That's what it has done, but it stores the words in the same file we use to coordinate settings. Maybe there is an alternative location for spelling?

TaranRallings: Fixed!

virtual_rainforest/models/animals/scaling_functions.py
TaranRallings: Alrighty, I've done most of this but will leave the totally-bespoke and allometric_laws portions for now.

davidorme (approved):
LGTM - I think simplifying the allometric functions is probably sensible, but this looks fine for the next step.

.vscode/settings.json
alexdewar: Sorry I'm so late to the party, but, yes, @davidorme there is an alternative location for saving spellings. You just have to choose "add to user settings" rather than "add to workspace settings", which makes it global.

PR #194, opened 2023-03-22
Basic data output functionality

### LINKED ISSUE #102: Basic outputting of data
The high level function that runs the simulation is going to generate data that should be saved somewhere. Need to develop a basic procedure to do this.

We need to write out data somewhere. To begin with this could be a simple time series in `csv` format that get saved to a location specified in the configuration.

Could save a more relevant format like `netcdf`, but feel that might introduce complications as we are only really looking for basic saving functionality here

jacobcook1995:
# Description

This PR adds a method to output the full data object as a NetCDF file. It also adds this into the main simulation flow, which now outputs both the initial and final Virtual Rainforest model states.

The current implementation reflects how the `Data` object is currently being used, i.e. contains just `numpy` arrays, doesn't have a time axis. So, if you have any ideas of how to make it more extensible for the types of data you are currently working with let me know.

Fixes #102 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/core/test_config.py
+++ b/tests/core/test_config.py
@@ -41,57 +41,6 @@ def test_check_dict_leaves(d_a: dict, d_b: dict, overlap: list) -> None:
 
 
 @pytest.mark.parametrize(
-    "out_path,expected_log_entries",
-    [
-        (
-            "./complete_config.toml",
-            (
-                (
-                    CRITICAL,
-                    "A file in the user specified output folder (.) already makes use "
-                    "of the specified output file name (complete_config.toml), this "
-                    "file should either be renamed or deleted!",
-                ),
-            ),
-        ),
-        (
-            "bad_folder/complete_config.toml",
-            (
-                (
-                    CRITICAL,
-                    "The user specified output directory (bad_folder) doesn't exist!",
-                ),
-            ),
-        ),
-        (
-            "pyproject.toml/complete_config.toml",
-            (
-                (
-                    CRITICAL,
-                    "The user specified output folder (pyproject.toml) isn't a "
-                    "directory!",
-                ),
-            ),
-        ),
-    ],
-)
-def test_check_outfile(caplog, mocker, out_path, expected_log_entries):
-    """Check that an error is logged if an output file is already saved."""
-    from virtual_rainforest.core.config import check_outfile
-
-    # Configure the mock to return a specific list of files
-    if out_path == "./complete_config.toml":
-        mock_content = mocker.patch("virtual_rainforest.core.config.Path.exists")
-        mock_content.return_value = True
-
-    # Check that check_outfile fails as expected
-    with pytest.raises(ConfigurationError):
-        check_outfile(Path(out_path))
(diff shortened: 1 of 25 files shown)


virtual_rainforest/core/config.py
```diff
@@ -279,9 +280,9 @@ def check_outfile(merge_file_path: Path) -> None:
     for file in Path(parent_fold).iterdir():
         if file.name == f"{out_file_name}":
             to_raise = ConfigurationError(
```
davidorme, line 282: Is this not just:

```suggestion
    if merge_file_path.exists():
        to_raise = ConfigurationError(
```

virtual_rainforest/main.py
```diff
@@ -232,13 +229,13 @@ def vr_run(
...
-    # TODO - Initialise the set of configured models
+    # TODO - A model spin up might be needed here in future
 
-    # TODO - Spin up the models
+    # Find parent folder for output for merge_file_path
+    parent_folder = merge_file_path.parent
 
-    # TODO - Save model state
+    # Save the initial state of the model
+    data.save_to_netcdf(Path(f"{parent_folder}/initial.nc"))
```
davidorme, line 238: This seems overkill - but it will be useful during testing. I think these names should be user configurable though and should be optional, so maybe look for`save_initial_data_path` and `save_final_data_path` or something similar in the configuration.

Ultimately, we may want to be able to specify which variables to save, but that can come later 😄

davidorme (approved):
LGTM -  a couple of suggested changes.

virtual_rainforest/main.py
jacobcook1995: Adding these as configuration options seems like a good idea!

vgro (approved):
LGTM.
The only thing I did not fully understand is what `merge_file_path` points to, could you put a comment on that somewhere? Or did I miss the description?

jacobcook1995:
Requesting re-reviews as I've added a `core.exceptions` module, which is probably useful for people to be aware of as it will break some imports

tests/core/test_config.py
```diff
@@ -36,57 +35,9 @@
...
-                (
-                    CRITICAL,
-                    "The user specified output folder (pyproject.toml) isn't a "
-                    "directory!",
-                ),
-            ),
-        ),
-    ],
-)
-def test_check_outfile(caplog, mocker, out_path, expected_log_entries):
```
vgro, line 77: just double-checking, did you replace `check_outfile` with the `save_to_netcdf` functionality?

jacobcook1995: I moved `check_outfile` to `core.utils` (from `core.config`). This is basically because I started using the function in `core.data` as well so I thought it made sense to move it to the utility function module as it does something generally useful rather than something only useful to the `core.config` module

jacobcook1995: This specific test hasn't been deleted it has just been moved to `tests/core/test_utils.py`

PR #195, opened 2023-03-29
Documentation for main simulation flow

### LINKED ISSUE #107: Developer docs for v0.1

jacobcook1995:
# Description

I've written some documentation for the main simulation flow (basically the `vr_run` function in `main.py`). Let me know whether what I have written makes sense, and also whether you think the way I've set it out is a good format or not.

This documentation will almost certainly gets revised soon, when we change things like how the simulation timing is handled.

Fixes #107

## Type of change

- [x] Documentation

## Key checklist

## Further checks

### DIFF
--- a/docs/source/virtual_rainforest/main_simulation.md
+++ b/docs/source/virtual_rainforest/main_simulation.md
@@ -38,15 +38,17 @@ simulating a particular aspect of the rainforest ecosystem. The models used for 
 specific simulation run can be set in the configuration. This will typically be the four
 standard models ([`AbioticModel`](../api/abiotic.md), `AnimalModel`, `PlantModel` and
 [`SoilModel`](../api/soil.md)), but this can be extended to include new models or
-different combinations of models. Once a list of models to configure has been extracted
-from the configuration, they are then all configured.
+different combinations of models. For more information about implementing new models,
+see [this page](../development/defining_new_models.md) about the required module
+structure and the model API. Once a list of models to configure has been extracted from
+the configuration, they are then all configured.
 
 ## Extracting simulation timing details
 
 The configuration contains a start time, an end time and an time interval for checking
-whether modules need to be updated. Once these details are extracted from the
+whether models need to be updated. Once these details are extracted from the
 configuration, a check is performed to ensure that the simulation update interval works
-suitably with the time steps for individual models, i.e. that no module has a shorter
+suitably with the time steps for individual models, i.e. that no model has a shorter
 time step than the overall simulation update interval.
 
 ## Saving the initial state
@@ -59,7 +61,7 @@ a single data file of the initial model state.
 
 The previously extracted timing details are used to setup a timing loop, which runs from
 the start time to the end time with a time step set by the update interval. At each
-step a check is performed to determine if any modules need to be updated. Any that do
+step a check is performed to determine if any models need to be updated. Any that do
 need to be updated are then updated.
 
 ## Saving the final state
--- a/docs/source/virtual_rainforest/main_simulation.md
+++ b/docs/source/virtual_rainforest/main_simulation.md
@@ -7,10 +7,12 @@ flow will now be set out step by step.
 
 As a first step, configuration files are loaded in and validated. These files are of
 `toml` format and are found based on paths provided by the user. When these files are
-loaded in they are validated to ensure that they are valid `toml`, and that no tags are
-repeated. In cases where a required tag is not present in any file, the validation will
-fail unless a default value for it has been set. Further details can be found in the
-[configuration documentation](./core/config.md).
+loaded in they are validated to ensure that they are valid `toml`, and that all the
+required settings are present and not duplicated. The Virtual Rainforest code provides
+default settings for some configuration options and these will be used automatically if
+that setting is not found in the configuration, but if no default is set then the
+validation will fail. Further details can be found in the [configuration
+documentation](./core/config.md).
 
 ## Grid creation
 
--- a/docs/source/virtual_rainforest/main_simulation.md
+++ b/docs/source/virtual_rainforest/main_simulation.md
@@ -1,3 +1,65 @@
 # Virtual Rainforest simulation flow
 
-This document describes the main simulation flow of the Virtual Rainforest model.
+This document describes the main simulation flow of the Virtual Rainforest model. This
+flow will now be set out step by step.
+
+## Loading configuration files
+
+As a first step, configuration files are loaded in and validated. These files are of
+`toml` format and are found based on paths provided by the user. When these files are
+loaded in they are validated to ensure that they are valid `toml`, and that no tags are
+repeated. In cases where a required tag is not present in any file, the validation will
+fail unless a default value for it has been set. Further details can be found in the
+[configuration documentation](./core/config.md).
+
+## Grid creation
+
+The next step is creating a grid, which defines the area and resolution of the
+simulation. The [grid documentation](./core/grid.md) describes this object further. This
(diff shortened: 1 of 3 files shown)


vgro (approved):
The overall structure looks good to me and the flow makes sense. At the moment, the document gives a good brief overview. I think it could be a bit more detailed in places, but this will probably happen over several iterations.
One point I noticed (and always struggle myself) is the use of 'models' or 'modules'. I don't know if we use them interchangeably or if we want to agree on what describes what?

jacobcook1995:
Yeah I trip over the module vs model thing a lot, we generally talk about our respective areas as modules, but they are all stored in a folder called `models` and are wrapped within a child of the `BaseModel` class. I currently switch between using module for high level description and model for when talking about the code, but gets really tricky here as the overall simulation flow incorporates both levels.

One solution would just be to talk about models (instead of modules) at all levels, but I don't know how others feel about that?

davidorme:
I think:

*  A _model_ is basically anything that is a child of `BaseModel`. It is an object that implements that API to implement a particular part of the VR.

* A module is just `python` speak for [code files that can be imported](https://docs.python.org/3/tutorial/modules.html). You could have a python _script_, which you can just run using `python file.py`, but modules contain objects that can be imported and used. Within packages, you then have a bunch of modules (one or more) that are contained within directories. If those are nested, then you get submodules and hence `virtual_rainforest.core.data`, where `data` points to the objects that you can import from the submodule file at `virtual_rainforest/core/data.py`.

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -0,0 +1,65 @@
...
+This document describes the main simulation flow of the Virtual Rainforest model. This
+flow will now be set out step by step.
+
+## Loading configuration files
+
+As a first step, configuration files are loaded in and validated. These files are of
+`toml` format and are found based on paths provided by the user. When these files are
+loaded in they are validated to ensure that they are valid `toml`, and that no tags are
+repeated. In cases where a required tag is not present in any file, the validation will
+fail unless a default value for it has been set. Further details can be found in the
```
davidorme, line 12: Unclear what tags are and where defaults come from:
```suggestion
loaded in they are validated to ensure that they are valid `toml`, and that all the 
required settings are present and not duplicated. The Virtual Rainforest code
provides default settings for some configuration options and these will be used 
automatically if that setting is not found in the configuration, but if no default is
set then the validation will fail. Further details can be found in the
```

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -0,0 +1,65 @@
...
+[configuration documentation](./core/config.md).
+
+## Grid creation
+
+The next step is creating a grid, which defines the area and resolution of the
+simulation. The [grid documentation](./core/grid.md) describes this object further. This
```
davidorme, line 18:
```suggestion
The next step is creating a grid, which defines the spatial structure of the simulation: 
the area, coordinate system and geometry of the individual cells that will be used in the 
simulation. The [grid documentation](./core/grid.md) describes this object further. This
```

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -0,0 +1,65 @@
...
+grid is then used to create an empty `Data` object of the correct size to describe the
+simulation.
+
+## Loading and validation of input data
+
+This `Data` object now needs to be populated with data. The data used in the simulation
+is stored in a set of data files which are specified in the configuration. This data
```
davidorme, line 25: This will change, I think. I think `Grid` should be part of a `Config` object, and `Data` should just be data. But that's for a future PR.

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -0,0 +1,65 @@
...
+is then loaded using the `data.load_data_config` method, which has built in validation
+procedures to ensure that the loaded data can be mapped onto the grid. Further details
+can be found in the [data system](./core/data.md) and [core axes](./core/axes.md)
```
davidorme, line 28:
```suggestion
procedures to ensure that the loaded data can be mapped onto the spatial grid and also
other core dimensions for the configured simulation. Further details
can be found in the [data system](./core/data.md) and [core axes](./core/axes.md)
```

jacobcook1995:
It gets a bit tricky when talking about things at a higher level, like I'm not sure which of the options in the following would be correct: "the simulation now determines which models/modules it needs to configure". As if say the `soil` module is configured the simulation will configure the `SoilModel` class and in doing so will import the `soil` module

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -0,0 +1,65 @@
...
+documentation.
+
+## Configuration of specific modules
+
+The models used for the specific simulation run can be set in the configuration. This
+will generally be the four standard modules (`abiotic`, `animals`, `plants` and `soil`),
+but it is also useful to have the ability to run novel modules or different combinations
+of the modules. Once a list of modules to configure has been extracted from the
+configuration, they are then all configured.
```
davidorme, line 37:
```suggestion
## Configuration of specific models

The Virtual Rainforest is implemented as model objects, each of which is responsible for simulating
a particular aspect of the rainforest ecosystem and the models used for a specific simulation run
can be set in the configuration. This will typically be the four standard models
(`AbioticModel`, `AnimalModel`, `PlantModel` and `SoilModel`),
but this can be extended to include new models or different combinations
of models. Once a list of models to configure has been extracted from the
configuration, they are then all configured.
```

davidorme:
> It gets a bit tricky when talking about things at a higher level, like I'm not sure which of the options in the following would be correct: "the simulation now determines which models/modules it needs to configure". As if say the `soil` module is configured the simulation will configure the `SoilModel` class and in doing so will import the `soil` module

I think this is at least partly because we're not using the actual model name in the config files and registry, which I think perhaps we should do.  We _don't_ configure the `soil` module, the thing we are configuring is the `SoilModel` model. Sure, we import that from `models.soil.soil_model`, but the actual namespace of the code isn't really relevant to the end user.

Maybe we should switch the config and registry to just using the actual class names directly rather than the module name?

docs/source/virtual_rainforest/main_simulation.md
davidorme: And maybe also link to the new model docs. "For more information about implementing new models, see this page about the required module structure and the model API."

jacobcook1995:
> LGTM - some suggested rewordings, but nothing critical. The only thing we could maybe add here is description of the config settings for stuff that happens in the body of the loop, like the data file stuff.

What are the config options that are relevant to the body of the loop?

davidorme:
I meant the options to save the data. We do need to think about some kind of a system to describe all of the config settings. I guess ideally, we add descriptions to the  JSONSchema and generate that automatically.

jacobcook1995:
Ahh right, at the moment the data only gets saved at the end of the loop (and then everything gets saved). I guess it makes sense to document this once there are options here? I could add something like "In future we envision options to X and Y" for the meantime?

And yes very much agree that we need to think of a documentation system for configuration options, as they will get hard to keep track of (I'll make an issue for that)

TaranRallings (approved):
LGTM!

PR #197, opened 2023-03-30
Feature/coordinates abiotic tools

vgro:
I started adding more dimensions, names, and coordinates to the DataArrays in the abiotic model. Here is the `abiotic_tools` script that holds constants and functions that are used across the abiotic model. The time dimension is still missing, this will follow in the next iteration. It would be great to hear your feedback on this.

For some reason, some of the tests don't work when I use the dummy_data fixture and I cannot figure out what the problem is. Ideas?

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/abiotic/test_abiotic_tools.py
+++ b/tests/models/abiotic/test_abiotic_tools.py
@@ -12,23 +12,18 @@ def dummy_data():
 
     from virtual_rainforest.core.data import Data
     from virtual_rainforest.core.grid import Grid
-    from virtual_rainforest.models.abiotic.abiotic_tools import set_layers_function
 
     # Setup the data object with two cells.
     grid = Grid(cell_nx=3, cell_ny=1)
     data = Data(grid)
-
-    # define layers along coordinates
-    # TODO from abiotic_model
-    layers_function = set_layers_function(canopy_layers=3, soil_layers=1)
 
     # Add the required data.
     data["temperature_2m"] = DataArray(
         [[30, 20, 30]],
         dims=["layers", "cell_id"],
         coords={
-            "layers": [len(layers_function) - 1],
-            "layers_function": ("layers", [layers_function[-1]]),
+            "layers": [6],
+            "layer_roles": ("layers", ["above"]),
             "cell_id": [0, 1, 2],
         },
     )
@@ -43,10 +38,10 @@ def dummy_data():
         ],
         dims=["layers", "cell_id"],
         coords={
-            "layers": np.arange(1, (len(layers_function))),
-            "layers_function": (
+            "layers": np.arange(1, 7),
+            "layer_roles": (
                 "layers",
-                layers_function[1 : (len(layers_function))],
+                ["surface", "below", "canopy", "canopy", "canopy", "above"],
             ),
             "cell_id": [0, 1, 2],
         },
@@ -55,10 +50,10 @@ def dummy_data():
         [[0.1, 0.1, 0.1], [1, 0.5, 0.5], [2, 2, 5], [3, 4, 10], [5, 7, 12]],
         dims=["layers", "cell_id"],
         coords={
-            "layers": np.arange(1, (len(layers_function) - 1)),
-            "layers_function": (
+            "layers": np.arange(1, 6),
+            "layer_roles": (
                 "layers",
-                layers_function[1 : (len(layers_function) - 1)],
+                ["surface", "below", "canopy", "canopy", "canopy"],
             ),
             "cell_id": [0, 1, 2],
         },
@@ -68,8 +63,8 @@ def dummy_data():
         [[101, 102, 103]],
         dims=["layers", "cell_id"],
         coords={
-            "layers": [len(layers_function) - 1],
-            "layers_function": ("layers", [layers_function[-1]]),
+            "layers": [6],
+            "layer_roles": ("layers", ["above"]),
             "cell_id": [0, 1, 2],
         },
     )
@@ -83,7 +78,6 @@ def test_calc_molar_density_air(dummy_data):
     from virtual_rainforest.models.abiotic import abiotic_tools
 
     data = dummy_data
-    layers_function = abiotic_tools.set_layers_function(canopy_layers=3, soil_layers=1)
 
     result = abiotic_tools.calc_molar_density_air(
         temperature=data["temperature_2m"],
@@ -95,8 +89,8 @@ def test_calc_molar_density_air(dummy_data):
             [[120.618713, 119.436176, 118.276602]],
             dims=["layers", "cell_id"],
             coords={
-                "layers": [len(layers_function) - 1],
-                "layers_function": ("layers", ["above"]),
+                "layers": [6],
+                "layer_roles": ("layers", ["above"]),
                 "cell_id": [0, 1, 2],
             },
             name="molar_density_air",
@@ -110,7 +104,6 @@ def test_calc_specific_heat_air(dummy_data):
     from virtual_rainforest.models.abiotic import abiotic_tools
 
     data = dummy_data
-    layers_function = abiotic_tools.set_layers_function(canopy_layers=3, soil_layers=1)
 
     result = abiotic_tools.calc_specific_heat_air(temperature=data["temperature_2m"])
 
@@ -120,8 +113,8 @@ def test_calc_specific_heat_air(dummy_data):
             [[29.214, 29.202, 29.214]],
             dims=["layers", "cell_id"],
             coords={
-                "layers": [len(layers_function) - 1],
-                "layers_function": ("layers", [layers_function[-1]]),
+                "layers": [6],
+                "layer_roles": ("layers", ["above"]),
                 "cell_id": [0, 1, 2],
             },
             name="specific_heat_air",
@@ -138,7 +131,6 @@ def test_calculate_latent_heat_vaporisation(dummy_data):
     from virtual_rainforest.models.abiotic import abiotic_tools
 
     data = dummy_data
-    layers_function = abiotic_tools.set_layers_function(canopy_layers=3, soil_layers=1)
 
     # test without dummy
     air_temperature = DataArray(
@@ -152,10 +144,10 @@ def test_calculate_latent_heat_vaporisation(dummy_data):
         ],
         dims=["layers", "cell_id"],
         coords={
-            "layers": np.arange(1, (len(layers_function))),
-            "layers_function": (
+            "layers": np.arange(1, 7),
+            "layer_roles": (
                 "layers",
-                layers_function[1 : (len(layers_function))],
+                ["surface", "below", "canopy", "canopy", "canopy", "above"],
             ),
             "cell_id": [0, 1, 2],
         },
@@ -171,7 +163,7 @@ def test_calculate_latent_heat_vaporisation(dummy_data):
             dims=["layers", "cell_id"],
             coords={
                 "layers": [6],
-                "layers_function": ("layers", [layers_function[-1]]),
+                "layer_roles": ("layers", ["above"]),
                 "cell_id": [0, 1, 2],
             },
             name="latent_heat_vaporisation",
@@ -190,10 +182,10 @@ def test_calculate_latent_heat_vaporisation(dummy_data):
             ],
             dims=["layers", "cell_id"],
             coords={
-                "layers": np.arange(1, len(layers_function)),
-                "layers_function": (
+                "layers": np.arange(1, 7),
+                "layer_roles": (
                     "layers",
-                    layers_function[1 : len(layers_function)],
+                    ["surface", "below", "canopy", "canopy", "canopy", "above"],
                 ),
                 "cell_id": [0, 1, 2],
             },
@@ -210,16 +202,14 @@ def test_calculate_bulk_aero_resistance():
 
     from virtual_rainforest.models.abiotic import abiotic_tools
 
-    layers_function = abiotic_tools.set_layers_function(canopy_layers=3, soil_layers=1)
-
     wind_below_canopy = DataArray(
         [[0.1, 0.1, 0.1], [1, 0.5, 0.5], [2, 2, 5], [3, 4, 10], [5, 7, 12]],
         dims=["layers", "cell_id"],
         coords={
-            "layers": np.arange(1, (len(layers_function) - 1)),
-            "layers_function": (
+            "layers": np.arange(1, 6),
+            "layer_roles": (
                 "layers",
-                layers_function[1 : (len(layers_function) - 1)],
+                ["surface", "below", "canopy", "canopy", "canopy"],
             ),
             "cell_id": [0, 1, 2],
         },
@@ -241,10 +231,10 @@ def test_calculate_bulk_aero_resistance():
             ],
             dims=["layers", "cell_id"],
             coords={
-                "layers": np.arange(1, (len(layers_function) - 1)),
-                "layers_function": (
+                "layers": np.arange(1, 6),
+                "layer_roles": (
                     "layers",
-                    layers_function[1 : (len(layers_function) - 1)],
+                    ["surface", "below", "canopy", "canopy", "canopy"],
                 ),
                 "cell_id": [0, 1, 2],
             },
@@ -253,12 +243,12 @@ def test_calculate_bulk_aero_resistance():
     )
 
 
-def test_set_layers_function():
+def test_set_layer_roles():
     """Test layers string created correctly."""
 
     from virtual_rainforest.models.abiotic import abiotic_tools
 
-    result = abiotic_tools.set_layers_function(canopy_layers=3, soil_layers=2)
+    result = abiotic_tools.set_layer_roles(canopy_layers=3, soil_layers=2)
 
     assert result == [
         "soil",
--- a/tests/models/abiotic/test_abiotic_tools.py
+++ b/tests/models/abiotic/test_abiotic_tools.py
@@ -1,5 +1,6 @@
 """Test abiotic_tools.py."""
 
+import numpy as np
 import pytest
 import xarray as xr
 from xarray import DataArray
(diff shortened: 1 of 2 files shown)


virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -65,6 +76,71 @@ def calc_specific_heat_air(
...
+
+    return DataArray(
+        (heat_transfer_coefficient / np.sqrt(wind_speed)),
+        dims=wind_speed.dims,
+        coords=wind_speed.coords,
+        name="areo_resistance",
+    )
+
+
+def set_layers_function(canopy_layers: int, soil_layers: int) -> List[str]:
```
vgro, line 130: This functionality should probably go somewhere in the `AbioticModel` to make share it is applied consistently across all functions.

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -40,14 +46,19 @@ def calc_molar_density_air(
...
+    molar_density_air = DataArray(
+        (
+            standard_mole
+            * (temperature_kelvin / atmospheric_pressure)
+            * (celsius_to_kelvin / temperature_kelvin)
+        ),
+        dims=temperature.dims,
+        coords=temperature.coords,
+        name="molar_density_air",
     )
```
dalonsoa, line 62: The result of all those operations should be a DataArray with the same dimensions and coordinates as the temperature and the atmospheric pressure, so you jus need to rename it, not create a new one:
```suggestion
    molar_density_air =  (
            standard_mole
            * (temperature_kelvin / atmospheric_pressure)
            * (celsius_to_kelvin / temperature_kelvin)
        ).rename("molar_density_air")
```

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -65,6 +76,71 @@ def calc_specific_heat_air(
...
-        specific heat of air at constant pressure (J mol-1 K-1)
+        specific heat of air at constant pressure, [J mol-1 K-1]
     """
-    return 2e-05 * temperature**2 + 0.0002 * temperature + molar_heat_capacity_air
+    return DataArray(
+        (2e-05 * temperature**2 + 0.0002 * temperature + molar_heat_capacity_air),
+        dims=temperature.dims,
+        coords=temperature.coords,
+        name="specific_heat_air",
+    )
```
dalonsoa, line 86: And the same here:
```suggestion
    return (2e-05 * temperature**2 + 0.0002 * temperature + molar_heat_capacity_air).rename("specific_heat_air")
```

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -65,6 +76,71 @@ def calc_specific_heat_air(
...
+    Returns:
+        latent heat of vaporisation, [J kg-1]
+    """
+
+    latent_heat_vaporisation = DataArray(
+        (-42.575 * temperature + 44994),
+        dims=temperature.dims,
+        coords=temperature.coords,
+        name="latent_heat_vaporisation",
+    )
```
dalonsoa, line 104: And here :) 
```suggestion
    latent_heat_vaporisation = (-42.575 * temperature + 44994).rename("latent_heat_vaporisation")
```

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -65,6 +76,71 @@ def calc_specific_heat_air(
...
+    Returns:
+        aerodynamic resistance
+    """
+
+    return DataArray(
+        (heat_transfer_coefficient / np.sqrt(wind_speed)),
+        dims=wind_speed.dims,
+        coords=wind_speed.coords,
+        name="areo_resistance",
+    )
```
dalonsoa, line 127: And here!
```suggestion
    return (heat_transfer_coefficient / np.sqrt(wind_speed)).rename("areo_resistance")
```

dalonsoa (approved):
It looks good! I've just pointed out that you do not need to create a DataArray when you're operating with DataArrays. The result of the operations will be automatically a DataArray with the correct dimensions and coordinates - or the operation will fail if there's any incompatibility.

Also, there is some inconsistency on how you create/return the values. Sometimes you do:
```python
an_array = DataArray(...)
return an_array
```
And in others, directly:
```python
return DataArray(...)
```
Pick one and stick to it, but I would go for the second option as you're not doing more operations with the array once created.

dalonsoa:
By the way, all tests seem to be passing, so I'm not sure what is the problem you're mentioning.

tests/models/abiotic/test_abiotic_tools.py
```diff
@@ -31,24 +82,191 @@ def test_calc_molar_density_air(dummy_data):
...
+            },
+            name="specific_heat_air",
+        ),
+    )
+
+
+def test_calculate_latent_heat_vaporisation(dummy_data):
+    """Test.
+
+    TODO check why it does not work with dummy_data.
```
jacobcook1995, line 135: Has this TODO been resolved?

jacobcook1995 (approved):
LGTM! I just had one query about a TODO that seemed outdated to me, but nothing substantial

tests/models/abiotic/test_abiotic_tools.py
jacobcook1995: Oh right I see what is happening now

tests/models/abiotic/test_abiotic_tools.py
```diff
@@ -31,24 +82,191 @@ def test_calc_molar_density_air(dummy_data):
...
 def test_calc_specific_heat_air(dummy_data):
-    """Test canclualte specific heat of air."""
+    """Test calculate specific heat of air."""
+
+    from virtual_rainforest.models.abiotic import abiotic_tools
+
+    data = dummy_data
+    layers_function = abiotic_tools.set_layers_function(canopy_layers=3, soil_layers=1)
+
+    result = abiotic_tools.calc_specific_heat_air(temperature=data["temperature_2m"])
```
davidorme, line 115: All the test functions have the same lines to create `data` and `layers_function`. I'd make `layers_function` a fixture and just use `dummy_data` directly.

```python
def test_calc_specific_heat_air(dummy_data, layers_function):
    """Test calculate specific heat of air."""

    from virtual_rainforest.models.abiotic import abiotic_tools

    result = abiotic_tools.calc_specific_heat_air(temperature=dummy_data["temperature_2m"])
```

tests/models/abiotic/test_abiotic_tools.py
davidorme: OK - just noting that the fixture is included in the sig but then not used.

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -65,6 +76,71 @@ def calc_specific_heat_air(
...
+
+    Args:
+        temperature: temperature, [C]
+
+    Returns:
+        latent heat of vaporisation, [J kg-1]
+    """
+
+    latent_heat_vaporisation = DataArray(
+        (-42.575 * temperature + 44994),
```
davidorme, line 100: And again - values into `AbioticConstants`?

virtual_rainforest/models/abiotic/abiotic_tools.py
davidorme: It should probably be defined in the config and turn into core axes! But, fine where it is for the moment.

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -65,6 +76,71 @@ def calc_specific_heat_air(
         molar_heat_capacity_air: molar heat capacity, [J mol-1 C-1]
 
     Returns:
-        specific heat of air at constant pressure (J mol-1 K-1)
+        specific heat of air at constant pressure, [J mol-1 K-1]
     """
-    return 2e-05 * temperature**2 + 0.0002 * temperature + molar_heat_capacity_air
+    return DataArray(
+        (2e-05 * temperature**2 + 0.0002 * temperature + molar_heat_capacity_air),
```
davidorme, line 82: The coefficients should probably go in `AbioticConstants`?

davidorme (changes requested):
All the functions look good, but there is a problem with the dimensions, which is what is stopping your tests from using `dummy_data`. The issue is that an `xarray.Dataset` has a set of shared common dimensions and any variable in the dataset that uses a named dimension needs to have the same length along that dimension. 

Here, you are putting lots of variables into the dataset using the same dimension names of `layers` and `layers_function`, but the lengths of those dimensions differ depending on what vertical slice of the total layers is applicable. The dataset doesn't complain about this, it just reduces them all to the common coordinates along those dimensions - so if you look at what the `dummy_data` fixture _actually_ hands out, it is this:

```python
In [16]: data.data
Out[16]: 
<xarray.Dataset>
Dimensions:               (layers: 1, cell_id: 3)
Coordinates:
  * layers                (layers) int64 6
  * cell_id               (cell_id) int64 0 1 2
    layers_function       (layers) <U5 'above'
Data variables:
    temperature_2m        (layers, cell_id) int64 30 20 30
    air_temperature       (layers, cell_id) int64 32 32 32
    wind_below_canopy     (layers, cell_id) float64 nan nan nan
    atmospheric_pressure  (layers, cell_id) int64 101 102 103
```

So, your first variable `temperature_2m` has coords `layers=6` and `layers_function = above` and then as the other data sets are added, the data for those coordinates are extracted. The `wind_below_canopy` data doesn't extend up to `above`, so it gets `np.nan` values.

I can think of two approaches at the moment:

1. We keep the full `layers_function` dimension length, but then all the data has to be padded with `np.nan` for non-applicable layers. That is simpler in the data object and you can see which layers are set for a particular variable. You would then need to always be `.isel()`-ing the correct layers for each variable though.

2. We identify the set of different slices that are needed and then each of them gets a separate dimension name with a fixed length. With this, you're passing arrays of the right size into functions, but there's more complexity in the data object. I think with this we can also set up the global coordinate system for the vertical axis and then provide coordinates on the sub dimensions that allow them to be mapped back together. There's a toy example below, - this doesn't handle the fact that the dimension names are the same and these can't exist in the same dataset as is - but it does show how they can be combined along a wider axis:

```python
from xarray import DataArray, combine_by_coords
import numpy as np

layers = [
    "soil_2",
    "soil_1",
    "ground",
    "below_canopy",
    "canopy_1",
    "canopy_2",
    "canopy_3",
    "above_canopy",
]
layers = {idx: val for idx, val in enumerate(layers)}

canopy = DataArray(
    np.ones((3, 6)),
    dims=["layers", "cell_id"],
    coords={
        "layers": [4,5,6],
        "cell_id": np.arange(6),
    },
)

soil_and_ground = DataArray(
    np.zeros((3, 6)),
    dims=["layers", "cell_id"],
    coords={
        "layers": [0, 1, 2],
        "cell_id": np.arange(6),
    },
)

combine_by_coords([canopy, soil_and_ground])
```

Note the gaps along the layer index though.

```python
Out[23]: 
<xarray.DataArray (layers: 6, cell_id: 6)>
array([[0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0.],
       [0., 0., 0., 0., 0., 0.],
       [1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1.],
       [1., 1., 1., 1., 1., 1.]])
Coordinates:
  * layers   (layers) int64 0 1 2 4 5 6
  * cell_id  (cell_id) int64 0 1 2 3 4 5
```

tests/models/abiotic/test_abiotic_tools.py
vgro: This is still open as the test does not work with the data object, @davidorme explained well why this is the case below. We will need to find a solution for vertical dimensions.

vgro:
> By the way, all tests seem to be passing, so I'm not sure what is the problem you're mentioning.

see problem arising with data object here https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/197#pullrequestreview-1366502806

tests/models/abiotic/test_abiotic_tools.py
```diff
@@ -1,27 +1,78 @@
...
     from virtual_rainforest.core.grid import Grid
+    from virtual_rainforest.models.abiotic.abiotic_tools import set_layers_function
 
     # Setup the data object with two cells.
     grid = Grid(cell_nx=3, cell_ny=1)
     data = Data(grid)
 
+    # define layers along coordinates
+    # TODO from abiotic_model
+    layers_function = set_layers_function(canopy_layers=3, soil_layers=1)
```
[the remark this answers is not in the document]
vgro: good point, I changed it to `layer_roles`

tests/models/abiotic/test_abiotic_tools.py
```diff
@@ -31,24 +82,191 @@ def test_calc_molar_density_air(dummy_data):
...
+    result = abiotic_tools.calc_specific_heat_air(temperature=data["temperature_2m"])
+
+    xr.testing.assert_allclose(
+        result,
+        DataArray(
+            [[29.214, 29.202, 29.214]],
+            dims=["layers", "cell_id"],
+            coords={
+                "layers": [len(layers_function) - 1],
+                "layers_function": ("layers", [layers_function[-1]]),
```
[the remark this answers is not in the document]
vgro: done :-)

tests/models/abiotic/test_abiotic_tools.py
vgro: I hard coded the layers that are expected, so this is outdated. For other scripts I will definitely follow this suggestion.

vgro:
> All the functions look good, but there is a problem with the dimensions, which is what is stopping your tests from using `dummy_data`. The issue is that an `xarray.Dataset` has a set of shared common dimensions and any variable in the dataset that uses a named dimension needs to have the same length along that dimension.
> 
> Here, you are putting lots of variables into the dataset using the same dimension names of `layers` and `layers_function`, but the lengths of those dimensions differ depending on what vertical slice of the total layers is applicable. The dataset doesn't complain about this, it just reduces them all to the common coordinates along those dimensions - so if you look at what the `dummy_data` fixture _actually_ hands out, it is this:
> 
> ```python
> In [16]: data.data
> Out[16]: 
> <xarray.Dataset>
> Dimensions:               (layers: 1, cell_id: 3)
> Coordinates:
>   * layers                (layers) int64 6
>   * cell_id               (cell_id) int64 0 1 2
>     layers_function       (layers) <U5 'above'
> Data variables:
>     temperature_2m        (layers, cell_id) int64 30 20 30
>     air_temperature       (layers, cell_id) int64 32 32 32
>     wind_below_canopy     (layers, cell_id) float64 nan nan nan
>     atmospheric_pressure  (layers, cell_id) int64 101 102 103
> ```
> 
> So, your first variable `temperature_2m` has coords `layers=6` and `layers_function = above` and then as the other data sets are added, the data for those coordinates are extracted. The `wind_below_canopy` data doesn't extend up to `above`, so it gets `np.nan` values.
> 
> I can think of two approaches at the moment:
> 
>     1. We keep the full `layers_function` dimension length, but then all the data has to be padded with `np.nan` for non-applicable layers. That is simpler in the data object and you can see which layers are set for a particular variable. You would then need to always be `.isel()`-ing the correct layers for each variable though.
> 
>     2. We identify the set of different slices that are needed and then each of them gets a separate dimension name with a fixed length. With this, you're passing arrays of the right size into functions, but there's more complexity in the data object. I think with this we can also set up the global coordinate system for the vertical axis and then provide coordinates on the sub dimensions that allow them to be mapped back together. There's a toy example below, - this doesn't handle the fact that the dimension names are the same and these can't exist in the same dataset as is - but it does show how they can be combined along a wider axis:
> 
> 
> ```python
> from xarray import DataArray, combine_by_coords
> import numpy as np
> 
> layers = [
>     "soil_2",
>     "soil_1",
>     "ground",
>     "below_canopy",
>     "canopy_1",
>     "canopy_2",
>     "canopy_3",
>     "above_canopy",
> ]
> layers = {idx: val for idx, val in enumerate(layers)}
> 
> canopy = DataArray(
>     np.ones((3, 6)),
>     dims=["layers", "cell_id"],
>     coords={
>         "layers": [4,5,6],
>         "cell_id": np.arange(6),
>     },
> )
> 
> soil_and_ground = DataArray(
>     np.zeros((3, 6)),
>     dims=["layers", "cell_id"],
>     coords={
>         "layers": [0, 1, 2],
>         "cell_id": np.arange(6),
>     },
> )
> 
> combine_by_coords([canopy, soil_and_ground])
> ```
> 
> Note the gaps along the layer index though.
> 
> ```python
> Out[23]: 
> <xarray.DataArray (layers: 6, cell_id: 6)>
> array([[0., 0., 0., 0., 0., 0.],
>        [0., 0., 0., 0., 0., 0.],
>        [0., 0., 0., 0., 0., 0.],
>        [1., 1., 1., 1., 1., 1.],
>        [1., 1., 1., 1., 1., 1.],
>        [1., 1., 1., 1., 1., 1.]])
> Coordinates:
>   * layers   (layers) int64 0 1 2 4 5 6
>   * cell_id  (cell_id) int64 0 1 2 3 4 5
> ```

Well spotted @davidorme !
I assumed that the `data` object could hold `DataArrays` of different sizes. I would tend to go with the first option you suggested (max array size with nan for unused layers), although I don't know if this needs a lot of storage/time when handling? The alternative seems sensible but like you say, more complicated and prone to error.
@dalonsoa and @dalonsoa what would you suggest?

dalonsoa:
As we discuss during out pair programming session, I think this calls for a single dimension named `layers`, an index coordinate also called `layers` (just the sequence of numbers), and a non-index coordinate that defines a role and helps to easily select between canopy and ground (or others) - the `layers_function` you defined. Then, we use @davidorme 's option 1, which I think is the simplest to implement and deal with, computationally speaking.

PR #198, opened 2023-03-31
Feature/Animal_Module: Basic dispersal function and tracking of cohort positions on grid.

TaranRallings:
# Description

This PR introduces basic spatial functionality for `AnimalCohort`. The location of all cohorts is tracked as an attribute of `AnimalModel`. The cohorts themselves can now move, randomly, to neighbouring grids via a `disperse()` function. Birth and cohort_death were also changed to add or remove cohorts from the tracking dict.

I do a couple of things here that I'm not convinced about and would like feedback on:

1) Does the storage and handling of `AnimalCohort` positions in a dict within `AnimalModel` make sense?
I am fairly confident that I will need this positional data stored somewhere for trophic behaviour. Would it make more sense
to do this within `Data` in some other way?
     
2) I ran into a circular import issue and moved the import of `AnimalCohort` inside the init call of `AnimalModel`. 
Is this a terrible idea? 

I've tagged everyone as its important we're all on the same page about how space is used and how location information is stored. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


davidorme (changes requested):
The main functionality is fine, but I think it is cleaner to have dispersal just update the cohort's position attribute and not have `disperse` try to modify the `AnimalModel` instance attributes. If you then need a dictionary of cohorts keyed by cell id, then you can run something like `itertools.groupby` over the list. This will also sort out your circular import issue - I think importing within the method is fine but it is probably easier to avoid it altogether.

dalonsoa:
Hi @TaranRallings , I'm going away on holidays for 2 weeks and cannot review this before leaving. So, do not wait for my input on this PR before merging because it might take a while!

jacobcook1995 (approved):
LGTM!

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -55,6 +55,12 @@ def __init__(
         **kwargs: Any,
     ):
         super().__init__(data, update_interval, start_time, **kwargs)
+        from virtual_rainforest.models.animals.dummy_animal_module import AnimalCohort
+
+        self.cohort_positions: dict[int, list[AnimalCohort]] = {
+            k: [] for k in self.data.grid.cell_id
+        }
```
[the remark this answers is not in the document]
TaranRallings: Let me know if I'm understanding you correctly, you're suggesting replacing the dictionary with an unstructured list of all of the animal cohorts in the model?

The dictionary structure isn't really about dispersal, it's about how trophic interactions will work. Every tick there will be many calls to a search to figure out what other animal cohorts are possible food sources or other interactions. I wanted to keep that spatial info organized to make that process easier. 

Does that make sense?

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -242,4 +245,27 @@ def birth(self) -> AnimalCohort:
...
+    def disperse(self, model: AnimalModel) -> None:
+        """The function to move a cohort between grid squares.
+
+        This is a simple implementation of what will need to be a much more complex
+        function. Right now it moves a cohort into a random neighboring grid, including
+        the current grid. Later functionality will need to involve bodysize, locomotion,
+        and other factors.
+
+        """
+        model.data.grid.set_neighbours(distance=sqrt(model.data.grid.cell_area))
```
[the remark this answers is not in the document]
TaranRallings: It may end up being useful to construct a neighbour matrix and store it as an animal cohort attribute as the experience of space is a function of the functional-type and body-size. In it's current form, I'm happy to store it in AnimalModel.__init__.

dalonsoa (approved):
Keeping aside some potential performance bottlenecks highlighted by @davidorme , The code looks ok. I don't have any problem with importing things within functions. It is not idea, but sometimes it is unavoidable - although sometimes it is just annoying, because things are imported just for type annotation and not other functionality. 

About using a dictionary to store the positions, to be honest, I don't know enough how it will be used in the future to decide if that's the best strategy, to be honest. @davidorme comments make sense to me, but so do @TaranRallings justification for using dict :)

virtual_rainforest/models/animals/animal_model.py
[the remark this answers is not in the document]
TaranRallings: I've committed a basic version of what I think you're talking about. It still needs a couple of tests and, if we agree about the handling of births/deaths for those handler functions within `AnimalModel`. Let me know if this is, in fact, what you are talking about and if you have any thoughts about the trophic issue.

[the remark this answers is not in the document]
davidorme: It might be worth having a group chat about how to structure this. One option that might be worth exploring is to avoid having an additional layer in the class structure:

* **AnimalModel**: 
  * handles the setup
  * creates a list of AnimalCommunity objects, one per grid cell.
  * provides any additional setup for connectivity between grid cells if needed.
* **AnimalCommunity**:
  * is the main workhorse containing all of the behavioural functionality
  * has a single list of AnimalCohorts, which are all of the cohorts currently present in the grid cell
  * has an `immigrate` method, which allows an AnimalCohort instance to be popped out of one instance and added to the cohort list for another AnimalCommunity
* **AnimalCohort**:
   * maintains the functional type and status of the individual cohorts.
 
This explicitly separates the within cell ecology happening to a set of cohorts, and the cohorts within a cell are always up to date, as the cohort list updates in real time with immigration.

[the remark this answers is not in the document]
davidorme: I think most of what is in `AnimalModel` now would move into `AnimalCommunity`?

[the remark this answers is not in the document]
TaranRallings: Interesting idea! I think that will be a cleaner way to organize things. I would rather have those function calls in something like an `AnimalCommunity` class than cluttering up `AnimalModel`. I'll throw something together and we can see what we think.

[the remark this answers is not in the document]
davidorme: It might make sense if each `AnimalCommunity` had a dictionary of cohorts keyed by functional type, rather than just a flat list? So:

```python
my_animal_community.cohorts = {
    FuncGroup1: [fg1cohort1, fg1cohort2], 
    FuncGroup2: [fg2cohort1, fg2cohort2, fg2cohort3],
}
```

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -55,6 +58,21 @@ def __init__(
...
     ):
         super().__init__(data, update_interval, start_time, **kwargs)
+        self.data.grid.set_neighbours(distance=sqrt(self.data.grid.cell_area))
+        # run a new set_neighbours with cohort specific dispersal distance (temp sol.)
+        # from virtual_rainforest.models.animals.dummy_animal_module import AnimalCohort
+
+        # imports AnimalCohort w/in init to dodge circular import problems: allowed??
+        # self.cohort_positions: dict[int, list[AnimalCohort]] = {
+        #    k: [] for k in self.data.grid.cell_id
+        # }
```
alexdewar, line 68: Is this commented-out code still needed?

```suggestion
```

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -55,6 +58,21 @@ def __init__(
...
+        """This stores the AnimalCohorts as values keyed to their grid positions."""
```
alexdewar, line 69: Is this a docstring for a member that's been removed?

virtual_rainforest/models/animals/animal_model.py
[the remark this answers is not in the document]
TaranRallings: After a bit of battling with naming, I think I've got things smoothed out! Take a look and let me know if this is what you had in mind with an intermediate 'AnimalCommunity' class. It seems to have worked well.

TaranRallings:
> Keeping aside some potential performance bottlenecks highlighted by @davidorme , The code looks ok. I don't have any problem with importing things within functions. It is not idea, but sometimes it is unavoidable - although sometimes it is just annoying, because things are imported just for type annotation and not other functionality.
> 
> About using a dictionary to store the positions, to be honest, I don't know enough how it will be used in the future to decide if that's the best strategy, to be honest. @davidorme comments make sense to me, but so do @TaranRallings justification for using dict :)

I have changed the code significantly since this time so I am going to request a re-review to make sure I didn't do anything wild.

dalonsoa:
Will try to review later today...

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -78,6 +82,75 @@ class PalatableSoil:
...
+            cohort: The AnimalCohort moving between AnimalCommunities.
+            destination: The AnimalCommunity the cohort is moving to.
+
+        """
+        immigrant = self.cohorts[cohort.name].pop(
+            self.cohorts[cohort.name].index(cohort)
+        )
+        # Remove target cohort from source community
+        destination.cohorts[immigrant.name].append(immigrant)
+        # Add cohort to destination community
```
dalonsoa, line 117: Personally, I would do this in one go to avoid intermediate variables, but it is totally up to you. 

```suggestion
        destination.cohorts[immigrant.name].append(
            self.cohorts[cohort.name].pop(
                self.cohorts[cohort.name].index(cohort)
            )
        )
```

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -78,6 +82,75 @@ class PalatableSoil:
...
+            cohort: The AnimalCohort instance which is producing a new AnimalCohort.
+
+        Returns:
+            A new age 0 AnimalCohort.
+
+        """
+        new_cohort = AnimalCohort(cohort.functional_group, cohort.mass, 0.0)
+        self.cohorts[cohort.name].append(new_cohort)
+
+        return new_cohort
```
dalonsoa, line 151: And similarly here:
```suggestion
        self.cohorts[cohort.name].append(AnimalCohort(cohort.functional_group, cohort.mass, 0.0))

        return self.cohorts[cohort.name][-1]
```

dalonsoa (approved):
I've made a couple of suggestions - too almost purely aesthetic - but the code looks good. It seems simpler than last time, which is good! 

I've noticed the "temporary solution". It might be worth to open an issue about that, so we don't forget and the "temporary solution" becomes permanent.

tests/models/animals/test_dummy_animal_module.py
davidorme, line 1: This is getting quite big - I'd be tempted to split it up into files for each class? But not a big deal.

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -78,6 +82,75 @@ class PalatableSoil:
...
+
+        Args:
+            cohort: The AnimalCohort instance that has lost all individuals.
+
+        """
+
+        if cohort.is_alive:
+            cohort.is_alive = False
+            LOGGER.info("An animal cohort has died")
+            self.cohorts[cohort.name].remove(cohort)
```
davidorme, line 130: I'd consider making this a `DEBUG` log entry? We're going to need to do a systematic review of logging entries at some point, but this seems like `DEBUG`. It would be good to report more specific data too - cohort type and maybe even grid id if available?

I think there's some redundancy here too - `self.cohorts` is a list of alive cohorts and if there isn't a conceivable mechanism where dead cohort could be there, then setting and testing flags might be superfluous? For example, setting `is.alive = False` here is immediately followed by removing and discarding that object.

virtual_rainforest/models/animals/dummy_animal_module.py
```diff
@@ -78,6 +82,75 @@ class PalatableSoil:
...
+        elif not cohort.is_alive:
+            LOGGER.exception("An animal cohort which is dead cannot die.")
+
+    def birth(self, cohort: AnimalCohort) -> AnimalCohort:
```
davidorme, line 134: Should this be a method of `AnimalCohort`? A cohort will have a number of individuals in a given (average) state that might be capable of reproducing and generating a new cohort. Then at the `AnimalCommunity` level, the update might do something like:

``` python
for this_name, these_cohorts in self.cohorts.items():
    # generate new cohorts for this functional group
    new_cohorts_for_name = [chrt.reproduce() for chrt in these_cohorts if chrt.can_reproduce()]
    # potentially merge those new cohorts (all age=0) if we don't care
    # (currently/ever) about maternal and paternal effects.
    self.cohorts[this_name].extend(new_cohorts_for_name)
```

I don't think there's any issue with the content of this method, but it logically feels like it should be a step further in?

davidorme: I guess, similarly, `death` feels like an `AnimalCohort` method. So then the update would be:

``` python
for this_name, these_cohorts in self.cohorts.items():
    self.cohorts[this_name] = [chrt for chrt in these_cohorts if not chrt.just_died]
```

davidorme: The order matters too! Also semelparity would need a mechanism - unless death happens before reproduction and then a semelparous cohort could set a `die_next_check` flag?

davidorme (commented):
LGTM - I think that the logic of which classes implement which methods could be refined. That is largely just moving code - it isn't critical but it does feel like a more natural way to structure the flow.

davidorme (approved):

PR #200, opened 2023-04-17
Remove usage of `pip install codecov` from CI action

### LINKED ISSUE #199: Code coverage checking fails as it can no longer be downloaded using pip
**Describe the bug**
`codecov` has been removed from `PyPi` so can longer be downloaded using `pip`. This means that all our recent `ubuntu 3.9` tests fail, as this test attempts to download `codecov` using `pip` in order to calculate the coverage.

**To Reproduce**

1. Open a terminal window
2. Type `pip install codecov`
3. See error

**Additional context**
A [migration guide](https://docs.codecov.com/docs/deprecated-uploader-migration-guide) for how to use `codecov` has been produced, and there is also [an example CI action file](https://github.com/codecov/example-python/blob/main/.github/workflows/ci.yml) showing how to use `codecov` without `pip`

jacobcook1995:
# Description

This fixes the issue that arises from the removal of `codecov` from PyPi. I did this following the [template](https://github.com/codecov/example-python/blob/main/.github/workflows/ci.yml) provided by `codecov`. This involves calculating the coverage as part of the `pytest` run and then uploading the generated coverage using the `codecov` GitHub action. 

Fixes #199

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,7 @@ jobs:
 
     - name: Run tests + codecov
       if: (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9')
-      run:
-        poetry run pytest --cov app
+      run: poetry run pytest --cov app
 
     - name: Upload coverage to Codecov
       if: (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9')
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,10 @@ jobs:
       run:
         poetry run pytest --cov app
 
+    - name: Upload coverage to Codecov
+      if: (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9')
+      uses: codecov/codecov-action@v3
+
   docs_build:
     needs: qa
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,14 +36,14 @@ jobs:
     - name: Install dependencies
       run: poetry install
 
-    - name: Run tests
+    - name: Run tests (no codecov)
+      if: false == (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9')
       run: poetry run pytest
 
-    - name: Run codecov
-      if: success() && (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.9')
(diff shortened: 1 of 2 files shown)


davidorme:
Looks sane to me 😄 Is there any mileage in just running the `codecov` upload on `pull_request` and not `push` just to cut down on the frequency? Seems like the kind of stat that we care about at PR. And it would be good if this PR added the `codecov` badge too, so that we've got a link to that uploaded coverage and the badge?

davidorme (approved):
Forgot to make that comment a review! Fine as it is, but it would be good to hear thoughts on the comment.

jacobcook1995:
> Looks sane to me 😄 Is there any mileage in just running the `codecov` upload on `pull_request` and not `push` just to cut down on the frequency? Seems like the kind of stat that we care about at PR. And it would be good if this PR added the `codecov` badge too, so that we've got a link to that uploaded coverage and the badge?

Making coverage `pull_request` (and I guess `release`) only definitely seems sensible to me. Happy to take a look at adding the `codecov` badge as well!

jacobcook1995:
Only thing is if we wanted to look at code coverage for specific branches before they are pull requested (which the badge would be helpful for), we would need to calculate code coverage for each push. If that is something we are interested in?

davidorme:
I guess that is true! Are the badges associated with individual commits?

jacobcook1995:
As far as I can tell the badges just link to a codecov.io page for the repo in question, on this page you can compare coverage between branches, but the badge always displays the coverage for the default branch (so I `feature/my_feature` would show the coverage for `develop`). There might be a way to override this behaviour though

davidorme:
I suspect it doesn't make much difference, but this might be cleaner?

```yaml
    - name: Run tests
      run: poetry run pytest

    - name: Run codecov
      if: success() && (matrix.os == 'ubuntu-latest' && matrix.python-version == 3.9)
      uses: codecov/codecov-action@v3
```

jacobcook1995:
Yeah did consider that, the thing that worried me is that the code coverage would have to be computed for every single test (as it is computed using `poetry run pytest --cov app`). I wasn't sure whether the step to compute the coverage or the step to upload the coverage was more demanding. If it's the upload I'm happy to use your approach as it is a lot cleaner

davidorme:
That's a good point - I'm not calculating any coverage for any of my tests!

PR #201, opened 2023-04-20
Feature/data object toml

vgro:
This is a list that contains (eventually all) variables that could potentially be stored in the data object. This list should be used to check for consistent variables names and properties across all modules. At the moment, it is just a collection where everyone can add their variables in a central location.

In the future, we might want to split the list according to the source or use of the variables, but this can be discussed when needed. Further, we already spoke about adding a flag for optional variables, but this is also not urgent.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/data_variables.toml
+++ b/virtual_rainforest/data_variables.toml
@@ -2,7 +2,8 @@
 name = "air_temperature"
 description = "air temperature profile"
 unit = "C"
-source = "abiotic"
+initialised_by = "abiotic"
+updated_by = "abiotic"
 used_by = [ "abiotic", "plants", "animals"]
 
 [[variable]]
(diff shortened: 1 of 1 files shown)


virtual_rainforest/data_variables.toml
```diff
@@ -0,0 +1,385 @@
...
+name = "atmospheric_pressure_ref"
+description = "atmospheric pressure at reference height (2m)"
+unit = "kPa"
+source = ["external"] 
+used_by = [ "abiotic"]
+
+[[variable]]
+name = "bulk_aerodynamic_resistance"
+description = "bulk aerodynamic resistance"
+unit = "s m-1"
```
jacobcook1995, line 60: We might have to think about how we express this type of unit, I think calling this "m^-1" might be better. Might be somewhat tricky to address now though as it very much depends on a what can be read correctly (which I'm not totally sure on)

virtual_rainforest/data_variables.toml
```diff
@@ -0,0 +1,385 @@
...
+description = "Clay as a percentage of total soil"
+unit = "%"
+source = "external"
+used_by = [ "abiotic", "soil"]
+
+[[variable]]
+name = "pH"
+description = "Soil pH"
+unit = "pH"
+source = "external + soil"
```
jacobcook1995, line 383: We probably also need to make sure we can express this consistently. "source1 + source2" definitely would be parseable downstream, but we would need to make sure we use it in every case

jacobcook1995 (approved):
LGTM! I added a couple of comments for areas where we might need to settle on a consistent style

virtual_rainforest/data_variables.toml
vgro: oh yes, there was something similar with an abiotic variables, I wanted to put a comment here but forgot. I think it might be best to split in into two variables?

davidorme: I'm not quite sure what it means when there is external _and_ a model? If it means that soil modifies external pH then it isn't a source, but it could mean that soil provides a default? This could be a list of sources? `source = ["external", "soil"]`

virtual_rainforest/data_variables.toml
```diff
@@ -0,0 +1,385 @@
...
+description = "CO2 concentration at reference height (above canopy)"
+unit = "ppm"
+source = "external"
+used_by = [ "abiotic"]
+
+[[variable]]
+name = "atmospheric_pressure"
+description = "atmospheric pressure profile"
+unit = "kPa"
+source = ["abiotic"]
```
davidorme, line 47: There is an inconsistent structure for `source`: mostly it is a string, but sometimes it is a list of string(s). Given @jacobcook1995 's comment about "external + soil", they should probably all be lists?

davidorme (approved):
LGTM

virtual_rainforest/data_variables.toml
davidorme: Or if it even needs to be parsed. I'm not sure how we'd actually use parsed units unless we attach units to everything.

virtual_rainforest/data_variables.toml
jacobcook1995: I mean I wasn't sure how to notate this one, basically pH will be drawn from data but down the line we might introduce a pH model to modify pH as the simulation proceeds (either based on changes in N or due to disturbance scenarios).

jacobcook1995: This could maybe become "modified_by" rather than "source" as I guess that's what we want to capture here

davidorme: Or "updated_by" to align with the BaseModel terminology. And I guess this should only be a single model, so just one string.

virtual_rainforest/data_variables.toml
vgro: Good point, I think it is still useful at the moment while we use it as a lookup table, too

virtual_rainforest/data_variables.toml
vgro: There is a difference between variables that come from an external source and others that are updated internally. I think they should have different names, such as `pH_input` or something like that. I used `_ref` for variables at reference height.

jacobcook1995: Wouldn't this basically double the number of variables? At least in the soil module pretty much every variable stored in the `data` object will be initialised using external data, duplicating everything seems like it could get pretty confusing. I guess the problem can be avoided once we store variables over time (rather than replacing everything at each time step) as the initial state will be preserved.

jacobcook1995: Though maybe I'm confused and you are referring to continuous data from external sources, which the soil module doesn't have any of at the moment (I doubt it ever will)

jacobcook1995: Ahhh right I misunderstood then 🙃. Completely agree that we should not be altering continuous input data!

jacobcook1995 (approved):
All the soil ones are correct!

vgro:
This is an intermediate solution, however, I merge so that others can add their variables easily. Need to discuss units and use in VR in the future, see #202

PR #206, opened 2023-04-24
Extend carbon pool from 2 to 3 pools

### LINKED ISSUE #161: Soil organic matter binding sub-model poorly parameterised
Having looked into the model parameterisation in detail I am fairly sure that there is a unit error in the mineral association of soil organic matter model that I am using.

However, it's not a simple fix as if I change the units the binding of organic matter to soil minerals becomes orders of magnitudes too fast. For this reason I am sticking with the current parameterisation for now. However, once I start validating the model this must be addressed.

I may be misreading something in the paper or missing something so I first should double check that I am correct. If I am, I need to look elsewhere to see if I can pin down a better parameterisation.

Other models in the literature use a different structure for binding of soil organic matter to minerals, which limits the rate at which organic matter can bind. This could suggest that there is a problem with the model structure which is masked by a parameterisation error. If this is the case I will have to change the model structure.

**Additional context**
This isn't a problem for our current milestone (where we just want to setup and run a set of interacting models), but for subsequent milestones it will be a pressing issue

jacobcook1995:
# Description

This pull request adds a microbial biomass pool to the soil carbon model. This means the model now has three pools which carbon can be exchanged between. I've taken parameters from a previous work (Abramoff et al. 2018), and verified that running the model with these parameters results in plausible long term behaviour.

The structure hasn't really changed from when it was a soil pool models, it's just had more functions hooked into it. It would be good to gets thoughts on whether the overall function structure (etc) makes sense.

Fixes #161

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/soil/constants.py
```diff
@@ -50,3 +55,54 @@ class TempScalar:
...
+LITTER_INPUT_RATE = 0.172 / 365.25
+"""Rate of litter input to the system [kg C m^-2 day^-1].
+
+This definitely is not a constant for our purposes. However,
+:cite:t:`abramoff_millennial_2018` use a constant litter input rate, so we shall also
+use one initially.
+"""
+
+
+@dataclass
```
alexdewar, line 96: I only learned about this recently, but apparently you can make `dataclass`es immutable, which I think would make sense here:

```suggestion
@dataclass(frozen=True)
```

alexdewar (approved):
The science is rather over my head, but the code looks good to me!

virtual_rainforest/models/soil/constants.py
davidorme: Yup - I did the same thing over in `pyrealm` in a fit of paranoia to stop people altering constants mid-run!

jacobcook1995: Ah yeah that definitely seems sensible!

davidorme: This might be a wider issue to apply to other constants classes?

vgro (approved):
LGTM. I can also not comment much on the science without digging a bit deeper in my memory, but the structure and functions look sensible to me.

davidorme (approved):
The new content all LGTM but there is a wider issue with constants that I think needs addressing. I'm not sure it needs addressing now, so I've approved this, but if it isn't fixed in this PR, I think it needs an issue.

You have a set of constants and dataclasses of coefficients which are then used in the functions. Many of those are not 'real' constants and it is very likely that someone (probably us!) will want to look at sensitivity. There are two issues:

1.  Being able to alter the values. It is possible to create an instance of a frozen dataclass  with alternative parameters (you just can't change it after creation), but it isn't clear to me that users can alter `Final` typed  constants at all. 
2. Being able to use alternative values. At the moment, alternative parameterisations cannot be passed down.
    
    For example: `calculate_microbial_carbon_uptake` has the argument `max_uptake_rate: float = MAX_UPTAKE_RATE_LABILE_C`. When it is called in `calculate_soil_carbon_updates`, that argument is not provided and there is no argument to `calculate_soil_carbon_updates` that would allow an alternative value to be substituted. 
    
There isn't an obvious answer here. In `pyrealm`, the high level functions take a single `dataclass` of constants that are passed down to all functions. That makes for a big ugly list of parameters, but means that the values can be adjusted. We don't yet have a way for users to populate that alternative constants class when running from the command line, but then there is  a route to adjust values in code.

virtual_rainforest/models/soil/constants.py
jacobcook1995: Have switched to using `@dataclass(frozen=True)`. I've also used [`typing.Final`](https://docs.python.org/3/library/typing.html#typing.Final) for the constants that aren't stored in `data` classes, this won't stop them from being altered but will cause a `mypy` error if they are altered

jacobcook1995:
Yeah definitely agree it needs an issue, pretty much everything I've put in `soil/constants.py` is only constant in the sense that it shouldn't be changed during a run (or across space). They are all things that we would want to be able to alter once we start looking at model sensitivity (though it is unlikely they all will have survived to that point).

Issue #204, opened 2023-04-21
Time scale bounds for each model class


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
We plan for the simulation time step to be configurable by the user. However, this runs into the problem that certain models will only be valid at a certain update frequency, e.g. when modelling plants there are sub-daily processes which can be ignored at monthly time scales but not at daily time scales.

**Describe the solution you'd like**
It would be good if each model had to define a minimum and maximum time scale for which the processes it describes can sensibly be run for. A system could then be set up to warn (stop?) the user if any of the models they have chosen to configure are not valid for the desired time step.

**Describe alternatives you've considered**
The only other option is to enforce bounds on the time step the user can select but this feels overly restrictive.

PR #211, opened 2023-04-26
Introduce common update time step and timescale bounds for all models

### LINKED ISSUE #203: Common time step for all models
We recently agreed that the simulation should use a common time step, that is that all models should be updated with the same frequency. Internal updates to models can still occur at a faster rate, but information will only be written to the `data` object (i.e. shared between models) with the frequency set by the common time step.

The current behaviour of each model having a different update interval should be removed in favour of a single common update interval at which all models are updated.

We agreed that a common time step is the most sensible way forward, but are open to the fact that exceptions may have to be made.

jacobcook1995:
# Description

This pull request removes all model specific time steps in favour of a common time step at which all models are updated.

In addition to this, it adds time bound attributes to `BaseModel`. These are required properties for all inheriting classes, and set the longest and shortest update frequency that the model makes sense to be run at. For example, it does not make sense to update a plant model which ignores sub-daily processes every hour. The intention here is to prevent the user from selecting a set of models with incompatible time scales.

@TaranRallings, I made my best guess on sensible bounds on the `AnimalModel`. Could you let me know what they should actually be, cheers.

Fixes #203, #204 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,12 @@ The research is based at Imperial College London:
 ## Project Team
 
 * Professor Robert Ewers
-* Olivia Daniels
+* Olivia Daniel
 * Dr. Jaideep Joshi
 * Dr. David Orme
 * Dr. Vivienne Groner
 * Dr. Jacob Cook
-* Taran Rallings
+* Dr. Taran Rallings
 
 The research team are supported by the Imperial College London
 [Research Software
--- a/docs/source/virtual_rainforest/main_simulation.md
+++ b/docs/source/virtual_rainforest/main_simulation.md
@@ -48,7 +48,9 @@ the configuration, they are then all configured.
 The configuration contains a start time, an end time and an time interval for the update
 of every model. These details are extracted from configuration, with a check performed
 to ensure that the simulation will update at least once between the start and end time
-of the simulation.
+of the simulation. It is important to note that because months and years are of
+inconsistent length they are currently averaged over. This means that 1 month becomes
+30.4375 days, and 1 year becomes 365.25 days.
 
 ## Saving the initial state
 
--- a/docs/source/virtual_rainforest/main_simulation.md
+++ b/docs/source/virtual_rainforest/main_simulation.md
@@ -45,11 +45,10 @@ the configuration, they are then all configured.
 
 ## Extracting simulation timing details
 
-The configuration contains a start time, an end time and an time interval for checking
-whether models need to be updated. Once these details are extracted from the
-configuration, a check is performed to ensure that the simulation update interval works
-suitably with the time steps for individual models, i.e. that no model has a shorter
-time step than the overall simulation update interval.
+The configuration contains a start time, an end time and an time interval for the update
+of every model. These details are extracted from configuration, with a check performed
+to ensure that the simulation will update at least once between the start and end time
+of the simulation.
 
 ## Saving the initial state
 
@@ -61,8 +60,7 @@ a single data file of the initial model state.
 
 The previously extracted timing details are used to setup a timing loop, which runs from
 the start time to the end time with a time step set by the update interval. At each
-step a check is performed to determine if any models need to be updated. Any that do
-need to be updated are then updated.
+step all models are updated.
 
 ## Saving the final state
 
--- a/tests/core/data/all_config.toml
+++ b/tests/core/data/all_config.toml
@@ -26,6 +26,3 @@ maxh = 1.0
 [[plants.ftypes]]
 pft_name = "broadleaf"
 maxh = 50.0
-
-[soil]
-model_time_step = "2 weeks"
--- a/tests/core/data/all_config.toml
+++ b/tests/core/data/all_config.toml
@@ -7,7 +7,7 @@ ny = 10
 
 [core.timing]
 start_date = "2020-01-01"
-update_interval = "10 minutes"
+update_interval = "2 weeks"
 run_length = "50 years"
 
 [core.data_output_options]
@@ -28,4 +28,4 @@ pft_name = "broadleaf"
 maxh = 50.0
 
 [soil]
-model_time_step = "5 days"
+model_time_step = "2 weeks"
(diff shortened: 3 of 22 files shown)


docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -45,11 +45,12 @@ the configuration, they are then all configured.
 
 ## Extracting simulation timing details
 
-The configuration contains a start time, an end time and an time interval for checking
-whether models need to be updated. Once these details are extracted from the
-configuration, a check is performed to ensure that the simulation update interval works
-suitably with the time steps for individual models, i.e. that no model has a shorter
-time step than the overall simulation update interval.
+The configuration contains a start time, an end time and an time interval for the update
```
vgro, line 48:
```suggestion
The configuration contains a start time, an end time and a time interval for the update
```

tests/core/data/all_config.toml
```diff
@@ -7,7 +7,7 @@ ny = 10
 
 [core.timing]
 start_date = "2020-01-01"
-update_interval = "10 minutes"
+update_interval = "2 weeks"
```
vgro, line 10: Why did you chose 2 weeks instead of 1 months here?

vgro (approved):
LGTM, I made a few minor suggestions.

tests/core/data/all_config.toml
jacobcook1995: This file is used for testing, basically wanted something here that was different from the default value, which is 1 month. I basically made everything in this file different to their respective default values to check that overwriting of defaults works. I don't really need to do this for every single tag though, so could change this to a month if that makes things clearer

davidorme: I'd just add a comment to note the deliberate changes from default. Another advantage of TOML over JSON 😄

dalonsoa (approved):
It looks good! I think it is a good idea to have a single update_interval and just check that it is within the models meaningful values to be updated.

README.md
```diff
@@ -31,12 +31,12 @@ The research is based at Imperial College London:
...
 
 * Professor Robert Ewers
-* Olivia Daniels
+* Olivia Daniel
 * Dr. Jaideep Joshi
 * Dr. David Orme
 * Dr. Vivienne Groner
 * Dr. Jacob Cook
-* Taran Rallings
+* Dr. Taran Rallings
```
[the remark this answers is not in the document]
TaranRallings: This is the best PR request we've ever done!

virtual_rainforest/core/base_model.py
```diff
@@ -262,6 +290,81 @@ def _check_required_init_vars(cls) -> None:
             LOGGER.error(to_raise)
             raise to_raise
 
+    @classmethod
+    def _check_time_bounds_units(cls) -> None:
+        """Check that the time bounds defined by each model have time units.
```
davidorme, line 295: This is repetitive - we could have:

```suggestion
    def _check_time_bounds_units(cls, which: str = "lower_bound_on_time_scale") -> Quantity
        """Check that the time bounds defined by each model have time units..."""
        # extract the property to test
        val = cls.get(which)
```

virtual_rainforest/core/base_model.py
```diff
@@ -279,6 +382,8 @@ def __init_subclass__(cls) -> None:
         try:
             cls._check_model_name()
             cls._check_required_init_vars()
+            cls._check_time_bounds_units()
+            cls._check_time_bounds_values()
```
davidorme, line 386: And then here we could do:

```python
lwbnd = cls._check_time_bounds_units('lower_bound_on_time_scale')
upbnd = cls._check_time_bounds_units('upper_bound_on_time_scale')
```

And then outside of the try just run the separate check for `lwbnd <= upbnd`.

virtual_rainforest/core/base_model.py
```diff
@@ -262,6 +290,81 @@ def _check_required_init_vars(cls) -> None:
...
+    def _check_time_bounds_values(cls) -> None:
+        """Check that the lower time bound is less than the upper bound.
+
+        Raises:
+            ValueError: If the lower bound is not less than the upper bound
+        """
+
+        # Find upper and lower bound
+        upper_bound = pint.Quantity(cls.upper_bound_on_time_scale)
+        lower_bound = pint.Quantity(cls.lower_bound_on_time_scale)
```
davidorme, line 359: Repeats what the `units` method does.

virtual_rainforest/core/utils.py
```diff
@@ -9,41 +9,59 @@
...
 from virtual_rainforest.core.exceptions import ConfigurationError, InitialisationError
 from virtual_rainforest.core.logger import LOGGER
 
 
-def extract_model_time_details(
-    config: dict[str, Any], model_name: str
-) -> tuple[datetime64, timedelta64]:
+def extract_update_interval(
+    config: dict[str, Any], lower_bound: str, upper_bound: str
+) -> timedelta64:
```
davidorme, line 20: Isn't this now largely a method of `BaseModel` that gets called in `BaseModel.__init__`? We should parse the `pint` Quantity for `update_interval` once from the config in setup in `main` and then it gets passed in to the BaseModel. So then we'd have this as:

```python
def _check_update_speed(self, update_interval: Quantity) -> Quantity:
```
and then in `BaseModel.__init__` do:

```python
self.update_interval = self._check_update_speed(update_interval)
```

davidorme (changes requested):
This all looks good - I think some things can be simplified and moved around a bit, but nothing is ringing any warning bells.

virtual_rainforest/core/base_model.py
davidorme: That probably makes `_check_time_bounds_values()` small enough to simply move the actual check back into `__init_subclass__`?

jacobcook1995:
@davidorme did what you suggested and moved `extract_update_interval()` into `BaseModel` (+ renamed it). Took a while to get all the unit tests to play ball

davidorme (approved):
LGTM

davidorme (approved):
LGTM

PR #218, opened 2023-05-17
Fixs for `pint` problem in docs build action

jacobcook1995:
# Description

It seems like `pint` have changed how they do typing, which caused documentation build failures. I have set `intersphinx` to read from an earlier version of the `pint` documentation to avoid this problem.

This will have to change in future if we update `pint`, but at the moment it seems like they are still trying to pin down their typing, so doesn't seem sensible to refactor the code at the moment.

## Type of change

## Key checklist

## Further checks


vgro (approved):
LGTM :-)

PR #219, opened 2023-05-17
Feature/abiotic simple model

vgro:
`abiotic_simple_model` (see #216) which includes the new way of handing the update interval across models. It also includes a 'counter' to allow the model to access the correct index from a time series of climate data input. The `setup` and `update` functions will be part of a separate PR on the `simple_regression`.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/abiotic_simple/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+the microclimate and hydrology for the Virtual Rainforest.
+
+Each of the abiotic sub-modules has its own API reference page:
+
+* The :mod:`~virtual_rainforest.models.abiotic_simple.abiotic_simple_model` submodule
+  instantiates the AbioticSimpleModel class which consolidates the functionality of the
+  abiotic simple module into a single class, which the high level functions of the
+  Virtual Rainforest can then use.
+
+* The virtual_rainforest.models.abiotic_simple.simple_regression submodule
```
jacobcook1995, line 12: Maybe stick a TODO in somewhere around here to remind yourself to make it a link once that submodule is defined? It's the sort of thing that I'm always convinced I will remember but always forget

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -0,0 +1,232 @@
...
+        update_interval: Time to wait between updates of the model state.
+        soil_layers: The number of soil layers to be modelled.
+        canopy_layers: The initial number of canopy layers to be modelled.
+        initial_soil_moisture: The initial soil moisture for all layers.
+    """
+
+    model_name = "abiotic_simple"
+    """The model name for use in registering the model and logging."""
+    lower_bound_on_time_scale = "1 day"
+    """Shortest time scale that soil model can sensibly capture."""
```
jacobcook1995, line 49: "that the abiotic simple model" I guess? Also same comment for the doc string below

jacobcook1995 (approved):
LGTM. Just a couple of minor comments

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -0,0 +1,232 @@
...
+    ):
+        # sanity checks for soil and canopy layers
+        if soil_layers < 1:
+            to_raise = InitialisationError(
+                "There has to be at least one soil layer in the abiotic model!"
+            )
+            LOGGER.error(to_raise)
+            raise to_raise
+
+        if soil_layers != int(soil_layers):
```
davidorme, line 81: It might be better to test here that `soil_layers` is an `int` object? This allows floats with no fractional part through - and that seems functionally fine - but it is probably best to ensure that only actual `int` values make it here?

```suggestion
        if not isinstance(soil_layers, int):
```

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -0,0 +1,232 @@
...
+    def update(self) -> None:
+        """Placeholder function to update the abiotic simple model."""
+
+        # output_variables = simple_regression.run_simple_regression(
+        #     data=self.data,
+        #     layer_roles=self.layer_roles,
+        #     time_index=self.time_index,
+        # )
+        # update_data_object(data=self.data, output_dict=output_variables)
+        self.time_index += 1
```
davidorme, line 182: Ohhhh. I get this now. At the moment, the `BaseModel.update` method just says update - it doesn't say this is the update for 20th June 2014. So _any_ module that needed to refer to data with a temporal dimension would need to keep an internal tracker. At the moment, that is just `abiotic` but that is likely to change at some point.

So... @jacobcook1995 @dalonsoa I think we need to amend the `update` method to pass in the timepoint from the `vr_run` loop? That could be a datetime object that can then be used to find the matching time slice, or it could be that `vr_run` tracks the indices of the timesteps alonf the temporal axis and simply passes in that index.

davidorme (approved):
LGTM - I think your counter is a local solution to something that needs a general solution (and I think that general solution should be quite simple). I don't think pulling this needs to wait for that though - we can replace the local solution later.

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
jacobcook1995: Yeah this is something we definitely need to give thought to down the line. (I would personally favour something that regularises the data such that a simple index can be used throughout).

I don't think it can be resolved in the near future as it requires us to all agree on the exact details of the timing system we will use, e.g. are we using 1 month time stamps (i.e. Jan = 31 days, Feb = 28/29 days, March = 31 days, etc), or an average length month (1 month = 30.437), or simplifying to a standard 30 day month.

I don't think we can reach a consensus on what is the best approach is until we know the needs of each model. So, I think this is definitely something to be resolved down the line.

davidorme: The temporal axis of the data drives it, I think? If we are working off a common time step, then that should be synchronised with whatever the increment is in the data. Or perhaps more pragmatically, we can use `isel` within the models to find the nearest match to the simple timestamp from the `vr_run` loop. That potentially leads to using data from the future, so we'd need to be careful. Alternatively `vr_run` just increments over the temporal axis from the data rather than using a separate update interval calculation?

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
vgro: upsi, yes, thanks :-)

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
vgro: Thanks for your comments on this. I agree that in the long-term we should have a time axis and counter outside the abiotic model but for the first tests it should work. I  added a `time_index` coordinate to my dummy dataset to try this as soon as the model is complete in develop.

PR #221, opened 2023-05-18
Feature/simple regression

vgro:
This PR reintroduces the simple regression functions back in the `abiotic_simple_model` as in #216.

Based on previous comments, I simplified the code where possible and added sanity checks and tests for exceptions.

The time looping is still to be discussed in the group but we agreed to leave it as it is for now.

I temporarily hard-coded two variables in the `setup_simple_regression` that I expect to be initialized by the plant model: `layer_heights` and `leaf_area_index`.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+        .rename({"dim_0": "layers", "dim_1": "cell_id"})
+        .assign_coords(
+            {
+                "layers": np.arange(0, 15),
+                "layer_roles": ("layers", layer_roles),
+                "cell_id": data.grid.cell_id,
+            }
+        )
+    )
+
```
vgro, line 131: This could possibly be shorter but it causes errors to just do all in one go, same with `leaf_area_index` below

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+MicroclimateParameters: Dict[str, float] = {
+    "soil_moisture_capacity": 90,
+    "water_interception_factor": 0.1,
+    "saturation_vapour_pressure_factor1": 0.61078,
+    "saturation_vapour_pressure_factor2": 7.5,
+    "saturation_vapour_pressure_factor3": 237.3,
+}
+"""Parameters for simple abiotic regression model."""
+
+Bounds: Dict[str, float] = {
```
jacobcook1995, line 41: What happens when these bounds are exceeded?

jacobcook1995 (approved):
LGTM! I just had some queries/comments about some low level stuff

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+                layers=np.arange(0, len(layer_roles) - layer_roles.count("soil"))
+            ),
+            soil_temperature_only,
+        ],
+        dim="layers",
+    )
+
+    # Mean atmospheric pressure profile, [kPa]
+    output["atmospheric_pressure"] = data["atmospheric_pressure"].where(
+        data["atmospheric_pressure"].coords["layer_roles"] == "soil",
```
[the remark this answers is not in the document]
vgro: It is weird and counterintuitive looking but this function does what it is supposed to.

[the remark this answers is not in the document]
davidorme: It really is counter-intuitive - and the opposite of `np.where`.

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+    Args:
+        layer_roles: list of layer roles (from top to bottom: above, canopy, subcanopy,
+            surface, soil)
+        data: Data object
+        initial_soil_moisture: initial soil moisture
+
+    Returns:
+        Dict of DataArrays for air temperature [C], relative humidity [-], vapour
+        pressure deficit [kPa], soil temperature [C], atmospheric pressure [kPa],
+        atmospheric :math:`\ce{CO2}` [ppm], soil moisture [-], and surface runoff [mm]
```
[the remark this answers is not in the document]
vgro: I raises an issue #222

virtual_rainforest/models/abiotic_simple/simple_regression.py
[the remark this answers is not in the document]
vgro: That's what I thought but for some reason this is what gives the correct result; when I try `!=` it fills the wrong parts of the DataArray. Maybe I am missing something?

With this:
```
atmospheric_pressure = data["atmospheric_pressure"].where(
        data["atmospheric_pressure"].coords["layer_roles"] == "soil",
        (data["atmospheric_pressure_ref"].isel(time=0)),
    )
```
I get this (which is what I want):
```xarray.DataArray'atmospheric_pressure' layers: (15 cell_id: 3)
array([[1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5],
       [nan, nan, nan],
       [nan, nan, nan]])
Coordinates:
cell_id
(cell_id)
int64
0 1 2
layers
(layers)
int64
0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
layer_roles
(layers)
<U9
'above' 'canopy' ... 'soil' 'soil'
Indexes: (2)
Attributes: (0)
```

But with `!=` I get this:
```
xarray.DataArray'atmospheric_pressure' (layers: 15cell_id: 3)
array([[nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [nan, nan, nan],
       [1.5, 1.5, 1.5],
       [1.5, 1.5, 1.5]])
Coordinates:
cell_id
(cell_id)
int64
0 1 2
layers
(layers)
int64
0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
layer_roles
(layers)
<U9
'above' 'canopy' ... 'soil' 'soil'
Indexes: (2)
Attributes: (0)
```

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+        dim="layers",
+    )
+    output["surface_run_off"] = surface_run_off
+
+    return output
+
+
+# supporting functions
+def logarithmic(x: DataArray, gradient: float, intercept: float) -> DataArray:
+    """Logarithmic function.
```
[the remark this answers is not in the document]
vgro: good point. I noticed that the function is not longer used, but I will remember for the future :-)

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+
+    Args:
+        x: x values
+        gradient: gradient
+        intercept: intercept
```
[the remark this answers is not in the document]
vgro: good point. I noticed that the function is not longer used, but I will remember for the future :-)

virtual_rainforest/models/abiotic_simple/simple_regression.py
vgro: When the bounds are exceeded, the value is reduced to the bound value, for example the maximum relative humidity is 100%, so if it gets higher than that, the value is set to 100; equally if it is below 0 it is set to 0. This does obviously not conserve energy or matter, I will add a comment on that and a TODO

virtual_rainforest/models/abiotic_simple/simple_regression.py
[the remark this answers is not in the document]
vgro: @dalonsoa do you have an idea why this happens?

davidorme (changes requested):
I think there are some structural issues - clearly nothing that stops it working 😄 but I think it is worth revising some of it. I do have more comments to add, but I've got a problem with out of date diffs that is blocking them.

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,571 @@
...
+    MicroclimateGradients: Dict[str, float] = MicroclimateGradients,
+    Bounds: Dict[str, float] = Bounds,
+    water_interception_factor: Union[DataArray, float] = MicroclimateParameters[
+        "water_interception_factor"
+    ],
+    soil_moisture_capacity: Union[DataArray, float] = MicroclimateParameters[
+        "soil_moisture_capacity"
+    ],
+) -> Dict[str, DataArray]:
+    r"""Calculate simple microclimate.
```
davidorme, line 169: This function feels like it is the content of `update` but outsourced into a stand alone function. I think it is probably better if the main sequence stays in the model, possibly in extra methods, and external static functions are only used for smaller steps.

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,571 @@
...
+    output["air_temperature"] = log_interpolation(
+        data=data,
+        reference_data=data["air_temperature_ref"].isel(time=time_index),
+        leaf_area_index_sum=leaf_area_index_sum,
+        layer_roles=layer_roles,
+        layer_heights=data["layer_heights"],
+        upper_bound=Bounds["air_temperature_max"],
+        lower_bound=Bounds["air_temperature_min"],
+        gradient=MicroclimateGradients["air_temperature_gradient"],
+    ).rename("air_temperature")
```
davidorme, line 253: Several of these blocks are repeated. I'd be tempted to restructure/rename elements so that you can use a single variable name in a loop. Something like:

```python
for vars in ['air_temperature', ...]:

    output[var] = log_interpolation(
        data=data,
        reference_data=data[var + "_ref"].isel(time=time_index),
        leaf_area_index_sum=leaf_area_index_sum,
        layer_roles=layer_roles,
        layer_heights=data["layer_heights"],
        upper_bound=Bounds[var].max,
        lower_bound=Bounds[var].min,  # Or even just bounds = Bounds[var]
        gradient=MicroclimateGradients[var],
    ).rename(var)

Comment (davidorme) [reply virtual_rainforest/models/abiotic_simple/simple_regression.py:253]: That function is also unpacking nearly everything it needs from `data` into arguments, but still passing in `data`. You could just pass in data and the `var` name and extract internally, or pass in all the required arguments ready extracted - I think it is just `cell_id` you're getting from `data`?

**Review by davidorme — CHANGES_REQUESTED** [review]
Some extra comments on the structure of `run_simple_regression`. Sorry this came in as two bits!

File: virtual_rainforest/models/abiotic_simple/simple_regression.py (hunk 1 above)
Comment (vgro) [reply virtual_rainforest/models/abiotic_simple/simple_regression.py:134]: What exactly is the `.T` for? Doe this assign temperature as an attribute?

Comment (davidorme) [reply virtual_rainforest/models/abiotic_simple/simple_regression.py:134]: No - sorry that is obscure. It is a `numpy` property that returns the matrix transpose. It just rotate the 3x15 matrix created by broadcasting the vertical layers into a 15x3 matrix as you had in the original inputs. There are a million ways we could create this matrix in `numpy` but this seemed most compact for now.

Comment (vgro) [reply virtual_rainforest/models/abiotic_simple/simple_regression.py:134]: Ah ok, that makes a lot of sense. Thank you :-)

File: virtual_rainforest/models/abiotic_simple/simple_regression.py [hunk 9]
```diff
@@ -0,0 +1,588 @@
+r"""The ``models.abiotic_simple.simple_regression`` module uses linear regression
+and logarithmic interpolation to calculate atmospheric temperature and humidity profiles
+as a function of leaf area index and canopy layer height. The relationships are derived
+from :cite:t:`hardwick_relationship_2015` and :cite:t:`jucker_canopy_2018`.
```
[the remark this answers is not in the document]
vgro: updated :-)

virtual_rainforest/models/abiotic_simple/simple_regression.py
dalonsoa: Yeah, maybe. I'm never sure with this things, to be honest. I tend to think of methods as relatively small functions that either update the state of the object or provide some output to be used elsewhere that make use of the object state. This function do the first thing, but is pretty large, so it does not fit into that definition (a personal definition). When in doubt, I tend to choose the option that makes testing easier.

dalonsoa (changes requested):
I have mostly agreed with David's comments, just adding my opinion in a couple of places. Otherwise, it looks really good and mature!

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -160,25 +158,25 @@ def from_config(
...
-        #     layer_roles=self.layer_roles,
-        #     data=self.data,
-        #     initial_soil_moisture=self.initial_soil_moisture,
-        # )
-        # update_data_object(data=self.data, output_dict=setup_variables)
+        setup_variables = simple_regression.setup_simple_regression(
+            layer_roles=self.layer_roles,
+            data=self.data,
+            initial_soil_moisture=self.initial_soil_moisture,
+        )
```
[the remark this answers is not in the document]
vgro: I am not sure I fully understand what you mean here.

First, `super().__init_()` is called in the `__init__` and I tested that all required variables are loaded. Did I miss something here?

Second, the variables that are created in the `setup_simple_regression` are not inputs so I don't see how they would be coming from the initial data. Also, they might be different depending on the `abiotic` model selected, so I think their initialization it should stay within the model somewhere. If the setup function is not a good place, what would be an alternative? Maybe config?

Lastly, I see how the inputs that come from the `plants` should be in `conftest` for testing, which in fact they are (jn the `dummy_climate_data` fixture.). However, in order for us to try `vr_run` we will need them available in `data`. An alternative could be to have a `plants` dummy that just created these variables so we can move on. Would you agree?

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+}
+"""Upper and lower bounds for abiotic variables."""
+
+
+def setup_simple_regression(
+    layer_roles: List[str],
+    data: Data,
+    initial_soil_moisture: float = 50,
+) -> Dict[str, DataArray]:
+    r"""Set up abiotic environment variables.
```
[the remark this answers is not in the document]
vgro: Again, I am not sure I fully understand. All variables that are needed for the `abiotic_simple_model` are initialized in the `AbioticSimpleModel.setup()` with the `setup_simple_regression` function, including temporarily the `layer_heights` and `leaf_area_index` which should be initialized by the `plants` model which doesn't currently exist. This is why I don't see the point in moving them to the tests.

virtual_rainforest/models/abiotic_simple/simple_regression.py
vgro: Could you please explain this a bit more? I was trying to keep the `abiotic_simple_model` as tidy as possible and only call very few functions that do all the bits. Do you mean that I should leave only the 'helper functions' in the `simple_regression` and move all the content of `run_simple_regression` to the `AbioticSimpleModel.update()`?

virtual_rainforest/models/abiotic_simple/simple_regression.py
vgro: I combined `air_temperature`, `relative_humidity` and `vapour_pressure_deficit` in a loop. To make it work, I moved the calculation of the reference vapour pressure deficit to the setup (it can be done for all time steps in one go)

tests/core/test_data.py
```diff
@@ -707,3 +708,60 @@ def test_save_to_netcdf(mocker, caplog, fixture_data, out_path, raises, exp_log)
         Path(out_path).unlink()
 
     log_check(caplog, exp_log)
+
+
+def test_Data_replace_from_dict(dummy_climate_data):
+    """Test reading from dictionary."""
```
vgro, line 714: I'm not sure this is the correct way of testing this, could you please check?

virtual_rainforest/models/abiotic_simple/simple_regression.py
```diff
@@ -0,0 +1,588 @@
...
+        gradient: gradient of regression from :cite:t:`hardwick_relationship_2015`
+
+    Returns:
+        vertical profile of provided variable
+    """
+
+    # Calculate microclimatic variable at 1.5 m as function of leaf area index
+    lai_regression = DataArray(
+        leaf_area_index_sum * gradient + reference_data, dims="cell_id"
+    )
```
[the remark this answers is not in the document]
davidorme: Oh right, because otherwise it changes from `numpy` to `DataArray` and the return type is already fixed. It's things like this that I find a little frustrating about `mypy`!

virtual_rainforest/models/abiotic_simple/simple_regression.py
davidorme: This is _definitely_ a personal choice. 

My feeling is that `run_simple_regression` does far more than running a single simple regression and that this function is basically the `update` method, but operating a bit indirectly. It has to pass back a list of updated outputs and then copy them into the object, rather being able to access `self.data` directly. 

We don't want big complex methods, but my comment was motivated by wanting to be able to read the `update` method to get a high level view of the steps being taken. The `run_simple_regression` function  runs a relatively small number of calls to existing low level functions that are easy to test, so packing them all up in this function and then feeding that into `update` felt like an extra layer. The `update` method would not be huge if it just worked through this sequence, directly calling the low level functions and assigning the results into `self.data`.

That's where I was coming from, but as @dalonsoa says, this is very much a personal style thing.

tests/core/test_data.py
davidorme: LGTM

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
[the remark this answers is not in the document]
vgro: Phew, that happens to me all the time with the code that is folded away, so all good :-) 

About initializing nan DataArrays and updating the data object, I think this might have party evolved from the class structure and the initialization of class attributes, and partly from the need to avoid updating the data object directly. Looking now at how I implemented updating the data object, it is not necessary for a variable to be there in the first place, so I can drop all variables from setup that don't contain values.

For the `plants` variables, I think a NetCDF file is a very good solution. I'm happy to create that 👍

virtual_rainforest/models/abiotic_simple/simple_regression.py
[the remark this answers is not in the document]
vgro: oh yes, mypy...

virtual_rainforest/models/abiotic_simple/simple_regression.py
vgro: Your comment makes a lot of sense. I think I could have chosen a better name for the function, it is more like `run_simple_model` which is, like you say, basically `update()`. I think I chose this approach based on the process-based model where I had multiple files that contained functions for one type of process, which is not necessary here. I'll have a think about it and do the other parts first :-)

virtual_rainforest/models/abiotic_simple/abiotic_simple_model.py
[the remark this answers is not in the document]
vgro: I removed the `setup_regression` function and added the initialization of soil moisture and vpd reference to the `abiotic_simple_model`.

virtual_rainforest/models/abiotic_simple/simple_regression.py
vgro: I keep this as it is for now because there will likely be changes to the structure and for now, I find it easier to navigate. I added a comment in the `abiotic_simple_model `that this routine could be unnested and moved to `update()`

virtual_rainforest/core/data.py
```diff
@@ -342,6 +342,22 @@ def save_to_netcdf(self, output_file_path: Path) -> None:
...
+        Args:
+            output_dict: dictionary of variables from submodule
+
+        Returns:
+            an updated data object for the current time step
+        """
+
+        for variable in output_dict:
+            self.data[variable] = output_dict[variable]
+
```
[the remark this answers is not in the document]
vgro: Based on our in person discussion, I renamed the function to `add_from_dict` and added a comment that we might want to split that in the future to have a `replace` option

vgro:
Thank you for your comments.

After a number of iterations, I now dropped the `setup_regression` function completely and instead initialize `soil_moisture` and `vapour_pressure_deficit_ref` in the `abiotic_simple_model`. They are assigned directly to `data`. The DataArrays filled with `np.nan` don't need to be created here. They are assigned directly to `data`.

I keep the `run_simple_regression` for now outside of the `abiotic_simple_model` but added a comment that one layer of nesting could be removed. This means that I also keep the separation of calculations and updating the data object.

Updating the data object is moved to `core.data` and allows to add and replace, I added a comment that this might need to be more selective in the future.

davidorme (approved):
LGTM

dalonsoa (approved):
And to me!

PR #225, opened 2023-05-26
Use abiotic soil layers correctly in soil model

jacobcook1995:
# Description

I've changed the soil model so that it uses the soil layers defined by the abiotic model correctly (well at all 🙂). This involves giving the `SoilModel` class a `top_soil_layer_index` attribute which is calculated using the `data` object used to init the overall class. This attribute is then used as an index to extract the top soil layer from `data` when required.

@vgro, there's no rush to review this. I just wanted to get the pull request in before I clock off for the long weekend.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/models/soil/soil_model.py
+++ b/virtual_rainforest/models/soil/soil_model.py
@@ -101,7 +101,7 @@ class SoilModel(BaseModel):
         # Find first soil layer from the soil temperature in data object
         self.top_soil_layer_index = next(
             i
-            for i, v in enumerate(data["soil_temperature"].coords["layer_roles"])
+            for i, v in enumerate(data["soil_moisture"].coords["layer_roles"])
             if v == "soil"
         )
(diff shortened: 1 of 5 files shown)


vgro (approved):
LGTM. One minor point about `xr.concat` the dummy data fixture and assigning coordinates, please double check :-)

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -98,6 +98,15 @@ def __init__(
         """A Data instance providing access to the shared simulation data."""
         self.update_interval
         """The time interval between model updates."""
+        # Find first soil layer from the soil temperature in data object
+        self.top_soil_layer_index = next(
+            i
+            for i, v in enumerate(data["soil_temperature"].coords["layer_roles"])
+            if v == "soil"
```
vgro, line 105: Better take index from `soil_moisture` which will definitely be initialized in the `setup()`

PR #230, opened 2023-05-31
Feature/animal module - Ectothermic metabolism

TaranRallings:
# Description

This adds a temperature sensitive ecothermic metabolism and a temperature insensitive endothermic metabolism to the functional groups. This also works as trial approach to having temperature sensitive functions for ectothermic animals broadly.  Let me know if you think there is a better way to be implementing the idea of temperature input as I intend to reuse the structure. 

The temperature variable I am currently using will be replaced with a reading of the actual temperature data from the grid square and vertical layer the animal occupies. 

I have split the ever-growing dummy_animal_module file into separate files focusing on the 'AnimalCohort' class, the 'AnimalCommunity' class, and the group of dummy plant and soil classes. 

I have also added new input validity checks and revised the testing for 'AnimalCohort'. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


jacobcook1995 (commented):
I had a query about the units, but in general it seems like a sensible approach to me!

virtual_rainforest/models/animals/__init__.py
```diff
@@ -1,15 +1,15 @@
-"""The :mod:`~virtual_rainforest.models.animal` module is one of the component models
+"""The :mod:`~virtual_rainforest.models.animals` module is one of the component models
 of the Virtual Rainforest. It is comprised of a number of submodules.
 
 Each of the animal sub-modules has its own API reference page:
 
-* The :mod:`~virtual_rainforest.models.soil.animal_model` submodule instantiates the
+* The :mod:`~virtual_rainforest.models.animals.animal_model` submodule instantiates the
```
jacobcook1995, line 6: You probably should mention the following api pages in here: animal_communities.md, animal_cohorts.md, functional_group.md, scaling_functions.md, dummy_plants_and_soil.md

TaranRallings: Thanks for the reminder! They were on the list and then I got very wrapped up in the weird errors I was getting.

jacobcook1995: Yes `sphinx` errors are an all consuming nightmare

virtual_rainforest/models/animals/dummy_plants_and_soil.py
```diff
@@ -0,0 +1,52 @@
...
+from virtual_rainforest.models.animals.constants import ENERGY_DENSITY
+
+
+class PlantCommunity:
+    """This is a dummy class of plant cohorts for testing the animal module."""
+
+    def __init__(self, mass: float, position: int) -> None:
+        """The constructor for Plant class."""
+        self.mass = mass
+        """The mass of the plant cohort [kg]."""
```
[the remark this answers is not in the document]
TaranRallings: I'm trying to stick with SI units where I can, so kg for mass. The issue is that the allometric equations I have are in a wild tangle of forms. When we get to dialling in the model parameters, I'll get to standardizing them. Right now, the actual scaling functions account for the discrepancies.

virtual_rainforest/models/animals/constants.py
```diff
@@ -10,13 +10,18 @@
...
+ENDOTHERMIC_METABOLIC_RATE_TERMS: dict[str, tuple[float, float]] = {
+    "mammal": (0.75, 0.018),
+    # Mammalian metabolic rate scaling from Metabolic Ecology p210. [g]
     "bird": (0.75, 0.05),
     # Toy values.
 }
 
+ECTOTHERMIC_METABOLIC_RATE_TERMS: dict[str, tuple[float, float]] = {
+    "insect": (0.75, 0.08)
+    # Insect metabolic rate scaling from Metabolic Ecology p210. [g]
```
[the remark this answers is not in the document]
TaranRallings: Sloppy comment on my part. Some of those unit references are actually reminders to myself about what mass unit the source is assuming. It's all handled in scaling_functions right now but it will save me a lot of annoying time when I am ironing out the parameters later on. I have just added a commit that clarifies what those unit references actually are.

virtual_rainforest/models/animals/dummy_plants_and_soil.py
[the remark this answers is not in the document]
dalonsoa: I think this was already discussed, but sticking to `pint` objects whenever you're dealing with scalars that have units will solve this issue. Not something to change without further discussion, but just an option to consider to avoid possible errors due to the inconsistency of units. I don't know if that would affect the performance significantly, though.

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -73,10 +97,30 @@ def import_functional_groups(fg_file: str) -> list[FunctionalGroup]:
...
+                )
+
+            name, taxa, diet, metabolic_type = row
             # create the FG instance and append it to the list
-            functional_group_list.append(FunctionalGroup(name, taxa, diet))
+            # It's expected that the FunctionalGroup __init__ method
+            # will handle further error checking for the values
+            functional_group_list.append(
+                FunctionalGroup(name, taxa, diet, metabolic_type)
+            )
```
dalonsoa, line 124: This can be removed.

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -73,10 +97,30 @@ def import_functional_groups(fg_file: str) -> list[FunctionalGroup]:
 
     with open(fg_file, newline="") as csv_file:
         reader = csv.reader(csv_file)
```
dalonsoa, line 99: This can be removed.

tests/models/animals/test_animal_model.py
```diff
@@ -63,10 +63,10 @@ def test_animal_model_initialization(
...
                     "model_time_step": "12 hours",
                     "functional_groups": (
-                        ("carnivorous_bird", "bird", "carnivore"),
-                        ("herbivorous_bird", "bird", "herbivore"),
-                        ("carnivorous_mammal", "mammal", "carnivore"),
-                        ("herbivorous_mammal", "mammal", "herbivore"),
+                        ("carnivorous_bird", "bird", "carnivore", "endothermic"),
+                        ("herbivorous_bird", "bird", "herbivore", "endothermic"),
+                        ("carnivorous_mammal", "mammal", "carnivore", "endothermic"),
+                        ("herbivorous_mammal", "mammal", "herbivore", "endothermic"),
```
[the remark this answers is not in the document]
TaranRallings: Endotherms are animals that maintain a steady internal body-temperature (birds/mammals), as opposed to ecotherms that have large swings in body-temperature as a function of environmental temperature (insects/reptiles/etc). 

Endotherms do run into problems at temperature extremes but this needs to be handled later and only if we are looking to run the model in rather nasty climate change scenarios.

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -160,6 +44,18 @@ def __init__(
...
+            raise TypeError("functional_group must be an instance of FunctionalGroup.")
+        """Check if functional_group is an instance of FunctionalGroup."""
+
+        if not isinstance(age, (int, float)) or age < 0:
+            raise ValueError("Age must be a positive number.")
+        """Check if age is a positive number. """
+
+        if not isinstance(mass, (int, float)) or mass < 0:
+            raise ValueError("Mass must be a positive number.")
+        """Check if mass is a positive number."""
```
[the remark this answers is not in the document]
TaranRallings: Ah, fair enough! Here I was, proud that I was validating my data inputs more rigorously haha. I'll try and keep a more explicit eye on this.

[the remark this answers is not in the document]
TaranRallings: Worth noting: the plan is for user-defined inputs of the data to generate functional groups. So they could fill out a csv, for example, and load it up to populate the list of functional groups that goes into generating the model. I think you're right that there isn't a going to a direct interface to put wrong-items into cohort initialization but some of the other checks are about vetting user input.

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -28,15 +27,38 @@ class FunctionalGroup:
 
     """
 
-    def __init__(self, name: str, taxa: str, diet: str) -> None:
+    def __init__(self, name: str, taxa: str, diet: str, metabolic_type: str) -> None:
         """The constructor for the FunctionalGroup class."""
+        # Check for valid inputs
+        valid_taxa = ["mammal", "bird", "insect"]
```
dalonsoa, line 33: This, as well as `valid_diets` and `valid_metabolic_types` is hardcoded here, but would be better to set them as class constants, so they are easy to spot and to update if needed. Also, making them tuples, not list, as they will be inmutable. 

If they are here because this is a work in progress, that's fine, but then indicate somewhere that this will be moved, so you don't forget.

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -71,12 +93,19 @@ def import_functional_groups(fg_file: str) -> list[FunctionalGroup]:
...
     functional_group_list: list[FunctionalGroup] = []
 
-    with open(fg_file, newline="") as csv_file:
-        reader = csv.reader(csv_file)
-        next(reader, None)  # skip the header
-        # unpack the row directly in the head of the for-loop
-        for name, taxa, diet in reader:
-            # create the FG instance and append it to the list
-            functional_group_list.append(FunctionalGroup(name, taxa, diet))
+    import pandas as pd
```
dalonsoa, line 96: Import the package a the top of the module.

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -71,12 +93,19 @@ def import_functional_groups(fg_file: str) -> list[FunctionalGroup]:
...
+
+    fg = pd.read_csv(fg_csv_file)
+
+    expected_header = ["name", "taxa", "diet", "metabolic_type"]
```
dalonsoa, line 100: Indicate in the docstring what are the required headers in the file.

dalonsoa (approved):
Looks good! I've just made a couple of comments/suggestions. 

I'm wondering how many taxa, diet and metabolic types you are expected to have. For taxa,  and maybe diet, that is probably really large - and growing - but for metabolic type, I cannot imaging having more than the two options you already have. If that the case, I would consider making that one [an `Enum`](https://docs.python.org/3/library/enum.html) that only allow those two possibilities. That will help with the validation, typing, as `mypy` will know that only those 2 options are possible, and save you - and everyone else from errors related to typos.

So, if you define somewhere at the top of your module:

```python
from enum import Enum

class MetabolicType(Enum):
    ENDOTHERMIC = "endothermic"
    ECTOTHERMIC = "ectothermic"
```
then in the constructor of `FunctionalGroup` you can do:
```python
self.metabolic_type = MetabolicType(metabolic_type)
```
which will automatically validate the value of `metabolic_type` (you can enclose it in a `try...except...` to provide a more meaningful error message).

When you do this, then in functions like `metabolic_rate` in `scaling_functions.py`, you can annotate `metabolic_type: MetabolicType` and you don't need to handle the case of `metabolic_type` not having a valid value because that will be impossible, by design. It will be something like:

```python
    match metabolic_type:
        case MetabolicType.ENDOTHERMIC:
             return terms[1] * mass_g ** terms[0]
        case MetabolicType.ECTOTHERMIC:
             b0, exponent = terms
             return (
                 b0 * mass_g**exponent * exp(-0.65 / (BOLTZMANN_CONSTANT * temperature_k))
             )
```
The [`match...case` control structure ](https://datagy.io/python-switch-case/)is something that was introduced in Python 3.10 and that I feel we should use more, as it makes the code more expressive under some conditions than `if` statements. 

This is a suggestion for you to discuss, not something to change now, but I think `Enum`s are a good option when you have a finite set of options to choose from.

virtual_rainforest/models/animals/functional_group.py
TaranRallings: I think diet and taxa are actually going to be short enough lists that enumerate would make sense for them as well. I went ahead and built the basics of this out but I'm curious about the match/case structure, would that be a function that replaces the dictionaries in constants.py?

Something like:
Replacing:
`ENDOTHERMIC_METABOLIC_RATE_TERMS: dict[str, tuple[float, float]] = {
    "mammal": (0.75, 0.018),
    # Mammalian metabolic rate scaling from Metabolic Ecology p210. [assumes g mass]
    "bird": (0.75, 0.05),
    # Toy values.
}`

With:
`def calculate_metabolic_rate(taxa_type: TaxaType, mass: float) -> float:
    return match taxa_type:
        case TaxaType.MAMMAL:
            return 0.75 * mass ** 0.018
        case TaxaType.BIRD:
            return 0.75 * mass ** 0.05
        case _:
            raise ValueError(f"Invalid taxa type: {taxa_type}")`

TaranRallings: I put together an implementation of what I think you're recommending. Let me know if I understood correctly.

dalonsoa: Using the dictionary with the enum as keys makes more sense.

TaranRallings: I got the docs issue sorted out. Can you let me know if the changes make sense?

virtual_rainforest/models/animals/animal_traits.py
```diff
@@ -0,0 +1,82 @@
...
+        Returns:
+            The corresponding `MetabolicType` enumeration member.
+
+        Raises:
+            ValueError: If the provided string does not match any valid metabolic type.
+        """
+        try:
+            return cls[value.upper()]
+        except KeyError:
+            raise ValueError(f"Invalid metabolic type: {value}")
```
dalonsoa, line 31: Mmmmm... I sorry for this. I've just noticed that the `StrEnum` I used in my example is only available from Python 3.11. Being that the case, you can explicitly set the values instead of using `auto`:

```suggestion
class MetabolicType(Enum):
    """Enumeration for metabolic types."""

    ENDOTHERMIC = "endothermic"
    ECTOTHERMIC = "ectothermic"
```

This should work fine when calling `MetabolicType("endothermic")`.

virtual_rainforest/models/animals/animal_traits.py
```diff
@@ -0,0 +1,82 @@
...
+        Returns:
+            The corresponding `DietType` enumeration member.
+
+        Raises:
+            ValueError: If the provided string does not match any valid diet type.
+        """
+        try:
+            return cls[value.upper()]
+        except KeyError:
+            raise ValueError(f"Invalid diet type: {value}")
```
dalonsoa, line 56:
```suggestion
    HERBIVORE = "herbivore"
    CARNIVORE = "carnivore"
```

virtual_rainforest/models/animals/animal_traits.py
```diff
@@ -0,0 +1,82 @@
...
+        Returns:
+            The corresponding `TaxaType` enumeration member.
+
+        Raises:
+            ValueError: If the provided string does not match any valid taxa type.
+        """
+        try:
+            return cls[value.upper()]
+        except KeyError:
+            raise ValueError(f"Invalid taxa type: {value}")
```
dalonsoa, line 82:
```suggestion
    MAMMAL = "mammal"
    BIRD = "bird"
    INSECT = "insect"
```

virtual_rainforest/models/animals/functional_group.py
```diff
@@ -28,36 +33,40 @@ class FunctionalGroup:
...
-        self.taxa = taxa
-        """The taxa of the functional group ("mammal" or "bird")."""
-        self.diet = diet
-        """The diet of the functional group ("herbivore" or "carnivore")."""
-        self.metabolic_rate_terms = ENDOTHERM_METABOLIC_RATE_TERMS[taxa]
+        self.taxa = TaxaType.from_str(taxa)
+        """The taxa of the functional group."""
+        self.diet = DietType.from_str(diet)
+        """The diet of the functional group."""
+        self.metabolic_type = MetabolicType.from_str(metabolic_type)
```
dalonsoa, line 45:
```suggestion
        self.taxa = TaxaType(taxa)
        """The taxa of the functional group."""
        self.diet = DietType(diet)
        """The diet of the functional group."""
        self.metabolic_type = MetabolicType(metabolic_type)
```

virtual_rainforest/models/animals/scaling_functions.py
```diff
@@ -31,23 +34,33 @@ def damuths_law(mass: float, terms: tuple) -> int:
...
+
+    if metabolic_type == MetabolicType.ENDOTHERMIC:
+        return terms[1] * mass_g ** terms[0]
+    elif metabolic_type == MetabolicType.ECTOTHERMIC:
+        b0, exponent = terms
+        return (
+            b0 * mass_g**exponent * exp(-0.65 / (BOLTZMANN_CONSTANT * temperature_k))
+        )
+    else:
+        raise ValueError("Invalid metabolic type: {metabolic_type}")
```
dalonsoa, line 63: Any reason why not using `match... case`? It was only a suggestion, but I'm wondering if it is just a choice or if you were having any technical problem with it. In principle, `mypy>=1.3` with Python 3.10 should work well with this.

```suggestion
    match metabolic_type:
        case MetabolicType.ENDOTHERMIC:
             return terms[1] * mass_g ** terms[0]
        case MetabolicType.ECTOTHERMIC:
             b0, exponent = terms
             return (
                 b0 * mass_g**exponent * exp(-0.65 / (BOLTZMANN_CONSTANT * temperature_k))
             )
```

dalonsoa (approved):
It looks really great. I think it is now more explicit and less prone to typos. I've made a couple of suggestions to make this work nicely with Python 3.10. My original suggestions where not available.

virtual_rainforest/models/animals/scaling_functions.py
TaranRallings: Oh, I tried it at first and then mypy got very angry about it so I shelved it for "later". The browsing I did online looked like it was a known problem but I may have misunderstood.

TaranRallings: It works now so it must have been something else I was doing! I'll add it in.

TaranRallings: I need to stop posting like things have been resolved before actually doing the push. 

Locally, flake8 and mypy accept the match/case structure. They get angry about it on github.

jacobcook1995: Is that because the GitHub CI action uses python 3.9 for the CI build step?

davidorme: Yes. That is why 😄

We _could_ abandon support for `3.9` - there is then some other stuff that we could adopt, like the use of `None | str` in typing and the like. But I think we should do that in a systematic way as a Project to update the code base to 3.10. I'd stick with what works with 3.9 here?

dalonsoa: Ups. I didn't realised that we were officially supporting 3.9. Sorry for that. Well, 3.10 and 3.11 come with a lot of improvements and speedups - and this Autumn we will have 3.12. So it will not be a bad move to abandone something that was already "old" (1.5 years old) when the project started.

But I totally agree with @davidorme , I would stick with 3.9 compatible stuff for now and make the transition in an organised way.

TaranRallings: Alright, thanks folks!

robewers01:
Hi Taran,
I saw earlier in this thread that Diego had mentioned setting up universal constants in core. Same issue applies to the scaling factor (-0.65) that’s in this equation – we want to avoid hard-coding that.

Regards,
Rob

PR #236, opened 2023-06-07
Various bug fixes to allow `vr_run` to work with `soil` and `abiotic_simple` models

### LINKED ISSUE #235: Move number of soil and canopy layers to core config
At the moment the number of soil and canopy layers is part of the config for the `abiotic_simple` model. @vgro and I don't think this really makes sense.

Number of soil layers and canopy layers should be stored in the core config rather than the abiotic model config as they define a common physical environment rather than something specific to the abiotic model.

N/A

jacobcook1995:
# Description

This pull request is mainly small bug fixes to make the `vr_run` function runable for the `abiotic_simple` and `soil` models (using data stored on RDS).

The most substantive change is moving the definition of layer roles from the `abiotic_simple` model into the core utilities. This was necessary as the `soil` model also needed to make use of these layers roles (to find the index associated with the top soil layer).

This PR also updates the documentation to explain to users how to access the data (if they are part of the VR project), and updates the simulation flow documentation to reflect recent changes to the flow.

Fixes #235

## Type of change

## Key checklist

## Further checks


jacobcook1995:
Adding @vgro as a reviewer as I'm not sure of best practice when a pull request is cowritten

vgro (approved):
LGTM

davidorme (approved):
LGTM - minor comment on using admonitions.

I think we should structure the main simulation document a bit more (and introduce all of the defined steps, even if they are currently lightly used). I was going to make suggestions, but instead I'm going to commit a suggested update. I've updated the Sphinx conf to include a [mermaid](https://mermaid.js.org/) diagram. I think that is useful here and it is probably something we'll want to use more of.

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -1,80 +1,158 @@
...
-## Loading and validation of input data
+### Model configuration
 
-This `Data` object now needs to be populated with data. The data used in the simulation
-is stored in a set of data files which are specified in the configuration. This data is
-then loaded using the `data.load_data_config` method, which has built in validation
-procedures to ensure that the loaded data can be mapped onto the spatial grid and also
-other core dimensions for the configured simulation. Further details can be found in the
-[data system](./core/data.md) and [core axes](./core/axes.md) documentation.
+The core configuration will include the configuration details for each individual
```
jacobcook1995, line 109: These details won't be in the core configuration

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -1,80 +1,158 @@
...
 
-As a first step, configuration files are loaded in and validated. These files are of
-`toml` format and are found based on paths provided by the user. When these files are
-loaded in they are validated to ensure that they are valid `toml`, and that all the
-required settings are present and not duplicated. The Virtual Rainforest code provides
-default settings for some configuration options and these will be used automatically if
-that setting is not found in the configuration, but if no default is set then the
-validation will fail. Further details can be found in the [configuration
-documentation](./core/config.md).
+* [`AbioticModel`](../api/abiotic.md),
```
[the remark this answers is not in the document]
davidorme: Good point - I've included both links since I think they will both be here to stay.

docs/source/virtual_rainforest/main_simulation.md
davidorme: Yeah - clumsily phrased.

vgro:
> LGTM - minor comment on using admonitions.
> 
> I think we should structure the main simulation document a bit more (and introduce all of the defined steps, even if they are currently lightly used). I was going to make suggestions, but instead I'm going to commit a suggested update. I've updated the Sphinx conf to include a [mermaid](https://mermaid.js.org/) diagram. I think that is useful here and it is probably something we'll want to use more of.

I like the idea of using the diagrams, however, they don't come out very nicely at the moment (e.g. the lines go over the text).

davidorme:
Yeah - it definitely isn't as I'd like and I am normally really fussy about figures. That is mostly the tool: `mermaid` doesn't really do fine controls -  or at least not without lots of tweaking of things. It does produce diagrams quickly though! We could replace this with a hand crafted image that we're happier with when things are stable, but this is quick to edit and update for now?

Issue #208, opened 2023-04-24
Clarify roles and structure of Grid, Core and Data objects and use in models and axis validation.


davidorme:
**Is your feature request related to a problem? Please describe.**

The current implementation of the core axis and axis validation passes in the `Grid` object and then `kwargs`. This was only ever really a placeholder for giving the axis validation setup access to more information from the configuration object to the axis definitions. More broadly, the Grid object is rather arbitrarily stuffed into Data and is passed around hidden in that. 

The aim of this PR is therefore have consistent use of the three high level objects (`Data`, `Config`, `Grid`) in signatures. The Axis validators can then accept (`Config`, `Grid`) arguments and models can expect to use (`Data`, `Config`, `Grid`). We _could_ shove `Grid` into `Config` - it is a much more logical home than `Data` - but from a quick chat we think having three objects with clear roles is easier to understand.

**Describe the solution you'd like**

We want the whole of the configuration and the grid to be passed in to axis validators. That _should_ give access to all of the information required to configure the model and hence everything that an axis validator could want. We also need to update model code to accept `Data`, `Config` and `Grid` objects.

davidorme:
@jacobcook1995 @TaranRallings @vgro - Does that seem reasonable as a restructure?

hrlai:
Sorry if this has been mentioned somewhere / posted in the wrong place, but how do we change the run_length of ve_run? I tried the following config
```
[core]
[core.timing]
run_length = "10 years"

[hydrology]
[abiotic_simple]
[animal]
[litter]
[soil]
```

But it still runs to step 24 (the default) and then gave the following error:
```
IndexError: index 24 is out of bounds for axis 0 with size 24
```

davidorme:
That is related. Because we don't yet validate the time axis on inputs the model hasn't said on startup ' hey! You've asked for a 120 month run but this variable only provides 24 values!" 

Which it needs to..

hrlai:
Ah I see... so is there a place where I should supply/define the 120 values?

hrlai:
Ah ok, I am guessing that `time` in `generation_scripts/common.py` needs to match the run length. Is there an internal script that runs these generation scripts and output them to the `data` folder?

Edit: I'll regenerate the data by hand for now (i.e., running the python scripts one by one).

PR #237, opened 2023-06-08
Continuous output of the `data` object

### LINKED ISSUE #231: Basic output of continuous time data
At the moment the main simulation can only output the initial and final state of the data object (whether it outputs either of these  is controlled by config options). It is becoming increasingly important to have continuous time output so that we can check whether models are behaving as expected.

At every time step the contents of the data object can be appended to a netcdf file. We should be able to turn this behaviour off using a config option. We also need to ensure that input data with a time axis (basically the climate data inputs) are not added, as this would result in a 

The above suggestion is basically the simplest way to get analysable output. It will however slow the simulation process (because frequent write operations are costly), and will result in unnecessarily large files (as we are unlikely to care about every single variable). However, addressing these issues adds complications, so I will instead add these as separate issues once the basic system is setup.

**Additional context**
N/A

jacobcook1995:
# Description

This pull request adds functionality for output of the state of the `data` object after every `update` step. This output contains every variable that does not have a `time_index` dimension defined (to prevent repeated output of climate input data). At the first step this output is used to create a new output file, and then every subsequent step appends to this file. This is done as we want a singular file to make plotting the simulation trajectory easier. Whether or not to save continuous output and the location to save them to are both core configuration options. 

One issue I have is that the current implementation does not feel very efficient. I suspect this is partly due to writing out data at every time step just not being a very efficient thing to do generally, but if you spot anything that could be made less intensive let me know.

Another thing that probably needs some thought is the fact that the main simulation loop now can save 3 different things, the initial state, the final state, and the continuous data output. There's overlap between what each of these contain, with both the initial state and final state containing the input climate data, and the continuous data output containing the initial and final states (excluding the climate input data). It's not a massive issue, but there might be someway to streamline this.

Fixes #231

## Type of change

## Key checklist

## Further checks


jacobcook1995:
> I also think we should add another `BaseModel` property that contains a list of variables that are updated by that model. It then becomes much easier for the continuous and final data saves to only extract the variables included in that list?

Yeah that seems a good solution to me in that I was concerned about how the system would work if some variables had `time_index` dimensions.

I could hard code these variables in to begin with, and then the `toml` file @vgro added in #201 could be use to automatically populate the attribute in a future PR (to stop this one getting unwieldy).

jacobcook1995:
> On the efficiency question, I wonder if it would be better for now if the `update` data capture simply writes out a single file for the time index rather than a continually growing blob of data that has to be loaded, `xr.concat`d and then saved. One really handy thing then is that [`xarray.open_mfdataset`](https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html) provides a really seamless way to open all of those files as if they were one file. We could provide a simple `compile_outputs` tool to wrap this, but for the proof of concept that sounds faster?

If it's easy to read in a large number of NetCDF files as then that definitely seems like a potential solution. My only hesitation is the number of files this would generate (a 50 year run would generate 600 files), which might be hassle to deal with. But if there isn't an efficient to combine them into a single file then I guess we will just have to deal with that for now

davidorme:
On the list of variables, if each model has `required_vars` and then a separate list of `updated_vars`, then that is programatically accessible from each model instance and could be used to autogenerate the manually created list from #201.

vgro:
> On the efficiency question, I wonder if it would be better for now if the `update` data capture simply writes out a single file for the time index rather than a continually growing blob of data that has to be loaded, `xr.concat`d and then saved. One really handy thing then is that [`xarray.open_mfdataset`](https://docs.xarray.dev/en/stable/generated/xarray.open_mfdataset.html) provides a really seamless way to open all of those files as if they were one file. We could provide a simple `compile_outputs` tool to wrap this, but for the proof of concept that sounds faster?
> 
> I also think we should add another `BaseModel` property that contains a list of variables that are updated by that model. It then becomes much easier for the continuous and final data saves to only extract the variables included in that list?
> 
> Suggestions, obviously - but that seems sleeker?

Which `update` would do that? Do you mean in each individual model? Does that not mean that there are intermediate stages where some of the data is updated and others not? For example, the abiotic model updates temperature and writes it in a file which has everything else from the previous time step, or would everything get a time index?

davidorme:
No - badly phrased. I meant the stage in the workflow after all models have run `update` - that then writes out a single file containing all of the required variables, with a time index, which `xarray.open_mfdataset` should then be able to use in gluing them back together again.

virtual_rainforest/core/data.py
```diff
@@ -334,19 +335,92 @@ def load_data_config(self, data_config: dict[str, Any]) -> None:
...
-        # If the file path is okay then write the model state out as a NetCDF
-        self.data.to_netcdf(output_file_path)
+        # If the file path is okay then write the model state out as a NetCDF. Should
+        # check if all variables should be saved or just the requested ones.
+        if variables_to_save:
+            self.data[variables_to_save].to_netcdf(output_file_path)
+        else:
+            self.data.to_netcdf(output_file_path)
+
+    def append_to_netcdf(
```
dalonsoa, line 363: I think this comment goes along the lines already discussed, but having to load the existing dataset, appending new data and saving it again is probably not that efficient but, specially, rather cumbersome.

I'd simplify things as much as possible and save each step in its own file and then, at the very end of the run, compile the data in a single file or in chunks. The fewer IO operations you have during the execution of the calculation, the better.

dalonsoa (approved):
I think the overall approach looks good and it is a much needed step the one of saving data, but I'd keep that as simple as possible and sabe individual files all the time, given that appending data is quite convoluted.

jacobcook1995:
I've changed the system to output a new file at every step, with a final step after the simulation is finished to collect them all into a single file. The variables to be written out are currently hardcoded into each model, so we will be able to manipulate this as needed during development. (I'll make an issue about making this configurable down the line).

I delete the single step files once the complete file is saved, in order to reduce the file bloat a bit. Sanity checking on this section would be appreciated as I'm always a bit nervous about programatic deletion of files.

@vgro I made changes to your new `HydrologyModel` class to make it compatible with the data output system, if you could check that I haven't done anything stupid with it that would be great!

vgro (approved):
LGTM, good job! I didn't spot any problems and I think the models I wrote should still work with the update.
One suggestion, maybe for the future, would be to give the output file a more specific name, even if it is just an index, to avoid overwriting it. As soon as we start varying anything in the model we will likely want to save more than one simulation output. Or were you thinking of different output folders? Anyways, can be a separate PR :-)

jacobcook1995:
> LGTM, good job! I didn't spot any problems and I think the models I wrote should still work with the update.
> One suggestion, maybe for the future, would be to give the output file a more specific name, even if it is just an index, to avoid overwriting it. As soon as we start varying anything in the model we will likely want to save more than one simulation output. Or were you thinking of different output folders? Anyways, can be a separate PR :-)

That's a good point! I tried to avoid configurable file names for the temporary files as there's potential there for people to do something stupid. There isn't the same risk with the final output file so I think that could be a configuration option. When I get time post holiday I'll make that an option

virtual_rainforest/main.py
```diff
@@ -171,6 +173,85 @@ def extract_timing_details(
...
+    # Find all relevant files in the given folder and convert to list
+    continuous_files = out_folder.rglob("continuous_state*.nc")
+
+    # Open all files as a single dataset
+    all_data = open_mfdataset(list(continuous_files))
+
+    # Check that output file doesn't already exist
+    out_path = Path(f"{out_folder}/all_continuous_data.nc")
+    try:
+        check_outfile(out_path)
```
dalonsoa, line 241: Maybe this check should be done before collecting all files, otherwise among the collected files you will have one containing all collected data. That will save you from having to close `all_data` as it won't be open, to start with.

virtual_rainforest/main.py
```diff
@@ -171,6 +173,85 @@ def extract_timing_details(
...
+        # Close dataset to prevent access problems on windows
+        all_data.close()
+        raise e
+    # Save and close complete dataset
+    all_data.to_netcdf(out_path)
+    all_data.close()
+
+    # Iterate over all continuous files and delete them
+    for file_path in list(out_folder.rglob("continuous_state*.nc")):
+        file_path.unlink()
```
dalonsoa, line 252: It seems OK, so I don't think you need to worry about this :)

virtual_rainforest/main.py
```diff
@@ -171,6 +173,85 @@ def extract_timing_details(
...
+
+    # Only variables in the data object that are updated by a model should be outputted
+    all_variables = [
+        models_cfd[model_nm].vars_updated for model_nm in models_cfd.keys()
+    ]
+    # Then flatten the list
+    variables_to_save = [item for sublist in all_variables for item in sublist]
+    # Create output file path for specific time index
+    out_path_name = (
+        f"{data_options['out_folder_continuous']}/continuous_state{time_index}.nc"
```
dalonsoa, line 208: I would pad the `time_index` with zeros, so the files are ordered. I know they are going to be deleted, but it is a question of principles :) . For example, to write the index in a field with 5 digits, you would do:

```suggestion
        f"{data_options['out_folder_continuous']}/continuous_state{time_index:05}.nc"
```
Resulting in filenames:
```output
continuous_state00001.nc
continuous_state00002.nc
continuous_state00003.nc
...
continuous_state00345.nc
```

dalonsoa (approved):
I've made a couple of small comments, but it looks good. And specially, it looks safe enough.

docs/source/virtual_rainforest/main_simulation.md
```diff
@@ -150,10 +152,29 @@ used to output a single data file of the initial simulation state.
...
 
 ### Saving the final state
 
 After the full simulation loop has been completed, the final simulation state held in
 the `Data` object can be optionally be saved to a path provided in the configuration,
 defaulting to saving the data.
+
+### Combining continuous data
+
+If the model has been set up to output continuos time data, then there is a final step
```
davidorme, line 166:
```suggestion
If the model has been set up to output continuous time data, then there is a final step
```

virtual_rainforest/main.py
```diff
@@ -171,6 +173,85 @@ def extract_timing_details(
...
+    )
+
+    # Save the required variables by appending to existing file
+    data.save_timeslice_to_netcdf(Path(out_path_name), variables_to_save, time_index)
+
+
+def merge_continuous_data_files(data_options: dict[str, Any]) -> None:
+    """Merge all continuous data files in a folder into a single file.
+
+    It is important to note that this function deletes all of the original continuos
```
davidorme, line 218:
```suggestion
 This function deletes all of the original continuous
```

virtual_rainforest/main.py
```diff
@@ -227,12 +308,26 @@ def vr_run(
...
+            models_cfd[mod_nm].update(time_index)
+
+        # With updates complete increment the time_index
+        time_index += 1
+
+        # Append updated data to the continuous data file
+        if config["core"]["data_output_options"]["save_continuous_data"]:
+            output_current_state(
+                data, models_cfd, config["core"]["data_output_options"], time_index
+            )
```
davidorme, line 326: I think this should return the Path of the created file and store that path in a list, which can then be used as an input to `merge_continuous_data_files`. So before the timing loop starts: `continuous_data_files = []` and then:

```python
continuous_data_files.append(
    output_current_state(
                data, models_cfd, config["core"]["data_output_options"], time_index
            )
)
```
This way, the merge operation is fed an explicit list of files to handle. I would also be tempted to separate the delete operation from the merge operation.

davidorme (changes requested):
This looks good to me apart from the "delete all files called XXX" part. I think we should internally track the output file paths and explicitly pass those to the merge operation. That guarantees that we don't delete files we haven't written (and we should set it up so it _will not_ overwrite existing files) and has the other advantage that we can vary the path internally to manage directory sizes (like writing annual directories or something) and just have a list of files to handle, not a directory search.

In the long run, both the `output` and `merge` need to be methods of `Data` rather than floating around in `main` but that is waiting for some core refactors to hand config around (#208). Then we could just do `Data.output_state_at_time(t=1)` and then `Data.merge_saved_states('myfile.nc')` and `Data.delete_saved_states()`

jacobcook1995:
Yep I agree that deleting based on a list is safer (and more extensible) so I have made that change!

PR #245, opened 2023-07-04
Forced module order to be fixed based on config input

jacobcook1995:
# Description

This pull request changes `select_models` to use a list rather than a set to find the non-duplicated model names. This means that the inputted order of models is preserved, so that models are always called in the order the user specifies them.

This doesn't fix the broader issue of how we should handle model ordering, where the information should be stored, etc. It does fix the issue of `vr_run` sometimes randomly stalling when a valid model order is entered.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -5,7 +5,7 @@ defined in main.py that it calls.
 """
 
 from contextlib import nullcontext as does_not_raise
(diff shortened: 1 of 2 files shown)


vgro (approved):
LGTM

PR #246, opened 2023-07-04
Remove soil model specific function to update data object

jacobcook1995:
# Description

The `soil` model had defined a function to update the data object. However, @vgro recently added a function to the `Data` object to do this. So, it makes sense to remove the `soil` specific function in favour of a function defined within `core`.

I also changed the `add_from_dict` slightly so that the relevant `LOGGER.info` output gets generated when data is saved using it.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -123,7 +123,7 @@ file.
 """  # noqa: D205, D415
 
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import Any, Optional
 
 import numpy as np
 from xarray import DataArray, Dataset
@@ -348,7 +348,7 @@ class Data:
         # If the file path is okay then write the model state out as a NetCDF
         self.data.to_netcdf(output_file_path)
 
-    def add_from_dict(self, output_dict: Dict[str, DataArray]) -> None:
+    def add_from_dict(self, output_dict: dict[str, DataArray]) -> None:
         """Update data object from dictionary of variables.
 
         This function takes a dictionary of updated variables to replace the
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -364,7 +364,7 @@ class Data:
         """
 
         for variable in output_dict:
-            self.data[variable] = output_dict[variable]
+            self[variable] = output_dict[variable]
 
 
 class DataGenerator:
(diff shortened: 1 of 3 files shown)


virtual_rainforest/core/data.py
```diff
@@ -348,7 +348,7 @@ def save_to_netcdf(self, output_file_path: Path) -> None:
         # If the file path is okay then write the model state out as a NetCDF
         self.data.to_netcdf(output_file_path)
 
-    def add_from_dict(self, output_dict: Dict[str, DataArray]) -> None:
+    def add_from_dict(self, output_dict: dict[str, DataArray]) -> None:
```
vgro, line 351: What is the difference between `dict` and `Dict `(just out of curiosity)?

virtual_rainforest/core/data.py
```diff
@@ -364,7 +364,7 @@ def add_from_dict(self, output_dict: Dict[str, DataArray]) -> None:
         """
 
         for variable in output_dict:
-            self.data[variable] = output_dict[variable]
+            self[variable] = output_dict[variable]
```
vgro, line 367: Why is the variable assigned to the `model` rather than `model.data `?

vgro (approved):
LGTM, a few minor questions.

virtual_rainforest/core/data.py
jacobcook1995: Their behaviour is identical. `Dict` was originally defined in `typing` as a type hint for a dictionary, but python 3.9 was changed to allow collections (e.g. `dict`) to be used to directly give type hints. So older versions of python have to import `Dict` from `typing` to type hint, but in more recent versions it's considered to be deprecated as you can just use `dict` instead without importing

virtual_rainforest/core/data.py
jacobcook1995: `self.data` selects the dictionary called `data` stored within the `Data` object, which can be updated directly.

`self` selects the `Data` object, which updates via its `__setitem__` method. This is a method that David wrote that does some sanity checking of the updated data (i.e. saving only `DataArray`), and gives `LOGGER` output for whether a variable is being updated or added.

This method is the main source of the logger output that floods the screen when you run `vr_run`, so it might need to be altered in future, but it's probably sensible to ensure that every update passes through this function for consistencies sake.

PR #248, opened 2023-07-04
Feature/animal module - Foraging functions

TaranRallings:
# Description

This is a first draft of the foraging system. In the Madingley model, all of the searching and sorting of animal foraging was their largest computational bottleneck and I am expecting something similar for us. I am very open to dramatically rewriting this entire system. I want this part to work as efficiently as possible. 

The draft assumptions are that each animal searches one grid square and consumes just one type of resource per time-step and that community foraging is unordered. Multi-grid search will be added after we've sorted multi-grid occupancy. The single-resource assumption will be fixed soon after we decide on the overall foraging structure. The ordering of foragers is a bigger discussion and the choice had significant outcomes for Madingley. 

I have broken the cohort.eat() function into separate functions for herbivory and predation. I expect these to be supplemented soon with scavenging and possibly soil consumption. 
I have added a cohort.cohort_forage() function that can be called to handle what kind of foraging should take place.
I have added a community.community_forage() function that loops through all animals in the communities and calls cohort_forage() for each of them.

Questions I'd like to get sorted:
- Does the distribution of these functions between AnimalCohort and AnimalCommunity make sense? 
       - And/or how could they be improved?
- Does the separation of different foraging types (predation/herbivory/etc) into different functions make sense?
- There is a lot of list manipulation and related shenanigans going on in here, how can this be tightened up?

I have kept the testing minimal as I'm expecting to demolish this thing and rebuild. I will write out a full testing suite once we're settled on the structure and distribution of functions.

**EDIT**

I have reworked the foraging system. The new form includes a few major changes:

- `forage_community()` was broken into `forage_community()` + `collect_prey()`
- `herbivory()` + `predation()` were collapsed into a single `eat()` method
- a `get_eaten()` method was added to `AnimalCohort` and `PlantCommunity`
- I have implemented Protocols to group classes by status as `Consumer`, `Resource`, or `Pool`
- a host of smaller changes to naming conventions and function logic

I tried to work in the various suggestions to improving the code but the broader rework has made it difficult to tell how successful this was in some cases. Testing is still in shambles and will be dialled in once we've settled on the structure. 

I think the code is much better now. It's easier to see what is happening and I think between `get_eaten()` and Protocols it will be easier to build on the code going forward.

Questions!
- what do we think about using Protocols? Is is acceptable? Does this need a system of supplementary checks to make sure nothing is awry?
- is the looping and grouping logic in `forage_community()` and `collect_prey()` sane?
- would any of the methods be better decomposed into smaller methods?
- have I completely overlooked one of your previous suggestions?

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
With the massive caveat that my knowledge of how to efficiently structure animal models is basically non-existent, LGTM!

Made a couple of comments about minor things.

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -180,4 +226,45 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+
         carcass_pool.energy += number_dead * self.mass * ENERGY_DENSITY["meat"]
+
+    def forage_cohort(
+        self,
+        plant_list: list[PlantCommunity],
+        animal_list: list[AnimalCohort],
+        carcass_pool: CarcassPool,
+        soil_pool: PalatableSoil,
+    ) -> float:
```
davidorme, line 238: I would try and turn this into a registry approach (probably not right now, but in the longer term). That would need a generic signature of all possible food sources and then each registered diet method would have body code that targeted the diet specific sources. So, you might have a registry dictionary like:

```python
{
    'herbivore': herbivory(self, plants, animals, carcasses, soil,...),
    'carnivore': carnivory(self, plants, animals, carcasses, soil,...), 
}
```

That then is easier for us or other users to extend with different diets. In addition, you can configure the foraging function when the cohort is created in `init`.

```python
self.forage = DIET_REGISTRY[self.functional_group.diet.value]
```

And then `self.forage` should automatically target the right diet by calling the appropriate function from under that `forage` label.

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -93,3 +99,31 @@ def birth(self, cohort: AnimalCohort) -> AnimalCohort:
...
+        carcass_pool = self.carcass_pool
+        soil_pool = self.soil_pool
+
+        animal_list = []  # List to collect all animals
+
+        for functional_group, cohorts in self.cohorts.items():
+            animal_list.extend(cohorts)  # Add cohorts from the current functional group
+
+        for cohort in animal_list:
+            other_animals = [animal for animal in animal_list if animal != cohort]
```
davidorme, line 122: You're iterating over a list, so you just use indexing here to omit the target cohort rather than a more expensive list comprehension and comparison:
```suggestion
        for idx, cohort in enumerate(animal_list):
            other_animals = animal_list[0:idx] + animal_list[idx+1:]
```

virtual_rainforest/models/animals/animal_cohorts.py
davidorme: That's basically just replacing `self.forage_cohort` with a direct link to the right function rather than having to maintain a lookup system within `self.forage_cohort`.

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -93,3 +99,31 @@ def birth(self, cohort: AnimalCohort) -> AnimalCohort:
...
+
+        """
+        plant_list = [self.plant_community]
+        carcass_pool = self.carcass_pool
+        soil_pool = self.soil_pool
+
+        animal_list = []  # List to collect all animals
+
+        for functional_group, cohorts in self.cohorts.items():
+            animal_list.extend(cohorts)  # Add cohorts from the current functional group
```
davidorme, line 119: If you don't need the cohort names, you can do:
```suggestion
        for cohorts in self.cohorts.values():
            animal_list.extend(cohorts)  # Add cohorts from the current functional group
```

davidorme: Or even:

```python
animal_list = [each_cohort for fg_cohorts in self.cohorts.values() for each_cohort in fg_cohorts]
```

I'm not sure a list comprehension is any faster - avoids having to declare `animal_list` first.

virtual_rainforest/models/animals/animal_communities.py
davidorme: Having said that doesn't this prevent within-cohort cannibalism?

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -105,55 +108,98 @@ def metabolize(self, dt: timedelta64) -> None:
...
-        Currently, this wraps the acts of consuming plants and excreting wastes. It will
-        later wrap additional functions for selecting a food choice and navigating
-        predation interactions.
+        # Increase predator's stored energy with the energy gained from eating
+        self.stored_energy += (
+            consumed_energy * self.functional_group.conversion_efficiency
+        )
+        prey.stored_energy -= consumed_energy
+
+        return consumed_energy * self.functional_group.conversion_efficiency
```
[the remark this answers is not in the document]
davidorme: The same is true for `birth` - appends to the cohort and returns the object?

davidorme (commented):
The main thing here for me is that I'm not sure that collapsing your `community.cohorts` structure into a flat list in `forage_cohorts` is helpful. I can see the intention is to remove the foraging cohort. The first issue is that (rarely!) things in a cohort might eat each other. 

The more important thing is that at some point soon you are going to want the foraging functions to filter the available cohorts down to what that cohort actually eats. That is going to be a pain to set up but I can imagine the first slices being functional group and then mass. If you pass the `community.cohorts` over _as is_ then it is a great deal easier to apply filters. Given an `AnimalCohort.prey_groups` attribute like `{'small_mammals': (12.0, 150), ...}` that sets possible targets and sizes handled,  the `AnimalCohort.carnivory` method could internally do something like:

```python
# filter targets
prey = []
for func_gp, cohorts in animal_cohorts:
    # Exclude non target cohorts
    if func_gp not in self.prey_groups:
        continue

    # Reduce to cohorts within handled size range
    min_size, max_size = self.prey_groups[func_gp]
    right_sized = [c for c in cohorts  if (c.mass >= min_size) or (c.mass <= max_size)]
    prey.extend[right_sized]
```

With the flattened list, you lose a lot of useful information.

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -105,55 +108,98 @@ def metabolize(self, dt: timedelta64) -> None:
...
-            A float containing the amount of energy consumed in the foraging bout.
+            A float containing the amount of energy consumed in the herbivory bout.
         """
-        consumed_energy = min(food.energy, self.intake_rate * food.energy_density)
+        consumed_energy = (
+            min(
+                plant_food.energy,
+                self.intake_rate * plant_food.energy_density * self.individuals,
+            )
+            * self.individuals
```
dalonsoa, line 131: Aren't you considering self.individuals twice if the winner option is the second one?

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -105,55 +108,98 @@ def metabolize(self, dt: timedelta64) -> None:
...
+                * self.functional_group.mechanical_efficiency
+            ),
+            prey.stored_energy,
+        )
 
-    def forage(self, food: PlantCommunity, soil: PalatableSoil) -> None:
-        """The function to enact multi-step foraging behaviors.
+        # Reduce the number of individuals in the prey cohort and add energy to predator
+        prey.individuals -= number_eaten
+        carcass_pool.energy += consumed_energy * (
```
dalonsoa, line 180: If the explanation is about adding energy to predator, why is it added to the carcass pool?

dalonsoa (commented):
Sorry, I haven't finished yet, but of of leaving a couple of comments.

virtual_rainforest/models/animals/animal_communities.py
TaranRallings: > Having said that doesn't this prevent within-cohort cannibalism?

It was intended to prevent within-cohort cannibalism but not within-fg cannibalism. The situation where you are predating on an individual that is both your "species" and identical body-mass would be extremely rare. Maybe I should build something in where this occurs when there are no other remaining animal cohorts.

virtual_rainforest/models/animals/animal_communities.py
dalonsoa: It is faster, actually. https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Loops

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -41,6 +44,9 @@ def __init__(
             k.name: [] for k in self.functional_groups
         }
         """Generate a dictionary of functional groups within the community."""
+        self.plant_community: PlantCommunity = PlantCommunity(10000.0, 1)
+        self.carcass_pool: CarcassPool = CarcassPool(10000.0, 1)
+        self.soil_pool: PalatableSoil = PalatableSoil(10000.0, 1)
```
dalonsoa, line 49: Why 10000.0 specifically?

dalonsoa (approved):
Besides my previous comments, just a minor question.

virtual_rainforest/models/animals/animal_communities.py
TaranRallings: With your above suggestion for a self.prey_groups, all of this will be moot i think. I'll be able to specify cannibalism and its conditions clearly in that attribute.

virtual_rainforest/models/animals/animal_communities.py
alexdewar, line 119: I see your list comprehension and I raise you an `itertools` :wink:

If you do `from itertools import chain`, then you can do:

```suggestion
        animal_list = list(chain.from_iterable(self.cohorts.values()))
```

alexdewar (changes requested):
Sorry for the really late response to this...

It looks good to me apart from what I think is a bug found by @dalonsoa in `herbivory()`. Other than that I'd be happy to merge.

I have made a few small suggestions elsewhere though.

virtual_rainforest/models/animals/animal_communities.py
davidorme: Welcome back! Hope all is going well with the little one. 

That solution is also less mystifying than `[item for sublist in l for item in sublist]`. I'm never convinced it isn't magic somehow.

virtual_rainforest/models/animals/animal_cohorts.py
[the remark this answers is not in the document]
TaranRallings: Ah, good find! I had planned to do the actual energy updating in forage_cohort but lost track of that along the way. I've updated it now.

Birth, on the other hand, is just a mistake. I think probably from debugging.

virtual_rainforest/models/animals/animal_cohorts.py
TaranRallings: Oops!

virtual_rainforest/models/animals/animal_communities.py
TaranRallings: It's a completely toy value that was round enough to be easy and big enough to not always be going to zero from a trophic interaction.

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -105,55 +108,98 @@ def metabolize(self, dt: timedelta64) -> None:
...
+        Returns:
+            A float containing the amount of energy consumed in the herbivory bout.
 
         """
-        waste_energy = consumed_energy * self.functional_group.conversion_efficiency
-        soil.energy += waste_energy * self.individuals
-        # The amount of waste by the average cohort member * number individuals.
+        # Calculate the number of individuals that can be eaten based on intake rate
+        # Here we assume predator can consume prey mass equivalent to its daily intake
+        number_eaten = int(min(self.intake_rate / prey.mass, prey.individuals))
```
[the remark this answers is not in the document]
TaranRallings: If I do this will the result not be a float? Would it be useful to do the floor division and call int()?

virtual_rainforest/models/animals/animal_cohorts.py
TaranRallings: > It's probably more complicated than that, but it was just a thought.

This is a good point. In the not so distant future get_eaten() will sometimes have to be handled by other modules in some way anyhow. I the registry point but I'll try get_eaten() for now and build towards the registry when things are more fleshed out.

dalonsoa (approved):
It looks good to me!

alexdewar (changes requested):
I've made some really minor suggestions, but I think your refactor turned out really nicely. It's def clearer now. Good job!

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -180,4 +148,98 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+        Args:
+            food: An object of a Resource class (currently: AnimalCohort, Plant
+                  Community)
+            pool: An object of a Pool class, which could represent depositional pools
+                  like soil or carcass pools.
+
+        """
+        # get the per-individual energetic gain from the bulk value
+        energy = food.get_eaten(self, pool) / self.individuals
+        self.stored_energy += energy
```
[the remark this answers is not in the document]
TaranRallings: Good point! When this is actually resolved into a simulation cycle a cohort with 0 individuals will "die" and be removed. I should probably check for this though.

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -85,11 +93,69 @@ def birth(self, cohort: AnimalCohort) -> AnimalCohort:
...
+                continue
+
+            # Get the size range of the prey this predator eats
+            min_size, max_size = consumer_cohort.prey_groups[prey_functional_group]
+
+            # Filter the potential prey cohorts based on their size
+            right_sized_prey = [
+                c for c in potential_prey_cohorts if min_size <= c.mass <= max_size
+            ]
+            prey.extend(right_sized_prey)
```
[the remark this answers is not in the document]
TaranRallings: That's a really interesting distinction, thanks for pointing it out.

virtual_rainforest/models/animals/protocols.py
```diff
@@ -0,0 +1,32 @@
...
+class Pool(Protocol):
+    """This is a protocol for defining dummy abiotic pools containing energy."""
+
+    stored_energy: float
+
+
+class Resource(Protocol):
+    """This is the protocol for defining what classes work as trophic resources."""
+
+    stored_energy: float
```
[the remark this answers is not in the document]
TaranRallings: It may end up working that way but it's not clear to me exactly what that is going to look like once things are stoichiometrically explicit. Some of these protocols may be abandoned as toy pools are trashed and animals are drawing on pools in other modules, which are handled in different ways. I will revisit the idea once we are further along.

alexdewar (approved):
LGTM :smile:

PR #251, opened 2023-07-07
Fix dask warning about saving "layer_roles" as an object

jacobcook1995:
# Description

I took the simplest approach I could think of to fix this warning. I just converted the "layer_roles" from being objects to being string with 9 characters ("S9"), right before the combined data is saved. With this change everything works without warnings, so I think that's enough to fix the bug.

I also changed the dummy_plant data recipe slightly so that the data set will show up as part of the documentation (which was already the case for the other recipes)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/data_recipes/plant_dummy.md
+++ b/docs/source/data_recipes/plant_dummy.md
@@ -23,9 +23,8 @@ section provides a recipe to create such an input conform with the dummy data fo
 [climate](./ERA5_preprocessing_example.md) and soil (add link here).
 
 ```{code-cell} ipython3
-import xarray as xr
 import numpy as np
-from xarray import DataArray
+from xarray import DataArray, Dataset
 
 # set layer roles
 layer_roles = ["above"] + 10 * ["canopy"] + ["subcanopy"] + ["surface"] + 2 * ["soil"]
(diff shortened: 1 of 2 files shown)


davidorme:
This suggests to me that the original `layer_roles` is created oddly - I can't see why these wouldn't just be string labels?

vgro (approved):
I think it works as a bug fix, we might want to find out at some point why this happens.

jacobcook1995:
It's something about the saving and reloading of netcdf files, like in the `data` object they are in fact string labels of "<U9" type. But when the data set is loaded back in they have become `object` type, which I presume is a catch all "this could be anything" type.

I *think* it's that `xarray`'s `Dataset.to_netcdf` function isn't able to save "U" type (for strings of indeterminate length) and instead can only save "S" (strings of fixed length), but couldn't see an option to override this.

jacobcook1995:
It would be a bit weird with them padded to constant length as you would then have to search for `"soil     "` (or `"     soil"`, or `"  soil   "` depending how the padding is done) when selecting layers.

I think a better solution for this is probably something that can wait until `layers` becomes a validated axis

PR #252, opened 2023-07-10
Updated poetry to latest versions

jacobcook1995:
# Description

The CI GitHub Action has a problem where occasionally a run will fail during the install dependancies step (see [here](https://github.com/ImperialCollegeLondon/virtual_rainforest/actions/runs/5505115190/jobs/10032099985)). This generally seems to happen on Windows with python version 3.10.

This PR updates `poetry` to latest version in case this is a `poetry` bug which has already been patched out. If it isn't then we'll have to investigate further.

(Updating `poetry` from `1.2.2` to `1.5.1` seems to halve dependancy install times, so I think this change is worthwhile regardless of if it fixes the bug)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,9 +29,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Install Poetry
-      uses: abatilo/actions-poetry@v2.1.6
+      uses: abatilo/actions-poetry@v2.3.0
       with:
-        poetry-version: 1.2.2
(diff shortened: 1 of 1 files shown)


davidorme (approved):
That's a lot of minor version changes in quite a short space of time! Still - they've been deprecating a load of old pre 1.00 options and updating stuff! Seems fine to me.

davidorme:
And it looks like one of those changes is explicitly updating to a 'modern' installer for packages.

PR #253, opened 2023-07-11
Adding a particulate organic matter pool

jacobcook1995:
# Description

This pull request adds an extra carbon pool to the soil carbon model. The pool added is the particulate organic matter (POM) pool. This pool is for organic matter that is too complex to be immediately taken up by microbes but that isn't protected through mineral association. The motivation for adding this pool is that the carbon transfers from litter to soil go to this pool, so adding it is a prerequisite for setting up a `litter` model/module.

I've just added @vgro as a reviewer here, as this PR just extends existing functionality and doesn't really add anything novel, so I don't feel it needs a really thorough going over. If others want to add comments that's completely fine.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/soil/test_carbon.py
+++ b/tests/models/soil/test_carbon.py
@@ -50,9 +50,10 @@ def test_calculate_soil_carbon_updates(dummy_carbon_data, top_soil_layer_index):
     from virtual_rainforest.models.soil.carbon import calculate_soil_carbon_updates
 
     change_in_pools = [
-        [1.44475655e-03, 1.01162673e-02, 7.04474125e-01, -5.43915134e-05],
+        [0.00525045055, 0.0205448757, 0.7136941904, 0.00141217],
         [0.13088391, 0.05654771, -0.39962841, 0.00533357],
         [-0.33131188, -0.16636299, -0.76078599, -0.01275669],
+        [-0.00349175378, -0.01011466818, -0.00890612528, -0.00115262138],
     ]
 
     delta_pools = calculate_soil_carbon_updates(
@@ -71,7 +72,8 @@ def test_calculate_soil_carbon_updates(dummy_carbon_data, top_soil_layer_index):
     # Check that the updates are correctly calculated
     assert np.allclose(delta_pools[:4], change_in_pools[0])
     assert np.allclose(delta_pools[4:8], change_in_pools[1])
-    assert np.allclose(delta_pools[8:], change_in_pools[2])
+    assert np.allclose(delta_pools[8:12], change_in_pools[2])
+    assert np.allclose(delta_pools[12:], change_in_pools[3])
 
 
 def test_calculate_mineral_association(dummy_carbon_data, moist_temp_scalars):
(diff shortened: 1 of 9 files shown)


tests/models/soil/test_carbon.py
```diff
@@ -70,7 +72,8 @@ def test_calculate_soil_carbon_updates(dummy_carbon_data, top_soil_layer_index):
     # Check that the updates are correctly calculated
     assert np.allclose(delta_pools[:4], change_in_pools[0])
     assert np.allclose(delta_pools[4:8], change_in_pools[1])
-    assert np.allclose(delta_pools[8:], change_in_pools[2])
+    assert np.allclose(delta_pools[8:12], change_in_pools[2])
+    assert np.allclose(delta_pools[12:], change_in_pools[3])
```
vgro, line 77: Do you have to make sure here that the pools are always in the right order or is this just for the test? Would it be worth giving them names and select by name?

tests/models/soil/test_carbon.py
```diff
@@ -311,12 +344,30 @@ def test_calculate_labile_carbon_leaching(dummy_carbon_data, moist_temp_scalars)
...
     )
+    from virtual_rainforest.models.soil.constants import LITTER_INPUT_RATE
 
-    actual_input = calculate_direct_litter_input_to_lmwc()
+    actual_input_lmwc, actual_input_pom = calculate_direct_litter_input_to_pools()
 
-    assert np.isclose(actual_input, 0.00015697011)
+    assert np.isclose(actual_input_lmwc, 0.00015697011)
+    assert np.isclose(actual_input_pom, 0.00031394022)
+    assert np.isclose(actual_input_lmwc + actual_input_pom, LITTER_INPUT_RATE)
```
vgro, line 373: Have you tested any of these functions with DataArrays? There might be some changes to the code required, like `.any()` or `.all()`

virtual_rainforest/models/soil/carbon.py
```diff
@@ -427,7 +528,10 @@ def calculate_direct_litter_input_to_lmwc(
             pools [kg C m^-2 day^-1].
 
     Returns:
-        Amount of carbon directly added to LMWC pool from litter.
+        Amount of carbon directly added to LMWC and POM pools from litter.
     """
```
vgro, line 532: Will this differ between grid cells at some point?

vgro (approved):
LGTM. 
The only thing I noticed is that a lot of functions operate with `float` which I guess assumes that this is the same across all grid cells. Are you planning on extending this to account for all cell_ids?

tests/models/soil/test_carbon.py
jacobcook1995: Yeah the pools should follow the order of the input dictionary, think this should be explicitly tested.

virtual_rainforest/models/soil/carbon.py
jacobcook1995: Yeah this is something that's a place holder until the `litter` module/model is added. Felt easier to keep this as a constant across all cells until a link with the litter module is made

tests/models/soil/test_carbon.py
jacobcook1995: I've changed the test setup so that it will fail if the output order does not match the order of the input dictionary

tests/models/soil/test_carbon.py
jacobcook1995: I think basically all of these functions return either `numpy` arrays or floats (as discussed the functions returning floats will go soon). It's only the highest level functions in the `soil` module that return `DataArray`s

PR #254, opened 2023-07-11
Feature/correct dummy climate

vgro:
I discovered some issues with the 2m temperature in the dummy climate data set and updated the documentation. I also changes the input precipitation from m to mm.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/data_recipes/ERA5_preprocessing_example.md
+++ b/docs/source/data_recipes/ERA5_preprocessing_example.md
@@ -64,11 +64,11 @@ dataset = xr.open_dataset("./ERA5_land.nc")
 dataset
 ```
 
-### 2. Convert temperatures to Celsius
+### 2. Convert temperatures units
 
 The standard output unit of ERA5-Land temperatures is Kelvin which we need to convert
-to degree Celsius for the Virtual Rainforest. This includes `2m air temperature` and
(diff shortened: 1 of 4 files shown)


docs/source/data_recipes/ERA5_preprocessing_example.md
```diff
@@ -64,11 +64,11 @@ dataset = xr.open_dataset("./ERA5_land.nc")
 dataset
 ```
 
-### 2. Convert temperatures to Celsius
+### 2. Convert temperatures units
 
 The standard output unit of ERA5-Land temperatures is Kelvin which we need to convert
-to degree Celsius for the Virtual Rainforest. This includes `2m air temperature` and
-`2m dewpoint temperature` which are used to calculate relative humidity in next step.
+to degree Celsius for the Virtual Rainforest. This includes 2m air temperature and
```
jacobcook1995, line 70: Is it worth switching from "m" to "meters" here, as you do that in the section below

jacobcook1995 (approved):
LGTM! Had a one minor comment.

docs/source/data_recipes/ERA5_preprocessing_example.md
vgro: I only changed it in the next section because it looked funny on its own without a number, here it is used as a unit.

PR #255, opened 2023-07-12
fix time_index

vgro:
There was an issue with the `time_index` as a coordinate only, we need it to be a dimension to be accessible by `isel`. I tried to add a 4th dimension but that caused problems, so as an intermediate solution I renamed the time dimension and create an issue.

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

PR #258, opened 2023-07-17
System to allow constants to be altered based on configuration

### LINKED ISSUE #208: Clarify roles and structure of Grid, Core and Data objects and use in models and axis validation.

### LINKED ISSUE #209: Allowing "constants" to vary
A lot of the constants that we have defined are not true constants. In the case of the soil module a lot of them are regression outputs from previous empirical studies. The sensitivity of these "constants" will therefore obviously needed to be tested, so they will need to be alterable in some way.

I don't really have a clear idea of a solution to this beyond redefining what we call a "constant" once it comes time to do sensitivity analysis. I think this is something we should discuss in detail at some point.

**Additional context**
[David's description of the problem.](https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/206#pullrequestreview-1400640673)

### LINKED ISSUE #210: Prevent changes to the value of constants
Both `dataclass`'s and simple constants defined using `ALL_CAPS` notation can be altered during the run of the model. We don't really want this to be possible if these are true constants (e.g. Boltzmann's constant).

Data classes can be frozen by specifying `@dataclass(frozen=True)`.

Simple constants can be annotated using [`typing.Final`](https://docs.python.org/3/library/typing.html#typing.Final) as follows:

```python
from typing import Final

EXAMPLE_CONSTANT: Final[float] = 0.036
```

This does not stop the constant from being altered, but altering the constant will cause `mypy` to raise an error. This should reduce the chance of altering constants incorrectly.

We might decide that this is a bit paranoid, but personally feel that it's better to be careful.

**Additional context**
Issue #209 basically pertains to the opposite problem, so the two issues are very much linked.

jacobcook1995:
# Description

This PR adds a system for reading constants from the configuration, which is necessary for sensitivity analysis. 

The idea is for each model to define either one or a small number of data classes in its `constants.py` module, with units and default values recorded there. Input arguments for these data classes can be stored in the configuration, and supplied in order to overwrite default values. This takes place within the model's `from_config` factory method, generating _instances_ of the constants classes to be fed to `__init__`, either using the default values, or configured with user provided values as needed.

The model `__init__` arguments are then extended to include any constants classes required and supplied constants class instances are then stored as attributes of the model instance and can be passed on to downstream functions either as entire constants classes or unpacked to individual members as required. 

The new system is only implemented here for the `abiotic_simple` model. The plan is that once it's merged further pull requests should be made to switch the other models (as well as core see #129) over to this system.

Anyway let me know what you think of this first draft.

Fixes #209
Fixes #210

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -44,7 +44,7 @@ to add other python modules containing different parts of the module functionali
 * A JSON Schema file defining the model configuration, called
   `{model_name}_schema.json`.
 * A python module  `constants.py` that will contain the constants relevant to the model
-  (in one or more dataclasses).
+  (see [here](../virtual_rainforest/parameterisation.md) for details).
 
 For example:
 
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -43,8 +43,7 @@ to add other python modules containing different parts of the module functionali
   object.
 * A JSON Schema file defining the model configuration, called
   `{model_name}_schema.json`.
-* A python module  `constants.py` that will contain the constants relevant to the model
-  (see [here](../virtual_rainforest/parameterisation.md) for details).
+* A python module  `constants.py` that will contain the constants relevant to the model.
 
 For example:
 
@@ -53,6 +52,40 @@ touch virtual_rainforest/models/freshwater/__init__.py
 touch virtual_rainforest/models/freshwater/freshwater_model.py
 touch virtual_rainforest/models/freshwater/freshwater_schema.json
 touch virtual_rainforest/models/freshwater/constants.py
+```
+
+## Defining constants and their default values
+
+Each model should define a `constants.py` module. Constants and their default values
+should be defined in this module using {func}`dataclasses.dataclass`. These constants
+can be stored in a single data class or spread over multiple data classes. However,
+having a large number of data classes is likely to make the downstream code messier, so
+constants should only be split across multiple classes when there's a strong reason to
+do so.
+
+It's also important that every constant is given an explicit type hint. If a type hint
+is not provided then `dataclass` treats the constant as a class attribute rather than an
+instance attribute. This means that its value cannot be changed when a new instance is
+created.
+
+An example `constants.py` file is shown below:
+
+```python
+from dataclasses import dataclass
+
+# The dataclass must be frozen to prevent constants from being accidentally altered
+# during runtime
+@dataclass(frozen=True)
+class ExampleConsts:
+    """Dataclass to store all constants for the `example_model` model."""
+    
+    # Each constant must be given a type hint, otherwise its default value cannot be
+    # changed
+    example_constant_1: float = -1.27
+    """Details of source of constant and its units."""
+
+    example_constant_2: float = 5.4
+    """Details of source of constant and its units."""
 ```
 
 ## Defining the new model class
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -310,10 +310,10 @@ method should raise an `InitialisationError` if the configuration fails.
 The `from_config` method should also check if any constants have been provided as part
 of the configuration. If they haven't a default set of constants is generated. If they
 have been supplied they are used to generate a custom set of constants. The
-{func}`~virtual_rainforest.core.utils.check_constants` utility function is used to check
-that no constant has been supplied with an incorrect name. At least one constants class
-should be created, but it's fine to split constants across more classes if that makes
-for clearer code.
+{func}`~virtual_rainforest.core.utils.check_valid_constant_names` utility function is
+used to check that no constant has been supplied with an incorrect name. At least one
+constants class should be created, but it's fine to split constants across more classes
+if that makes for clearer code.
 
 As an example:
 
@@ -340,7 +340,7 @@ def from_config(
     # Check if any constants have been supplied
     if "freshwater" in config and "constants" in config["freshwater"]:
         # Checks that constants is config are as expected
-        check_constants(config, "freshwater", "FreshwaterConsts")
+        check_valid_constant_names(config, "freshwater", "FreshwaterConsts")
         # If an error isn't raised then generate the dataclass
         constants = FreshwaterConsts(
             **config["freshwater"]["constants"]["FreshwaterConsts"]
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -34,7 +34,7 @@ directory.
 mkdir virtual_rainforest/models/freshwater
 ```
 
-You will need to create at least three files within this folder, although you may choose
+You will need to create at least four files within this folder, although you may choose
 to add other python modules containing different parts of the module functionality.
 
 * An `__init__.py` file, which tells Python that the folder is a submodule within the
@@ -43,6 +43,8 @@ to add other python modules containing different parts of the module functionali
   object.
 * A JSON Schema file defining the model configuration, called
   `{model_name}_schema.json`.
+* A python module  `constants.py` that will contain the constants relevant to the model
+  (in one or more dataclasses).
 
 For example:
 
@@ -50,6 +52,7 @@ For example:
 touch virtual_rainforest/models/freshwater/__init__.py
 touch virtual_rainforest/models/freshwater/freshwater_model.py
 touch virtual_rainforest/models/freshwater/freshwater_schema.json
+touch virtual_rainforest/models/freshwater/constants.py
 ```
 
 ## Defining the new model class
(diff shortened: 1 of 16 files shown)


jacobcook1995:
> Is there a way (or really a need) to provide some kind of a base model for constants classes? We could even build a registry and have some kind of checking at startup? This isn't critical for this proof of concept at all, but it would be nice to be able to enforce some kind of structure on the constants classes.

I'm not sure about a base model, because would the different constants classes share any attributes or methods? It might be worth adding them to a registry though for ease of access though

davidorme:
Yeah - it isn't an ABC, something simpler that just checks for "frozen dataclass"-ness and  registers them. But that really isn't needed here or even for rolling this out to other models.

docs/source/development/defining_new_models.md
```diff
@@ -300,6 +307,14 @@ The method then uses those parsed arguments to actually call the `__init__` meth
...
 method should raise an `InitialisationError` if the configuration fails.
 
+The `from_config` method should also check if any constants have been provided as part
+of the configuration. If they haven't a default set of parameters is generated. If they
+have been supplied they are used to generate a custom set of parameters. The
+{func}`~virtual_rainforest.core.utils.check_constants` utility function is used to check
+that no constant has been supplied with an incorrect name. At least one parameter class
+should be created, but it's fine to split parameters across more classes if that makes
+for clearer code.
+
```
vgro, line 317: Here again, to be clear, I would stick to 'constants' or 'parameters' (whichever you chose)

docs/source/virtual_rainforest/core/config.md
```diff
@@ -50,6 +50,21 @@ optional. A list of modules to be configured can be specified using the tag
...
+analysis can find and alter them straightforwardly. Each model should define its own set
+of constants under `model_name.constants`. Within this one or more groups of constants
+should be defined, these groups are each input for a particular dataclass in the relevant
+`constants.py` module, and should use the same name as this dataclass. Individual
+constants should then be defined within these groups. The full tag structure would take
+the following pattern `model_name.constants.SomeVars.constant_1`. It is important to
+note that **only** the constants you wish to change need to be included in the
+configuration, any constant that isn't given will just use the default value set in the
+relevant `constants.py` module.
+
```
vgro, line 67: Should we mention here the `core.constants` or will this be considered separately?

docs/source/development/defining_new_models.md
```diff
@@ -177,6 +180,7 @@ def __init__(
     data: Data,
     update_interval: pint.Quantity,
     no_of_ponds: int,
+    parameters: FreshwaterParams,
```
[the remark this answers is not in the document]
davidorme: We should also have a clear statement about whatever distinction we make in the docs - not just the model creation docs but also to clarify it to users. 

We have a bunch of "parameters" that must be kept constant across a simulation and they are defined in params/constants classes and configured via the params/constants TOML sections. I'd argue for constants - within the context of particular simulation.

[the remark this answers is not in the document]
jacobcook1995: Yeah think I probably should set aside my squeamishness about calling them constants as it makes better sense to call them all that

virtual_rainforest/core/utils.py
```diff
@@ -119,3 +121,49 @@ def set_layer_roles(canopy_layers: int, soil_layers: int) -> List[str]:
         + ["soil"] * soil_layers
     )
     return layer_roles
+
+
+def check_constants(config: dict[str, Any], model_name: str, class_name: str) -> None:
+    """Check that the constants defined in the config are expected.
+
```
vgro, line 128: It is not clear to me where the defaults are stored and how this function checks the inputs, is there a list of allowed variables names somewhere?

jacobcook1995: The defaults are stored within the constants class, e.g. `AbioticSimpleConsts` in `abiotic_simple`'s `constants.py` module. It's not really documented anywhere that this is the case though, so might be worth me making a new documentation page for this.

One thing this function does do is print the full list of possible names which should hopefully make it obvious for the user where they have gone wrong

docs/source/virtual_rainforest/core/config.md
jacobcook1995: I don't think that any of this will be different for `core.constants`, provided we are prepared to call `core` a model name. Unless we have to settle on a radically different way of handling `core` constants

davidorme: I think it will as `core` isn't a model. This doesn't matter until we actually move shared constants into core but I think the core configuration will use a different mechanism. It isn't immediately clear to me where those config instances end up, but it's part and parcel of restructuring the `Data`/`Grid`/`Config` information.

vgro (approved):
LGTM.
I agree that the documentation could be a bit clearer but that can be a separate PR.

jacobcook1995:
@vgro I've just added a new docs page that hopefully addresses some of the problems you pointed out. Let me know if it looks sensible to you, and if there is anything you think is missing

docs/source/virtual_rainforest/parameterisation.md
```diff
@@ -0,0 +1,59 @@
...
+is fine to use different values for them across different simulations.
+
+## Defining constants and their default values
+
+Each model should define a `constants.py` submodule. Constants and their default values
+should be defined in this submodule using {func}`dataclasses.dataclass`. These constants
+can be stored in a single data class or spread over multiple data classes. However,
+having a large number of data classes is likely to make the downstream code messier, so
+constants should only be split across multiple classes when there's a strong reason to
+do so. It's also important that every constant is given an explicit type hint, otherwise
```
vgro, line 22: and a unique name

docs/source/virtual_rainforest/parameterisation.md
```diff
@@ -0,0 +1,59 @@
...
+class they belong to. An example of this can be seen below:
+
+```toml
+[example_model.constants.ExampleConsts]
+example_constant_1 = 23.0
+example_constant_2 = -7.7
+```
+
+Any values supplied in this way will be used to override the default values for the data
+class in question. Only constants for which non-default values are supplied will be
```
vgro, line 57: maybe add the link to the config documentation here

jacobcook1995: Ahh I probably should clarify that these aren't the default values set by the config system, they are the defaults set in the data classes

davidorme (changes requested):
Looking good - I think we need to be clearer about docs roles and I think there is a smoother way to check bad constants names that doesn't require creating multiple instances.

virtual_rainforest/core/utils.py
```diff
@@ -143,26 +143,19 @@ def check_constants(config: dict[str, Any], model_name: str, class_name: str) ->
...
     if unexpected_names:
         LOGGER.error(
-            "Incorrect constant names supplied for %s dataclass: %s"
-            % (class_name, unexpected_names)
+            "Unknown names supplied for %s constants: %s"
+            % (class_name, ', '.join(unexpected_names))
         )
-        # Find all valid names and inform the user about them
-        valid_names = [fld.name for fld in dataclasses.fields(ConstantsClass)]
         LOGGER.info("Valid names are as follows: %s" % (valid_names))
```
davidorme, line 161: Sorry - should have fixed this to remove those set braces in the logger message.

```python
LOGGER.info("Valid names are as follows: %s" % (", ".join(valid_names)))
```

dalonsoa (approved):
Nothing to add. It looks a good approach easy to roll out to the other models. 

My only concern is that the constants that a model requires and their default values are going to be hard to find except if the interested party dives into the code or the documentation for the API, which is less than ideal.

Probably not for this PR, but exploiting the idea of the constants registry suggested by @davidorme , I guess it should be possible to extract all the constants and their values and put them in a dedicated section in the documentation when this one is built.  

@alexdewar , here you have all you need to start working on the sensitivity analysis code!

jacobcook1995:
Ah great! I'll add an issue for the constants registry + documentation

jacobcook1995:
@davidorme I've addressed you requested changes aside the constants registry one, which I've made an issue for (#259)

davidorme:
LGTM

davidorme (approved):

PR #260, opened 2023-07-24
Feature/update hydrology constants

vgro:
This PR updates the use of constants as agreed for the `abiotic_simple` in #258 in the `hydrology` model.

In the course of this, I unpacked the `update_soil_moisture()` function in `update()` because the multiple levels of nesting made the use of constants and documentation very messy (this was an open TODO which is now resolved). I updated the doc strings accordingly. I kept a `dict` to update the data object at the end of `update()`, this might no longer be required. Please let me know what you think about that.

I also added some comments on the likely replacement by the SPLASH model in the model doc strings and the model description on the model overview page.

There is currently a `NotImplementedError()` in the update function because I only allow monthly time steps. This is one of the next things to update.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/hydrology/constants.py
```diff
@@ -1,27 +1,77 @@
...
+    """
+
+    hydraulic_gradient: float = 0.01
+    """The hydraulic gradient (m/m) is a measure of the change in hydraulic head
+    (pressure) per unit of distance in a particular direction within a fluid or porous
+    medium, such as soil or an aquifer. It represents the driving force behind the
+    movement of water and indicates the direction in which water will flow.
+    """
+
+    seconds_to_month: float = 2.628e6
```
jacobcook1995, line 51: This is probably a candidate for a universal constant (though obviously only when we get round to that)

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -204,242 +222,228 @@ def spinup(self) -> None:
...
-        HydroConsts["hydraulic_conductivity"]
-    ),
-    hydraulic_gradient: Union[float, DataArray] = (HydroConsts["hydraulic_gradient"]),
-    timestep_conversion_factor: float = (HydroConsts["seconds_to_month"]),
-    nonlinearily_parameter: Union[float, DataArray] = (
-        HydroConsts["nonlinearily_parameter"]
-    ),
-    meters_to_millimeters: float = HydroConsts["meters_to_millimeters"],
+    soil_moisture_capacity: Union[float, DataArray],
+    soil_moisture_residual: Union[float, DataArray],
```
jacobcook1995, line 441: Are these `Union`'s because you can either supply them with real data or just a constant fixed across all grid cells?

jacobcook1995 (approved):
LGTM!

If you wanted to retain the nesting then there isn't a problem with supplying `constants` unpacked to nested functions, and then just unpacking when it becomes appropriate. But if you already had problems with the nesting then it seems fine to get rid of it.

virtual_rainforest/models/hydrology/constants.py
vgro: yes, absolutely.

virtual_rainforest/models/hydrology/hydrology_model.py
davidorme: Actually - this is a thing I went through in `pyrealm` and asked @dalonsoa and @alexdewar about. Having this `Union` typing for scalar versus arrays inputs makes things really long-winded, compared to simply asserting that it is an array (which could be a scalar array).

In my case, I was dealing with scalars and np.arrays and had lots of Union statements, but I ended up ripping all of that out and just asserting that all things were going to be arrays. I'm not sure it works quite so cleanly with DataArrays, but I think it is worth thinking about insisting on one kind of input and simplifying this?

virtual_rainforest/models/hydrology/constants.py
davidorme: And several below too 😄

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -127,12 +130,27 @@ def from_config(
...
+        # Check if any constants have been supplied
+        if "hydrology" in config and "constants" in config["hydrology"]:
+            # Checks that constants is config are as expected
+            check_valid_constant_names(config, "hydrology", "HydroConsts")
+            # If an error isn't raised then generate the dataclass
+            constants = HydroConsts(**config["hydrology"]["constants"]["HydroConsts"])
+        else:
+            # If no constants are supplied then the defaults should be used
+            constants = HydroConsts()
+
```
davidorme, line 142: When we set up the `constants` registry, that should include the config path in the registry details. That should then allow us to have a shared general function in `from_config` methods:

constants = load_constants(config, HydroConsts)

and move all of this into that shared function. We can key the registry dictionary using the classes, I think.

vgro: I guess this is something for a separate PR because it will need to be adjusted across all models?

virtual_rainforest/models/hydrology/hydrology_model.py
vgro: I had a look into it and the problem with `DataArray` seems to be that the dimensions have to match, so the constants have to have at least the `cell_id `dimension. It seems to me like more of a clutter to introduce that than having the `Union` statement.

PR #261, opened 2023-07-25
Change soil constants to use new system

jacobcook1995:
# Description

This PR changes puts all the soil constants into a single data class which is then stored as an attribute of the soil model. This follows the new system for constants, that has so far only be applied to the `abiotic_simple` and `hydrology` models.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/soil/test_carbon.py
+++ b/tests/models/soil/test_carbon.py
@@ -23,16 +23,16 @@ def moist_temp_scalars(dummy_carbon_data, top_soil_layer_index):
 
     moist_scalars = convert_moisture_to_scalar(
         dummy_carbon_data["soil_moisture"][top_soil_layer_index],
-        SoilConsts().moisture_scalar_coefficient,
-        SoilConsts().moisture_scalar_exponent,
+        SoilConsts.moisture_scalar_coefficient,
+        SoilConsts.moisture_scalar_exponent,
     )
     temp_scalars = convert_temperature_to_scalar(
         dummy_carbon_data["soil_temperature"][top_soil_layer_index],
-        SoilConsts().temp_scalar_coefficient_1,
-        SoilConsts().temp_scalar_coefficient_2,
-        SoilConsts().temp_scalar_coefficient_3,
-        SoilConsts().temp_scalar_coefficient_4,
-        SoilConsts().temp_scalar_reference_temp,
+        SoilConsts.temp_scalar_coefficient_1,
+        SoilConsts.temp_scalar_coefficient_2,
+        SoilConsts.temp_scalar_coefficient_3,
+        SoilConsts.temp_scalar_coefficient_4,
+        SoilConsts.temp_scalar_reference_temp,
     )
 
     return moist_scalars * temp_scalars
@@ -85,7 +85,7 @@ def test_calculate_soil_carbon_updates(dummy_carbon_data, top_soil_layer_index):
         dummy_carbon_data["soil_temperature"][top_soil_layer_index],
         dummy_carbon_data["percent_clay"],
         pool_order,
-        constants=SoilConsts(),
+        constants=SoilConsts,
     )
 
     # Check that the updates are correctly calculated. Using a loop here implicitly
@@ -109,7 +109,7 @@ def test_calculate_mineral_association(dummy_carbon_data, moist_temp_scalars):
         dummy_carbon_data["bulk_density"],
         moist_temp_scalars,
         dummy_carbon_data["percent_clay"],
-        constants=SoilConsts(),
+        constants=SoilConsts,
     )
 
     # Check that expected values are generated
@@ -127,7 +127,7 @@ def test_calculate_equilibrium_maom(dummy_carbon_data):
         dummy_carbon_data["pH"],
         Q_max,
         dummy_carbon_data["soil_c_pool_lmwc"],
-        constants=SoilConsts(),
+        constants=SoilConsts,
     )
     assert np.allclose(equib_maoms, output_eqb_maoms)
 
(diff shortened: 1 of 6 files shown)


virtual_rainforest/models/soil/constants.py
```diff
@@ -4,117 +4,141 @@
...
-from typing import Final
 
 # TODO - Need to figure out a sensible area to volume conversion
 
 
 @dataclass(frozen=True)
-class BindingWithPH:
-    """From linear regression :cite:p:`mayes_relation_2012`."""
+class SoilConsts:
+    """Dataclass to store all constants for the `abiotic_simple` model."""
```
vgro, line 13:
```suggestion
    """Dataclass to store all constants for the `soil` model."""
```

jacobcook1995: Oh shit, the risks of copy + paste

virtual_rainforest/models/soil/constants.py
```diff
@@ -4,117 +4,141 @@
...
 
-    slope: float = -0.186
-    """Units of pH^-1."""
-    intercept: float = -0.216 + 3.0
-    """Unit of log(m^3 kg^-1). n.b. +3 converts from mg^-1 to kg^-1 and L to m^3"""
+    binding_with_ph_slope: float = -0.186
+    """Units of [pH^-1]. From linear regression :cite:p:`mayes_relation_2012`."""
```
vgro, line 16: Could you add a bit more information about the parameters, especially the ones from linear regression?

virtual_rainforest/models/soil/constants.py
```diff
@@ -4,117 +4,141 @@
 """  # noqa: D205, D415
```
vgro, line 4: Why is there a " after constants?

jacobcook1995: I think I had it as "constants", then decided against but forgot to delete the second "

virtual_rainforest/models/soil/constants.py
```diff
@@ -4,117 +4,141 @@
...
+    """Units of [pH^-1]. From linear regression :cite:p:`mayes_relation_2012`."""
 
+    binding_with_ph_intercept: float = -0.216 + 3.0
+    """From linear regression :cite:p:`mayes_relation_2012`.
 
-@dataclass(frozen=True)
-class MaxSorptionWithClay:
+    Unit of [log(m^3 kg^-1)]. n.b. +3 converts from mg^-1 to kg^-1 and L to m^3
+    """
     """From linear regression :cite:p:`mayes_relation_2012`."""
```
vgro, line 23: I think you duplicated the doc string here?

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -50,7 +51,9 @@ class SoilModel(BaseModel):
         data: The data object to be used in the model.
         start_time: A datetime64 value setting the start time of the model.
         update_interval: Time to wait between updates of the model state.
```
vgro, line 53: I just noticed that the start_time is not in the `AbioticModel`, is it missing there or did we drop it altogether?

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -117,6 +121,8 @@ def __init__(
         """The layer in the data object representing the first soil layer."""
```
vgro, line 121: I noticed that in the `AbioticModel`, the layer roles are assigned as an attribute (`self.layer_roles`), does that make a difference?

virtual_rainforest/models/soil/carbon.py
```diff
@@ -76,22 +73,25 @@ def calculate_soil_carbon_updates(
         bulk_density,
         moist_temp_scalar,
         percent_clay,
+        constants,
     )
     microbial_uptake = calculate_microbial_carbon_uptake(
-        soil_c_pool_lmwc, soil_c_pool_microbe, moist_temp_scalar, soil_temp
+        soil_c_pool_lmwc, soil_c_pool_microbe, moist_temp_scalar, soil_temp, constants
```
vgro, line 79: Do you import `constants` when you use more than one?

virtual_rainforest/models/soil/soil_model.py
jacobcook1995: It should be here actually, the docstring is out of date with the arguments of the `__init__`. `AbioticSimpleModel` does something with `start_time` (iirc), but models don't need it by default anymore. It being included here is a hangover from a while ago

tests/models/soil/test_carbon.py
```diff
@@ -21,10 +22,17 @@ def moist_temp_scalars(dummy_carbon_data, top_soil_layer_index):
     )
 
     moist_scalars = convert_moisture_to_scalar(
-        dummy_carbon_data["soil_moisture"][top_soil_layer_index]
+        dummy_carbon_data["soil_moisture"][top_soil_layer_index],
+        SoilConsts().moisture_scalar_coefficient,
```
vgro, line 26: What are the brackets for? Or in other words, why `SoilConsts().bla` and not `SoilConsts.bla`?

virtual_rainforest/models/soil/soil_model.py
jacobcook1995: I chose not to assign them as an attribute as I didn't think I would use them elsewhere, i.e. I would just make use of `self.top_soil_layer_index`. If that changes I would definitely take the approach you've used in your abiotic models

virtual_rainforest/models/soil/carbon.py
jacobcook1995: I messed this up a bit, but the idea was that `SoilConsts` is only importing for type hinting. Then `constants` is passed down the function stack as an argument, starting with `SoilModel.update` which passes `self.constants` avoiding any need for constants to be imported. (if that make sense?)

vgro (approved):
LGTM, I added a few minor comments and questions for clarification.

virtual_rainforest/models/soil/carbon.py
vgro: makes sense 👍

tests/models/soil/test_carbon.py
jacobcook1995: Turns out it doesn't do anything 😐

jacobcook1995: I'll get rid of all instances of the brackets in tests then!

virtual_rainforest/models/soil/constants.py
jacobcook1995: I've added extra details for the parameters from linear regression.

For the temperature and moisture scalars, I'm honestly not completely sure what they are. I've added a TODO about this because it's a much bigger problem that the description of constants

PR #262, opened 2023-07-26
Feature/add streamflow

vgro:
This PR adds a super simple stream flow equation to the hydrology model: Q=P-ET-dS . This requires the plant model to run first, but as the p-model does currently not need soil moisture as an input, that should work. I added ET to the dummy plant data set and made a small adjustment to the input precipitation. Once we add p-hydro or SPLASH, we will have to rejig this; maybe split the stream flow from the hydrology model.
For now, the new features gives us the stream flow Q for each grid cell which can then in a next step be summed up for the whole area or catchments within the study area. 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

Issue #239, opened 2023-06-09
Implement sensitivity analysis preparation


dalonsoa:
VR depends on many parameters that, in isolation or collectively, define the evolution of the simulation. A sensitivity analysis will help to identify which parameters have more impact in the simulation, which ones result in small changes when vary, and which might be correlated with each other. 

This umbrella issue implements the necessary tooling and configuration files to create a collection of input files that can be run massively in parallel at Imperial's HPC using array jobs.

- [x] #240
- [x] #241
- [x] #242

davidorme:
I think the sensitivity analysis tool is likely to be widely used by end users and is going to be an important component of many of the end user stories we anticipate for `vr`. Not all of those end users are going to have access to an HPC (or alternatively the specific flavour of HPC we use, although differences there are likely cosmetic). I think we want to support SA runs on a personal computer (ideally allowing parallel runs on different cores) so I think we also need to provide an approach to batch run and compile SA runs on a local machine.

I think that also alters how we approach configuring the runs? If we're using array jobs on the HPC, then having a potentially large folder of config files and then just chunking them into array job sized blocks and then looping over a set of inputs to `vr_run` makes sense. If this is something running on a personal computer, I worry about the sea of largely identical TOML files - which could get very large if more than one parameter is varying and we need combinations.

Could we have a `SensitivityAnalysisManager`, with methods to:
* `ingest` the SA input file and work out a list of different permutations of the varying parameters.
* outputs the permutation list as TOML, optionally divided into `n` chunks.

Then we have a `SensitivityAnalysisRunner`, which takes a broader config file and a (possibly divided) permutation list and runs `vr`, overwriting the broader config with each setting in the permutation list in turn.

This still works with the HPC - we can `divide` into job array chunks and then parallelise those chunks, but it also provides a framework to run locally. We could potentially then add a `parallelise` method to `SensitivityAnalysisRunner` which passes jobs to different cores?

dalonsoa:
The actual implementation is something we will need to consider, but I agree that running a SA in the perfectly capable, multicore computers we all have is a perfectly sensible use case. We can certainly implement something that can serve both of them.

About the `the sea of largely identical TOML files`, I feel that's the least of your problems: if you are running a sensitivity analysis with many permutations in your laptop, the killing factor is going to be the collection of output data that will be produced. The TOML files can be discarded after using them, as you suggest, but the output files will need to stay, and that might represent a barrier.

[...]

PR #267, opened 2023-07-31
Allow setting params with command-line arguments

### LINKED ISSUE #266: `vr_run`: Allow for setting parameter values from command line
The idea for this is to be able to allow users to set values from single parameters in addition to the values provided via config files, e.g.:

```sh
vr_run --param hydrology.initial_soil_moisture=0.3 --param some.other.parameter=4 dummy_data/
```

This would be particularly useful for #239. Currently, the only way to pass parameters into `vr_run` is via one or more config files and if you're attempting to do a parameter sweep, this will mean generating a whole bunch of input files. Having these extra files lying around isn't such a big deal (though it is a somewhat inelegant solution), but having to generate them before being able to invoke `vr_run` means having another step in the SA pipeline, which makes things more complex.

My suggestion is that the user specifies a value via the command line that is already set in a config file, then we should raise an error, i.e. this shouldn't be a mechanism for overriding existing config values.

The implementation of this should be fairly simple, because you can parse strings with `tomllib` directly, e.g.:

```
>>> tomllib.loads("hydrology.initial_soil_moisture=0.3")
{'hydrology': {'initial_soil_moisture': 0.3}}
```

alexdewar:
# Description

This PR adds the option to set additional parameters when invoking `vr_run` by using command-line arguments, without having to create additional config files.

For example, it allows you to do something like this:

```sh
vr_run --param hydrology.initial_soil_moisture=0.2 --param some.other.param=value dummy_data/
```

The main motivation for this is that I want to be able to invoke `vr_run` in parallel with different parameters as part of my work on the SA (see #239). While that is currently possible, it requires generating config files as an intermediate step, which makes the process substantially more complex.

Closes #266.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -298,7 +298,7 @@ class Config(dict):
     Args:
         cfg_paths: A string, Path or list of strings or Paths giving configuration
             file or directory paths.
-        extra_params: Extra parameters provided by the user.
+        override_params: Extra parameters provided by the user.
         auto: flag to turn off automatic validation.
 
     """
@@ -306,7 +306,7 @@ class Config(dict):
     def __init__(
         self,
         cfg_paths: Union[str, Path, list[Union[str, Path]]],
-        extra_params: list[dict[str, Any]] = [],
+        override_params: dict[str, Any] = {},
         auto: bool = True,
     ) -> None:
         # Standardise cfg_paths to list of Paths
@@ -316,8 +316,6 @@ class Config(dict):
             self.cfg_paths = [Path(p) for p in cfg_paths]
 
         # Define custom attributes
-        self.extra_params = deepcopy(extra_params)
-        """A list of dictionaries of extra parameters supplied by the user."""
         self.toml_files: list[Path] = []
         """A list of TOML file paths resolved from the initial config paths."""
         self.toml_contents: dict[Path, dict] = {}
@@ -336,6 +334,7 @@ class Config(dict):
             self.resolve_config_paths()
             self.load_config_toml()
             self.build_config()
+            self.override_config(override_params)
             self.build_schema()
             self.validate_config()
 
@@ -438,7 +437,7 @@ class Config(dict):
         """
 
         # Get the config dictionaries
-        input_dicts = list(self.toml_contents.values()) + self.extra_params
+        input_dicts = list(self.toml_contents.values())
 
         if len(input_dicts) == 0:
             # No input dicts, Config dict is empty
@@ -601,3 +600,23 @@ class Config(dict):
         with open(outfile, "wb") as toml_file:
             tomli_w.dump(self, toml_file)
         LOGGER.info("Saving config to: %s", outfile)
+
+    def override_config(self, override_params: dict[str, Any]) -> None:
+        """Override any parameters desired.
+
+        Args:
+            override_params: Extra parameter settings
+        Raises:
+            ValidationError: if an invalid value is provided for any configuration
+                             settings.
+        """
+        updated, conflicts = config_merge(self, override_params, conflicts=tuple())
+
+        # Conflicts are not errors as we want users to be able to override parameters
+        if conflicts:
+            LOGGER.info(
+                "The following parameter values were overridden: "
+                + ", ".join(conflicts)
+            )
+
+        self.update(updated)
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -606,9 +606,6 @@ class Config(dict):
 
         Args:
             override_params: Extra parameter settings
-        Raises:
-            ValidationError: if an invalid value is provided for any configuration
-                             settings.
         """
         updated, conflicts = config_merge(self, override_params, conflicts=tuple())
 
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -321,7 +321,7 @@ class Config(dict):
         self.config_errors: list[tuple[str, Any]] = []
         """Configuration errors, as a list of tuples of key path and error details."""
         self.merged_schema: dict = {}
-        """The merged schema for the core and modules present in the configutation."""
+        """The merged schema for the core and modules present in the configuration."""
         self.validated: bool = False
         """A boolean flag indicating successful validation."""
 
--- a/virtual_rainforest/core/config.py
+++ b/virtual_rainforest/core/config.py
@@ -31,7 +31,7 @@ import json
 import sys
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Iterator, Union
+from typing import Any, Iterator, Optional, Union
 
 import dpath.util  # type: ignore
 import tomli_w
@@ -298,12 +298,16 @@ class Config(dict):
     Args:
         cfg_paths: A string, Path or list of strings or Paths giving configuration
             file or directory paths.
+        extra_params: Extra parameters provided by the user.
         auto: flag to turn off automatic validation.
 
     """
 
     def __init__(
-        self, cfg_paths: Union[str, Path, list[Union[str, Path]]], auto: bool = True
+        self,
+        cfg_paths: Union[str, Path, list[Union[str, Path]]],
+        extra_params: Optional[list[dict[str, Any]]] = None,
+        auto: bool = True,
     ) -> None:
         # Standardise cfg_paths to list of Paths
         if isinstance(cfg_paths, (str, Path)):
@@ -312,6 +316,8 @@ class Config(dict):
             self.cfg_paths = [Path(p) for p in cfg_paths]
 
         # Define custom attributes
+        self.extra_params = deepcopy(extra_params) if extra_params else []
+        """A list of dictionaries of extra parameters supplied by the user."""
         self.toml_files: list[Path] = []
         """A list of TOML file paths resolved from the initial config paths."""
         self.toml_contents: dict[Path, dict] = {}
@@ -432,7 +438,7 @@ class Config(dict):
         """
 
         # Get the config dictionaries
-        input_dicts = list(self.toml_contents.values())
+        input_dicts = list(self.toml_contents.values()) + self.extra_params
 
         if len(input_dicts) == 0:
             # No input dicts, Config dict is empty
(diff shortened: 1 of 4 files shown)


virtual_rainforest/core/config.py
```diff
@@ -432,7 +438,7 @@ def build_config(self) -> None:
         """
 
         # Get the config dictionaries
-        input_dicts = list(self.toml_contents.values())
+        input_dicts = list(self.toml_contents.values()) + self.extra_params
```
jacobcook1995, line 441: Am I right in thinking that this step adds the extra parameters to the existing config dictionaries? If so is there anything here that handles the case where the config files define a parameter which is then supplied as an extra parameter? I'm not actually 100% sure what the desired behaviour would be in that case?

jacobcook1995 (commented):
Looks sensible to me, just had a query about how parameters defined in both config files and the command line will be handled.

virtual_rainforest/core/config.py
```diff
@@ -298,12 +298,16 @@ class Config(dict):
...
+        extra_params: Extra parameters provided by the user.
         auto: flag to turn off automatic validation.
 
     """
 
     def __init__(
-        self, cfg_paths: Union[str, Path, list[Union[str, Path]]], auto: bool = True
+        self,
+        cfg_paths: Union[str, Path, list[Union[str, Path]]],
+        extra_params: Optional[list[dict[str, Any]]] = None,
```
davidorme, line 309: I'm curious as to why this uses `Optional` and `None` when the custom attribute is immediately  set to be an empty list if it isn't provided. As opposed to doing:

```suggestion
        extra_params: list[dict[str, Any]] = [],
```

That is then a mutable default, which is dangerous, but it could be a tuple of dicts instead.

virtual_rainforest/core/config.py
alexdewar: If a parameter is set in one of the config files and you then *also* try to set it via a command-line arg, then you'll get an error, just as you would if you tried to set twice in different config files.

I can see that there might be a case where a user wants to override a parameter set in a config file, but I don't think that should be the default behaviour as it's a bit error prone. If we want to add this feature later we can do (e.g. we could have a separate `--param-override` flag for this case).

jacobcook1995: Ahh yeah that makes sense! Also agree that overriding parameters seems error prone

davidorme: I'm not sure it is dangerous and I think there is a case for making the override behaviour the explicit mechanism for this approach. 

* The configuration process at the moment takes a bunch of files and uses those to build the configuration tree, substituting default values for any settings that aren't explicitly mentioned. That gives us a base configuration, with everything set. 

* If we want to tweak those settings, we shouldn't need to worry about whether the base files set a value or if it is just being populated by default. That would put us in a position of having to maintain multiple sets of base configuration files, each one omitting the setting that we want to tweak. It would be neater to have a default config that can just be adjusted as needed.

* Explicitly providing extra params would then be a way to substitute values in a particular run for a sensitivity analysis, by updating the `Config` object created from the base configuration for a given run. Those extra params aren't then simply added to the TOML contents, but are fed into a separate `Config` method. That does then need some kind of a validation step though.

davidorme: I think that method would be something like:

```python
    def override_config(self, extra_params) -> None:
        """Adjust an already created config to update settings

        Raises:
            ValidationError: if duplicate config definitions occur across files.
        """

        # Copy the current dictionary
        master = deepcopy(self.__dict__)
        
        # Update that dictionary
        # TODO - don't know if this works to replace values in master even when conflicts found, 
        #               or whether we need an explicit switch or method to do override.
        master, conflicts = config_merge(master, extra_params, conflicts=tuple())
        
        # Ignore any conflicts - explicitly overriding settings - so now revalidate against the 
        # schema defined in the base config.
        self.update(master)
        self.validate_config()
```

davidorme (commented):
This works - and that is probably the thing that matters now - but I think having to avoid duplicating configuration settings in TOML files and `extra_params` is another route to piles of slightly differing configuration TOML files.  This approach definitely avoids having multiple files that differ in just one (or two or three...) values, but does require unique config files for each param set being altered in a specific sensitivity analysis. 

What do you think? @jacobcook1995 - what do you think the dangers here? We'd _have_ to explicitly validate any overridden values, but we have the mechanism to do that.

Incidentally, @alexdewar, my son Thomas has just gotten quite into Exploding Kittens - the Beard Cat icon was instantly familiar 😄

virtual_rainforest/core/config.py
jacobcook1995: At the moment, none of the constants are populated by default in the config (i.e. not using default values set within model schema). It's only after config validation when a set of constants is used to create a model specific constants class that default values come in, These defaults are hardcoded into the various constants classes, and just aren't replaced if there is no value stored in the config to replace them with.

So I don't really see a danger the current approach, unless we change how default constants are handled

davidorme: The `config_merge` does indeed return the updated dictionary when conflicts are present:

```python
In [11]: from virtual_rainforest.core.config import Config, config_merge
In [12]: cfg = Config('tests/core/data/all_config.toml')
In [13]: updated, conflicts = config_merge(c, {'core': {'layers': {'soil_layers': 3}}})
In [14]: updated['core']['layers']['soil_layers']
Out[14]: 3
```

So, this should work as:

```python
    def override_config(self, extra_params) -> None:
        """Adjust an already created config to update settings

        Raises:
            ValidationError: if duplicate config definitions occur across files.
        """
        
        # Update the dictionary
        updated, conflicts = config_merge(self, extra_params, conflicts=tuple())
        
        # Ignore any conflicts - explicitly overriding settings - so now revalidate against the 
        # schema defined in the base config.
        self.update(updated)
        self.validate_config()
```

jacobcook1995: Obviously, if changing the grid size or another part of the config that does have a default setting through this method would be risky, but I don't think we would want to do that?

jacobcook1995:
Is there any real danger in terms of the validation? Because the new parameter gets added to the list of config files to validate, i.e. the change is made before any validation takes place

virtual_rainforest/core/config.py
davidorme: So configuration settings that are then checked against other inputs/data could cause all sorts of chaos. 

But this process should only (?) ever be invoked when the `Config` is built and validated and bad overrides would cause exactly the same errors as bad initial settings - they can be JSONSchema invalid and get trapped early - or bad/inconsistent in setting up the rest of the simulation and cause a later error. 

Exposing a `Config.override_config` does raise a route where programmatic use _could_ change the `Config` half way through a simulation (and we should very clearly say that this is extremely ill-advised!), but in `vr_run` it would only ever be something like:

```python
config = Config(toml_files)
config.override_config(extra_params)
# Now leave config alone!
```

jacobcook1995:
@davidorme I guess my view is that overriding existing config settings introduces a fair bit of complexity to this pull request without really addressing any immediate problem?

Like if down the line it becomes very annoying to have to ensure that options have not been populated in config files when doing sensitivity analysis then we can always take the approach you've suggested here, but I'm not sure there's any great need to address it now as a potential future problem (because there are an awful lot of those).

davidorme:
This method is a vast improvement on having to write multiple files to run variations for a given SA but I think having to create a separate compatible set of base config files for any given sensitivity analysis _is an immediate problem_.  If there isn't a danger to the override behaviour and if the resulting user experience is better, then I'm not sure why we don't just implement it, given that we already have the mechanisms in place to do so.

davidorme:
To give a concrete example:

* Let's say we run a big long analysis that saves a whole bunch of stuff with some non-default configured settings like `plants.beta_value = 0.5` and `soil.zeta_value=0.95`.
* Now we want to run an SA to see how much GPP varies if we change `plants.beta_value = range(0.4, 0.6, 0.01)` and `soil.zeta_value= range(0.85, 1.05, 0.1)`.
* Those SA runs have to have **different config files** to our base run, because they _cannot_ contain `plants.beta_value = 0.5` and `soil.zeta_value=0.95`.
* That is less clear and less reproducible.

virtual_rainforest/core/config.py
alexdewar: Good point. We're `deepcopy`ing it anyway, so it doesn't matter if it's mutable.

alexdewar:
@davidorme I take your point that that it could be useful to be able to override params so you can reuse config files between simulation runs. I think it would be a mistake to silently override the value the user has put in a config file, though, in case it was accidental; we should at least warn them.

I'm happy to have a go at this, but if it turns out to be finicky, then perhaps we could merge this as is and open a separate issue for the "allow users to override parameters" feature? My main goal atm is to get a minimal version of the SA working so that we can all have a discussion about it. Once that's done we can of course polish things.

davidorme:
That sounds good to me - we shouldn't lose the bigger SA picture - but I think this is worth fixing here if we can.

We can warn users - no harm in that - but the context of sensitivity analyses and also the documentation for `extra_params` should be very clear that we are adjusting configs. Maybe calling it `alter_params` or `override_params` to give a clear hint that this is what it will do!

davidorme (approved):
That implementation looks good to me. I'd thought of `Config.override_config` as something to call post `__init__` but embedding it in `__init__`  also makes sense. I briefly considered suggesting we make it 'private' as `Config._override_config` to further discourage use, but honestly none of those `Config` methods are expected to be used except via `__init__` in 99% of use cases.

I hadn't thought that the `conflicts` object provides an automatic list of altered settings - so that is pretty seamless for reporting changes. I guess the thing there is that the conflicts will be against things set in files and things set by default - but honestly the whole point of this mechanism is to make it easy to change settings, so I think that distinction is largely artificial and recording in logs _which_ settings have been changed is useful.

alexdewar:
Thanks @davidorme.

I think we could probably safely make all of those methods private and remove the `auto` parameter to `__init__`. That way we wouldn't have to worry about users using them at all. There aren't any current uses of them outside of `__init__` and anyone will want to invoke them directly anyway.

jacobcook1995 (approved):
LGTM!

PR #268, opened 2023-08-01
Adding a litter model

jacobcook1995:
# Description

This pull request adds a litter model to the Virtual Rainforest. The purpose of this model is to handle the decomposition of complex biomass lying on or just under the forest floor.

I made this a separate model from the `soil` model for a couple of reasons:

1. There are plausible use cases (mainly animal focused) where you would want to track the litter (as a resource for consumption) but not the full soil nutrient cycles.
2. There's an alternative microbially explicit approach to modelling litter decay that I would like to try (if I get the time). I feel like this would be easier to implement with the litter model separated from the soil model.

At the moment, the model only has two pools "above_ground_metabolic" and "above_ground_structural" which represent the fast and slow decomposing above ground litter, respectively. In future, below ground pools and wood pools will also be added. The model will also get quite a lot more complex when we add leaf chemistry in (lignin, N and P).

I originally planned to add links from this model to the soil model in, but I thought that pull request was already getting large enough, so decided to save that for a later pull request.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -188,6 +188,10 @@ def dummy_litter_data(layer_roles_fixture):
     # manner
     data["litter_pool_above_metabolic"] = DataArray([0.3, 0.15, 0.07], dims=["cell_id"])
     """Above ground metabolic litter pool (kg C m^-3)"""
+    data["litter_pool_above_structural"] = DataArray(
+        [0.5, 0.25, 0.09], dims=["cell_id"]
+    )
+    """Above ground structural litter pool (kg C m^-3)"""
 
     return data
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -193,6 +193,52 @@ def dummy_litter_data(layer_roles_fixture):
     )
     """Above ground structural litter pool (kg C m^-3)"""
 
+    data["soil_temperature"] = xr.concat(
+        [DataArray(np.full((13, 3), np.nan)), DataArray(np.full((2, 3), 20))],
+        dim="dim_0",
+    )
+    data["soil_temperature"] = (
+        data["soil_temperature"]
+        .rename({"dim_0": "layers", "dim_1": "cell_id"})
+        .assign_coords(
+            {
+                "layers": np.arange(0, 15),
+                "layer_roles": ("layers", layer_roles_fixture),
+                "cell_id": data.grid.cell_id,
+            }
+        )
+    )
+
+    data["air_temperature"] = xr.concat(
+        [
+            DataArray(
+                [
+                    [30.0, 30.0, 30.0],
+                    [29.844995, 29.844995, 29.844995],
+                    [28.87117, 28.87117, 28.87117],
+                    [27.206405, 27.206405, 27.206405],
+                ],
+                dims=["layers", "cell_id"],
+            ),
+            DataArray(np.full((7, 3), np.nan), dims=["layers", "cell_id"]),
+            DataArray(
+                [
+                    [22.65, 22.65, 22.65],
+                    [16.145945, 16.145945, 16.145945],
+                ],
+                dims=["layers", "cell_id"],
+            ),
+            DataArray(np.full((2, 3), np.nan), dims=["layers", "cell_id"]),
+        ],
+        dim="layers",
+    ).assign_coords(
+        {
+            "layers": np.arange(0, 15),
+            "layer_roles": ("layers", layer_roles_fixture[0:15]),
+            "cell_id": data.grid.cell_id,
+        }
+    )
+
     return data
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -113,10 +113,11 @@ def dummy_carbon_data(layer_roles_fixture):
     grid = Grid(cell_nx=4, cell_ny=1)
     data = Data(grid)
 
-    # The required data is now added. This includes the three carbon pools: mineral
-    # associated organic matter, low molecular weight carbon, and microbial carbon. It
-    # also includes various factors of the physical environment: pH, bulk density, soil
-    # moisture, soil temperature, percentage clay in soil.
+    # The required data is now added. This includes the four carbon pools: mineral
+    # associated organic matter, low molecular weight carbon, microbial carbon and
+    # particulate organic matter. It also includes various factors of the physical
+    # environment: pH, bulk density, soil moisture, soil temperature, percentage clay in
+    # soil.
     data["soil_c_pool_lmwc"] = DataArray([0.05, 0.02, 0.1, 0.005], dims=["cell_id"])
     """Low molecular weight carbon pool (kg C m^-3)"""
     data["soil_c_pool_maom"] = DataArray([2.5, 1.7, 4.5, 0.5], dims=["cell_id"])
@@ -173,9 +174,34 @@ def dummy_carbon_data(layer_roles_fixture):
 
 
 @pytest.fixture
+def dummy_litter_data(layer_roles_fixture):
+    """Creates a dummy litter data object for use in tests."""
+
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    # Setup the data object with four cells.
+    grid = Grid(cell_nx=3, cell_ny=1)
+    data = Data(grid)
+
+    # These values are taken from SAFE Project data, albeit in a very unsystematic
+    # manner
+    data["litter_pool_above_metabolic"] = DataArray([0.3, 0.15, 0.07], dims=["cell_id"])
+    """Above ground metabolic litter pool (kg C m^-3)"""
+
+    return data
+
+
+@pytest.fixture
 def top_soil_layer_index(layer_roles_fixture):
     """The index of the top soil layer in the data fixtures."""
     return next(i for i, v in enumerate(layer_roles_fixture) if v == "soil")
+
+
+@pytest.fixture
+def surface_layer_index(layer_roles_fixture):
+    """The index of the top soil layer in the data fixtures."""
+    return next(i for i, v in enumerate(layer_roles_fixture) if v == "surface")
 
 
 @pytest.fixture
--- a/virtual_rainforest/models/litter/litter_model.py
+++ b/virtual_rainforest/models/litter/litter_model.py
@@ -85,10 +85,6 @@ class LitterModel(BaseModel):
             LOGGER.error(to_raise)
             raise to_raise
 
-        self.data
-        """A Data instance providing access to the shared simulation data."""
-        self.update_interval
-        """The time interval between model updates."""
         self.constants = constants
         """Set of constants for the litter model"""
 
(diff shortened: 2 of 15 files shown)


tests/conftest.py
```diff
@@ -172,12 +173,87 @@ def dummy_carbon_data(layer_roles_fixture):
...
 
+@pytest.fixture
+def dummy_litter_data(layer_roles_fixture):
+    """Creates a dummy litter data object for use in tests."""
+
+    from virtual_rainforest.core.data import Data
+    from virtual_rainforest.core.grid import Grid
+
+    # Setup the data object with four cells.
+    grid = Grid(cell_nx=3, cell_ny=1)
```
alexdewar, line 184: Would it be worth having `cell_ny > 1` here? I can imagine a case where the code only works with `cell_ny == 1`.

virtual_rainforest/models/litter/litter_model.py
```diff
@@ -0,0 +1,178 @@
...
+            to_raise = InitialisationError(
+                "Initial litter pools contain at least one negative value!"
+            )
+            LOGGER.error(to_raise)
+            raise to_raise
+
+        self.data
+        """A Data instance providing access to the shared simulation data."""
+        self.update_interval
+        """The time interval between model updates."""
```
alexdewar, line 91: Type hints?

virtual_rainforest/models/litter/litter_model.py
```diff
@@ -0,0 +1,178 @@
...
+            update_interval: Frequency with which all models are updated
+        """
+
+        # Find number of soil and canopy layers
+        soil_layers = config["core"]["layers"]["soil_layers"]
+        canopy_layers = config["core"]["layers"]["canopy_layers"]
+
+        # Check if any constants have been supplied
+        if "litter" in config and "constants" in config["litter"]:
+            # Checks that constants is config are as expected
```
alexdewar, line 130:
```suggestion
            # Checks that constants in config are as expected
```

virtual_rainforest/models/litter/litter_model.py
```diff
@@ -0,0 +1,178 @@
...
+        Args:
+            time_index: The index representing the current time step in the data object.
+        """
+
+        # Find all litter pools
+        litter_pools = {
+            str(name): self.data[str(name)].to_numpy()
+            for name in self.data.data.keys()
+            if str(name).startswith("litter_pool_")
+        }
```
alexdewar, line 162: Why are you casting `name` to `str`? Aren't the keys always strings?

virtual_rainforest/models/litter/litter_pools.py
```diff
@@ -0,0 +1,162 @@
...
+    constants: LitterConsts,
+) -> dict[str, DataArray]:
+    """Calculate updates for all litter pools.
+
+    Args:
+        surface_temp: Temperature of soil surface, which is assumed to be the same
+            temperature of the above ground litter [C]
+        litter_pool_above_metabolic: Above ground metabolic litter pool [kg C m^-2]
+        litter_pool_above_structural: Above ground structural litter pool [kg C m^-2]
+        update_interval: Interval that the litter pools are being updated for [days]
```
alexdewar, line 31: I remember being involved in some discussion about this a while back... but did you guys arrive at an agreement about using a common time unit throughout?

virtual_rainforest/models/soil/carbon.py
```diff
@@ -1,7 +1,7 @@
 """The ``models.soil.carbon`` module  simulates the soil carbon cycle for the Virtual
-Rainforest. At the moment only two pools are modelled, these are low molecular weight
-carbon (LMWC) and mineral associated organic matter (MAOM). More pools and their
-interactions will be added at a later date.
+Rainforest. At the moment four pools are modelled, these are low molecular weight carbon
+(LMWC), mineral associated organic matter (MAOM), microbial biomass, and particulate
+organic matter (POM).
```
alexdewar, line 4: I'm assuming this is just a correction based on the existing soil module code?

alexdewar (approved):
I've got some minor queries/suggestions, but otherwise LGTM!

tests/conftest.py
davidorme: It actually should be fine, because the 'grid' is always a one-dimensional array axis indexed by `grid_id`.

virtual_rainforest/models/litter/litter_pools.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+    return new_litter_pools
+
+
+def calculate_temperature_effect_on_litter_decomp(
+    temperature: NDArray[np.float32],
+    reference_temp: float,
+    offset_temp: float,
+    temp_response: float,
```
davidorme, line 82: Should these default to the `LitterConsts` defaults?

```suggestion
    reference_temp: float = LitterConsts.litter_decomp_reference_temp,
    offset_temp: float = LitterConsts.litter_decomp_offset_temp,
    temp_response: float = = LitterConsts.litter_decomp_temp_response,
```

davidorme (approved):
Not a lot to add to @alexdewar. LGTM

virtual_rainforest/models/litter/litter_model.py
jacobcook1995: If I don't cast them I get a `mypy` error because apparently `name` is of type `Hashable` when the dictionary expects the key to be of type `str`. There might be a smarter way around this though?

virtual_rainforest/models/litter/litter_model.py
jacobcook1995: I actually think these are redundant anyway. Like both are supplied to the `super().__init__()` above so I'm not sure why they are even included here (type hints for both come through the arguments to this `__init__` function

alexdewar: True.

```suggestion
```

virtual_rainforest/models/soil/carbon.py
jacobcook1995: Yeah I added some updates to the soil module documentation, which I maybe should have done as a separate pull request

alexdewar: Nah, I think it's fine. Just wanted to check it wasn't an accident.

virtual_rainforest/models/litter/litter_model.py
davidorme: I assume you've done it like this because the data being used by in the Litter Pools might expand. It does seem odd to make a kind of 'mini-Data' object though. I personally would probably explicitly unpack the expected variables into `calculate_litter_pool_updates` and skip this completely:

```python
updated_litter_pools = calculate_litter_pool_updates(
        surface_temp=self.data["air_temperature"][
            self.surface_layer_index
        ].to_numpy(),
        constants=self.constants,
        update_interval=self.update_interval.to("day").magnitude,
    litter_pool_above_metabolic= self.data['litter_pool_above_metabolic'].to_numpy(),
    litter_pool_above_structural=self.data['litter_pool_above_metabolic'].to_numpy(),
)
```

Alternatively, you could pass in `data = self.data` to keep the API simpler, but I think an explicit list of inputs is cleaner.

davidorme: I mean it's not like the function is going to do anything with `data['litter_pools_fiz_buz`]` apart from raise `TypeError: calculate_litter_pool_updates() got an unexpected keyword argument`

virtual_rainforest/models/litter/litter_pools.py
jacobcook1995: I moved away from doing that with the new constants system. It seemed a bit easy for my liking to forget to give all the arguments to the function and then inadvertently supply the default as defined in as defined in the original data class (apresuming that `LitterConsts.constant_name` supplies that default). An oversight like this would mean that sensitivity analysis wasn't actually changing the parameter that it was claiming to change

virtual_rainforest/models/litter/litter_pools.py
```diff
@@ -0,0 +1,162 @@
...
+    ) * update_interval
+
+    # Construct dictionary of data arrays to return
+    new_litter_pools = {}
+    new_litter_pools["litter_pool_above_metabolic"] = DataArray(
+        litter_pool_above_metabolic + change_in_metabolic_above, dims="cell_id"
+    )
+    new_litter_pools["litter_pool_above_structural"] = DataArray(
+        litter_pool_above_structural + change_in_structural_above, dims="cell_id"
+    )
```
davidorme, line 73:
```suggestion
    new_litter_pools = {
    "litter_pool_above_metabolic":  DataArray(
        litter_pool_above_metabolic + change_in_metabolic_above, dims="cell_id"
    ),
    "litter_pool_above_structural": DataArray(
        litter_pool_above_structural + change_in_structural_above, dims="cell_id"
    )
    }
```

virtual_rainforest/models/litter/litter_pools.py
jacobcook1995: This might work different with a constants registry though?

davidorme: I sort of see that, but it does seem extremely strange to have an explicit set of defaults for a function like this and then not to connect those up so that they get used in the function. I need to think about this one but it doesn't feel right.

virtual_rainforest/models/litter/litter_pools.py
jacobcook1995: I don't know if we ever reached an agreement on common time units, we agreed that all models would use the same time step for updates (i.e. that no model should update the data object faster than the other models). I don't think we've settled on common units for anything, though we probably need to as part of sorting #129

alexdewar: That's fine. Just wanted to check. You've been explicit about what your time units are, which is all you really can do in this case.

virtual_rainforest/models/litter/litter_pools.py
jacobcook1995: The alternative feels strange as well though as you are providing default values that should never be used, unless the functions are called individually outside the main simulation flow. I can't see this happening much for this level of function (ones that calculate individual variables for a specific model)? Even for plotting simulation results you probably want to make sure you supply the constants used by the simulation, rather than the defaults

virtual_rainforest/models/litter/litter_model.py
jacobcook1995: Yeah this is copied over from the `soil` module where there's more complex ordering of pools going on (i.e. into a 1D `numpy` array and back again). Kept it in because it saved a small amount of effort, but that's not really a good enough reason to keep it

virtual_rainforest/models/litter/litter_pools.py
jacobcook1995: Yeah think this works best as an issue, as I think we need to decide on a consistent approach to this across models and then document it

TaranRallings (approved):
LGTM!

PR #269, opened 2023-08-01
Treat paths in config files as relative to config dir

### LINKED ISSUE #264: In config files, paths are relative to CWD
Really, they need to be relative to the config file itself, otherwise you have to `cd` to the correct dir before invoking `vr_run` in order to make things work as expected.

I'll give an example. Currently I have a config file with this in:

```toml
[[core.data.variable]]
file = "dummy_data/ERA5_land_dummy.nc"
var_name = "air_temperature_ref"
```

The `ERA_land_dummy.nc` file is in the same directory as the config file, but I also have to put in the name of the subfolder (`dummy_data`) in order for VR to be able to find it. Note that if `vr_run` is invoked from a different folder (e.g. the user runs `vr_run some/parent/folders/dummy_data`) it won't work at all.

This makes it quite annoying to share datasets with people, as they will either have to a) change a bunch of paths in the config files or b) make sure the path on their machine (either relative or absolute) is the same. Both of these options are a bit grim.

This is of particular interest to me as I want to be able to invoke `vr_run` from an arbitrary location as part of my work on #239.

alexdewar:
# Description

In some places in config files, paths to other (netCDF) files are used, e.g.:

```toml
[[core.data.variable]]
file = "dummy_data/ERA5_land_dummy.nc"
var_name = "air_temperature_ref"
```

Currently these paths can be absolute or relative. The problem with the relative paths is that they are relative to wherever VR is being invoked from (i.e. the process's CWD) rather than the config file itself. This means that users have to always invoke `vr_run` from the same path relative to the config file otherwise things will break or manually edit the config file. Among other things, this presents difficulties in sharing data.

This PR fixes this problem by adding another step to the config loading which looks for file paths in the configuration and, if they are relative paths, adjusts them to be relative to the config file they were loaded from. This step happens after loading the TOML files, but before validation.

At present, this involves looking for keys called `"file"` as these were the only examples of file paths in the config files I could find. Admittedly, this is a bit of a hack, but unfortunately there are a limited number of ways we can solve this problem as fixing up the file paths has to occur *before* the various config dicts are merged together, so that we still know which config file each value came from (different config files may be in different folders and so their relative paths will have a different root).

I haven't written tests for this yet, as I wanted to get feedback first, but can do if people are happy with the implementation.

Fixes #264.

## Type of change

(Though technically this is a breaking change!)

## Key checklist

## Further checks

(No tests added yet, but can do if people are happy with the implementation.)

### DIFF
--- a/tests/core/test_config.py
+++ b/tests/core/test_config.py
@@ -6,7 +6,9 @@ test that a complete configuration file passes the test, which will have to be k
 to date.
 """
 
+import sys
 from contextlib import nullcontext as does_not_raise
+from itertools import repeat
 from logging import CRITICAL, ERROR, INFO, WARNING
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


jacobcook1995 (approved):
LGTM!

I don't know if the iteration is needed actually, as I think we were planning on supplying all file paths within the `[[core.data.variable]]` table of tables. But I guess it's easier to have a general solution in place in case we change our minds

davidorme (approved):
Hmmm. This is a pain to do cleanly, isn't it. Have to admit I don't love the `file` test.

Having said that, the other way I can see is to introduce an `isPath = true` property on appropriate JSONSchema entries. That way, the code can unambiguously identify paths that need to be made absolute using an existing canonical source of validation information. 
However, I think that then requires two passes and a new Validator:

1. Merge the original config content to identify the modules and hence build the schema.
2. Run a PathResolverValidator using that schema over the individual files in `self.toml_contents`: this validates contents but also look for entries with `isPath=true` and updates the value with the absolute paths of the `toml_contents` keys. It doesn't add defaults though, because has to happen on the last validation pass to avoid creating clashes.
3. Now merge the updated `toml_contents` again and then run this through the current default filling validation.

This is a lot more complex but is also explicit about which inputs are paths. That cutdown Validator

alexdewar:
Done. I've also added tests.

jacobcook1995 (approved):
LGTM!

davidorme (approved):
LGTM too

dalonsoa (approved):
LGTM!

Issue #278, opened 2023-08-03
Importing `virtual_rainforest` package always causes models to be loaded


alexdewar:
Currently if you import the `virtual_rainforest` package, all models are loaded (which is slow) and lots of log output is produced, even though the models are not necessarily needed. For example, if you run `vr_run --help`, you get this output.

This isn't currently a problem (though it would be nice if the output of `vr_run --help` weren't so ugly), but we may wish to add additional commands to `vr_run` or make `virtual_rainforest` usable as a library in some way, in which case this will become annoying.

davidorme:
The intention behind this was to auto discover all models and make sure registries are populated and schema are discovered. There is very little that can be done with the model until that has happened, so sticking it in `__init__.py` made a lot of sense, but as the suite of tools and entry points we support increases, that is going to be happening prematurely.

I think we basically need to wrap this step up into a `discover_models` function - but having that as a mandatory step before you can run a model also seems a bit clunky. 

The immediate logical place to put that step might be within the `Config` set up:

- You _have_ to have a `Config` object to run a model.
- You have to have loaded all the models to validate the configuration.
- It delays model loading until a step that logically precedes running a model.

So if this happens automatically as a method of `Config`? The thing I can't remember is whether the model tests necessarily go through the config process? The sequence is awkward.

alexdewar:
That makes sense.

Would it work to have the model discovery step be done explicitly in `vr_run()`? In the short term, it may be sufficient to just move various `import`s into the body of `vr_run()`.

davidorme:
We could do that but I think it would be better if the model loading happened as a part of standard operation - like creating the `Config` - rather than as part of a longer recipe. That is primarily so that programmatic use of the library is more logical/cleaner. Short term, we could use a `vr_run` recipe, but I don't think that's a natural place for it.

davidorme:
That makes sense but I was wondering if we can make SCHEMA_REGISTRY itself a bit smarter and import modules? The advantage to that is that we can just use `SCHEMA_REGISTRY["my_model"]` to trigger the import.  I think the argument for this is that then we don't even have to create a Config to ensure that a model is loaded, which I think might be useful in some of the tests?

Something like this? I haven't tested it at all 😄 

```python
class SchemaRegistry(dict):

    def __getitem__(self, __key: Any) -> Any:

        if __key in self:
            return super().__getitem__(__key)

        try:
            import_module(f"virtual_rainforest.models.{__key}")
[...]

PR #279, opened 2023-08-03
Add example data

### LINKED ISSUE #265: Add example config files to the repo
As part of my work on #239, I want to be able to produce an example script (or snakemake file) to show how to run a SA. The problem is that the data is not part of this repo (and contains hardcoded paths; see #264) and ideally I don't want to produce an example that requires getting data files from elsewhere.

Given that the data doesn't seem sensitive, it might make sense to just add these config files to this repo. Yes, there are some binary files in there, but they are tiny, so the filesize isn't worth worrying about IMO.

It's nice for users to be able to run an example directly from a repo directly, without having to try to obtain the data from elsewhere or construct their own based on the docs.

alexdewar:
# Description

This PR adds some example data as a subpackage inside `virtual_rainforest`, as suggested by @davidorme. This allows end users to try VR out without having to construct their own dataset or get access to the RDS. This does include some binary files, though they are small in size (< 100Kb total).

I based the dataset on the one at `/rds/project/virtual_rainforest/live/dummy_data`, with the paths in config files changed to be relative (now that #269 is merged).

I added a new flag to `vr_run` (`--example`) to allow users to run a simulation with the example data. I'm not entirely sure about this solution -- it would be nicer to have users just point `vr_run` at wherever the data is located -- but it might not be easy to find if e.g. VR was installed via `pip`. Suggestions welcome.

There is also now an `example_data_path` member in the `virtual_rainforest` package for convenience, though if you import it then all the models will also be loaded (see #278).

Closes #265.

## Type of change

## Key checklist

## Further checks


davidorme:
I was thinking more of having a completely parallel package within the same repo, rather than embedding it within `vr`. So that we would have a separate package structure, an extra entry in `pyproject.toml` and then have `import virtual_rainforest_example_data as vr_data` or use `importlib.resources`.

This search page shows how the example from `pyrealm` is structured through the repo:
https://github.com/search?q=repo%3AImperialCollegeLondon%2Fpyrealm%20pyrealm_build_data&type=code

jacobcook1995 (approved):
LGTM!

Just had a query, how much of a problem will updating those binary files be. Like I know we don't want to be changing them super frequently, but they will need to be altered every time model inputs are changed, which is still going to happen with reasonable (weekly?) frequency. Is the best approach just not to worry and to update them every time a new model input gets added (or removed or changed)?

davidorme:
We don't want to update them too much, because you end up with complete copies of each binary file in the repo history. The files aren't huge, so not a big deal, but it wouldn't take too many updates to bloat the repo.

One thing here might be to develop the `csv` reader to allow example data with incremental not binary updates - but that is a bit of a sledgehammer/nut solution.

dalonsoa:
From time to time, you can purge the repo from binary files in the history if it becomes too big.

dalonsoa (approved):
LGTM!

alexdewar:
@davidorme I hear what you're saying, but I'm not sure that we want the data in a separate repo (for now), for a couple of reasons.

1. Downloading an additional dataset repo is an extra (unnecessary) job to do before users can actually try out VR. For the sake of the minimal amount of data savings, it's just not worth it, and just makes the project less accessible
2. VR is currently under very active development and there's nothing like a stable format for data files atm. This means that the dataset repo will also need to be regularly updated in order to be made compatible with the latest version of VR and it means that if you want to try VR with the example data you'll have to make sure you get *exactly* the right version of the data for *exactly* the right version of VR or it just won't work

If VR had been around for years and there was a stable data format and you had, e.g. 1GB of example model data, then it might make sense to store that somewhere else (e.g. using git LFS). Even then though, I think it's still nice to have a minimal example dataset to go along with code, so that users can test it out.

Yes, you shouldn't be committing loads and loads of binary data to your git repo as a matter of course, but it's fine to have *some* small binary files in there, even if they change semi-regularly. People routinely use git to store websites, even where there are image files in there that change all the time. My copy of the VR repo is currently at ~5MB. Even if it ends up being >100MB a few years down the line, that's honestly totally fine too. Let's not make our lives unnecessarily difficult here.

davidorme:
I agree that the binary files here are not likely to be an issue - mostly responding to @jacobcook1995's query 😄

I take your point about keeping the data in sync while `vr` is in alpha/beta. Ultimately I'm not bothered about the minor increase in the `virtual_rainforest` package size and just think the separate package is cleaner, but that shouldn't stop us doing this now. I think it is worth adding a long range data package issue to revisit this at some point.

alexdewar:
@davidorme Sounds good to me.

PR #280, opened 2023-08-03
Systematise various output file parameters

### LINKED ISSUE #271: Allow for easily setting the output path for simulation files
Currently, VR produces a number of output files for a simulation and allows the user to choose the paths for these files like so:

```toml
[core.data_output_options]
save_initial_state = true
save_continuous_data = true
save_final_state = true
out_path_initial = "./initial_state.nc"
out_folder_continuous = "."
continuous_file_name = "all_continuous_data"
out_path_final = "./final_state.nc"
```

(These are just the default values.)

In addition to these, there is also the path to the merged config file, which is not settable via a parameter (see #270). Note that in order to make sure that all files are output to the same folder, the user has to set a bunch of paths explicitly. Also note that `continuous_file_name` is unique in a) not including a path (because the value of `out_folder_continuous` is used) and b) not including a file extension (`.nc` is used). In future, in the event that new output file types are added to VR, the user will presumably have to supply the whole paths for these too. I'm using different output folders for different simulations as part of my work on #239 and I'm finding the current mechanism rather cumbersome.

I propose that we make the output folder a separate parameter, so that the user can just set this once in order to get all the output files in one place. Furthermore, I suggest that we let the user choose the file names for the initial and final state files, but not the whole path. (We should also probably just put `all_continuous_data.nc` in the output folder too.) This means users would lose the ability to put the initial and final state files into different folders (for example), but this is a rather peculiar thing to want to do in any case, so I don't think it's much of a loss. We *may* want to keep `out_folder_continuous` as a separate option, in case e.g. the user wants to write these files to `/tmp`, but in any case it should probably default to the common output folder too (otherwise you get file conflicts if you're trying to run multiple simulations from the same folder at the same time).

What does everyone think?

alexdewar:
# Description

This PR systematises the various output file parameters, as discussed in #271.

Commit summary:

- Add a `core.data_output_options.out_path` parameter, which represents the output folder for all output files
- Replace `out_path_initial` and `out_path_final` with `out_initial_file_name` and `out_final_file_name`, respectively. They are now file names rather than full paths
- Replace `continuous_file_name` with `out_continuous_file_name` for consistency and don't add `.nc` extension to name in VR now
- Add an `--outpath` flag to `vr_run` to set the output path

Fixes #271.

## Type of change

## Key checklist

## Further checks


alexdewar:
NB: This will cause merge conflicts with #275, so we should merge that first, then I'll fix up the conflicts on this branch.

alexdewar:
While we're here, I think some of these options could be simplified further.

Possibilities:

- Drop `save_initial_state` and `save_output_state` options and just always save these files
- **or** still give users the choice of whether they want to save these files, but instead of having separate boolean flags for this, just make `out_{initial,final}_file_name` optional and then don't write the files if the parameters aren't given
- Don't bother letting the user change these file names (when would you need to?) and just use default values of `initial_state.nc` etc.

virtual_rainforest/core/core_schema.json
```diff
@@ -132,36 +132,42 @@
...
-                     "default": "./initial_state.nc"
+                     "default": "."
+                  },
+                  "out_initial_file_name": {
+                     "description": "File name for initial state output file",
+                     "type": "string",
+                     "default": "initial_state.nc",
+                     "pattern": "^[^/\\\\]+$"
                   },
                   "out_folder_continuous": {
```
jacobcook1995, line 146: Did you mean to keep this option? If you did why was the default option for it removed?

virtual_rainforest/core/core_schema.json
```diff
@@ -132,36 +132,42 @@
...
+                     "pattern": "^[^/\\\\]+$"
                   }
                },
                "default": {},
                "required": [
                   "save_initial_state",
                   "save_continuous_data",
                   "save_final_state",
-                  "out_path_initial",
-                  "out_folder_continuous",
```
jacobcook1995, line 162: "out_folder_continuous" shouldn't be removed from here if it's still being used

jacobcook1995 (approved):
LGTM!

jacobcook1995:
Definitely seems sensible to me to remove the options to change the file names. Like the only use case I could see would be if you were saving to a folder that already has files with those names it. But that seems like very risky behaviour that should be discouraged. (or maybe could be done as part of an automated workflow, but shouldn't be done manually).

At some point the initial and final files might get quite large as they both contain input climate data for the entire run. I'm not sure how large though (Vivienne would know but she's on holiday this week). I don't know if this would be a reason to not save them by default though? At present this data is duplicated between the files, i.e. both initial and final files contain the full climate input data for the simulation.

I don't think we would want the saving of the initial and final state to be controlled by whether or not a name was set, as this feels quite easy to forget and to end up with missing data preventing analyses from being carried out.

dalonsoa (approved):
It looks good, but like @jacobcook1995 , I'm a bit confused about `out_folder_continuous`, if it is needed, or not, or if it is used at all.

alexdewar:
> Definitely seems sensible to me to remove the options to change the file names. Like the only use case I could see would be if you were saving to a folder that already has files with those names it. But that seems like very risky behaviour that should be discouraged. (or maybe could be done as part of an automated workflow, but shouldn't be done manually).

Agreed!

> At some point the initial and final files might get quite large as they both contain input climate data for the entire run. I'm not sure how large though (Vivienne would know but she's on holiday this week). I don't know if this would be a reason to not save them by default though? At present this data is duplicated between the files, i.e. both initial and final files contain the full climate input data for the simulation.

Maybe this is a reason to allow users not to save one or both of these files (at least for now). Longer term, maybe it would be better to come up with some way of not duplicating large amounts of data between the initial and final state files?

> I don't think we would want the saving of the initial and final state to be controlled by whether or not a name was set, as this feels quite easy to forget and to end up with missing data preventing analyses from being carried out.

You're probably right that this approach would be too subtle and error prone for users.

alexdewar:
`out_folder_continuous` currently *is* still used. While VR is running, it continually saves the simulation state in files called `continuous_state00001.nc` etc. in the `out_folder_continuous` folder. Once the simulation is completed, all these files are merged into one (which is now saved in the output folder with everything else).

I've changed the code so that if the user doesn't specify `out_folder_continuous` (the usual case), these temporary files are saved in the output path too, hence why `out_folder_continuous` is no longer a required parameter. The reason why we don't want to just save these files into the current directory (which was the default behaviour before) is because if you have multiple simulations running at the same time they'll all be trying to write files with the same name to the same folder at the same time, unless you do some nasty gymnastics.

vgro:
> Definitely seems sensible to me to remove the options to change the file names. Like the only use case I could see would be if you were saving to a folder that already has files with those names it. But that seems like very risky behaviour that should be discouraged. (or maybe could be done as part of an automated workflow, but shouldn't be done manually).
> 
> At some point the initial and final files might get quite large as they both contain input climate data for the entire run. I'm not sure how large though (Vivienne would know but she's on holiday this week). I don't know if this would be a reason to not save them by default though? At present this data is duplicated between the files, i.e. both initial and final files contain the full climate input data for the simulation.
> 
> I don't think we would want the saving of the initial and final state to be controlled by whether or not a name was set, as this feels quite easy to forget and to end up with missing data preventing analyses from being carried out.

Depending on the length of the simulation and the temporal resolution, the input climate data could get quite large. I don't think we need to write the full time series in initial and final state because the user has the data stored somewhere; it might make more sense to save the variables that are created in the first time step, like the first vertical temperature profile.

jacobcook1995:
> > Definitely seems sensible to me to remove the options to change the file names. Like the only use case I could see would be if you were saving to a folder that already has files with those names it. But that seems like very risky behaviour that should be discouraged. (or maybe could be done as part of an automated workflow, but shouldn't be done manually).
> > At some point the initial and final files might get quite large as they both contain input climate data for the entire run. I'm not sure how large though (Vivienne would know but she's on holiday this week). I don't know if this would be a reason to not save them by default though? At present this data is duplicated between the files, i.e. both initial and final files contain the full climate input data for the simulation.
> > I don't think we would want the saving of the initial and final state to be controlled by whether or not a name was set, as this feels quite easy to forget and to end up with missing data preventing analyses from being carried out.
> 
> Depending on the length of the simulation and the temporal resolution, the input climate data could get quite large. I don't think we need to write the full time series in initial and final state because the user has the data stored somewhere; it might make more sense to save the variables that are created in the first time step, like the first vertical temperature profile.

Ahh yeah that makes sense!

I guess for now we could just change the option to save initial and final states by default, and then make a separate issue for not saving time series input data in the initial state and final state?

alexdewar:
> I guess for now we could just change the option to save initial and final states by default, and then make a separate issue for not saving time series input data in the initial state and final state?

That sounds sensible, though I'd probs do this in a separate PR (which I'm happy to open).

@vgro That's interesting. I'm wondering if maybe the approach of writing out the entire simulation's data store at every timestep (which includes everything, even immutable data) is the right one. Maybe we could just add an option to the schemas to tag the variables that we actually *want* to be saved at every timestep and then just write those? Similarly for the final state file, we could only write these values and then if e.g. the user wants to find the value for this large climate data set, they could look in the initial state file.

vgro:
> > I guess for now we could just change the option to save initial and final states by default, and then make a separate issue for not saving time series input data in the initial state and final state?
> 
> That sounds sensible, though I'd probs do this in a separate PR (which I'm happy to open).
> 
> @vgro That's interesting. I'm wondering if maybe the approach of writing out the entire simulation's data store at every timestep (which includes everything, even immutable data) is the right one. Maybe we could just add an option to the schemas to tag the variables that we actually _want_ to be saved at every timestep and then just write those? Similarly for the final state file, we could only write these values and then if e.g. the user wants to find the value for this large climate data set, they could look in the initial state file.

I think we started with just writing everything out keeping in mind that users might only be interested in certain variables. I'm not sure what the current status is.
@jacobcook1995 isn't there a section in each model that defines the output variables?

jacobcook1995:
In terms of the continuous data, only the variables that models are defined as updating get written out (basically to avoid writing out all the climate inputs at every time step). That's relatively crude at the moment, and certainly be refined. But it's only for the initial and final saves where everything in the data object gets output

alexdewar:
Ahh ok. That makes sense.

Maybe these are the only vars we want to write to final state too... What does everyone think?

jacobcook1995:
If we do that it probably renders saving the initial and final state redundant if continuous output is saved, as the continuous file will contain everything saved in the initial and final file. Don't know if this is actually a problem though?

alexdewar:
True. I don't know either.

PR #283, opened 2023-08-10
Feature/update hydro model

vgro:
After discussion the limitations of soil parameterizations, I decided to remove the vertical flow from the soil moisture update and only keep the function for a 'mean flow rate' with lots of warnings in the comments (potentially too many repetitions). We can find a solution later together with the soil and plant model.
I also simplified the code a little bit by calculating intermediate variables that are used several times at the start, and I corrected how available capacity is calculated (sum of layers rather than mean)

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
Looks sensible to me!

PR #284, opened 2023-08-11
Feature/animal module

TaranRallings:
# Description

This is a simple implementation of `AnimalModel.update()` with required handler functions and tests. It is bare bones but it gets us closer to having animals in `vr_run` and hooking the animal module into other modules.

Most of the new `AnimalCommunity` level methods take an `AnimalCohort` level method and run it for every cohort in the community. The `AnimalModel` level method takes these community level methods and runs them for every community in the model. At both levels this is overly simplistic but it gets us started.

I have also added the `birth_mass` and `adult_mass` functional traits required to populate the model with cohorts and run the `birth()` process correctly.

One of the next steps will be adding a basic reproduction system behind `birth()` to get us out of this flawless exponential growth regime.

If I have done anything mad or if there is a better first-pass way to manage this two-level looping, please let me know! 

*EDIT* 
On further consideration, I'm going to rework reproduction before I pull this through so that we can actually do vr_run stuff without creating unstoppable rat armies. May as well hold off review until that is done. 

*ADDITIONAL EDIT*
The birth rework is done and it should now be able to run without exploding. Please review.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM! (though bear in mind that I haven't done proper OOP in years)

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -94,7 +95,9 @@ def from_config(
 
         functional_groups = []
         for k in functional_groups_raw:
-            functional_groups.append(FunctionalGroup(k[0], k[1], k[2], k[3]))
+            functional_groups.append(
+                FunctionalGroup(k[0], k[1], k[2], k[3], k[4], k[5])
```
dalonsoa, line 99: Assuming that checking that the number of columns is the right one has been done somewhere else, then this could be simplified as:
```suggestion
                FunctionalGroup(*k)
```
This will handle possible future situations when the number of arguments changes.

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -110,11 +113,43 @@ def spinup(self) -> None:
...
+        - birth_community
+        - metabolize_community
+        - increase_age_community
+        - mortality_community
+
+
+        """
+        for community in self.communities.values():
+            method = getattr(community, community_method)
+            method(*args, **kwargs)
```
dalonsoa, line 155: I'm not sure of the reason for this implicit application of methods. Why not simply:
```suggestion
        for community in self.communities.values():
            community.forage_community()
            community.migrate_community()
            community.birth_community()
            community.metabolize_community(timedelta64(1, "D"))
            community.mortality_community()
            community.increase_age_community(timedelta64(1, "D"))

    def cleanup(self) -> None:
        """Placeholder function for animal model cleanup."""
```
Or even simpler, add an `update` method to the community class that calls all of the above and this is simply left as:

```python
        for community in self.communities.values():
            community.update(timedelta64(1, "D"))
```

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -110,11 +113,43 @@ def spinup(self) -> None:
         """Placeholder function to spin up the animal model."""
 
     def update(self, time_index: int) -> None:
-        """Placeholder function to solve the animal model.
+        """Function to step the animal model through time.
+
+        Currently this is a toy implementation.
```
dalonsoa, line 118: Even being a toy implementation, the implicit call of methods I point out seems odd - and unnecessary.

TaranRallings: My thinking was `apply_community_method` as a placeholder for what will be a variety of community-method specific model-level handlers. It's absolutely not necessary right now but the implicit approach may become necessary once I am implementing multi-grid occupancy and issues like "how does dispersal work?" become more complicated. If it's a problem in the interim I can always simplify right now and rebuild the specific ones I need down the road.

virtual_rainforest/models/animals/animal_model.py
davidorme: It's more robust though to provide these arguments as a dictionary and use `**k`.  That way the order and set of arguments in the signature can change. Don't know how easy it is to change `k` to be a `dict` - and I suspect they're might be performance implications. Thoughts?

dalonsoa: I doubt it - they are very few arguments to unpack, so overhead should be really minimal.

Keeping that aside, using kwargs would be more robust, that is true.

PR #285, opened 2023-08-14
Added three more litter pools to litter model

jacobcook1995:
# Description

This pull request adds three pools to the litter model. These are a `woody` pool for deadwood, a `below_ground_metabolic` pool for rapidly decaying below ground litter (e.g. dead fine roots), and a `below_ground_structural` pool for more recalcitrant below ground litter (e.g. coarse root debris). This is basically the full set of pools that I intend to add for the standard litter model, though in future it needs to track litter chemistry (i.e. lignin) and stoichiometry.

Below ground litter decay rates are impacted by the soil water potential, which the hydrology model doesn't calculate at present. I added a function to estimate soil water potential from soil moisture. This is very much a stop gap solution until we decide how this is going to be handled in the plant model as well, but works well enough for the time being.

## Type of change

## Key checklist

## Further checks


tests/conftest.py
```diff
@@ -212,6 +220,25 @@ def dummy_litter_data(layer_roles_fixture):
...
     )
 
+    # The layer dependant data has to be handled separately
+    data["soil_moisture"] = xr.concat(
+        [
+            DataArray(np.full((13, 3), np.nan), dims=["layers", "cell_id"]),
+            # At present the soil model only uses the top soil layer, so this is the
+            # only one with real test values in
+            DataArray([[0.5, 0.7, 0.6]], dims=["layers", "cell_id"]),
+            DataArray(np.full((1, 3), 0.1), dims=["layers", "cell_id"]),
```
vgro, line 230: Why/how did you set the value for the lowest soil layer?

tests/models/litter/test_litter_model.py
```diff
@@ -272,4 +338,28 @@ def test_update(litter_model_fixture, dummy_litter_data):
...
+
+def test_convert_soil_moisture_to_water_potential(
+    dummy_litter_data, top_soil_layer_index
+):
+    """Test that function to convert soil moisture to a water potential works."""
+    from virtual_rainforest.models.litter.litter_model import (
+        convert_soil_moisture_to_water_potential,
+    )
+
+    expected_potentials = [-568.76270, -50.105649, -152.49040]
```
vgro, line 357: These values seem a bit low, did you check that you used the right units for the result to be in kPa? I thought soil water potential was more in the range of 0 to -200 or so for soil moisture above wilting point, but I might be wrong

virtual_rainforest/models/litter/constants.py
```diff
@@ -43,6 +61,25 @@ class LitterConsts:
     Value is taken from :cite:t:`kirschbaum_modelling_2002`.
     """
 
+    litter_decay_constant_woody: float = 1.0 / 150.0
+    """Decay constant for the woody litter pool [day^-1].
+
```
vgro, line 66: Why did you chose to provide the values as fractions?

virtual_rainforest/models/litter/constants.py
```diff
@@ -73,3 +134,48 @@ class LitterConsts:
...
+    """Water potential at which soil pores begin to aerate [kPa].
+
+    The constant is used to estimate soil water potential from soil moisture. As this
+    estimation is a stopgap this constant probably shouldn't become a core constant. The
+    value is the average across all soil types found in
+    :cite:t:`cosby_statistical_1984`. In future, this could be calculated based on soil
+    texture.
+    """
+
+    water_retention_curvature: float = -7.22
```
vgro, line 175: I agree that this is an important issue, we should definitely make sure that we coordinate these calculations across models. It might be that some of the hydrology moves to soil or that this functionality moves to the hydrology model. Or the plants. Either way, we don't want to duplicate constants or use slightly different approaches to estimate variables

virtual_rainforest/models/litter/litter_pools.py
```diff
@@ -178,6 +286,111 @@ def calculate_litter_decay_structural_above(
...
+
+    Returns:
+        Rate of decay of the below ground metabolic litter pool [kg C m^-2 day^-1]
+    """
+
+    return (
+        litter_decay_coefficient
+        * temperature_factor
+        * moisture_factor
+        * litter_pool_below_metabolic
```
vgro, line 351: Is the `litter_chemistry_factor` not relevant here? I don't know much about the science but you used it for the other two new pools

jacobcook1995: Ahh right yes this could be clearer, the metabolic pools are for biological matter that doesn't contain recalcitrant polymers (i.e. lignin). The `litter_chemistry_factor` is a placeholder for actually tracking the recalcitrant polymers, so is not needed in either metabolic pool

virtual_rainforest/models/litter/litter_model.py
```diff
@@ -154,19 +168,62 @@ def update(self, time_index: int) -> None:
...
+    This function provides a coarse estimate of soil water potential. It is taken from
+    :cite:t:`campbell_simple_1974`.
+
+    TODO - This is a stopgap solution until we decide on a systematic way of handling
+    water potentials across the relevant models (`soil`, `litter`, `plants` and
+    `hydrology`).
+
+    Args:
+        soil_moisture: Volumetric relative water content [unitless]
+        air_entry_water_potential: Water potential at which soil pores begin to aerate
```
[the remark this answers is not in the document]
jacobcook1995: I think so yes

virtual_rainforest/models/litter/constants.py
jacobcook1995: This was how they were given in the paper I took them from, so I wanted to use the same to preserve numerical accuracy

tests/conftest.py
jacobcook1995: Uhh, I copied and pasted it from `dummy_carbon_data` 😅. I don't actually use the lowest soil layer for anything, would it be better for this layer to just be `nan`?

virtual_rainforest/models/litter/litter_pools.py
jacobcook1995: I'll add a description of the pools to the module docstring as there isn't all that much info in there at the moment

tests/models/litter/test_litter_model.py
jacobcook1995: So, the soil moisturise used to generate this are [0.5, 0.7, 0.6]. I'm guessing that 0.5 would still be expected to be above wilting point? I think something is definitely wrong in the expression, as you would expect the highest soil moisture to result in the most negative soil water potential right?

tests/conftest.py
vgro: maybe yes, at least for now

tests/models/litter/test_litter_model.py
vgro: I think high soil moisture has the least negative potential because it is easier to remove water, right? So the direction seems right, maybe it's just the parameter you used in the equation

vgro: The decrease is steep at some point, so you might be right

jacobcook1995: I'll look into it a bit more because it is very steep. Like a soil moisture of 0.1 results in a potential of -411224.11 kPa, which feels very high?

vgro: If you look at the curves, they are very flat when soil moisture is high and then increase exponentially at low soil moisture, so that might be correct but worth double-checking

vgro (approved):
LGTM.

PR #286, opened 2023-08-16
Fix unified path merge bug

alexdewar:
# Description

In PR #280, I inadvertently broke things while resolving merge conflicts introduced by @jacobcook1995's PR.

Now invoking `vr_run` gives errors like so:

```
  File "/home/alex/code/virtual_rainforest/.venv/bin/vr_run", line 6, in <module>
    sys.exit(_vr_run_cli())
             ^^^^^^^^^^^^^
  File "/home/alex/code/virtual_rainforest/virtual_rainforest/entry_points.py", line 152, in _vr_run_cli
    vr_run(cfg_paths, override_params, Path(args.merge_file_path))
  File "/home/alex/code/virtual_rainforest/virtual_rainforest/main.py", line 287, in vr_run
    merge_continuous_data_files(
  File "/home/alex/code/virtual_rainforest/virtual_rainforest/core/data.py", line 474, in merge_continuous_data_files
    check_outfile(out_folder / f"{data_options['continuous_file_name']}.nc")
                                  ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'continuous_file_name'
```

I forgot to update the attribute names in the merged code and clearly also forgot to retest that everything was still working before merging. Sorry!

In an attempt to stop this from happening again, I've added a simple integration test for `vr_run`, which just runs the program with the example data. It takes < 10s to run on my machine, but if you want to speed it up, you could reduce the example data to be more minimal.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/entry_points.py
+++ b/virtual_rainforest/entry_points.py
@@ -68,7 +68,7 @@ def _parse_command_line_params(
         raise to_raise
 
 
-def _vr_run_cli() -> None:
+def vr_run_cli() -> None:
     """Configure and run a Virtual Rainforest simulation.
 
     This program sets up and runs a simulation of the Virtual Rainforest model. At
@@ -88,7 +88,7 @@ def _vr_run_cli() -> None:
     """
 
     # Check function docstring exists, as -OO flag strips docstrings I believe
-    desc = textwrap.dedent(_vr_run_cli.__doc__ or "Python in -OO mode: no docs")
+    desc = textwrap.dedent(vr_run_cli.__doc__ or "Python in -OO mode: no docs")
     fmt = argparse.RawDescriptionHelpFormatter
     parser = argparse.ArgumentParser(description=desc, formatter_class=fmt)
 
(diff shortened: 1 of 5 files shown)


virtual_rainforest/entry_points.py
```diff
@@ -68,7 +68,7 @@ def _parse_command_line_params(
         raise to_raise
 
 
-def _vr_run_cli() -> None:
+def vr_run_cli() -> None:
```
jacobcook1995, line 71: Why is this function renamed? I will admit that I don't really no what was implied by it name starting with `_` in the first place?

jacobcook1995 (approved):
LGTM, I just had a quick question about the renaming of the function.

I think that adding a test of the overall `vr_run` function is sensible as this problem wasn't caught by the current unit tests and similar problems could well crop up again.

virtual_rainforest/entry_points.py
alexdewar: The underscore implies it's private to `entry_points.py`, which it isn't. Even before I wrote that test, it was still being invoked by the poetry's command-line wrapper tool.

This does mean you might need to rerun `poetry install` in order to get a working `vr_run`.

dalonsoa (approved):
LGTM!

PR #287, opened 2023-08-17
Linking animal excretion to the litter model

jacobcook1995:
# Description

This pull request links animal excretion (that takes place in the animal module) to the metabolic litter pool. The treatment of animal excrement decay is extremely simplistic (it builds up and then all decays at once), so this will need further work in future.

I mainly want to check that my changes to the animal model are sensible and that the way that the models are setup to communicate is sensible.

There's another couple of processes within the animal model that should feed into the litter model, so this pull request is also a prototype for those (slightly more complex) exchanges.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -41,6 +44,7 @@ def __init__(
         self.plant_community: PlantCommunity = PlantCommunity(10000.0, 1)
         self.carcass_pool: CarcassPool = CarcassPool(10000.0, 1)
         self.soil_pool: PalatableSoil = PalatableSoil(10000.0, 1)
+        self.excrement_pool: ExcrementPool = ExcrementPool(0.0)
```
TaranRallings, line 47: I think we're best off initializing at a non-zero value. If we've got small bodied excrement specialists and a call order that puts foraging before excretion, we could have complete extinction from starvation in the first monthly time-step.

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -127,5 +131,32 @@ def update(self, time_index: int) -> None:
...
+        # Find the size of all excrement pools
+        remaining_excrement = [
+            community.excrement_pool.stored_carbon(self.data.grid.cell_area)
+            for community in self.communities.values()
+        ]
+
+        # After an update all excrement that isn't consumed is assumed to enter the
+        # litter, so stored energy of each pool is reset to zero
+        for community in self.communities.values():
+            community.excrement_pool.stored_energy = 0.0
```
TaranRallings, line 156: If I'm understanding correctly, this ordering will create a situation where the excrement pool is always empty when foraging begins. Right now, excretion and foraging and intertwined processes. Perhaps this is a sign they should be untangled. In the meantime, if we could use a slightly different assumption, like 50% loss per tick or something, I think that would work better.

TaranRallings (commented):
The code looks good but I had a couple of requests on the animal-intersection front. Let's have a chat about this update ordering issue and get something together that will work well for both of us. Scarabs are likely going to be an important part of the system and we have to be careful they don't all die on us!

virtual_rainforest/models/animals/animal_model.py
jacobcook1995: Yeah you are correct. I basically did this setting to zero as a first pass as I was sure how best to handle competition between excrement scavengers and microbial decomposition. If a percentage loss per tick would work for you then happy to go with that

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -126,16 +126,16 @@ def metabolize(self, dt: timedelta64) -> None:
         energy_needed = self.metabolic_rate * float((dt / timedelta64(1, "s")))
         self.stored_energy -= min(self.stored_energy, energy_needed)
 
-    def excrete(self, soil: Pool, consumed_energy: float) -> None:
-        """The function to transfer waste energy from an animal cohort to the soil.
+    def excrete(self, excrement_pool: Pool, consumed_energy: float) -> None:
```
alexdewar, line 129: I'm wondering whether, instead of taking a `Pool` argument to modify, this function could just return the amount of waste energy and then the caller could update a `Pool` if it wants. What do you reckon?

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -53,8 +54,11 @@ class AnimalModel(BaseModel):
     """Longest time scale that soil model can sensibly capture."""
     required_init_vars = ()
     """Required initialisation variables for the animal model."""
-    vars_updated = []
-    """The animal model doesn't currently use the data object so nothing to update."""
+    vars_updated = ["excess_excrement"]
```
alexdewar, line 57: Assuming this value is never changed, I think it would be better as a `tuple`:

```suggestion
    vars_updated = ("excess_excrement",)
```

virtual_rainforest/models/litter/litter_model.py
```diff
@@ -158,6 +159,17 @@ def setup(self) -> None:
         # rate so that the soil model can be run before the litter model. Think we need
         # to decide how we are handling model order first though.
 
+        # TODO - This should be created by the animal model, but it is not yet linked
+        # into the full vr_run flow yet. Once it is this step should be deleted.
+        self.data["excess_excrement"] = DataArray(
+            np.full((len(self.data.grid.cell_id)), 0.0),
```
alexdewar, line 165:
```suggestion
            np.zeros_like(self.data.grid.cell_id),
```

tests/conftest.py
```diff
@@ -203,6 +203,10 @@ def dummy_litter_data(layer_roles_fixture):
         [0.6, 0.31, 0.02], dims=["cell_id"]
     )
     """Below ground structural litter pool (kg C m^-2)"""
+    data["excess_excrement"] = DataArray(
+        [8e-07, 8.42857e-07, 3.28571e-05], dims=["cell_id"]
+    )
+    """Rate of excrement input from the animal model [kg C m^-2 day^-1]."""
```
alexdewar, line 209: Out of interest, where did these numbers come from?

tests/models/animals/test_animal_model.py
```diff
@@ -173,6 +178,9 @@ def test_update_method_sequence(data_instance, functional_group_list_instance):
 
     # Assert the methods were called in the expected order
     assert call_sequence == method_names
+    # Check that excrement data is created, all elements are zero as no actual updates
+    # have occurred
+    assert all(element == 0.0 for element in model.data["excess_excrement"])
```
alexdewar, line 183: You need to be careful when doing floating-point comparisons.

Would this work?

```suggestion
    assert np.allclose(model.data["excess_excrement"], 0.0)
```

tests/models/animals/test_animal_model.py
```diff
@@ -189,3 +197,35 @@ def test_update_method_time_index_argument(
...
+    litter_additions = model.calculate_litter_additions()
+
+    # Check that litter addition pools are as expected
+    assert np.allclose(
+        litter_additions["excess_excrement"], [5e-08, 8e-07, 8.42857e-07, 3.28571e-05]
+    )
+
+    # Check that the function has reset the pools correctly
+    for community in model.communities.values():
+        community.excrement_pool.stored_energy = 0.0
```
alexdewar, line 231: This doesn't check anything currently.

How about:

```suggestion
    # Check that the function has reset the pools correctly
    assert np.allclose(
        (
            community.excrement_pool.stored_energy
            for community in model.communities.values()
        ),
        0.0,
    )
```

tests/models/animals/test_carcasses_and_poo.py
```diff
@@ -0,0 +1,31 @@
...
+
+class TestCarcassPool:
+    """Test the CarcassPool class."""
+
+    def test_initialization(self):
+        """Testing initialization of CarcassPool."""
+        from virtual_rainforest.models.animals.carcasses_and_poo import CarcassPool
+
+        c1 = CarcassPool(1000.7, 1)
+        assert c1.stored_energy == 1000.7
```
alexdewar, line 15: Again, you can't rely on floating-point comparisons always being correct because of rounding errors.

If you add an `import pytest`, you can do:

```suggestion
        assert pytest.approx(c1.stored_energy, 1000.7)
```

tests/models/animals/test_carcasses_and_poo.py
```diff
@@ -0,0 +1,31 @@
...
+        """Testing initialization of CarcassPool."""
+        from virtual_rainforest.models.animals.carcasses_and_poo import ExcrementPool
+
+        poo = ExcrementPool(25.0)
+        # Test that function to calculate stored carbon works as expected
+        assert poo.stored_energy == 25.0
+        assert poo.stored_carbon(1.0) == 2.5e-5
+        assert poo.stored_carbon(10.0) == 2.5e-6
+        assert poo.stored_carbon(25.0) == 1.0e-6
+        assert poo.stored_carbon(5000.0) == 5.0e-9
```
alexdewar, line 31: Ditto.

```suggestion
        assert pytest.approx(poo.stored_energy, 25.0)
        assert pytest.approx(poo.stored_carbon(1.0), 2.5e-5)
        assert pytest.approx(poo.stored_carbon(10.0), 2.5e-6)
        assert pytest.approx(poo.stored_carbon(25.0), 1.0e-6)
        assert pytest.approx(poo.stored_carbon(5000.0), 5.0e-9)
```

virtual_rainforest/models/animals/__init__.py
```diff
@@ -20,8 +20,8 @@
   module.
 * The :mod:`~virtual_rainforest.models.animals.constants` provides a set of dataclasses
   containing the constants required by the broader animal model.
-* The :mod:`~virtual_rainforest.models.animals.carcasses` provides a model for the
-  surface carcasses created by mortality.
+* The :mod:`~virtual_rainforest.models.animals.carcasses_and_poo` provides a model for
+  the both surface carcasses created by mortality and animal excrement.
```
alexdewar, line 24:
```suggestion
* The :mod:`~virtual_rainforest.models.animals.carcasses_and_poo` provides a model both
  for surface carcasses created by mortality and animal excrement.
```

alexdewar (changes requested):
I've made some small suggestions, but on the whole it seems sensible.

virtual_rainforest/models/animals/carcasses_and_poo.py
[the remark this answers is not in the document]
jacobcook1995: Would `decay_pools` work? Slightly hesitant to use `litter` as there's an entire model called that

tests/conftest.py
jacobcook1995: These numbers are basically just invented, I could try and estimate something from data if you think that would be better

virtual_rainforest/models/animals/carcasses_and_poo.py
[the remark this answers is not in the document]
davidorme: Could just be `animals.decay`. That's very clear about the processes handled and nice and short.

tests/conftest.py
alexdewar: Nah, that's fine. I might add a comment saying these numbers are totally made up.

virtual_rainforest/models/animals/animal_model.py
jacobcook1995: The value should never be changed in the course of a simulation but more elements will be added as the model develops. Is the advantage to use a tuple here that it is immutable?

alexdewar: Yes, exactly.

virtual_rainforest/models/animals/animal_cohorts.py
jacobcook1995: I'm not sure because I don't think the step to calculate waste energy would ever be invoked for any reason but to update the excrement pool, so I don't feel like there's a value in having it be a separate function?

alexdewar: I was thinking that it would make it more modular and testable. The problem with passing `excrement_pool` in and modifying it is that `excrete()` then depends on the state of `excrement_pool` in addition to `consumed_energy` and various attributes of `self`.

On a related note, I've had a look and it seems there's not a test for `excrete()` as well as various other smaller functions. (I know this isn't your doing in particular -- I'm not singling you out!) There are tests for higher-level functions, so the test coverage looks good, but this isn't the best practice for unit testing. Ideally you want separate tests for smaller functions, checking that certain inputs give the expected outputs, then you can (often) just test that higher-level functions call the lower-level ones in the correct way with mocking.

I've noticed that you've just changed `excrete()` to split the energy between two pools. I'm not entirely sure about the way you've implemented this. Is `decay_fraction` something you expect to change between calls to `excrete()`? If so, it should be settable via `forage_communities()` too. If not, then it should be set once for the `AnimalCohort` in `__init__`.

alexdewar: PS -- I have reviewed large chunks of this code without twigging that these smaller functions aren't being tested, so it's partly on me. I would suggest that you guys adopt a different approach in future though.

TaranRallings: I took a look and the only missing tests in the animal module appear to for the newly implemented `prey_group_selection` and `migrate_community` methods, which I'm still fiddling with. Some of these tests may need to be expanded in functionality but I can't find a significant hole in coverage. Perhaps Jacob removed some tests from this branch while he was changing the forms.

jacobcook1995: Yeah I'm pretty sure that `excrete()` is tested (within the `TestAnimalCohort` testing class) as I had to modify the test when I altered the `excrete()` function

jacobcook1995: On the `decay_fraction` question, for the meantime this is not something that should change between calls to `excrete()`. In future, we _might_ want to change the model such that it can change between (due to e.g. temperature changes). I guess it's not sensible to hedge against this potential future use and to just set it in the `__init__`?

alexdewar: My bad @TaranRallings. You're quite right. I totally missed it, but it does seem like there are plenty of tests.

virtual_rainforest/models/animals/animal_cohorts.py
```diff
@@ -107,6 +113,10 @@ def __init__(
             self.mass,
             self.functional_group.prey_scaling,
         )
+        # TODO - In future this should be parameterised using a constants dataclass, but
+        # this hasn't yet been implemented for the animal model
+        self.decay_fraction_excrement: float = DECAY_FRACTION_EXCREMENT
+        """The fraction of excrement which decays before it gets consumed."""
```
TaranRallings, line 119: Do I understand correctly that this decay_fraction_excrement would then (later) be a function of our time-step size?

TaranRallings (approved):
LGTM!

virtual_rainforest/models/animals/animal_cohorts.py
jacobcook1995: Nope, I set it up so that this would be time step size independent. At the end of the time step all excrement in the decay pool is transferred to the litter pool. The excrement in the scavengable subpool never gets transferred to the litter and if nothing scavenges it it just builds up in that pool (which is obviously a little odd). 

The reason for this slightly odd division is that if we say set 50% of the excrement pool to decay at the end of every time step, shorter time steps would result in a reduced steady state excrement pool unless we did something clever to control for that. We might want to do something like that in the future, but I thought having two totally independent excrement subpools covering the two destinations (litter, scavenging) was easier for now

tests/conftest.py
```diff
@@ -203,10 +203,13 @@ def dummy_litter_data(layer_roles_fixture):
...
     )
     """Below ground structural litter pool (kg C m^-2)"""
-    data["excess_excrement"] = DataArray(
+    data["decomposed_excrement"] = DataArray(
         [8e-07, 8.42857e-07, 3.28571e-05], dims=["cell_id"]
     )
-    """Rate of excrement input from the animal model [kg C m^-2 day^-1]."""
+    """Rate of excrement input from the animal model [kg C m^-2 day^-1].
+
+    These values are completely made up, so you should read anything into them.
```
alexdewar, line 211:
```suggestion
    These values are completely made up, so you should not read anything into them.
```

tests/models/animals/test_animal_model.py
```diff
@@ -214,18 +214,24 @@ def test_calculate_litter_additions(functional_group_list_instance):
     model = AnimalModel(data, pint.Quantity("1 week"), functional_group_list_instance)
 
     # Update the waste pools
-    excess_excrement = [3.5e3, 5.6e4, 5.9e4, 2.3e6]
+    decomposed_excrement = [3.5e3, 5.6e4, 5.9e4, 2.3e6]
     for ind, community in enumerate(model.communities.values()):
-        community.excrement_pool.stored_energy = excess_excrement[ind]
+        community.excrement_pool.decomposed_energy = decomposed_excrement[ind]
```
alexdewar, line 219: I *think* this could be:

```python
    for energy, community in zip(decomposed_excrement, model.communities.values()):
        community.excrement_pool.decomposed_energy = energy
```

tests/models/animals/test_decay.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestExcrementPool:
+    """Test the ExcrementPool class."""
+
+    def test_initialization(self):
+        """Testing initialization of CarcassPool."""
+        from virtual_rainforest.models.animals.decay import ExcrementPool
+
+        poo = ExcrementPool(77.7, 25.0)
```
alexdewar, line 27: I've seen people name variables `poo` before, but I think this is the first time where it's actually relevant :stuck_out_tongue:

alexdewar (approved):
I've made a couple of minor suggestions, but otherwise all good!

tests/models/animals/test_animal_model.py
jacobcook1995: Yep that works! Is it generally better to make use of `zip()` rather than enumerate + indexing?

alexdewar: In general it's best practice to avoid looping over indices if there's another way, because a) it's usually more readable and b) it stops you from making off-by-one errors etc.

I've realised that these two things aren't functionally identical though. `zip` is fine if you know your sequences have the same length, otherwise it'll just stop after either one of them is finished:

```python
>>> list(zip(range(2),range(3)))
[(0, 0), (1, 1)]
```

I guess that's fine here though?

jacobcook1995: Yeah I think that's fine here, like if the number of communities returned changes that is something that should have been caught by a different unit test

PR #290, opened 2023-08-17
Fix path merge bug, take 2

alexdewar:
# Description

@jacobcook1995 @dalonsoa @davidorme 

I had to unmerge my last PR, because it was causing CI failures, and I didn't clock this till after it was merged.

It's my fault for not checking this properly (sorry!), but it turns out that your branch protection rules weren't set up correctly, for either `main` or `develop`. You need to manually select what the required status checks are, otherwise GitHub will happily let you merge things that don't pass tests.

I've sorted it now, but for future reference this is the relevant section on the settings page:

image

It's worth knowing, because if you add a new runner (e.g. for a new version of Python), you also have to add it to the required status checks.

Let's try this again. I'll fix up the CI failures then request a review.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ authors = [
 ]
 
 [tool.poetry.scripts]
-vr_run = "virtual_rainforest.entry_points:_vr_run_cli"
+vr_run = "virtual_rainforest.entry_points:vr_run_cli"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
--- a/tests/core/test_cli_integration.py
+++ b/tests/core/test_cli_integration.py
@@ -1,0 +1,22 @@
+"""An integration test for the VR command-line interface."""
+import shutil
+from contextlib import nullcontext as does_not_raise
+from pathlib import Path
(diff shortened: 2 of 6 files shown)


alexdewar:
This is ready for review now and the tests are passing. The only thing I've changed since my last PR is this: 2f587bfcb300c894aecfbabd2af630690ed1d3cc

The issue was that a seemingly unrelated test `test_make_hex_grid` was failing, due to it sometimes timing out on Windows and macOS runners (it's a `hypothesis`-based test). As these runners can sometimes be slow, I've just disabled the deadline for this test.

As to why these failures only seemed to occur after my code changes, I have no idea. Perhaps the extra resource usage incurred by my new integration test meant that the runners were still running slower than usual even after this one test had finished (e.g. because some memory had been swapped to disk), which made this problem more likely to rear its head. Who knows. Either way, the fix seems harmless enough.

dalonsoa (approved):
LGTM- again :) !

jacobcook1995 (approved):
LGTM!

PR #291, opened 2023-08-18
Fix: Correctly merge configs recursively

alexdewar:
# Description

Currently, the code will clobber any "core" config options if an outpath is also provided on the command line. Fix by using the `config_merge()` function.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/entry_points.py
+++ b/virtual_rainforest/entry_points.py
@@ -134,7 +134,8 @@ def vr_run_cli() -> None:
     override_params: dict[str, Any] = {}
     if args.outpath:
         # Set the output path
-        override_params |= {"core": {"data_output_options": {"out_path": args.outpath}}}
+        outpath_opt = {"core": {"data_output_options": {"out_path": args.outpath}}}
+        override_params, _ = config_merge(override_params, outpath_opt)
     if args.params:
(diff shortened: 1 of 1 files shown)


jacobcook1995 (approved):
LGTM!

davidorme (approved):
The output path is likely to be changed frequently - and likely with the same config inputs - so having a dedicated CLI option seems reasonable, but having just looked at  #292 - which removes the `--merge` option - would it be more consistent to set this in `config` files and potentially override it using `--param`?

Basically LGTM - certainly while the CLI is in such an early state.

alexdewar:
@davidorme It would make the code a little simpler, but I do feel like having a short alias for this option is more useful than for choosing the merge path. The problem with keeping the `--merge` option as well as making it settable via these config options would be that you would end up with two cases with different code paths: one where the merged config file is in the same folder as all the other output files (i.e. it uses `core.data_output_options.out_path`) and one where it isn't. I figured the latter possibility probably wasn't something many people would want. By contrast, this just means you can write `--outpath some/path` rather than `--param core.data_output_options.out_path="some/path"`, which is a lot less memorable. I figured it's probably something people might want to do quite often.

Issue #234, opened 2023-06-06
`[[core.data.variable]]` tags cannot be used in more than one config file


jacobcook1995:
**Describe the bug**
`[[core.data.variable]]` tags cannot be used across multiple toml config files without causing `ConfigurationError: Duplicated entries in config files: core.data.variable` to be raised.

**To Reproduce**

1. Create a folder of config files with multiple files that define `[[core.data.variable]]` tags
2. From the terminal call `vr_run path/to/config/folder`
3. See error

**Expected behavior**
We probably want to be able to define data paths across multiple files, rather than forcing everyone to work with a single file. If we can't find a work around for this then it should be documented.

davidorme:
I don't think this is a bug. I think we'd typically have users define all the data sources in one configuration file that might be used with different model configurations. If models each define their own data sources, they should clash if duplicated. So this is docs, I think

jacobcook1995:
Probably should make clear that this isn't a clash where 2 files both define data sources for the same variable. This is a clash where more than one file defines (any) data sources. So, all datasources will have to be defined within a single toml file.

vgro:
I've added a note to the documentation, the PR will show if I missed something. Maybe we need a new issue to actually fir the problem of not being able to use the functionality across `config` files?

davidorme:
> Probably should make clear that this isn't a clash where 2 files both define data sources for the same variable. This is a clash where more than one file defines (any) data sources. So, all datasources will have to be defined within a single toml file.

I think this _is_ a bug then. Identical variable names should be an error, but not having data in different files.

PR #292, opened 2023-08-18
Remove `--merge` command-line option and replace with config option

### LINKED ISSUE #270: Allow setting path for merged config file with a parameter
Currently, the only way to choose the path for the merged config file which is output is via the `--merge` (or `-m`) command-line flag. It seems a bit odd that it cannot be set in config files like the paths for all the other output files.

Once #267 is merged, there will be a generic mechanism for passing additional parameter values to a simulation, meaning we could easily keep the current `--merge` flag without any extra complexity.

alexdewar:
# Description

Currently, setting the path for the merged config file is a different process to setting the path for other simulation output files:

1. It can only be set via a command-line argument
2. This command-line argument comprises the full path (i.e. including directory), meaning that the output path (e.g. set by the `--outpath` argument) is not used
3. There is no option to disable its generation

I have added a couple of config options to control the generation of merged config files:

- `core.data_output_options.save_merged_config` - for enabling/disabling generation
- `core.data_output_options.out_merge_file_name` - the name of the merged config file (**not** the whole path!)

If generated, merged config files will now always be saved to the output path along with other output files. I have removed the `--merge` option as keeping it around with the existing behaviour would have made the code more complex and I don't think it's necessary anyway; if a user wants to change the merged config file name then they can use the `--param` option, as for other parameters.

Fixes #270.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/core/test_cli_integration.py
+++ b/tests/core/test_cli_integration.py
@@ -1,7 +1,6 @@
 """An integration test for the VR command-line interface."""
 import shutil
 from contextlib import nullcontext as does_not_raise
-from pathlib import Path
 from unittest.mock import patch
 
 
@@ -14,8 +13,6 @@ def test_vr_run(shared_datadir):
         "--example",
         "--outpath",
         str(shared_datadir),
-        "--merge",
-        str(Path(shared_datadir) / "vr_full_model_configuration.toml"),
     ]
     with does_not_raise():
         with patch("virtual_rainforest.entry_points.sys.argv", args):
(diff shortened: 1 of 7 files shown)


jacobcook1995 (approved):
LGTM!

vgro (commented):
LGTM.
This reminds me that we had issues in the past with merging config files which is why we put all the config info in one file for the dummy experiment. @jacobcook1995  might remember the exact details of the error. I wonder if that has been addressed or if that could be solved here?

jacobcook1995:
That was issue #234, I don't think that one is solvable here. I _think_ that's an issue we either need to just accept, or need to find a clever way of tricking python's built in `toml` reader to accept arrays of tables that are spread across multiple `toml` files. I feel like the latter would be a bit out of scope for this PR

davidorme (approved):
LGTM too - that `--merge` option was a temporary mandatory step for testing and making it optional was always the plan. Making that a `config` setting and keeping the command line interface simple seems right to me.

PR #293, opened 2023-08-21
Feature/horizontal flow

vgro:
This PR introduces horizontal flow of water in the hydrology model (accumulated surface runoff only).

At the moment, this means that upstream neighbours are identified in `setup()`, and at each time step, the surface runoff from the upstream neighbours is added to the runoff of each grid cell. This should be working for runoff provided by SPLASH. Below ground flow is not yet implemented.

I added recipes to make dummy `elevation` and `surface_runoff` data sets. I purposely didn't add the surface runoff to climate data because this might come from SPLASH soon.

I also changed the description of soil moisture to 'Volumetric relative water content (unitless)' to be consistent with the soil model.

There are still a lot of TODOs regarding spinup, boundaries, and below ground flow which I will address later.

## Type of change

## Key checklist

## Further checks


vgro:
I accidentally made a bit of a mess with poetry, please help to get around merging this

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -224,24 +239,68 @@ def setup(self) -> None:
...
+
+        # set initial accumulated runoff to zero
+        accumulated_runoff = np.full_like(self.data["elevation"], 0)
+
+        # calculate accumulated runoff for each cell (sum of the upstream neighbours)
+        # TODO this requires a proper spin up!
+        # TODO check for negative numbers
+
+        for cell_id, upstream_id in enumerate(self.upstream_ids):
+            accumulated_runoff[cell_id] += np.sum(baseline_runoff[upstream_id])
```
alexdewar, line 276: Is there a reason why these todos can't be done in this PR too? In particular, it seems like it should be easy to check for negative numbers. If you're not sure exactly what you want to do when you detect one, you could just raise an error for now and figure out how to handle it more gracefully in future.

Again, I think this block should be separated out from the rest of this function, either into a function with the rest of the code above or into its own function.

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -390,7 +444,7 @@ def update(self, time_index: int) -> None:
             relative_humidity=subcanopy_humidity,
             atmospheric_pressure=subcanopy_pressure,
             soil_moisture=soil_moisture_infiltrated,
-            wind_speed=0.1,  # TODO wind_speed,
+            wind_speed=0.1,  # TODO wind_speed in data object,
```
alexdewar, line 447: Is this in m/s or something else? Might be worth mentioning the units in the comment

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -448,6 +502,21 @@ def update(self, time_index: int) -> None:
...
+        # Calculate accumulated surface runoff
+        # Get the runoff created by SPLASH or initial data set
+        single_cell_runoff = np.array(soil_hydrology["surface_runoff"])
+
+        # Get accumulated runoff from previous time step
+        accumulated_runoff = np.array(self.data["surface_runoff_accumulated"])
+
+        # Calculate accumulated runoff for each cell (me + sum of upstream neighbours)
+        for cell_id, upstream_id in enumerate(self.upstream_ids):
+            accumulated_runoff[cell_id] += np.sum(single_cell_runoff[upstream_id])
```
alexdewar, line 514: Given that you're repeating this code, it suggests it should live in its own function. I know it's only two lines of code, but it's not totally trivial and I think deserves a unit test.

alexdewar (changes requested):
I've given a few queries/suggestions, but it otherwise looks ok to me (notwithstanding that I don't understand the science!).

docs/source/data_recipes/create_dummy_elevation.md
```diff
@@ -0,0 +1,59 @@
...
+  display_name: vr_python3
+  language: python
+  name: vr_python3
+---
+
+# Elevation data for dummy module
+
+This section explains how to create a low resolution dummy elevation map from a
+digital elevation model ([SRTM](https://www2.jpl.nasa.gov/srtm/)) for use in
+`vr_run --example`. The initial data covers the region 4°N 116°E to 5°N 117°E, see
```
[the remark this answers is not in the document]
vgro: It's describing how the example data set is created. The original tif file is very large and I didn't want to commit this and do an interactive script. I see how that is confusing here, I'll change the wording.

alexdewar:
I forgot to comment on the `poetry.lock` problem. As far as I can tell, you should be able to just `git revert` all your changes to it (in reverse order).

If that doesn't work, you can just copy the version on `develop` to get it back to how it was, either manually or with:

```sh
git show develop:./poetry.lock > poetry.lock
```

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -338,19 +393,17 @@ def update(self, time_index: int) -> None:
...
         # flow between layers and a gradient of soil moisture and soil water potential
         total_soil_moisture_mm = (
-            self.data["soil_moisture"]
+            self.data["soil_moisture"].isel(layers=-1)
             * (-self.constants.meters_to_millimeters)
-            * self.data["layer_heights"]
-            .where(self.data["layer_heights"].layer_roles == "soil")
-            .dropna(dim="layers")
-        ).sum(dim="layers")
+            * self.data["layer_heights"].isel(layers=-1)
```
jacobcook1995, line 405: Is this simplification possible because you are basically treating all soil layers as a single "bucket"?

jacobcook1995 (approved):
LGTM! I had a couple of queries about the science/approach, but on the code quality front I didn't notice anything that Alex hadn't already brought up

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -224,24 +239,68 @@ def setup(self) -> None:
...
+        # Invert mapping of cell_id: downstream neighbour to cell_id: upstream_neighbors
+        upstream_ids: list = [[] for i in range(len(self.data.grid.cell_id))]
+        for down_s, up_s in enumerate(downstream_ids):
+            upstream_ids[up_s].append(down_s)
+
+        self.upstream_ids = upstream_ids
+        """Upstream ID's for the calculation of accumulated surface runoff."""
+
+        # Get the runoff created by SPLASH or initial data set as the baseline:
+        baseline_runoff = np.array(self.data["surface_runoff"])
```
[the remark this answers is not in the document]
vgro: The hydrology model needs some initial surface runoff to calculate the accumulated runoff in the setup because the update needs an input from the previous time step. At the moment, surface runoff is calculated in the hydrology model at each time step. If we implement SPLASH, the surface runoff will be provided and the calculation in the hydrology model can be removed/ made optional.
As far as I understand, the SPLASH model will be part of the pyrealm package but still somewhat separate from the plant model.

virtual_rainforest/models/hydrology/hydrology_model.py
vgro: yes, that's what it is.

vgro: In the previous version, I made an error when summing over all depths, I treated the layer depths like layer thickness. If I want one bucket I just need the depth of the lowest layer.

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -224,24 +239,68 @@ def setup(self) -> None:
...
+        ]
+        downstream_ids = []
+
+        for cell_id, neighbors_id in enumerate(mapping):
+            # Find lowest neighbour
+            downstream_id_loc = np.argmax(elevation[cell_id] - elevation[neighbors_id])
+            downstream_ids.append(neighbors_id[downstream_id_loc])
+
+        # Invert mapping of cell_id: downstream neighbour to cell_id: upstream_neighbors
+        upstream_ids: list = [[] for i in range(len(self.data.grid.cell_id))]
```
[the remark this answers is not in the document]
vgro: This suggestion does not work correctly and mypy gives me an error because it wants a type annotation. So I'll leave it as it is for now

virtual_rainforest/models/hydrology/hydrology_model.py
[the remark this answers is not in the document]
davidorme: That is correct. The intention is for `pyrealm` to provide a modern Python implementation of SPLASH v1.0, which could be then be imported and used in one of the `abiotic` models.

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -224,24 +239,68 @@ def setup(self) -> None:
...
         )
 
+        # Turn elevation data into np array
+        elevation = np.array(self.data["elevation"])
+
+        # map neighbours and identify downstream cell_ids
+        mapping = [
+            self.data.grid.neighbours[cell_id]
+            for cell_id in range(len(self.data.grid.cell_id))
+        ]
```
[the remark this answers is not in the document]
vgro: This works, however, I realized that this is unnecessary, I can use the `self.data.grid.neighbours` directly

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -224,24 +239,68 @@ def setup(self) -> None:
...
+            downstream_id_loc = np.argmax(elevation[cell_id] - elevation[neighbors_id])
+            downstream_ids.append(neighbors_id[downstream_id_loc])
+
+        # Invert mapping of cell_id: downstream neighbour to cell_id: upstream_neighbors
+        upstream_ids: list = [[] for i in range(len(self.data.grid.cell_id))]
+        for down_s, up_s in enumerate(downstream_ids):
+            upstream_ids[up_s].append(down_s)
+
+        self.upstream_ids = upstream_ids
+        """Upstream ID's for the calculation of accumulated surface runoff."""
```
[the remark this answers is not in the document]
davidorme: I'm wondering if the DEM / runoff / stream flow should all be part of the setup of `core.grid.Grid` so it can be initialised right at the start and then shared? 

That doesn't need doing right now - and it is all quite specific to hydrology.  But @TaranRallings I can imagine we might want a graph showing the relative heights of cells for calculating movement costs at some point? That would use a lot of the same information.

I definitely agree that this should become a separate function - that would also make it easy to move into `core.grid.Grid` in the future (possibly as one of a range of runoff models).

virtual_rainforest/models/hydrology/hydrology_model.py
davidorme: This is definitely jumping the gun, but we could have a set of [0,1] coefficients here to allow partial draining into different cells 😄

davidorme (commented):
LGTM too. It would be good to pop out those functions - for the first I think we want something like:

```python
def calculate_drainage_map(grid: Grid, dem: DataArray) dict[int, tuple[int]]:
```

That should return a mapping that allows us to lookup which cells drain into each cell. To future proof it, the tuple members could be tuples of cell id and runoff fraction, but that seems excessive now!

virtual_rainforest/models/hydrology/hydrology_model.py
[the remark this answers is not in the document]
vgro: I agree. I separated the functions and added tests. I also changed the variable names slightly to make in clearer what they represent.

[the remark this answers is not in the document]
vgro: There was initially a comment that this should all move to core which might have gone lost somewhere, but I totally agree. I put it here for easy testing but I'm very happy to move this to core when it is the right time :-)

virtual_rainforest/models/hydrology/hydrology_model.py
vgro: @davidorme partial drainage is something for the below-ground flow I guess, which is not yet implemented. We can discuss that at a later stage ;-)

vgro:
> LGTM too. It would be good to pop out those functions - for the first I think we want something like:
> 
> ```python
> def calculate_drainage_map(grid: Grid, dem: DataArray) dict[int, tuple[int]]:
> ```
> 
> That should return a mapping that allows us to lookup which cells drain into each cell. To future proof it, the tuple members could be tuples of cell id and runoff fraction, but that seems excessive now!

I saw that a bit too late for the last commit, but I think this is a good idea. I'll introduce something like this in the next iteration 👍

vgro:
> > LGTM too. It would be good to pop out those functions - for the first I think we want something like:
> > ```python
> > def calculate_drainage_map(grid: Grid, dem: DataArray) dict[int, tuple[int]]:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That should return a mapping that allows us to lookup which cells drain into each cell. To future proof it, the tuple members could be tuples of cell id and runoff fraction, but that seems excessive now!
> 
> I saw that a bit too late for the last commit, but I think this is a good idea. I'll introduce something like this in the next iteration 👍

I have now implemented something like this but kept it in the hydrology model for now (in `__init__`). It is a bit messy with a mix of lists and dicts...it works but could be prettier.

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -702,3 +691,113 @@ def calculate_soil_evaporation(
...
+        accumulated_runoff: accumulated surface runoff from previous time step, [mm]
+
+    Returns:
+        accumulated surface runoff, [mm]
+    """
+
+    upstream_id_list = [[i for i in drainage_map[x]] for x in drainage_map.keys()]
+
+    for cell_id, upstream_id in enumerate(upstream_id_list):
+        accumulated_runoff[cell_id] += np.sum(surface_runoff[upstream_id])
```
alexdewar, line 761: I think this could be:

```suggestion
    upstream_ids = (range(drainage_map[x]) for x in drainage_map.keys())

    for cell_id, upstream_id in enumerate(upstream_ids):
        accumulated_runoff[cell_id] += np.sum(surface_runoff[upstream_id])
```

Then you avoid creating extra lists unnecessarily.

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -702,3 +691,113 @@ def calculate_soil_evaporation(
...
+
+    for num in accumulated_runoff:
+        if num < 0:
+            to_raise = ValueError(
+                "The accumulated surface runoff should not be negative!"
+            )
+            LOGGER.error(to_raise)
+            raise to_raise
```
alexdewar, line 769:
```suggestion
    if (accumulated_runoff < 0.0).any():
        to_raise = ValueError(
            "The accumulated surface runoff should not be negative!"
        )
        LOGGER.error(to_raise)
        raise to_raise
```

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -702,3 +691,113 @@ def calculate_soil_evaporation(
...
+    return accumulated_runoff
+
+
+# TODO move this to core.grid once we decided on common use
+def calculate_drainage_map(grid: Grid, elevation: np.ndarray) -> dict[int, tuple[int]]:
+    """Calculate drainage map based on digital elevation model.
+
+    This function finds the lowest neighbour for eacj grid cell, identifies all upstream
+    IDs and creates a dictionnary that provides all upstream cell IDs for each grid cell
+    . This function currently supports only square grids.
```
alexdewar, line 780:
```suggestion
    This function finds the lowest neighbour for each grid cell, identifies all upstream
    IDs and creates a dictionary that provides all upstream cell IDs for each grid
    cell. This function currently supports only square grids.
```

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -702,3 +691,113 @@ def calculate_soil_evaporation(
...
+
+    Args:
+        Grid: grid object
```
alexdewar, line 783:
```suggestion
        grid: grid object
```

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -702,3 +691,113 @@ def calculate_soil_evaporation(
...
+        elevation: elevation, [m]
+
+    Returns:
+        tuple of cell IDs and their upstream neighbours
```
alexdewar, line 787: I don't think this is what's being returned... right?

virtual_rainforest/models/hydrology/hydrology_model.py
```diff
@@ -702,3 +691,113 @@ def calculate_soil_evaporation(
...
+
+    grid.set_neighbours(distance=sqrt(grid.cell_area))
+    lowest_neighbours = find_lowest_neighbour(grid.neighbours, elevation)
+    upstream_ids = find_upstream_cells(lowest_neighbours)
+
+    output = {}
+    for cell_id in range(grid.n_cells):
+        output[cell_id] = upstream_ids[cell_id]
+
+    return output
```
alexdewar, line 803: Is there a reason output is a `dict` rather than a `list`?

alexdewar (dismissed):
I've made some minor suggestions, but it's otherwise looking good.

The only other minor comment I've got is that I think some of the types you're using could be made more specific. Various args are specified as `dict`s or `list`s, without any extra info about what type of `dict` etc. they are. If you do know this, then it's better to specify that it's e.g. a `dict[int, str]`. It means you get better type-checking with `mypy` and it also makes it easier to read.

virtual_rainforest/models/hydrology/hydrology_model.py
vgro: yes, you're correct. I was looking at the tuple in the dict

virtual_rainforest/models/hydrology/hydrology_model.py
vgro: I initially used lists but generated a dict based on @davidorme 's  suggestion. I think it makes sense for easier selection, but I couldn't find a good solution for the internal functions. The suggestions you made to reduce the lists don't work unfortunately

alexdewar: It's just that atm you start with `upstream_ids`, which is a `list`, then you create a `dict`, which just has the numbers 0 to `n` as keys (i.e. the same as a `list` would). So I'm not clear on what you get by converting it to a `dict`. Presumably when you extend this function in future you'll sometimes be returning `dict`s without all of these keys?

If you really do want to convert your `list` to a `dict` you can do:

```suggestion
    return dict(enumerate(upstream_ids))
```

(I'm assuming that `grid.n_cells == len(upstream_ids)` here.)

virtual_rainforest/models/hydrology/hydrology_model.py
alexdewar: @vgro Is this the suggestion that isn't working for you? Can I ask why?

vgro: yes, @alexdewar , I get the following error message : TypeError: 'list' object cannot be interpreted as an integer

I also noticed that when I tried to make the typing more specific, for example `list[int]`, I get a lot of mypy error messages. I think you pointed out something very important here. I will look into it again with a fresh brain next week and see how I can tidy this up. Thank you :-)

virtual_rainforest/models/hydrology/hydrology_model.py
vgro: @alexdewar the intention was to move this to `core.grid` and have a `dict` that holds all upstream ids with they `cell_id` as key. However, I agree that if it is only me using this here, it is unnecessary to do this and have all these conversions in the middle. Like I said in the previous comment, there is still some clarity needed regarding the types, so I'll sort this out 👍

virtual_rainforest/models/hydrology/hydrology_model.py
alexdewar: Ahh ok. I think I misread this code. You're just trying to get the *values* out of `drainage_map` and `enumerate` over them, right?

In that case, this might work:

```suggestion
    for cell_id, upstream_ids in enumerate(drainage_map.values()):
        accumulated_runoff[cell_id] += np.sum(surface_runoff[upstream_ids])
```

But yes, more precise typing will help with figuring this out.

vgro: This solution is very neat and works well. I updated the types and I hope that mypy will pick up arising problems now

Issue #259, opened 2023-07-20
Constants registry to document default values


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
The default value of constants are currently only documented in the API docs, which fundamentally are not user facing. They should be documented in a way that most users can view 

**Describe the solution you'd like**
As suggested by @davidorme constant classes could be added to a unit registry. This could then be used to create a default constant values documentation page.

**Describe alternatives you've considered**
Not sure that there's an obvious alternative for documenting the constants

davidorme:
Using a registry also makes for a neater programmatic interface for validating user configuration of constants classes:
https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/258#discussion_r1267911971

davidorme:
And also make a helper function for validating constant classes:
https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/260#discussion_r1272501628

jacobcook1995:
A constants registry has been added so this can be closed

PR #295, opened 2023-08-24
Add a constants registry

jacobcook1995:
# Description

This pull request adds a constants registry that all constants classes should be registered in. It also adds a couple of helper functions that make use of the registry to remove boiler plate code. Let me know if the implementation looks sensible.

My original intention was to also write something to document the contents of the constants registry. However, I couldn't get the registry to play nicely with `sphinx.ext.autodoc`. Went down a bit of a rabbit hole with `Sphinx` custom directives, but in the end gave up on that as well (`feature/custom_directive` shows my progress there).

This partially addresses #259, though not completely as that issue mentions using the registry for documentation.

## Type of change

## Key checklist

## Further checks


davidorme (commented):
LGTM - replaces a load of boilerplate with a shared function. The only thing I wondered about is if we can simply run the registry on the constant class name alone, rather than model name and class name, but I can see that parallel implementations (e.g. abiotic) could easily clash.  They _shouldn't_ do because as of #278 being resolved only the required constants for models are loaded but it just seems safer to leave it as is.

I'm also wondering if we want to replace the model `__init__` code with a shared `register_model` core function? This works but we could bundle a lot of the code into one function and use `inspect` to autodiscover the components?

jacobcook1995:
> LGTM - replaces a load of boilerplate with a shared function. The only thing I wondered about is if we can simply run the registry on the constant class name alone, rather than model name and class name, but I can see that parallel implementations (e.g. abiotic) could easily clash. They _shouldn't_ do because as of #278 being resolved only the required constants for models are loaded but it just seems safer to leave it as is.

I might give this a try as I _think_ it could be a solution to the documentation problem. Like `autodoc` struggles with the current structure of `dict[str,dict[str,Class]]` but might be able to handle `dict[str,Class]`.

> I'm also wondering if we want to replace the model `__init__` code with a shared `register_model` core function? This works but we could bundle a lot of the code into one function and use `inspect` to autodiscover the components?

Would we be able to us inspect to discover all constants classes (from `virtual_rainforest.models.model_name.constants`) in cases where a model had more than one class?

davidorme:
So if we define this function:

```python
from types import ModuleType
import inspect
from importlib import resources
from dataclasses import is_dataclass

from virtual_rainforest.core.base_model import BaseModel
from virtual_rainforest.core.config import register_schema
from virtual_rainforest.core.constants import register_constants_class


def register_model(module: ModuleType, schema_filename: str) -> None:
    """Draft model registration function"""

    # Locate the BaseModel subclass for the model, which should be imported into the
    # module root for the model
    model_obj_found = [
        obj
        for nm, obj in inspect.getmembers(module)
        if inspect.isclass(obj) and issubclass(obj, BaseModel)
    ]

    # Check only one BaseModel subclass is found
    if len(model_obj_found) != 1:
        msg = f"Model {module} does not define a single model class"
        LOGGER.critical(msg)
        raise RuntimeError(msg)

    model = model_obj_found[0]

    # Register the schema
    with resources.path(module, schema_filename) as schema_file_path:
        register_schema(
            module_name=getattr(model, "model_name"), schema_file_path=schema_file_path
        )

    # Find and register the constant dataclasses
    constants_classes = [
        nm for nm, obj in inspect.getmembers(module.constants) if is_dataclass(obj)
    ]
    module_model_stem = module.__name__.split(".")[-1]
    for const_class in constants_classes:
        register_constants_class(module_model_stem, const_class)

```

Then I think the boilerplate in the model `__init__` can be:

```
import from virtual_rainforest.models.soil as this_module
from this_module.soil_model import SoilModel

register_model(this_module,  "soil_schema.json")
```

Another big advantage to this is that we only have _one_ place to adapt the model registration code when it changes, rather than having to adjust every model. Note that if the `xxx_model` submodule file was consistently named (e.g. `model.py`, we wouldn't even have to import the model here, we could inspect `module.model` in the function. At the moment, I'm using the convenience that it is imported to search from a single `BaseModel` subclass.

davidorme:
We could always use `pkgutil.iter_modules` to search all the model submodules for a BaseModel to avoid that import being needed!

alexdewar (commented):
I think I might be missing the point of this PR. Maybe you could give a bit more information about what the motivation for this is?

I think that global constants that users will never want to change, like acceleration due to gravity, could just be good old-fashioned Python variables. Make a file `constants.py` and have `GRAVITY_G = 9.81` in it or whatever. I don't see what advantage you get by having a value like that in some kind of registry (it also means you have to use an uglier syntax to retrieve it).

I don't think I understand what the problem for model-related constants is. Are these values going to be user configurable or fixed? If it's the latter, why can't they just live either in one of the files for the particular module they're used in or, if they're used by multiple modules, in the global `constants.py` file?

davidorme:
I think the decision is that they are all configurable (even `g`!). The model ones need to be because they are part of sensitivity analysis. Some recent PRs have been explicitly about exposing those values through config. There are then a bunch of shared constants, some of which are truly constant (like `g`) but some of which are subject to revision. So we use the same mechanisms to access them all, rather than having two different approaches.

The main point of this is to facilitate documentation, by providing somewhere to look up all the defined constants.

jacobcook1995:
@alexdewar this isn't really about core constants, I just added the registry into `core.constants` as it seemed like the most obvious place. The registry is populated with the default constants classes from each model, so that helper functions can access the classes more easily. These classes can be given non-default values (via the helper functions) when the models are set up.

As David mentioned the other reason is for documentation of constants, but as I mentioned I couldn't really get `CONSTANTS_REGISTRY` to document nicely with `Sphinx`, even when I reduced the nesting of the dictionary.

alexdewar:
@jacobcook1995 Ahh I see. That makes more sense. In that case though, I think the word "constant" is kind of misleading. If they are properties of the models that you want users to be able to change, then effectively they are parameters. How about calling them "model constants" to make the distinction clearer?

I'm still sceptical about making *all* constants modifiable though. Would any user really want to modify *g*? You don't get rainforests on the moon.

davidorme:
> You don't get rainforests on the moon.

Maybe we can get Elon to grow some on Mars 😄

We did discuss parameters/constants - these are all values that will be held constant within a single `vr_run`. While some of them might never vary, having a single kind of structure for holding "constants" seemed simpler to use and to document, rather than having to remember whether it is `CoreConst.variable_thing` or `GRAVITY_G`.

virtual_rainforest/core/constants.py
```diff
@@ -0,0 +1,132 @@
...
+from virtual_rainforest.core.logger import LOGGER
+
+CONSTANTS_REGISTRY: dict[str, dict[str, Callable]] = {}
+"""A registry for all the constants data classes.
+
+:meta hide-value:
+"""
+
+
+def register_constants_class(model_name: str, class_name: str) -> None:
```
alexdewar, line 25: Could users supply the class type to this function directly? E.g.:

```py
register_constants_class("my_model", MyConstants)
```

You can get a class's name from the `__name__` attribute.

That would mean you could get rid of the `import_module` shenanigans.

virtual_rainforest/core/constants.py
```diff
@@ -0,0 +1,132 @@
...
+    if model_name in config and "constants" in config[model_name]:
+        # Checks that constants in config are as expected
+        check_valid_constant_names(config, model_name, class_name)
+        # If an error isn't raised then generate the dataclass
+        constants = constants_class(**config[model_name]["constants"][class_name])
+    else:
+        # If no constants are supplied then the defaults should be used
+        constants = constants_class()
+
+    return constants
```
alexdewar, line 132:
```suggestion
        return constants_class(**config[model_name]["constants"][class_name])

    # If no constants are supplied then the defaults should be used
    return constants_class()
```

alexdewar (changes requested):

jacobcook1995:
> Then I think the boilerplate in the model `__init__` can be:
> 
> ```
> import from virtual_rainforest.models.soil as this_module
> from this_module.soil_model import SoilModel
> 
> register_model(this_module,  "soil_schema.json")
> ```

@davidorme I don't think this would work because the model then imports itself during it's import process

davidorme:
@jacobcook1995  Gaah - of course. However, we can probably make it work with `__name__` instead. That's a string not a module object but should work.

jacobcook1995:
Surprisingly, it seems fine with `import virtual_rainforest.models.soil as this_module` actually, it's `from this_module.soil_model import SoilModel` that causes the problem

davidorme:
Huh. If you use the full name does it work?

jacobcook1995:
Ahh that seems to work!

davidorme:
But... I think we can reduce it to something even simpler - give me 5 minutes.

davidorme:
The logger messages in that are confusing! I was using them to track that it worked. It would be nice to import the BaseModel within the `register_model` for a complete container of the process, but maybe that's the next PR?

jacobcook1995:
I've implemented the `register_model` helper function and made the changes that @alexdewar suggested. I didn't reduce the nesting of the dictionary though, as it contains potentially useful information for documenting  (i.e. which model each constants class belongs to).

virtual_rainforest/core/base_model.py
```diff
@@ -517,3 +523,57 @@ def check_init_data(self) -> None:
...
+    # Get a reference to the module
+    module = sys.modules[module_name]
+
+    # Locate the BaseModel subclass for the model, which should be imported into the
+    # module root for the model
+    model_obj_found = [
+        obj
+        for nm, obj in getmembers(module)
+        if isclass(obj) and issubclass(obj, BaseModel)
+    ]
```
alexdewar, line 554: Rather than doing this magic, why not just have the user pass in the model class directly? I think that would be clearer too, rather than importing the model classes into `__init__.py` then not using them. Explicit is better than implicit.

virtual_rainforest/core/base_model.py
```diff
@@ -517,3 +523,57 @@ def check_init_data(self) -> None:
...
+        )
+
+    # Find and register the constant dataclasses
+    constants_class_names = [
+        obj for nm, obj in getmembers(module.constants) if is_dataclass(obj)
+    ]
+    module_model_stem = module.__name__.split(".")[-1]
+    for const_class in constants_class_names:
+        # Import dataclass of interest
+        register_constants_class(module_model_stem, const_class)
```
alexdewar, line 579: Again, I would make this explicit rather than implicit. Make the user pass in the constants data classes.

alexdewar (approved):
It's getting there! I do have some a suggestion about `register_model`, but it's up to you whether you want to do it in this PR or later (or not at all), so I'm approving this.

My suggestion is: I think `register_model` should take the model class and constants classes as arguments, rather than relying on special magic to find them, because it's clearer and simpler that way.

I have some longer-term proposals:

1. Is there a reason we can't just standardise on a schema file name (e.g. `model_schema.json`) and use that everywhere? It seems redundant to have to pass in a file name every time you register a model.
2. Given that we're only allowing users to register one model class per module (which I think is sensible!), you could move the model registration code out of `BaseModel.__init_subclass__`...
3. ...and once you've done that you will have all the model registration code in one place (i.e. for model class, schema and constants). Then you can put all of this information into a single model registry object, which should simplify things:

```py
@dataclass
class ModelInfo:
    cls: type[BaseModel]
    schema: dict[str, Any]
    constants: WhateverTypeThisIs

model_registry: dict[str, ModelInfo]
"""Information about models, keyed by name."""
```

What do you think? And @davidorme?

jacobcook1995:
> 1. Is there a reason we can't just standardise on a schema file name (e.g. `model_schema.json`) and use that everywhere? It seems redundant to have to pass in a file name every time you register a model.

I would be in favour of using a standardised name. The reason we used different names for the `{model_name}_model.py` files is that there's often reason to look at multiple model's code simultaneously so them having named the same creates significant confusion. I honestly very rarely look at the schema for even one model, let alone look a multiple model's schema at once, so I really don't think there's a chance of significant confusion here

davidorme:
It isn't really _magic_ exactly - it's just auto-discovering the components. The advantage is that `__init__` doesn't need to be updated as much as the model code changes, but I do take your point about explicit being clearer.

On the file names, it would be more systematic to have `model_name/model_schema.json` (and  also `model_name/model_class.py` for that matter, but that is less important because we access it without the filename). I think the main reason we didn't was to avoid having VSCode tabs with identical titles. Isn't it also better to be explicit about the filename rather than using magic to discover it 😈 

I had been wondering about merging the registries too. I think it makes a lot of sense for the models but the `core` complicates things - it has constants and a configuration schema but isn't a model.

jacobcook1995:
> 2. Given that we're only allowing users to register one model class per module (which I think is sensible!), you could move the model registration code out of `BaseModel.__init_subclass__`...
> 3. ...and once you've done that you will have all the model registration code in one place (i.e. for model class, schema and constants). Then you can put all of this information into a single model registry object, which should simplify things:
> 
> ```python
> @dataclass
> class ModelInfo:
>     cls: type[BaseModel]
>     schema: dict[str, Any]
>     constants: WhateverTypeThisIs
> 
> model_registry: dict[str, ModelInfo]
> """Information about models, keyed by name."""
> ```

This approach seems sensible to me, far cleaner to have one registry rather than three implicitly associated registry. I guess it does tie us to having a single constants data class per model? But that's probably not an unreasonable restriction anyway

davidorme:
The constants slot could be list of constants classes, so that's fine. 

I think we can get this to work with `core` as well but it would mean that the model registry code has to have a special case where the `ModelInfo` for `core` doesn't have a model?  It really doesn't make sense to have to look in a different location for `core` constants and schema.

alexdewar:
> It isn't really _magic_ exactly - it's just auto-discovering the components. The advantage is that `__init__` doesn't need to be updated as much as the model code changes, but I do take your point about explicit being clearer.

On reflection, I think auto-discovering the constants isn't so bad if there are likely to be multiple constants classes, though personally I think it would be cleaner if models only had one constants class and explicitly passed it to `register_model()`.

For the model class though, as there's only going to be one and you have to import it into `__init__.py` anyway you might as well pass it to `register_model()` as that's where it's going to be used. Then you won't have to suppress that `flake8` warning.

> 
> On the file names, it would be more systematic to have `model_name/model_schema.json` (and also `model_name/model_class.py` for that matter, but that is less important because we access it without the filename). I think the main reason we didn't was to avoid having VSCode tabs with identical titles. Isn't it also better to be explicit about the filename rather than using magic to discover it 😈

Hey, I didn't say I was consistent :wink:. I do think it's slightly different for file names though. If someone's looking through your code to figure out where `SomeModelClass` is instantiated, it would help them to see that it's being passed into `register_model.py`. It seems more obvious (to me, anyway) that if there are a bunch of files called `model_schema.json`, that this is a convention and that the core code is doing something special with them. But you could leave it as is if you want to be extra explicit.

I personally don't think having the files called the same thing should be much of an issue -- how often do you have multiple models' schemas open at once anyway? -- but you could stick to the convention you're already using of calling them `{model_name}_schema.json` and still have them be auto-discovered.

> 
> I had been wondering about merging the registries too. I think it makes a lot of sense for the models but the `core` complicates things - it has constants and a configuration schema but isn't a model.

I'd thought about that. I'd be inclined to just have the core constants and schema be represented by standalone variables separate from the model repository. That way you don't have to have a pseudo-model called `"core"` or whatever. For when you want to iterate over all constants classes/schemas, you could add helper functions that look in the core as well as in the models repository.

jacobcook1995:
Yeah I terms of the `core` config schema, it is already treated as a special case (it has to be as it contains a list of model schemas to load). So, it would really just be a change to treating the core constants as a special case, which isn't ideal but doesn't feel like a massive problem either

jacobcook1995:
> 2. Given that we're only allowing users to register one model class per module (which I think is sensible!), you could move the model registration code out of `BaseModel.__init_subclass__`...
> 3. ...and once you've done that you will have all the model registration code in one place (i.e. for model class, schema and constants). Then you can put all of this information into a single model registry object, which should simplify things:
> 
> ```python
> @dataclass
> class ModelInfo:
>     cls: type[BaseModel]
>     schema: dict[str, Any]
>     constants: WhateverTypeThisIs
> 
> model_registry: dict[str, ModelInfo]
> """Information about models, keyed by name."""
> ```

Okay so I've renamed all the schema files, and started supplying the Model classes directly to `register_model`. Do we think it makes sense to tackle point 2 & 3 on this PR, or leave that for a later PR?

davidorme:
I'd say a later PR.

jacobcook1995:
Okay cool, if everyone's happy with the current state I'll merge it in, and make an issue for Alex's outstanding points + the documentation issue

davidorme (approved):
LGTM

PR #299, opened 2023-08-29
updated to include hydro and abiotic_simple

vgro:
I updated the `data_variables.toml` file to the current state of the `abiotic`, `abiotic_simple`, and `hydrology` model. Some initialization might need changing based on the order in which we run models. I often included both the `abiotic` and `abiotic_simple` model in the `initialised_by` and `updated_by` section; we will need to define rules to treat these lists as `OR` and the `used_by` list as `AND` once we actually use this file. For now, it is only a collection of variables.

## Type of change

## Key checklist


virtual_rainforest/data_variables.toml
```diff
@@ -304,23 +335,23 @@ used_by = [ "abiotic", "soil"]
```
jacobcook1995, line 335: Don't know what we want to do with this `soil_respiration` variable. It isn't currently produced by the `soil` model (though it could be fairly straightforwardly), but I guess it makes sense to keep it as it is something that the full `abiotic` model would want eventually?

virtual_rainforest/data_variables.toml
```diff
@@ -304,23 +335,23 @@ used_by = [ "abiotic", "soil"]
 
 [[variable]]
 name = "soil_temperature"
-description = "soil temperature profile"
+description = "Soil temperature profile"
 unit = "C"
 initialised_by = "abiotic"
```
jacobcook1995, line 340: I guess this is also initialised by `abiotic_simple`?

virtual_rainforest/data_variables.toml
```diff
@@ -366,29 +405,37 @@ initialised_by = "abiotic"
 updated_by = "abiotic"
 used_by = [ "abiotic"]
 
+[[variable]]
+name = "vertical_flow"
+description = "Vertical flow of water through soil column"
+unit = "mm"
```
jacobcook1995, line 411: If it's a flow should the unit be distance per time?

jacobcook1995 (approved):
LGTM! Just had a few questions

virtual_rainforest/data_variables.toml
vgro: A lot of the variables are stretch goals, including all the respiration. I kept them despite knowing that we will not use them in the near future.

virtual_rainforest/data_variables.toml
vgro: Good point. At the moment, it is mm/month, but we could convert that in whatever is useful, like mean flow in m3/s

davidorme (approved):
LGTM - once this is merged I will try and remember to update for PlantsModel.

PR #303, opened 2023-09-01
Adding canopy functionality to draft PlantsModel

davidorme:
# Description

This PR adds the canopy API to the PlantsModel implementation. 

* It defines functions (in `plants/functions.py`) to:
    * Convert the plant community in a cell to a canopy model (`generate_canopy_model`) although this is a placeholder with currently fixed layer heights and LAI.
    * Assemble those models across cells into arrays of `n_canopy_layers` by `grid.n_cells` (`build_canopy_arrays`).
    * Initialise the `layer_heights` and `leaf_area_index` variables in the `data` instance.

* It extends `PlantsModel` to:
    * Add an `update_canopy_layers` method to generate the canopy layers and insert them into the heights and LAI arrays. There is a reason that this is separated out into so many parts - the stuff in `functions` is likely to get absorbed into `pyrealm` at some point, whereas this method is specifically about inserting the canopy array into the data structures used within `virtual_rainforest`.
    * in `__init__`: initialise calculate the canopy variables and update them.
    * in `update`: update the canopy layers.

* It also restructures `PlantCommunities` to be `dict` subclass - the previous version had it as a kind of cut down dict that only had `__getitem__`, which was silly.

* Adds tests for all of the above.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/plants/functions.py
```diff
@@ -0,0 +1,170 @@
...
+
+    Args:
+        community: A list of plant cohorts.
+
+    Returns:
+        A tuple of one dimensional numpy arrays giving the canopy layer heights and leaf
+        area indices.
+    """
+
+    # TODO - actually calculate these
```
jacobcook1995, line 37: I don't know how long this TODO is likely to hang around, but if this is more than a temporary thing I think it would be good to mention it in the docstring. So, that people looking at the api docs don't assume that the function is actually doing the full calculation when it isn't

virtual_rainforest/models/plants/functions.py
```diff
@@ -0,0 +1,170 @@
...
+    layer_hght = np.array([30.0, 20.0, 10.0])
```
jacobcook1995, line 38: Feel like `layer_height` is quite a lot clearer as a name for minimal extra length

virtual_rainforest/models/plants/functions.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Initialise list of arrays
+    layer_heights: list = []
+    layer_lai: list = []
+    cell_has_too_many_layers: list = []
+
+    # Loop over the communities in each cell
+    for cell_id, community in communities.items():
+        # Calculate the canopy model for the cell and pad as needed
+        this_lyr_hght, this_lyr_lai = generate_canopy_model(community)
```
jacobcook1995, line 74: Couldn't this just be `layer_height`, `layer_lai`? as I don't think these are used elsewhere in the function

jacobcook1995 (approved):
LGTM! Had a few minor comments which were mostly about naming

virtual_rainforest/models/plants/canopy.py
```diff
@@ -0,0 +1,174 @@
...
+    # TODO - These layer roles desperately need to be set up in _one_ place!
+    layer_roles = set_layer_roles(n_canopy_layers, n_soil_layers)
+    layer_shape = (len(layer_roles), data.grid.n_cells)
+
+    # Set the layers
+    data["leaf_area_index"] = DataArray(
+        data=np.full(layer_shape, fill_value=np.nan),
+        coords={
+            "layer": np.arange(len(layer_roles)),
+            "cell_id": data.grid.cell_id,
```
vgro, line 147: Can you please add a `layer_roles` coordinate here, please, this is used a lot to select layers

virtual_rainforest/models/plants/canopy.py
```diff
@@ -0,0 +1,174 @@
...
+        },
+    )
+    data["layer_heights"] = DataArray(
+        data=np.full(layer_shape, fill_value=np.nan),
+        coords={
+            "layer": np.arange(len(layer_roles)),
+            "cell_id": data.grid.cell_id,
```
vgro, line 154: see comment above

tests/models/plants/test_functions.py
```diff
@@ -0,0 +1,85 @@
+"""Tests the handler functions in models.plants.canopy."""
+
```
vgro, line 2: I think this needs to be renames to `test_canopy.py`?

tests/models/plants/test_functions.py
```diff
@@ -0,0 +1,85 @@
...
+    # Use fixture communities for now - this may need parameterised communities in the
+    # future to try and trigger various warning - or might not.
+    data = initialise_canopy_layers(plants_data, n_canopy_layers=10, n_soil_layers=3)
+
+    assert "layer_heights" in data
+    assert "leaf_area_index" in data
+
+    exp_shape = (1 + 10 + 2 + 3, len(data.grid.cell_id))
+    assert data["layer_heights"].shape == exp_shape
+    assert data["leaf_area_index"].shape == exp_shape
```
vgro, line 85: Can you make sure here that everything has all the dimensions and coordinates named correctly for other models to use?

vgro (approved):
Looks overall good to me. We need to make sure that dimensions and coordinates of your outputs match with the other models, but that should be easy to test.

I agree that we should define the layer_roles somewhere central

vgro:
One thing I forgot to mention is that the `layer_heights` variable needs to include the soil layers as well. At the moment they are not changing during the simulation, so they could just be set at fixed values. at the moment, that would be 50 cm and 1m

virtual_rainforest/models/plants/canopy.py
```diff
@@ -0,0 +1,178 @@
...
+        area indices by cell id.
+    """
+
+    # TODO - this could be a method of PlantCommunities but creates circular import of
+    #        PlantCohorts
+
+    # Initialise list of arrays
+    layer_heights: list = []
+    layer_lai: list = []
+    cell_has_too_many_layers: list = []
```
alexdewar, line 73: Adding the `: list` annotations here doesn't add anything; `mypy` knows these variables are lists. What *would* be useful would be to annotate them with the `type` of list they are, e.g. `list[int]`, which will give better type-checking and intellisense.

virtual_rainforest/models/plants/canopy.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # Loop over the communities in each cell
+    for cell_id, community in communities.items():
+        # Calculate the canopy model for the cell and pad as needed
+        this_lyr_hght, this_lyr_lai = generate_canopy_model(community)
```
alexdewar, line 78: What does `lai` stand for?

I'd spell out your var names here, e.g. `this_lyr_hght` => `layer_height`.

virtual_rainforest/models/plants/plants_model.py
```diff
@@ -108,18 +122,29 @@ def from_config(
...
+        canopy_layers = config["core"]["layers"]["canopy_layers"]
+
         # Load in the relevant constants
         constants = load_constants(config, "plants", "PlantsConsts")
 
         # Generate the flora
         flora = Flora.from_config(config=config)
 
-        # Try and create the instance - safeguard against exceptions
+        # Try and create the instance - safeguard against exceptions from __init__
```
alexdewar, line 138: Out of interest, does the code that creates model instances not already check for errors raised during this process? If not, might that be a better place to catch errors than in model classes' factory methods?

alexdewar (approved):
I've made a couple of small suggestions. Otherwise, feel free to merge!

virtual_rainforest/models/plants/canopy.py
davidorme: @vgro and @jacobcook1995  also commented on this 😄  I'm not using `layer_height` because there is also `layer_heights`. Will find something!

virtual_rainforest/models/plants/community.py
```diff
@@ -28,7 +28,7 @@ class PlantCohort:
     n: int
 
 
-class PlantCommunities:
+class PlantCommunities(dict):
```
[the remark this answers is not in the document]
davidorme: Largely because I'm about 90% sure that it will get methods in the near future. Probably something to do with herbivory.

virtual_rainforest/models/plants/plants_model.py
davidorme: Hmm - I had a reason. It felt clever at the time, but I cannot remember it for the life of me. Might be the heat. 

I think it was along the lines of wanting users to understand that the process of turning their configuration into a model instance had borked, and hopefully giving them some context on where to fix things? @vgro @jacobcook1995 @TaranRallings ?

davidorme: That's handled for use from `vr_run`, but people might be creating these models from within a Python session. I agree that is going to be much less common, but catching errors here isn't a bad thing. @alexdewar - is that what you meant? That _upstream_ code creating a suite of model instances would catch them? I thought you meant _downstream_ code in the `__init__` or somewhere in `BaseModel`?

davidorme: That was basically my thinking - make sure _something_ gets written to the log if it falls over.

davidorme: Will leave as is for the moment.

PR #304, opened 2023-09-05
Updated `mypy` and `pytest-mypy`

jacobcook1995:
# Description

On my local setup (macOS python 3.11) the `mypy` part of the unit testing takes ages to run. When I deleted the `mypy` cache files the unit test took 87.84s to run. When I updated `mypy` and `pytest-mypy` this same test dropped to 25ish seconds.

I felt like this was a big enough change to be worth updating the dependancies for (even if it only affected my setup), provided that others agree.

## Type of change

## Key checklist

- [N/A] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [N/A] All tests pass: `$ poetry run pytest`

## Further checks

- [N/A] Code is commented, particularly in hard-to-understand areas
- [N/A] Tests added that prove fix is effective or that feature works


jacobcook1995:
Made this a draft as the newest version of `mypy` does not seem to like the constants registry. I'm going to test if upgrading to 1.0.1 improves the performance, and if not I'll just have to accept that `mypy` runs very slowly on my machine

jacobcook1995:
Updating `mypy` to v1.0.1 increases the speed by less (55ish s), but doesn't break the pre-commit hook, so might be worth doing?

davidorme:
What  is the issue with the constants registry?

alexdewar:
I think upgrading `mypy` is a good idea.

One other thing you could do is telling `mypy` to exclude the `.venv` folder. We do it on other projects, though I've never actually checked whether it is actually needed and speeds things up.

jacobcook1995:
@alexdewar given we are using `poetry` I don't think we have a `.venv` folder? Or at I can't see a folder called that within the `virtual_rainforest` folder on my local machine.

Is it worth working out how to exclude `poetry` equivalents of this folder, or is safe to assume that they will be excluded anyway?

alexdewar:
@jacobcook1995 Ah, well if that's the case, then that won't be why you're seeing slowdowns. Not sure why that is. It might be worth seeing if someone's raised an issue in the `mypy` repo and, if not, raising one of your own.

(I actually do have a `.venv` folder in my repo because I've set the [`virtualenvs.in-project`](https://python-poetry.org/docs/configuration/#virtualenvsin-project) option for `poetry`.)

davidorme:
I don't think we're in general using `virtualenvs.in-project = true`, so shouldn't have any Python environments floating around inside the project directory.

davidorme:
Superseded by changes in #323

PR #309, opened 2023-09-07
Update soil depth configuration

### LINKED ISSUE #308: Configure `soil_layer` depths.

At the moment, the number of soil layers is configured, but the depths are hardcoded within a model.

The config needs to be updated to use something like `soil_layer_depths = [0.5, 1.0]` to indicate two layers at the given depths

davidorme:
# Description

This PR:

* replaces `soil_layers = 2` with `soil_layers = [0.5, 1.0]` in the configuration, to describe the number and depths of soil layers in the configuration.
* updates `set_layer_roles` to validate this, adding extra tests to check from strictly increasing positive depths, and extending the existing testing a bit.
* Updates (nearly) all the existing models to use the new config setup and updates their tests.
* Does *not* update `abiotic` - this predates `set_layer_roles` but also has `soil_layers` in the model specific configuration so is unaffected by the other changes.

Fixes #308

## Type of change

## Key checklist

## Further checks


virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,89 @@ def check_outfile(merge_file_path: Path) -> None:
...
+        :widths: 5, 10, 30
+
+        0, "above", "Canopy top height + 2 metres"
+        1, "canopy", "Height of first canopy layer"
+        "...", "canopy", "Height of canopy layer ``n``"
+        10, "canopy", "Height of last canopy layer"
+        11, "subcanopy", "1.5 metres above ground level"
+        12, "surface", "0.1 metres above ground level"
+        13, "soil", "Depth of first soil layer"
+        .., "soil", "Depth of soil layer ``m``"
```
jacobcook1995, line 82: I read the `m` here as meters, is there a less ambiguous notation?

virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,89 @@ def check_outfile(merge_file_path: Path) -> None:
...
     Args:
-        canopy_layers: number of canopy layers soil_layers: number of soil layers
+        canopy_layers: the number of canopy layers
+        soil_layers: a list giving the depth of each soil layer
 
     Raises:
-        InitialisationError: If the number soil or canopy layers are not both positive
-            integers
+        InitialisationError: If the number of canopy layers is not a positive
+            integer or the soil depths are not a list of float values.
```
jacobcook1995, line 94: Should also mention `InitialisationError("Soil layer depths must be strictly increasing and positive.")` here I guess

jacobcook1995 (approved):
LGTM! Just had a couple of minor comments about docstrings

virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,91 @@ def check_outfile(merge_file_path: Path) -> None:
...
+        15, "soil", "Depth of last soil layer"
+
+    The number of canopy layers is taken from the canopy layer argument and the number
+    of soil layers is taken from the length of the soil_layers argument, which also
+    provides soil layer depths. Both are typically set in the model configuration.
 
     Args:
-        canopy_layers: number of canopy layers soil_layers: number of soil layers
+        canopy_layers: the number of canopy layers
+        soil_layers: a list giving the depth of each soil layer
```
vgro, line 92: It would be better if the soil layers had negative values to indicate the they go in a different direction

virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,91 @@ def check_outfile(merge_file_path: Path) -> None:
...
+    if not all([isinstance(v, (float, int)) for v in soil_layers]):
+        to_raise = InitialisationError("The soil layer depths are not all numeric.")
+        LOGGER.error(to_raise)
+        raise to_raise
+
+    np_soil_layer = np.array(soil_layers)
+    if not (np.all(np_soil_layer > 0) and np.all(np.diff(np_soil_layer) > 0)):
         to_raise = InitialisationError(
-            "There has to be at least one canopy layer in the Virtual Rainforest!"
+            "Soil layer depths must be strictly increasing and positive."
```
vgro, line 125: Soil layers should be decreasing and negative based on the convention :-)

virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,91 @@ def check_outfile(merge_file_path: Path) -> None:
...
+
+        0, "above", "Canopy top height + 2 metres"
+        1, "canopy", "Height of first canopy layer"
+        "...", "canopy", "Height of canopy layer ``i`` "
+        10, "canopy", "Height of last canopy layer"
+        11, "subcanopy", "1.5 metres above ground level"
+        12, "surface", "0.1 metres above ground level"
+        13, "soil", "Depth of first soil layer"
+        "...", "soil", "Depth of soil layer ``j``"
+        15, "soil", "Depth of last soil layer"
```
vgro, line 83: This numbering is only true for the default configuration of 10 canopy layers

tests/core/test_utils.py
```diff
@@ -64,57 +64,78 @@ def test_check_outfile(caplog, mocker, out_path, expected_log_entries):
...
+        pytest.param(
+            [-0.5, 1.0],
+            10,
             pytest.raises(InitialisationError),
             (
                 (
                     ERROR,
-                    "There has to be at least one canopy layer in the Virtual "
-                    "Rainforest!",
+                    "Soil layer depths must be strictly increasing and positive.",
```
vgro, line 99: Same here, layers should be increasingly negative

virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,91 @@ def check_outfile(merge_file_path: Path) -> None:
...
-    maximum number of canopy layers (defined in config) which are filled from the top
-    with canopy node heights from the plant module (the rest is set to NaN). Below the
-    canopy, we currently set one subcanopy layer (around 1.5m above ground) and one
-    surface layer (0.1 m above ground). Below ground, we include a maximum number of
-    soil layers (defined in config); the deepest layer is currently set to 1 m as the
-    temperature there is fairly constant and equals the mean annual temperature.
+    This function creates a list of strings describing the layer roles for the vertical
+    dimension of the Virtual Rainforest. The vertical dimension consists of the
+    following layers and roles.
+
```
vgro, line 70: maybe add here a comment that above ground values are positive and below ground negative

vgro (changes requested):
Looks overall good to me, the only important point is that soil depths should be increasingly negative, not positive. They represent a value along a z axis, not the absolute layer thickness. It would be good if you could change that because it affects how the values are used

virtual_rainforest/core/utils.py
vgro: [m] ?

davidorme: Yeah - will just describe all of this better.

virtual_rainforest/core/utils.py
davidorme: And isn't even true for the default soil layers of `[-0.5, -1.0]`...

virtual_rainforest/core/core_schema.json
```diff
@@ -188,10 +188,17 @@
...
-                     "type": "integer",
-                     "exclusiveMinimum": 0,
-                     "default": 2
+                     "description": "Depth and number of soil layers to simulate",
+                     "type": "array",
+                     "items": {
+                        "type": "number"
+                     },
+                     "minItems": 1,
+                     "uniqueItems": true,
```
alexdewar, line 197: I think these should be in snake_case like your other key names.

virtual_rainforest/core/core_schema.json
```diff
@@ -188,10 +188,17 @@
...
+                     "default": [
+                        -0.5,
+                        -1.0
+                     ]
```
alexdewar, line 201: Just a thought, but I'm wondering how much sense it really makes to have a default value at all for this param? Isn't it something v simulation specific that users will basically always want to change?

I appreciate that you've similarly got defaults for things like canopy layers already, so I'm not suggesting you remove it. I just wondered whether you had a plan to remove these defaults at some point when VR becomes more than just a skeleton.

virtual_rainforest/core/utils.py
```diff
@@ -59,62 +61,104 @@ def check_outfile(merge_file_path: Path) -> None:
...
         raise to_raise
 
     layer_roles = (
         ["above"]
-        + ["canopy"] * canopy_layers
+        + ["canopy"] * int(canopy_layers)
         + ["subcanopy"]
         + ["surface"]
-        + ["soil"] * soil_layers
+        + ["soil"] * len(soil_layers)
```
alexdewar, line 161: I think these values should be encapsulated in an enum:

```py
class LayerRole(Enum):
    ABOVE = 1
    CANOPY = 2
    # etc.
```

Converting to and from strings isn't arduous and you could write helper methods if you like:

```py
# to string
print(LayerRole.ABOVE.name.lower())

# from string
val = getattr(LayerRole, some_string.upper())
```

alexdewar (changes requested):
A couple of suggestions, but otherwise LGTM.

virtual_rainforest/core/core_schema.json
davidorme: Those are JSONSchema keywords - can't do anything 😄

virtual_rainforest/core/core_schema.json
davidorme: It's a good point and I think revising the schema defaults once we have all four models running is going to be really useful. These aren't especially simulation specific: the canopy layers is a maximum and just basically sets a maximum array size, and the soil depths are fairly standard.

One thing we may want to do with canopy layers is allow it to be dynamic because the number of actually _used_ layers varies with plant community growth. But that's a whole 'nuther PR.

virtual_rainforest/core/utils.py
davidorme: I can see the point in using the `Enum` to create a tightly controlled set of values to stop users making mistakes. Is that the main motivation for using them?

All of this should really be configured and set at a much deeper level - it's basically a core axis - so needs reworking when the data configuration is built. At that point, we'd have two core axes as DataArray dimensions: the layer number `(0, ..., 14)` and the layer roles `(1, 2, 2, ....)` or possibly `(above, canopy, canopy,...)`. 

The problem then is that you can create a DataArray with enum coordinate values but they don't serialise to file.

```python
In [35]: xarray.DataArray(
    ...:     [1,2,3,4],
    ...:     coords={
    ...:         'layer_role': [
    ...:             LayerRole['ABOVE'],
    ...:             LayerRole['ABOVE'],
    ...:             LayerRole['CANOPY'],
    ...:             LayerRole['CANOPY']
    ...:         ]
    ...:     }
    ...: ).to_netcdf("test.nc")
ValueError                                Traceback (most recent call last)

...
ValueError: unable to infer dtype on variable 'layer_role'; xarray cannot serialize arbitrary Python objects
```
At this point, we have to start translating to and from different representations to use `Enum`. That does sound possible (https://github.com/pydata/xarray/issues/2292) but I'm going to park it for now.

Unless I've missed something more fundamental about switching now?

vgro: @alexdewar  I'm not sure about the advantage of this, have you seen how we currently use the layer roles? Does that work with your suggestion?

virtual_rainforest/core/core_schema.json
davidorme: Is that info documented somewhere?

Feels like we ought to put this information into the JSONSchema or somewhere to help users populate it. I can imagine a tool _waaay_ in the future that does:

```text
> So, you want to create a VR Config?
...
> Step N: Set your soil layer depths: the top layer being where microbial activity
> happens and the second layer where soil temperature equals annual mean air 
> temperature. Edit the default below:
>>> [-0.5, -1.0]
```

vgro: The lower level is documented in the microclimate docstring, this is where it is used to calculate soil temperature profiles (will also be in the abiotic model at some point when I implement the thomas algorithm). I'm not sure about the top soil, might be in the soil model. I agree it should be somewhere central now that we know where it is initialized

virtual_rainforest/core/utils.py
alexdewar: The main reason would be for safer typing. The problem with using a string is that it can be anything, so you have to remember to check in all the relevant places that the string the user has provided is a valid one. Enums are also better from a documentation standpoint.

That said, it's really not a big deal and if it requires a bunch of other changes then it's definitely not worth it -- certainly not in this PR, anyway.

davidorme: Makes a lot of sense - and I can see how they've been working over in `animals`. I'll park it for now.

virtual_rainforest/core/core_schema.json
davidorme: For the moment, I've put it in the ``set_layer_roles`` docstring, so it exists.

davidorme:
@vgro and @alexdewar - Think it is all fixed up now. Main additions to the PR are:

* Uses the negative soil depth convention.
* Updates the set_layer_roles docstring to be easier to follow.

If all looks good, can you approve?

alexdewar (approved):
LGTM. For future reference, if you need someone to re-review something you can click the "re-request review" icon next to their name in the reviewers list and it'll send a request.

tests/models/abiotic/test_radiation.py
```diff
@@ -76,7 +76,7 @@ def test_simple_atmospheric_transmissivity():
         ),
         pytest.param(
             DataArray([100, 1000], dims="cell_id"),
-            DataArray([-0.5, 1.5], dims="cell_id"),
+            DataArray([0.5, 1.5], dims="cell_id"),
```
vgro, line 79: This was negative on purpose to raise an error when the sunshine fraction is negative.

vgro (approved):
Looks all go to me now, one mini comment about the radiation test, other than that, I'm happy for you to merge

tests/models/abiotic/test_radiation.py
davidorme: Did a too broad global replace and messed up this fix.

Issue #302, opened 2023-09-01
Single registry for all models 


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
At the moment there are 3 separate registries used for model setup: `SCHEMA_REGISTRY` which contains the schema required to validate configurations for each model, `CONSTANTS_REGISTRY` which contains the constants classes for each model, and `MODEL_REGISTRY` which contains uninitialised instances of each model. 

**Describe the solution you'd like**
@alexdewar suggested consolidating these registries into a single registry:

2. Given that we're only allowing users to register one model class per module (which I think is sensible!), you could move the model registration code out of `BaseModel.__init_subclass__`...
3. ...and once you've done that you will have all the model registration code in one place (i.e. for model class, schema and constants). Then you can put all of this information into a single model registry object, which should simplify things:

```py
@dataclass
class ModelInfo:
    cls: type[BaseModel]
    schema: dict[str, Any]
    constants: WhateverTypeThisIs

model_registry: dict[str, ModelInfo]
"""Information about models, keyed by name."""
```

_Originally posted by @alexdewar in https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/295#pullrequestreview-1595301552_

**Describe alternatives you've considered**
The main thing to think about is whether `core` belongs in this combined registry. `core` will have a constants class, and the `core` schema is currently stored in `SCHEMA_REGISTRY`, **but** it doesn't define a model (because it isn't a model).

davidorme:
Just to add that when this happens, it would be good to include a logger step in `register_model`, because at the moment the log doesn't seem to note when this happens and that would be useful.

PR #310, opened 2023-09-08
Update typing of `config` arguments.

### LINKED ISSUE #294: Incorrect typing in `BaseModel.from_config` args
**Describe the bug**
The `config` argument to `core.base_model.BaseModel.from_config` is typed as a dict:

https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/173c79130acfad2dd93ec83e8a6d690cdfcf6018/virtual_rainforest/core/base_model.py#L219-L224

With the updated `Config` object, we should probably type this as a `Config` instead.

**Expected behavior**

I think this just needs an import in `core.base_model.BaseModel.from_config` and then changing the types in the BaseModel and subclass methods.

@jacobcook1995 Does that sound right?

davidorme:
# Description

This PR:
* Updates the BaseModel to type the `config` argument as the `Config` object, not just a generic `dict[str, Any]`
* Updates all of the models to also use that typing
* Updates other uses of `config` args in `core`.
* In particular `data.load_from_config` gets an overhaul: it now takes a `Config` object rather than a section from the dict  (`cfg['core']['data']`) and does some more validation of the contents, including logging a warning if the data section is there (which it is by default) but does not define any sources.
* Updates tests, docstrings and the user docs for creating new model classes.

Edited to add:

* This PR also suggested that having a direct method to create `Config` objects, without having to keep files or write Mockers to fake `Paths` or content would simplify a lot of testing. I've added a `cfg_string` argument to `Config.__init__` that provides an alternate source of the configuration TOML data. `vr_run` now also accepts this argument. 
* Updated tests of `Config` to check it works.
* Updated tests of `data` to use simpler string and remove redundant files
* Updated tests of main to allow a full config process rather than Mocking it.

Fixes #294 

## Type of change

## Key checklist

## Further checks


davidorme:
@jacobcook1995 The current failure is due to one of the `vr_run` tests. The mocker was injecting a model `topsoil` and failing it with an initialisation error. With the test now using the full config process, this fails much earlier, because there is no `topsoil` model schema, so the config can't even be validated. I think this test is redundant?

jacobcook1995:
I think that test (and the behaviour it is testing) is redundant if generating a `Config` object now leads to an error if a schema can't be found. I don't think there's any need to add another check for model `__init__` failures as there's a few already and this test was designed to specifically test the step to catch missing models.

jacobcook1995 (approved):
The changes all look good to me. I'm not sure why the `vr_run` test is failing, but that definitely needs fixing

davidorme:
> The broken test could be replaced by a test that takes in a valid config as a string, and then mocks the `MODEL_REGISTRY` to not include one of the models provided in `[core.modules]`. This would then catch the test that models that have schemas but have not been registered properly in the `MODEL_REGISTRY` are handled correctly in the main loop.

But that is then a test of a model registry issue, not of failing to initialise a properly registered model due to configuration with data that meets the schema but not the fine details. I'll add a second test of the another config issue, just for variety,

jacobcook1995:
Yeah it is though the test is a test of the `vr_run` function not a configuration specific function, so having tests that test `vr_run`'s handling of various errors is probably sensible right?

davidorme:
> However, that might be an overly complex test to add given that `select_models` (which raises the relevant exception) is already tested, and that we are planning to combine the model, schema, and constants registries (see #302 )

Yeah - I think forcing a test of behaviour that would only happen if `select_models` breaks is too complex - as you say I think this whole API is likely to change shortly.

PR #311, opened 2023-09-08
Feature/animal vr run integration

TaranRallings:
# Description

This is a series of changes to allow the animal model to run with a `vr_run` call. This includes a toml file for functional groups, schema updates, testing, and small modifications to `from_config` in `AnimalModel`.

The other change was some additional calls to `die_cohort` and a `die_cohort_community` method to ensure cohorts with no individuals are removed before they are used other methods. 

Jacob also made some changes to the litter module as part of this PR.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -97,11 +97,16 @@ def die_cohort(self, cohort: AnimalCohort) -> None:
 
         if cohort.is_alive:
             cohort.is_alive = False
-            LOGGER.debug("An animal cohort has died")
+            # LOGGER.debug("An animal cohort has died")
```
alexdewar, line 100: Did you mean to leave this commented out?

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -97,11 +97,16 @@ def die_cohort(self, cohort: AnimalCohort) -> None:
...
             self.animal_cohorts[cohort.name].remove(cohort)
         elif not cohort.is_alive:
             LOGGER.exception("An animal cohort which is dead cannot die.")
 
+    def die_cohort_community(self) -> None:
+        """This handles die_cohort for all cohorts in a community."""
+        for cohort in chain.from_iterable(self.animal_cohorts.values()):
```
alexdewar, line 107: Given that you iterate over all animal cohorts in a couple of places, how about adding a helper method/property for it?

E.g.:

```py
@property
def all_animal_cohorts(self) -> Iterable[AnimalCohort]:
    return chain.from_iterable(self.animal_cohorts.values())
```

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -187,11 +192,14 @@ def collect_prey(self, consumer_cohort: AnimalCohort) -> list[AnimalCohort]:
...
-            )
-            prey.extend(right_sized_prey)
-
+            for cohort in potential_prey_cohorts:
+                if min_size <= cohort.mass <= max_size:
+                    # Check if the cohort has zero individuals
+                    if cohort.individuals <= 0:
+                        # Call the die_cohort method here
+                        self.die_cohort(cohort)
+                        continue  # Skip this cohort and move on to the next one
```
alexdewar, line 201: I don't quite understand the logic here. Shouldn't `die_cohort` already be called when `cohort.individuals` reaches zero? Otherwise you have to wait until something tries to eat the cohort before the cleanup is performed.

And `cohort.individuals` should never be less than zero... right?

virtual_rainforest/models/animals/animal_model.py
```diff
@@ -153,16 +153,16 @@ def from_config(
 
         functional_groups_raw = config["animals"]["functional_groups"]
 
-        functional_groups = []
+        animal_functional_groups = []
+
         for k in functional_groups_raw:
-            functional_groups.append(FunctionalGroup(*k))
-        """create list of functional group objects to initialize  communities with."""
+            animal_functional_groups.append(FunctionalGroup(**k))
```
alexdewar, line 159: How about:

```py
animal_functional_groups = [FunctionalGroup(**k) for k in functional_groups_raw]

File: virtual_rainforest/models/animals/model_schema.json [hunk 5]
```diff
@@ -22,12 +22,27 @@
                             "diet": {
                                 "type": "string",
                                 "default": ""
+                            },
+                            "metabolic_type": {
+                                "type": "string",
+                                "default": ""
+                            },
```
alexdewar, line 29: What are the possible values for metabolic type? What does a type of `""` mean?

virtual_rainforest/models/animals/model_schema.json
```diff
@@ -22,12 +22,27 @@
...
+                            "birth_mass": {
+                                "type": "number",
+                                "default": 0.0
+                            },
+                            "adult_mass": {
+                                "type": "number",
+                                "default": 0.0
                             }
```
alexdewar, line 37: What does it mean if these values are zero?

virtual_rainforest/models/animals/model_schema.json
```diff
@@ -22,12 +22,27 @@
...
                         },
                         "required": [
                             "name",
                             "taxa",
```
alexdewar, line 41: Pedantic point, but should this be "taxon"? :stuck_out_tongue: 

Obvs not important.

alexdewar (changes requested):
I've made a couple of small suggestions/queries, but otherwise all good!

virtual_rainforest/models/animals/animal_communities.py
TaranRallings: I did. Right now all of the animal cohorts die very quickly and the terminal gets clogged with these death reports. I've left it in as it becomes useful again once I'm parametrizing things.

davidorme: This reminded me to start a new issue on a group wide review of log output across model:  #316

davidorme: @TaranRallings - you can also change the LOGGER level to avoid having to comment out messages. Are you calling `vr_run` from the command line? At the moment, we don't have a switch to do that. If not, you can do `LOGGER.setLevel(logging.INFO)` to mute that kind of thing.

alexdewar: I think @davidorme's approach is the best one, but it's also fine to leave as is if that's more convenient. It is only one line, after all.

I also think it's useful to make the log level configurable via an environment variable, so you don't have to keep remembering to pass in an extra command-line argument. Here's how I implemented that for another project: https://github.com/ImperialCollegeLondon/FINESSE/blob/main/finesse/logger.py

virtual_rainforest/models/animals/animal_communities.py
TaranRallings: This was meant as a temporary hacky solution but I may as well get things sorted now. 

The main issue is about which class holds what information. An `AnimalCohort` handles both the `eat` and `get_eaten` methods, which would be the natural places to check `cohort.individuals` and call `die_cohort`. The problem is that `AnimalCohort` does not know what `AnimalCommunity` it is in (to avoid the circular referencing problems) and so cannot call `die_cohort`.  

I took another look at there is a spot in `forage_community` that I believe is a more sensible place to put the cleanup. The foraging `AnimalCohort` now returns its chosen prey and the community level method can check `cohort.individuals` and call `die_cohort` if necessary. 

Does this make more sense?

virtual_rainforest/models/animals/model_schema.json
TaranRallings: haha, yes it should!

virtual_rainforest/models/animals/model_schema.json
TaranRallings: The possible values are "endothermic" and "ectothermic". None of these functional group properties should have default values.

virtual_rainforest/models/animals/model_schema.json
TaranRallings: It means I've put in a default value without thinking it through!

jacobcook1995: Yeah if you provide default values missing values get filled when the schema gets validated, if you don't then the schema validation fails if missing values are encountered for that field. I'm guessing you probably want the validation to fail if someone supplies a taxon without a body mass? If so I would just delete the default field here

virtual_rainforest/models/animals/model_schema.json
TaranRallings: I thought about this a little more and remembered why it's plural in the first place! Right now, I'm handing it a single level of classification. I'm expecting this category to expand a bit into multiple levels of classification and so left it with a plural. I'm going to leave it plural for now and we'll see what the final use of a taxa category looks like.

virtual_rainforest/models/animals/model_schema.json
alexdewar: Agreed ^^

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -135,7 +147,7 @@ def birth_community(self) -> None:
         """This handles birth for all cohorts in a community."""
 
         # Create a snapshot list of the current cohorts
-        current_cohorts = list(chain.from_iterable(self.animal_cohorts.values()))
+        current_cohorts = list(self.all_animal_cohorts)
 
         # reproduction occurs for cohorts with sufficient energy
         for cohort in current_cohorts:
```
alexdewar, line 153: Unless I'm missing something, this can just be:

```suggestion
        # reproduction occurs for cohorts with sufficient energy
        for cohort in self.all_animal_cohorts:
```

alexdewar (changes requested):
Looks good, except that I think you should remove those default values from the JSON schema to force users to provide them.

virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -187,11 +206,14 @@ def collect_prey(self, consumer_cohort: AnimalCohort) -> list[AnimalCohort]:
...
-            prey.extend(right_sized_prey)
-
+            for cohort in potential_prey_cohorts:
+                if min_size <= cohort.mass <= max_size:
+                    # Check if the cohort has zero individuals
+                    # if cohort.individuals <= 0:
+                    # Call the die_cohort method here
+                    #    self.die_cohort(cohort)
+                    #    continue  # Skip this cohort and move on to the next one
+                    prey.append(cohort)
```
alexdewar, line 216: This is totally up to you, but you *could* just return the prey items lazily rather than copying them into a `list`. It would be faster in the case where the user just wanted the first prey item, for example.

To do this, you'd drop the `prey` list altogether and use this line instead:

```suggestion
                    yield cohort
```

The return type would then be `Iterable[AnimalCohort]`.

As I said, it's fine as is, but this would be an alternative way to do it.

TaranRallings:
> Looks good, except that I think you should remove those default values from the JSON schema to force users to provide them.

Unless I have misunderstood something, I have already done this. There are currently no default values for functional group properties in `animals/model_schema.json`

alexdewar (approved):
Sorry @TaranRallings, I must have been looking at the wrong thing. LGTM :smile:

virtual_rainforest/models/animals/animal_communities.py
TaranRallings: The reason I had done it this way is that the first pass just called a birth process with no energetic gatekeeping, which caused newly born cohorts to also give birth and so on. I'll change it now.

alexdewar: Ah I see. Makes sense.

PR #314, opened 2023-09-12
Update `vr_run_cli`

### LINKED ISSUE #313: Update `vr_run_cli`  entry point

There are a few usability issues with the current `vr_run_cli` interface to the `vr_run` function:

* The `--example` option runs using the `example_data` folder _inside_ the package architecture. That is a hard location for users to find and nothing should be writing data into that folder.
* The function is using LOGGER and raising exceptions. That isn't appropriate inside a CLI entry point - at this point we are before a LOGGER is configured at all and there's nothing to handle Exceptions - it just outputs a whole bunch of complex text.
* The function args can be updated to make it easier to test.
* The logging should be redirected to a file.

* Move to an `--install_example` option which duplicates the `example_data` folder to a user chosen location. This location can then be fed straight into a `vr_run /chosen/location/vr_example`.
* Simplify error reporting to `sys.stdout`/`sys.stderr` messages.
* Update the args to facilitate testing.
* Add a log destination and handler.

davidorme:
# Description

This PR:
* Updates `vr_run_cli`:
    * Replaces `--example` with a `--install_example` step. This does mean users have to do two steps:
    
        ```sh
        vr_run --install_example /path/to/use
        vr_run /path/to/use/vr_example
        ```
        but this means that the files being used are now outside of the package tree and can be easily inspected and modified by the user.
    * Updates the messaging to just write to `stdout`/`stderr`, rather than raising long winded exceptions and writing log messages.
    * Adds an optional `arg_list` argument that can be used to invoke the function directly from tests by passing in a `sys.argv`-like list of strings. This is cleaner and much easier to debug.

*  Updates the tests `vr_run_cli` testing to:
    *  Pass the command line args via `arg_list` and then test the expected `stdout`.
    * Use TemporaryDirectory rather than using that using the `shared_datadir` fixture to do that - copies a whole bunch of unused files.
    
* Provides a new function and arguments to set a log file via `vr_run_cli` and `vr_run`. This now diverts all logging to a text file (except for the current auto model loading, which will go when #278 is fixed).

* Updates docs.

Edited to add:

* This PR revealed some issues with a previous PR that updated the Config object to accept strings  #310. I have fixed those issues alongside these changes. 

* I have also update `Config.__init__` so that `export_config` is not an automated part of the `__init__` that is controlled by rather hard to follow settings in the inputs, and instead added an explicit step within `vr_run`. This separates the programmatic use of `Config.__init__` with what we want it do inside `vr_run`.

Fixes #313

## Type of change

## Key checklist

## Further checks


davidorme:
OK - I _think_ that calling the `set_file_logging` breaks the `caplog` fixture. The file logging uses `LOGGER.propogate=False` to avoid logging messages being propagated up from the `vr` logger to the base logger and the StreamHandler attached to that. The reason is to avoid having the messages directed to file from the `vr` logger _and_ to stdout by propagation.

virtual_rainforest/entry_points.py
```diff
@@ -67,31 +69,70 @@ def _parse_command_line_params(
...
+
+    The resolved complete configuration will then be written to a single consolidated
+    config file in the output path with a default name of
+    `vr_full_model_configuration.toml`. This can be disabled by setting the
+    `core.data_output_options.save_merged_config` option to false.
+
+    Args:
+        args_list: This is a developer and testing facing argument that is used to
+            simulate command line arguments, allowing this function to be called
+            directly.
```
dalonsoa, line 103: Maybe provide an example of how this should be used. For example, if I want to request to install the example this way, what should this list be? In other words, what's the equivalent to `--install_example /path/to/copy/the/example/to`, for example?

virtual_rainforest/entry_points.py
```diff
@@ -103,34 +144,39 @@ def vr_run_cli() -> None:
...
+
+        example_dir = example_path / "vr_example"
+        if example_dir.exists():
+            sys.stderr.write(
+                "VR example directory already present in --install_example path.\n"
+            )
+
+        copytree(example_data_path, example_dir, ignore=ignore_patterns("__*"))
+        sys.stdout.write(f"Example directory created at:\n{example_dir}\n")
+        return
```
dalonsoa, line 177: I would put this in a separate function, as it is its own workflow.

virtual_rainforest/main.py
```diff
@@ -190,9 +190,10 @@ def extract_timing_details(
 
 
 def vr_run(
-    cfg_paths: Optional[Union[str, Path, Sequence[Union[str, Path]]]] = None,
-    cfg_string: Optional[str] = None,
+    cfg_paths: Union[str, Path, Sequence[Union[str, Path]]] = [],
+    cfg_strings: Union[str, list[str]] = [],
```
dalonsoa, line 194: I'm not fully sure of the use case for this. It is just to simplify testing?

dalonsoa (approved):
I've made a couple of comments, but it looks good to me!

virtual_rainforest/main.py
davidorme: Yes. All over the place we have snippets of code where we want to create a particular config. This allows us to do this without having to mock files. It just makes sense to be able to create a config without using a file.

We don't expect to do this from the command line,  so this argument isn't used in `vr_run_cli`.

jacobcook1995 (approved):
LGTM, just had a small query

tests/core/test_cli_integration.py
```diff
@@ -1,19 +1,44 @@
...
-        str(shutil.which("vr_run")),
-        "--example",
-        "--outpath",
-        str(shared_datadir),
-    ]
-    with does_not_raise():
-        with patch("virtual_rainforest.entry_points.sys.argv", args):
-            vr_run_cli()
+    with TemporaryDirectory() as tempdir:
+        with does_not_raise():
```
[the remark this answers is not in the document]
davidorme: Legacy from when this was a parameterised test that included some failing param sets.

virtual_rainforest/entry_points.py
```diff
@@ -67,31 +69,99 @@ def _parse_command_line_params(
...
+
+    example_dir = install_dir / "vr_example"
+    if example_dir.exists():
+        sys.stderr.write(
+            "VR example directory already present in --install_example path.\n"
+        )
+
+    copytree(example_data_path, example_dir, ignore=ignore_patterns("__*"))
+
+    sys.stdout.write(f"Example directory created at:\n{example_dir}\n")
```
alexdewar, line 97: Why not use `print`?

```suggestion
    print(f"Example directory created at:\n{example_dir}\n")
```

virtual_rainforest/entry_points.py
```diff
@@ -141,4 +205,10 @@ def vr_run_cli() -> None:
...
 
     # Run the virtual rainforest run function
-    vr_run(cfg_paths=cfg_paths, override_params=override_params)
+    vr_run(
+        cfg_paths=args.cfg_paths,
+        override_params=override_params,
+        logfile=args.logfile,
+    )
+
+    sys.stdout.write("VR run complete.\n")
```
alexdewar, line 214:
```suggestion
    print("VR run complete.\n")
```

virtual_rainforest/core/config.py
```diff
@@ -500,31 +521,24 @@ def load_config_toml_string(self) -> None:
...
-            )
-            LOGGER.critical(to_raise)
-            raise to_raise
+        for index, cfg_string in enumerate(self.cfg_strings):
+            # Load the contents into the instance
+            try:
+                self.toml_contents[f"cfg_string_{index}"] = tomllib.loads(cfg_string)
+            except tomllib.TOMLDecodeError as err:
+                to_raise = ConfigurationError(
+                    f"TOML parsing error in cfg_strings: {str(err)}"
```
alexdewar, line 530: I just learned you can do this with f-strings:

```suggestion
                    f"TOML parsing error in cfg_strings: {err!s}"
```

virtual_rainforest/core/logger.py
```diff
@@ -76,10 +76,45 @@
...
+
+
+def remove_file_logger() -> None:
+    """Remove an existing file logger and return to stream logging."""
+
+    # Find the file logger by name and remove it
+    vr_logfile = [
+        handler for handler in LOGGER.handlers if handler.name == "vr_logfile"
+    ]
+    LOGGER.removeHandler(vr_logfile[0])
```
alexdewar, line 117: You don't need to make a list here:

```suggestion
    vr_logfile = next(
        handler for handler in LOGGER.handlers if handler.name == "vr_logfile"
    )
    LOGGER.removeHandler(vr_logfile)
```

alexdewar (changes requested):
I've made some minor suggestions, but otherwise looks good. I do personally think the flag should be called `--install-example`, but it's not a biggie.

virtual_rainforest/entry_points.py
davidorme: I guess I was explicitly differentiating between the `stderr` and non-error messages. Is there a cost to explicitly writing to `stdout`?

virtual_rainforest/entry_points.py
```diff
@@ -67,31 +69,99 @@ def _parse_command_line_params(
...
+    parser.add_argument(
+        "--version",
+        action="version",
+        version="%(prog)s {version}".format(version=vr.__version__),
+    )
+
     parser.add_argument("cfg_paths", type=str, help="Paths to config files", nargs="*")
+
+    parser.add_argument(
+        "--install_example",
```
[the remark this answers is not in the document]
davidorme: Agreed - I think it is non-canon practice.

virtual_rainforest/core/config.py
davidorme: I guess there's always one more formatting syntax! I think the existing code is clearer, but that might just be ignorance of the format tricks.

virtual_rainforest/entry_points.py
alexdewar: There's not a cost to writing to `stdout` directly, but it's a bit odd from a stylistic point of view. Unless you've overridden the default `print()` function, it'll do the same thing.

virtual_rainforest/core/config.py
alexdewar: Yeah I don't think it's much clearer. But I thought I'd show you the trick anyway :stuck_out_tongue:. You can also write `{thing!r}` instead of `{repr(thing)}`.

virtual_rainforest/entry_points.py
```diff
@@ -94,7 +94,7 @@ def install_example_directory(install_dir: Path) -> None:
 
     copytree(example_data_path, example_dir, ignore=ignore_patterns("__*"))
 
-    sys.stdout.write(f"Example directory created at:\n{example_dir}\n")
+    print(f"Example directory created at:\n{example_dir}\n")
```
alexdewar, line 97:
```suggestion
    print(f"Example directory created at:\n{example_dir}")
```

virtual_rainforest/entry_points.py
```diff
@@ -211,4 +211,4 @@ def vr_run_cli(args_list: Optional[list[str]] = None) -> None:
         logfile=args.logfile,
     )
 
-    sys.stdout.write("VR run complete.\n")
+    print("VR run complete.\n")
```
alexdewar, line 214:
```suggestion
    print("VR run complete.")
```

alexdewar (approved):
Aside from a couple of trailing `\n`s, it LGTM.

davidorme:
@alexdewar - I forgot to `return` after failure messages, so added those. Should those be `sys.exit(1)` though, given it is a CLI?

alexdewar:
@davidorme The cleanest way to do this is probably to return an `int` from your main function (e.g. `return 1` for failure) which will be used as the return code.

PR #318, opened 2023-09-27
soil evaporation constraint

vgro:
This PR makes sure that soil moisture does not fall below residual soil moisture. This was likely the cause of this error occurring during `vr_run` : 

path/virtual_rainforest/virtual_rainforest/models/hydrology/below_ground.py:79: RuntimeWarning: invalid value encountered in sqrt* np.sqrt(effective_saturation)

I didn't add a specific test or warning message.

## Type of change

## Key checklist

## Further checks


davidorme (approved):
LGTM - does that change in the test catch the case where the evaporation is constrained not to push SM below RSM? Might be useful to note that?

vgro:
> LGTM - does that change in the test catch the case where the evaporation is constrained not to push SM below RSM? Might be useful to note that?

It didn't because I made the change in the `update` function. I added the clip to the function that calculates evaporation  so I can test for it and it looks like it works ok.

PR #319, opened 2023-09-28
Feature/rainfall stochasticity

vgro:
This PR introduces stochasticity in the daily rainfall distribution. It is currently very simple and does not represent extreme events or similar, in the future this could be replaced by some distribution based on observed rainfall patterns.

I set the random seed in the function, we will need to move this to core at some point to make sure we all use the same. 

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/hydrology/above_ground.py
```diff
@@ -272,3 +272,36 @@ def estimate_interception(
...
+    rainfall patterns.
+
+    Args:
+        total_monthly_rainfall: Total monthly rainfall, [mm]
+        num_days: Number of days to distribute the rainfall over
+
+    Returns:
+        An array containing the daily rainfall amounts, [mm]
+    """
+    np.random.seed(42)  # TODO move this to core for all models to be the same
```
jacobcook1995, line 294: This feels a bit weird to me in that it means that the function will always have the same result (provided that it is called with the same input), which I don't think is really what you want. Does it make sense to move this seed to the test?

tests/models/hydrology/test_above_ground.py
```diff
@@ -215,3 +215,18 @@ def test_estimate_interception():
...
     np.testing.assert_allclose(result, exp_result)
+
+
+def test_distribute_monthly_rainfall():
+    """Test that randomly generated numbers are reproducible."""
+    from virtual_rainforest.models.hydrology.above_ground import (
+        distribute_monthly_rainfall,
+    )
+
+    monthly_rain = np.array([0.0, 20.0, 200.0])
```
jacobcook1995, line 226: If you moved the seed here then I think the following would test what you are currently testing:
```python
np.random.seed(42)
result = distribute_monthly_rainfall(monthly_rain, 10)
np.random.seed(42)
result1 = distribute_monthly_rainfall(monthly_rain, 10)
```

jacobcook1995 (changes requested):
I found the placement of the random seed set a bit odd and think it probably should be moved (unless I've misunderstood the intent of the function).

virtual_rainforest/models/hydrology/above_ground.py
vgro: It's very likely that I misunderstood, I thought we want simulations to be reproducible and the stochasticity to be somewhat deterministic? I just didn't want to have the exact same rainfall every day. Or do we accept that we always get slightly different results every time we run the vr?

jacobcook1995: Ahh yeah, I guess it's fine for what you are doing at the moment. Just in the end I think we want functions like this to generate a random result, but the overall simulation flow to be reproducible by setting the seed in (I think) `vr_run`

jacobcook1995 (approved):

vgro:
I agree, that's why I added the TODO here, based on our discussion yesterday.

virtual_rainforest/models/hydrology/above_ground.py
```diff
@@ -272,3 +272,36 @@ def estimate_interception(
...
+
+    daily_rainfall_data = []
+    for total_monthly_rainfall in total_monthly_rainfall:
+        daily_rainfall = np.zeros(num_days)
+
+        for _ in range(int(total_monthly_rainfall)):
+            day = np.random.randint(0, num_days)  # Randomly select a day
+            daily_rainfall[day] += 1.0  # Add 1.0 mm of rainfall to the selected day
```
davidorme, line 302: I think this is just going to converge on a uniform distribution? It doesn't matter greatly what happens here for now, but do we want something more episodic?

The `ingestr` package has a function to go fro CRU monthly to daily values, which they adopted from the LPX model:
https://github.com/geco-bern/ingestr/blob/master/R/ingest_globalfields.R#L1196

Honestly, the commenting on that (and the original LPX source) make it a little hard to follow what is going on!

virtual_rainforest/models/hydrology/above_ground.py
davidorme: Also, this is a legacy function. We probably should be using:

```python
rng = np.random.default_rng(seed=42)
```

And then:

```python
rng.random.randint(0, num_days)
```

davidorme (changes requested):
I don't think we should implement the full random number generator solution on this PR but I think it is worth moving closer to it? I think having `seed` as an optional argument to  `distribute_monthly_rainfall` adds that route to fix or randomise the output. 

Handily, the [default_rng](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng) constructor accepts `None` or an `int`, so the function could have a sig:

```python
def distribute_monthly_rainfall(
    total_monthly_rainfall: NDArray[np.float32],
    num_days: int,
    seed: Optional[int] = None,
)

and then use:

```python
rng = np.random.default_rng(seed)
```

The in testing we could do:

```
result = distribute_monthly_rainfall(monthly_rain, 10, 42)
result1 = distribute_monthly_rainfall(monthly_rain, 10, 42)
```
Ultimately, we'd want a way to feed that seed from config. The actual generated patterns feel like they'll be too uniform - I'm not sure what is best to do otherwise though.

vgro:
> I don't think we should implement the full random number generator solution on this PR but I think it is worth moving closer to it? I think having `seed` as an optional argument to `distribute_monthly_rainfall` adds that route to fix or randomise the output.
> 
> Handily, the [default_rng](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng) constructor accepts `None` or an `int`, so the function could have a sig:
> 
> ```python
> def distribute_monthly_rainfall(
>     total_monthly_rainfall: NDArray[np.float32],
>     num_days: int,
>     seed: Optional[int] = None,
> )
> 
> and then use:
> 
> ```python
> rng = np.random.default_rng(seed)
> ```
> 
> The in testing we could do:
> 
> ```
> result = distribute_monthly_rainfall(monthly_rain, 10, 42)
> result1 = distribute_monthly_rainfall(monthly_rain, 10, 42)
> ```
> 
> Ultimately, we'd want a way to feed that seed from config. The actual generated patterns feel like they'll be too uniform - I'm not sure what is best to do otherwise though.

This is a very good suggestion, I'll add the seed as an argument.

I am aware that the distribution is not realistic at the moment but I didn't want to do a deep dive into the SAFE data to extract the rainfall pattern and create a function that randomly generates rainfall based on some provided statistics, like rainfall days etc. The function you suggested looks interesting, but for now, I would just like to have something in place that gives some daily values.

virtual_rainforest/models/hydrology/above_ground.py
vgro: I agree that this is very simple, but I think for now it works and I can add a more advances method later. Would you agree?

davidorme (approved):
LGTM

PR #322, opened 2023-09-29
Feature/plants model

davidorme:
# Description

This integrates the `plants` model data into the `animals` model. The specific changes:

* Adds a placeholder `plants` model canopy data array containing fictitious `layer_leaf_mass` by canopy layer. This is _community wide_ and eventually we will likely need cohort specific `cohort_leaf_mass` to allow selective foraging. Either way, the `plants` model now populates something that can be eaten.
* There is now a `plant_resources` submodule providing `PlantResources`. This is very like the old `PlantCommunity` in that it implements the `Resource` protocol from `animal_model.protocols` but it no longer has `grow` and `die`  methods, because they are never used.
* `PlantResources` instances are created inside `AnimalCommunity.forage_community` method. They are no longer an instance attribute added at `__init__`, because they have to be updated from the `Data` instance at each time step.
* As a result `AnimalCommunity` now accepts the `data` argument and stores that `Data` instance as an attribute so that `forage_community` can access the data and hence the plant resources can be populated.
* At the moment, this only exposes that single `layer_leaf_mass` resource, but this will expand as the `plants` model adds more resources.  The leaf mass is currently summed to give a total mass to eat, but we can change that when foraging becomes vertically stratified.
* There is no feedback of herbivory to the plant model. This needs fixing but it is a Milestone 2 extension to the `plants` model.

I've also:
* Done a bit of docstring restructuring
* Fixed up the tests
* Moved the `animals` model specific test fixtures into `models/animals/conftest.py` to keep things tidy.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/animals/animal_communities.py
```diff
@@ -163,7 +177,12 @@ def forage_community(self) -> None:
 
 
         """
-        plant_list = [self.plant_community]
+        # Generate the plant resources for foraging.
+        plant_community: PlantResources = PlantResources(
+            data=self.data, cell_id=self.community_key
+        )
+
```
TaranRallings, line 184: Just making sure that I understand what is going on here - this is generating a new `PlantResource` object every time `forage_community` is called so that it can be initialized with the current state of data? 

Is this a workaround so that feedbacks can be avoided for now?

On runs, should I then expect animals to treat plants as unlimited until such a time where they are completely consumed each tick?

TaranRallings (approved):
The animal integration looks good to me. As I mentioned, I am in the middle of converting the animal module to run on mass instead of energy. Would you rather make the mass change on `PlantResources` yourself, or should I switch it over once you've pulled it through? I'm happy either way.

virtual_rainforest/models/animals/animal_communities.py
davidorme: That's what it is doing and it is not a workaround. The `plants` model is not accessible in an object oriented way to the `animals` model, so the `PlantResources` will always be an interface to the data. The `PlantResources` just exposes the current data in a way that is compatible with your protocols for `animals` model. What will need to happen is that `PlantResources` writes back data for herbivory, that the `plants` model can then integrate. 

There is a timing issue there, but we can probably estimate GPP as some kind of average of the original leaf mass and the remaining leaf mass. That's Milestone 2 though.

davidorme:
How far off are you on the mass re-org?

virtual_rainforest/models/animals/animal_communities.py
davidorme: > On runs, should I then expect animals to treat plants as unlimited until such a time where they are completely consumed each tick?

It isn't unlimited because the energy gets consumed as with your other resources. It's just that the consumption doesn't feedback to `plants` so that boatload of leaf mass will currently reappear magically at the next tick.

virtual_rainforest/models/plants/canopy.py
```diff
@@ -20,7 +20,7 @@
 
 def generate_canopy_model(
     community: list[PlantCohort],
-) -> tuple[NDArray, NDArray, NDArray]:
+) -> tuple[NDArray, NDArray, NDArray, NDArray]:
```
alexdewar, line 23: This return value is getting a bit cumbersome. Would it be worth creating a new dataclass object to house these values so they can be named? I know it's a bit more boilerplate, but I think it might make the interface cleaner.

virtual_rainforest/models/plants/canopy.py
```diff
@@ -45,28 +45,32 @@ def generate_canopy_model(
...
+    layer_leaf_mass = np.array([10000.0, 10000.0, 10000.0], dtype=np.float32)
 
-    return layer_heights, layer_leaf_area_indices, layer_fapar
+    return layer_heights, layer_leaf_area_indices, layer_fapar, layer_leaf_mass
 
 
 def build_canopy_arrays(
     communities: PlantCommunities, n_canopy_layers: int
-) -> tuple[NDArray, NDArray, NDArray]:
+) -> tuple[NDArray, NDArray, NDArray, NDArray]:
```
alexdewar, line 73: Ditto above.

virtual_rainforest/models/plants/canopy.py
```diff
@@ -79,17 +83,21 @@ def build_canopy_arrays(
         n_canopy_layers: The maximum number of permitted canopy layers.
 
     Returns:
-        A tuple of two dimensional numpy arrays giving the canopy layer heights and leaf
-        area indices by cell id.
+        A tuple of two dimensional numpy arrays giving the canopy layer heights and then
+        leaf area indices, fapar and leaf mass by cell id.
```
alexdewar, line 87: Am I right in thinking FAPAR is an acronym?

```suggestion
        leaf area indices, FAPAR and leaf mass by cell id.
```

alexdewar (changes requested):
A few small questions/suggestions, but otherwise LGTM. Can't comment on the science, but as long as @TaranRallings is happy, that's the main thing.

virtual_rainforest/models/plants/canopy.py
davidorme: Hah - yes it is. It is usually written as fAPAR because it is the fraction of APAR ($f_{APAR}$).

virtual_rainforest/models/plants/canopy.py
davidorme: Yeah - I think that makes sense, particularly since this is going to get more cumbersome as more resources (fruit mass, etc) are added.

davidorme: Actually - this probably is the right way to go, but things are going to change radically as this model introduces some actual science and doesn't  just assert values, and this would make the internals of the functions more complex. So I'm going to park it for now.

davidorme:
I've parked the canopy dataclass suggestion for now - I think it is likely to all change fairly soon.

alexdewar (approved):

PR #323, opened 2023-09-29
Moving to a single registry and importing on request not automatically

### LINKED ISSUE #124: Timing of schema parsing in config
**Describe the bug**

At present, the `@register_schema` decorator causes a schema file to be read in (more than once, as it happens) and then validated _at package startup_. Errors in a schema then cause `vr` to fail to import, which is not friendly and also crashes the test suite.

**To Reproduce**

1. Edit one of the schema json files and change e.g. `"type": "object"` to `"type": "bobject"`

1. Open python and `import virtual_rainforest`

```python
In [1]: import virtual_rainforest as vr
[CRITICAL] - logger - log_and_raise(79) - Module schema core not valid JSON!
SchemaError                               Traceback (most recent call last)
...
OSError: Module schema core not valid JSON!
In [2]: 
```

**Expected behavior**

This should throw an exception, but only when the faulty schema is _requested_, not at start time.  The schema registry should be populated with a set of paths (one for each model name) - and path validation and then schema validation should only happen when that file is needed.

One other config thing, when this is being addressed:

https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/develop/virtual_rainforest/core/config.py#L478-L480

This is a string that is guaranteed to be `properties` or `blah/blah/blah/properties`, there are simpler ways to truncate!

### LINKED ISSUE #278: Importing `virtual_rainforest` package always causes models to be loaded

davidorme:
# Description

Issues #278 and #302 request that modules are loaded on demand and not automatically and that the existing registries for models, constants and schema are integrated. Those are difficult to solve in isolation, so this does both.

This PR is currently a work in progress. I think it fully implements the requested functionality but it does nothing to update the tests or docs to match. If the changes makes sense then we can then roll out those additional commits.

This also fixes #124 - bad schemas will now fail at `register_module` rather than cause the package import to fail.

## Overview

* Two new `core` submodules are added and one is renamed:
    * `schema.py` simply moves the JSONSchema loading and merging code out of `config` to simplify the file structure.
    * `registry.py` collects and updates the existing registration functions from `base_model` (`register_model`), `constants` (`register_constant_class`) and `config` (`register_schema`). It then adds an umbrella function `register_module` that wraps those and the `MODULE_REGISTRY` itself.
    * In the models, the `constants.py` submodule is used as the home for constants dataclasses, and is used in registering those classes.  We expect to want to do that in `core` too,  so the existing functionality in`core.constants` has moved to `core.constants_loader`.

* The `MODULE_REGISTRY` includes `core` as well as models. The registry is structured as a dict keyed by module short  name, each of which contains a dict of with keys `['model', 'constants', 'schema']`. The entries for each of those elements is identical to the previous `..._REGISTRY` object. So, for example, the plants model is at `MODULE_REGISTRY['plants']['model']`. The `core` module does not have an entry for `model`.

* The `register_module` function automatically imports the requested module and then searches for a single BaseModel subclass (except in `core`), any data classes in `xyz.constants` and then the appropriate JSONSchema. Because it searches for the BaseModel subclass using `inspect`, individual models now only have to import that object in `__init__`, and all the registration code in `__init__` files is removed. 

* Within the `vr_run` flow, models cannot be registered until the model config is built and merged. At that point, we should have the model list, so modules need to be registered then before we can validate the config, which is inside `Config.build_schema`. So that method now calls `register_module('virtual_rainforest.core')` and then `register_module('virtual_rainforest.model.xyz')` for the requested models right before building the schema.

* I've then stripped out all of the previous autodiscovery code and all of the apparatus around the previous `SCHEMA_REGISTRY`, `MODEL_REGISTRY` and `CONSTANTS_REGISTRY`.

So - this does run using `vr_run` - the mechanisms seem to work as expected, but please give this WIP a solid kicking before we do anything rash like testing or documenting it. There are a lot of changes but a lot of it is just moving code locations.

Fixes #278 and #302 

## Type of change

## Key checklist

- [ VERY MUCH NO ] All tests pass: `$ poetry run pytest`

## Further checks

- [ AGAIN, NO ] Tests added that prove fix is effective or that feature works


davidorme:
Just to demonstrate the registry structure and mechanism, which we'll need to work into the tests:

```python
In [1]: from virtual_rainforest.core.registry import register_module, MODULE_REGISTRY
In [2]: register_module('virtual_rainforest.models.plants')
[INFO] - registry - register_module(72) - Registering virtual_rainforest.models.plants module components
[INFO] - registry - register_model(145) - Registered class for plants model: PlantsModel
[INFO] - registry - register_schema(179) - Schema registered for module plants: /Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/models/plants/model_schema.json 
[INFO] - registry - register_constants_classes(222) - Constants class plants.PlantsConsts registered
In [3]: register_module('virtual_rainforest.core')
[INFO] - registry - register_module(72) - Registering virtual_rainforest.core module components
[INFO] - registry - register_schema(179) - Schema registered for module core: /Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/core_schema.json 
In [4]: MODULE_REGISTRY
{'core': {
    'constants': {},
    'schema': {'properties': {'core': ...}},
    },
 'plants': {
     'constants': {'PlantsConsts': <class 'virtual_rainforest.models.plants.constants.PlantsConsts'>},
     'model': <class 'virtual_rainforest.models.plants.plants_model.PlantsModel'>,
     'schema': {'properties': {'plants': ... }}
}
```

davidorme:
## Point of concern

This revision does not populate the registry for a module until that `register_module` function is called - and importing a module doesn't call that.  That works fine within the context of `vr_run`, when the sequence of events is enforced, but I might have pulled this back too centrally, and we still want `register_module` to be run when a module is imported?

I think that would also likely reduce the amount of `register_module` calls required in testing too?

virtual_rainforest/models/abiotic_simple/__init__.py
```diff
@@ -19,9 +19,6 @@
 
 """  # noqa: D205, D415
 
-from virtual_rainforest.core.base_model import register_model
-from virtual_rainforest.models.abiotic_simple.abiotic_simple_model import (
+from virtual_rainforest.models.abiotic_simple.abiotic_simple_model import (  # noqa F401
```
jacobcook1995, line 22: Is this line still required?

virtual_rainforest/models/animals/__init__.py
```diff
@@ -27,7 +27,4 @@
   the animal module.
 """  # noqa: D205, D415
 
-from virtual_rainforest.core.base_model import register_model
-from virtual_rainforest.models.animals.animal_model import AnimalModel
-
-register_model(__name__, AnimalModel)
+from virtual_rainforest.models.animals.animal_model import AnimalModel  # noqa: F401
```
jacobcook1995, line 30: Same question for `AnimalModel` does it still need to be imported (and I assume for every other model `__init__.py`)?

jacobcook1995 (commented):
The change seems sensible to me!

I do wonder whether `core` should really be contained in the registry though as it kind of creates the impression that `virtual_rainforest.core` is a module that could be switched out for an alternative "core", which I don't think is something we would ever want to support? I guess the core constants existing outside the registry would be a bit weird though?

virtual_rainforest/models/animals/__init__.py
davidorme: That is basically my point of concern above - I've moved _all_ of the registration process inside the config `build_schema` method, and maybe that should just import the required modules and leave the `register_module` call in the model `__init__` so that they also get registered when imported?

davidorme:
> I do wonder whether `core` should really be contained in the registry though as it kind of creates the impression that `virtual_rainforest.core` is a module that could be switched out for an alternative "core", which I don't think is something we would ever want to support? I guess the core constants existing outside the registry would be a bit weird though?

I did think about having a separate CORE_REGISTRY, but really having a single place to look for module components seemed like less code to maintain. I have just been wondering if we want a `core` model to hold config validation stuff, but that would really be confusing the use cases!

virtual_rainforest/core/constants_loader.py
```diff
@@ -108,14 +66,25 @@ def load_constants(config: Config, model_name: str, class_name: str) -> Any:
...
-
-    # Check if any constants have been supplied
-    if model_name in config and "constants" in config[model_name]:
+    try:
+        constants_class = MODULE_REGISTRY[model_name]["constants"][class_name]
+    except KeyError as excep:
+        LOGGER.critical(f"Unknown constants class: {model_name}.{class_name}")
+        raise excep
+
+    # Check if any constants have been supplied for this constants class
```
alexdewar, line 75: Is it possible to create a constants class which doesn't supply constants?

virtual_rainforest/core/registry.py
```diff
@@ -0,0 +1,222 @@
...
+The registry is a dictionary keyed by the final part of the module name. The entry for
+each module is itself a dictionary with 'model', 'constants' and 'schema' entries. For
+example:
+
+MODULE_REGISTRY['plants']['model'] = <virtual_rainforest.models.plants.PlantsModel>
+MODULE_REGISTRY['plants']['constants'] = [
+    'PlantsConsts': <virtual_rainforest.models.plants.constants.PlantsConsts>
+]
+MODULE_REGISTRY['plants']['schema'] = {schema_dict}
+"""
```
alexdewar, line 44: If each element of `MODEL_REGISTRY` is a dict with the same members, couldn't you use a dataclass for this instead? That way you get better typing etc.

virtual_rainforest/core/schema.py
```diff
@@ -0,0 +1,148 @@
...
+
+from virtual_rainforest.core.logger import LOGGER
+
+
+def set_defaults(
+    validator: type[Draft202012Validator],
+    properties: dict[str, Any],
+    instance: dict[str, Any],
+    schema: dict[str, Any],
+) -> Iterator:
```
alexdewar, line 27: What is this an `Iterator` of? Can the type be made more specific?

virtual_rainforest/core/schema.py
```diff
@@ -0,0 +1,148 @@
...
+    p_paths = []
+    # Recursively search for all instances of properties in the schema
+    for path, value in dpath.util.search(comb_schema, "**/properties", yielded=True):
+        # Remove final properties instance from path so that additionalProperties ends
+        # up in the right place
+        p_paths.append("" if path == "properties" else path[:-11])
+
+    # Set additional properties to false everywhere that properties are defined
+    for path in p_paths:
+        dpath.util.new(comb_schema, f"{path}/additionalProperties", False)
```
alexdewar, line 146: Could these two loops be combined? Then you could drop `p_paths`

virtual_rainforest/main.py
```diff
@@ -68,7 +71,7 @@ def select_models(model_list: list[str]) -> list[Type[BaseModel]]:
         raise to_raise
 
     # Then extract each model from the registry
-    modules = [MODEL_REGISTRY[model] for model in unique_models]
+    modules = [MODULE_REGISTRY[model_name]["model"] for model_name in unique_models]
 
     return modules
```
alexdewar, line 76:
```suggestion
    return [MODULE_REGISTRY[model_name]["model"] for model_name in unique_models]
```

virtual_rainforest/models/animals/__init__.py
alexdewar: I think it might be nice to make this more explicit and keep the call to `register_model` (or whatever you want to call it).

I'd also make the model class type an argument, so that users have to actually *use* the imported class somewhere. This means a) they don't have to suppress any warnings (which is a bit gross) and b) they won't forget to import the model class, because they need it as an argument and c) it's obvious that the purpose of importing the class is for model registration.

alexdewar (changes requested):
Definitely an improvement. Good work!

I do have some suggestions and queries though.

virtual_rainforest/core/registry.py
```diff
@@ -0,0 +1,222 @@
...
+    # Try and get the constants submodule and simply leave if none is present
+    constants_submodule = getattr(module, "constants", None)
+
+    if constants_submodule is None:
+        return
+
+    # Otherwise get any constant dataclasses and register them
+    constants_classes = (
+        class_ for class_ in getmembers(constants_submodule) if is_dataclass(class_[1])
+    )
```
dalonsoa, line 209: What if a module have dataclasses that have nothing to do with constants in the VirtualRainforest sense?

virtual_rainforest/core/registry.py
```diff
@@ -0,0 +1,222 @@
...
+
+    for class_name, class_obj in constants_classes:
+        if class_name in MODULE_REGISTRY[module_name_short]["constants"]:
+            excep = ValueError(
+                f"Constants class {module_name_short}.{class_name} already registered"
+            )
+            LOGGER.critical(excep)
+            raise excep
```
dalonsoa, line 217: So, just to confirm I get this right: a module can only have one model (with a model_name equal to the short module name) but can have any number of dataclasses defining constants, correct?

dalonsoa (commented):
I've given a first pass and mentioned a couple of things than look a bit odd to me. Maybe I've missunderstood somehinght.

Other than that, I think loading models on-demand as needed by the simulation to be run is a good move and much cleaner than loading everything. So, while there might be things to polish on the implementation side, my 👍 on purpose of the PR.

virtual_rainforest/core/registry.py
davidorme: Meant to raise this point - you had this structure in an earlier post. The problem I ran into here is that importing those types to construct the dataclass causes some horrific circularity issues. Part of the reason I have separated `schema` and `registry` here is that they are then very low level and don't require much else from VR. Once you import `BaseModel` or things like it then I was getting circular imports all over the place.

It would certainly be much neater to type this more explicitly.

alexdewar: Couldn't the relevant types be put into their own `.py` files then? I can't imagine the `MODEL_REGISTRY` would be needed in those files, so there shouldn't be any circular import problems.

virtual_rainforest/core/constants_loader.py
davidorme: Badly phrased comment - the config might not supply values, in which case defaults are used.

virtual_rainforest/core/schema.py
davidorme: It iterates over the iterates over the entries in a JSONSchema document - it isn't clear from the `python-jsonschema` docs what the types are.

virtual_rainforest/core/registry.py
davidorme: That's a good point - the intention is that the `xyz.constants` submodules should only ever contain 'constants' dataclasses. However we could create a subclass of `dataclass` that makes this explicit?

dalonsoa: I think that would be the way to go. Forbidding users to have other dataclasses in their module to do other stuff so that VR works seems rather arbitrary.

virtual_rainforest/core/registry.py
dalonsoa: I thought I have mentioned this somewhere, but you can simply do `model_name = Path(__file__).stem` and you automatically get the model name equal to the module name without telling the user what to do neither having to check if they are the same, as it is done somewhere else.

virtual_rainforest/core/schema.py
davidorme: Merging the loops results in the code trying to edit the dictionary whilst it is being iterated over, and it objects. But it can be expressed probably more cleanly by generating the list first and then looping over doing the edits.

virtual_rainforest/core/config.py
```diff
@@ -625,18 +464,22 @@ def build_schema(self) -> None:
...
         all_schemas: dict[str, Any] = {"core": core_schema}
         for module in requested_modules:
             # Trap unknown model schemas
-            if module not in SCHEMA_REGISTRY:
+            if module not in MODULE_REGISTRY:
                 to_raise = ConfigurationError(
                     f"Configuration contains module with no schema: {module}"
                 )
                 LOGGER.error(to_raise)
                 raise to_raise
```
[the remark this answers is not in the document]
davidorme: That's correct. The `register_schema` code handles missing files, JSON coding errors and some basic schema checks, so this should never be needed.

virtual_rainforest/core/registry.py
```diff
@@ -0,0 +1,222 @@
...
+
+    if (
+        module_name_short in MODULE_REGISTRY
+        and "schema" in MODULE_REGISTRY[module_name_short]
+    ):
+        excep = ValueError(
+            f"The module schema for {module_name_short} is already registered"
+        )
+        LOGGER.critical(excep)
+        raise excep
```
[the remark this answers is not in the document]
davidorme: A good point - this is a hangover from the separate registries, which I kept as a kind of guard against incomplete registration. But - that shouldn't happen unless someone tries to run the schema function independently of the register_submodule.

I think this all gets a lot cleaner if I can get the dataclass working, because then register_module can atomically add a module dataclass, and the deeper structure is guaranteed.

virtual_rainforest/core/schema.py
```diff
@@ -0,0 +1,148 @@
...
+        comb_schema["properties"][this_module] = this_schema["properties"][this_module]
+        # Add module name to list of required modules
+        comb_schema["required"].append(this_module)
+
+    p_paths = []
+    # Recursively search for all instances of properties in the schema
+    for path, value in dpath.util.search(comb_schema, "**/properties", yielded=True):
+        # Remove final properties instance from path so that additionalProperties ends
+        # up in the right place
+        p_paths.append("" if path == "properties" else path[:-11])
```
dalonsoa, line 142:
```suggestion
        p_paths.append(path.replace("properties", ""))
```

virtual_rainforest/core/schema.py
dalonsoa: I think that will be much neater, to be honest.

virtual_rainforest/core/registry.py
davidorme: The issue is that `BaseModel` has a whole bunch of higher level imports:

```
from virtual_rainforest.core.axes import AXIS_VALIDATORS
from virtual_rainforest.core.config import Config
```

All of those structures come after configuration, so the circular imports from importing `BaseModel` seem really hard to avoid.

virtual_rainforest/core/registry.py
davidorme: Actually, we could do this in `BaseModel.__init_subclass`:

```python
setattr(cls, "model_name", cls.__module__.split(".")[2])
```

That takes, for example `virtual_rainforest.models.plants.plants_model` and enforces the class model_name without needing to add it to each model subclass.

virtual_rainforest/models/animals/__init__.py
davidorme: I'm not so bothered about all of that - the documentation is clear on what is needed and why - but I think the main thing for restoring that structure  is that it us just much cleaner if importing a component of a model triggers registration, as opposed to then having to run `register_module` as a second step.

virtual_rainforest/core/registry.py
dalonsoa: I'm not sure if this will be picked by `mypy`, though...

virtual_rainforest/core/registry.py
davidorme: I can't see any reason why you couldn't just put your dataclass in another submodule, but having an explicit ConstantsDataclass base class seems clearer and we can give it shared methods.

virtual_rainforest/core/registry.py
davidorme: I'll see if it works, but probably not in this PR!

virtual_rainforest/core/registry.py
dalonsoa: Filtering using generic Python objects, which is what you are doing in essence, is bound to create conflicts as chances are you will pick things that you don't want. It's like forbidding to use integers in a module because you are using integers for some very specific application of your tool. If you have a custom class and filter by that class, everyone is happier and has more freedom to do whatever they want.

virtual_rainforest/core/schema.py
davidorme: The `path` variable contains strings with multiple `properties` occurrences - e.g. `properties/core/properties` - so this doesn't work and we have a special case with the first occurrence missing the leading '/' so can't use `str.removesuffix` easily either.

dalonsoa: Why is there a special case?

davidorme: Because we want  the `properties` to be used to add `/additionalProperties` and (e.g.) `properties/core/properties` to add `properties/core/additionalProperties`, so in one case we remove 'properties' and in all other cases '/properties'. I mean we _could_ regex it, but that seems like overkill.

davidorme: This would be a whole bunch simpler if the JSONSchema spec or the `python-jsonschema` implementation allowed an easy way to enforce `additionalProperties=false` globally. I suspect we could extend the `Validator` to do that but I couldn't find an obvious recipe online.

davidorme:
## Round 2

OK - so this is a major revision of the WIP draft.

* The `core.registry` module now uses that `ModuleInfo` dataclass. Because those are added to the `MODULE_REGISTRY` only when all components have been checked, a _lot_ of boilerplate checking of dictionary key structures in the various `register_*` functions dropped away, and the whole thing reduces to a reasonably compact single `register_module` function.

* The module registration step has been moved back into the `MyModel.__init__.py` files, so that importing the module runs the registration without the need to also run `register_module`. This is _not_ true for `core` for the relatively trivial reason that it causes the `LOGGER` to emit registration messages before you can do pretty much _anything_ with `vr`. Having to explicitly call `register_module('virtual_rainforest.core')` allows those messages to be delayed until after log files are set up.

* There is now an explicit `ConstantsDataclass` base class, to allow unambiguous detection of constants dataclasses. It has a `from_config` method, which replaces the `check_valid_constant_names` function.

### Things to address

* As @alexdewar points out, we should be using `BaseModel` in typing in the new `ModuleInfo` definition, but I cannot figure out a way to do this without getting circular imports.

* There is a `mypy` issue where it doesn't think a dataclass is callable, and I cannot see why: https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/f2b69b7ad0af2fb1257902a1972550c3f3e65730/virtual_rainforest/core/constants_loader.py#L49

virtual_rainforest/core/constants_loader.py
```diff
@@ -0,0 +1,53 @@
...
+    try:
+        constants_config = config[model_name]["constants"][class_name]
+        return constants_class.from_config(constants_config)
+    except (KeyError, IndexError):
+        # No constants_config section, so return default instance.
+        return constants_class()  # type: ignore [operator]  # FIXME !!!!!
+    except ConfigurationError:
+        # The from_config call failed, so log and exit.
+        LOGGER.critical("Could not initialise {model_name}.{class_name} from config.")
+        raise
```
dalonsoa, line 53: I'm not sure why `mypy` complains, to be honest, but all in all, this does not look like the correct approach.

Exceptions should be used to deal with unexpected circumstances. Not having constants is not an unexpected circumstance, but a perfectly reasonable use case, so an Exception is not the right control structure here. 

In summary: if the configuration provides a section for this class, great! Use it. If not, just pass an empty dict:
```suggestion
    constants_config = config[model_name]["constants"][class_name] if "constants" in config[model_name] and class_name in config[model_name]["constants"] else {}
    try:
        return constants_class.from_config(constants_config)
    except ConfigurationError:
        # The from_config call failed, so log and exit.
        LOGGER.critical("Could not initialise {model_name}.{class_name} from config.")
        raise
```
As far as I can tell, `from_config" is absolutely happy ingesting an empty dict and there is no meaningful speed gain calling directly the constructor.

virtual_rainforest/core/constants_loader.py
dalonsoa, line 1: Why not keeping this and the constant class in the same module? Is it because of circular import issues? Can you avoid them by putting the problematic import (is it MODULE_REGISTRY) within the `load_constants` function?

virtual_rainforest/core/registry.py
```diff
@@ -0,0 +1,147 @@
...
+        constants_classes = {}
+    else:
+        # Get all subclasses of ConstantsDataclass, excluding the ABC
+        constants_classes = {
+            class_name: class_obj
+            for class_name, class_obj in getmembers(constants_submodule)
+            if is_dataclass(class_obj)
+            and issubclass(class_obj, ConstantsDataclass)
+            and not class_obj.__name__ == "ConstantsDataclass"
+        }
```
dalonsoa, line 136: So, just to confirm, you not only want for it to be a subclass of `ConstantsClass`, it also needs to be defined as `dataclass` (since a subclass of a dataclass is not a dataclass), correct?

virtual_rainforest/main.py
```diff
@@ -13,15 +13,16 @@
...
 from virtual_rainforest.core.config import Config
 from virtual_rainforest.core.data import Data, merge_continuous_data_files
 from virtual_rainforest.core.exceptions import ConfigurationError, InitialisationError
 from virtual_rainforest.core.grid import Grid
 from virtual_rainforest.core.logger import LOGGER, add_file_logger, remove_file_logger
+from virtual_rainforest.core.registry import MODULE_REGISTRY
 
 
-def select_models(model_list: list[str]) -> list[Type[BaseModel]]:
+def select_models(model_list: list[str]) -> list[Any]:  # FIXME -> list[type[BaseModel]]
```
dalonsoa, line 25: Why was this ok before and it is not now? What has changed that causes circular import erorrs?

dalonsoa (commented):
More comments, but overall is looking cleaner and robust with each iteration 👍

jacobcook1995 (approved):
LGTM!

I think the addition of the `ConstantsDataclass` _should_ fix the problem I was having with upgrading `mypy`, which is nice.

virtual_rainforest/core/constants_loader.py
davidorme: I'd disagree a bit that this is the wrong approach - that 'easier to ask forgiveness than permission' style is something commonly used in Python. It isn't a style I particularly like, but it is more compact than the long chain of iterative dictionary key checks. But yes - using the empty dictionary should sidestep that `mypy` issue.

virtual_rainforest/core/registry.py
davidorme: The issue is that `issubclass` chokes if `class_obj` is not a class (and plenty of members aren't), so really all I need is a filter so that only classes get to that point. We could generically test for is a class, but since they are all dataclasses, this is a more usefully specific way to get to the same point. And then we don't want the ABC that has been imported to create the constants classes - I don't know if there is cleaner way than checking the `__name__`

virtual_rainforest/core/constants_loader.py
dalonsoa: Well, if the check were really complicated or deep within the function, I would certainly agree, but the meaning of Exception itself suggest that using them for flow control when you are in normal operation is plainly wrong and can lead to confusion and masking other errors that have nothing to do with what you are trying to catch.

dalonsoa: Moreover, in this particular case, you do need to add those comments explaining what's going on or otherwise there's no obvious reason of why you're catching those exceptions. The code is way more explicit and needs no explanation if you check for the existence of the relevant constants dict in the config, as I suggest.

davidorme: But many examples of "idiomatic" python use exactly this kind of structure. And the Python glossary specifically mentions this kind of construct (emphasis mine):

> EAFP
> Easier to ask for forgiveness than permission. This common Python coding style **assumes the existence of valid keys or attributes and catches exceptions if the assumption proves false**. This clean and fast style is characterized by the presence of many [try](https://docs.python.org/3/reference/compound_stmts.html#try) and [except](https://docs.python.org/3/reference/compound_stmts.html#except) statements. The technique contrasts with the [LBYL](https://docs.python.org/3/glossary.html#term-LBYL) style common to many other languages such as C.

https://docs.python.org/3/glossary.html#term-EAFP 

I'm not sure how this example differs from that description. Again - it isn't a style I use widely but I do want to understand why what looks to me like a stylistic choice is problematic.

virtual_rainforest/main.py
davidorme: The problem is that `registry` is now imported, and if that imports `BaseModel` to type `ModuleInfo.model: BaseModel` then we get a circular import:

```python
 ~/vr_example/ --outpath ~/vr_example/out1 --log ~/vr_example/out1/log.log
  File "/Users/dorme/Library/Caches/pypoetry/virtualenvs/virtual-rainforest-Laomc1u4-py3.10/bin/vr_run", line 3, in <module>
    from virtual_rainforest.entry_points import vr_run_cli
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/entry_points.py", line 16, in <module>
    from virtual_rainforest.core.config import config_merge
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/config.py", line 41, in <module>
    from virtual_rainforest.core.registry import MODULE_REGISTRY, register_module
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/registry.py", line 27, in <module>
    from virtual_rainforest.core.base_model import BaseModel
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/base_model.py", line 83, in <module>
    from virtual_rainforest.core.axes import AXIS_VALIDATORS
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/axes.py", line 53, in <module>
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/grid.py", line 24, in <module>
    from virtual_rainforest.core.config import Config, ConfigurationError
ImportError: cannot import name 'Config' from partially initialized module 'virtual_rainforest.core.config' (most likely due to a circular import) (/Users/dorme/Research/Virtual_Rainforest/virtual_rainforest/virtual_rainforest/core/config.py)
```

virtual_rainforest/core/constants_class.py
```diff
@@ -0,0 +1,38 @@
...
+from virtual_rainforest.core.logger import LOGGER
+
+
+@dataclass(frozen=True)
+class ConstantsDataclass(ABC):
+    """XYZ."""
+
+    @classmethod
+    def from_config(cls, config: dict[str, Any]) -> ConstantsDataclass:
+        """MNO."""
```
alexdewar, line 19: These need proper docstrings

virtual_rainforest/core/registry.py
```diff
@@ -19,32 +19,38 @@
...
+    constants_classes: dict[str, ConstantsDataclass]
+    """A dictionary of module constants classes."""
+
+
+MODULE_REGISTRY: dict[str, ModuleInfo] = {}
+"""The module registry."""
 
 
-def register_module(module_name: str) -> None:
+def register_module(module_name: str, model: Optional[Any] = None) -> None:
```
alexdewar, line 53:
```suggestion
def register_module(module_name: str, model: Any = None) -> None:
```

virtual_rainforest/core/registry.py
```diff
@@ -19,32 +19,38 @@
...
-MODULE_REGISTRY['plants']['constants'] = [
-    'PlantsConsts': <virtual_rainforest.models.plants.constants.PlantsConsts>
-]
-MODULE_REGISTRY['plants']['schema'] = {schema_dict}
-"""
+    This dataclass is used to hold individual module registration details within the
+    module registry.
+    """
+
+    model: Optional[Any]  # Optional[type[BaseModel]]
```
alexdewar, line 41:
```suggestion
    model: Any  # Optional[type[BaseModel]]
```

virtual_rainforest/core/registry.py
```diff
@@ -67,156 +73,75 @@ def register_module(module_name: str) -> None:
...
+        raise RuntimeError(msg)
+    elif not is_core and model is None:
+        msg = "A model class is required to register model modules."
+        LOGGER.critical(msg)
+        raise RuntimeError(msg)
+    elif model is not None:
+        if module_name_short != model.model_name:
+            msg = f"The model_name attribute and module name differ in {module_name}"
+            LOGGER.critical(msg)
+            raise RuntimeError(msg)
```
alexdewar, line 97: You don't need `else`/`elif` if the preceding block raises an exception and I find all the `elif`s make things less readable. I'd also personally prefer `if model` to `if model is not None`, but that is partly a question of taste.

```suggestion
    if is_core and model is not None:
        msg = "No model should be registered for the core module."
        LOGGER.critical(msg)
        raise RuntimeError(msg)
    if not is_core and model is None:
        msg = "A model class is required to register model modules."
        LOGGER.critical(msg)
        raise RuntimeError(msg)
    if model is not None and module_name_short != model.model_name:
        msg = f"The model_name attribute and module name differ in {module_name}"
        LOGGER.critical(msg)
        raise RuntimeError(msg)
```

virtual_rainforest/core/registry.py
```diff
@@ -67,156 +73,75 @@ def register_module(module_name: str) -> None:
...
+
+        # Register the resulting single model class
+        LOGGER.info(
+            f"Registering model class for {module_name_short} model: {model.__name__}"
         )
 
     # Register the schema
-    schema_file_name = "model_schema.json" if is_not_core else "core_schema.json"
+    schema_file_name = "core_schema.json" if is_core else "model_schema.json"
```
alexdewar, line 105: Is there a reason the core model's schema has a different file name to the others?

virtual_rainforest/core/schema.py
```diff
@@ -13,7 +13,7 @@
 from pathlib import Path
 from typing import Any, Iterator
 
-import dpath.util  # type: ignore
+import dpath  # type: ignore
```
alexdewar, line 16: Why's the `type: ignore` needed?

virtual_rainforest/core/schema.py
```diff
@@ -134,15 +135,17 @@ def merge_schemas(schemas: dict[str, dict[str, Any]]) -> dict[str, Any]:
...
-
-    # Set additional properties to false everywhere that properties are defined
-    for path in p_paths:
-        dpath.util.new(comb_schema, f"{path}/additionalProperties", False)
+    # Recursively search for all instances of properties and insert
+    # additionalProperties=false to prevent undocumented properties in schema.
+    # It does seem odd that JSONSchema has no universal setting for this.
+    property_paths = [
+        path for path, _ in dpath.search(comb_schema, "**/properties", yielded=True)
+    ]
```
alexdewar, line 143: Unnecessary list creation:

```suggestion
    property_paths = (
        path for path, _ in dpath.search(comb_schema, "**/properties", yielded=True)
    )
```

virtual_rainforest/main.py
```diff
@@ -57,11 +57,7 @@ def select_models(model_list: list[str]) -> list[Type[BaseModel]]:
     LOGGER.info("Attempting to configure the following models: %s" % unique_models)
 
     # Make list of missing models, and return an error if necessary
-    miss_model = [
-        model
-        for model in unique_models
-        if model not in MODULE_REGISTRY or "model" not in MODULE_REGISTRY[model]
-    ]
+    miss_model = [model for model in unique_models if model not in MODULE_REGISTRY]
     if miss_model:
```
alexdewar, line 61: I think this should do the trick and is faster:

```suggestion
    if any(model not in MODULE_REGISTRY for model in unique_models):
```

alexdewar (changes requested):
A few small suggestions, but as @dalonsoa says, it's getting there! And I *much* prefer the stronger typing you're now using for the `MODULE_REGISTRY` :smile:.

I'm slightly suspicious about the circular import problem. I *suspect* it could be resolved by parcelling out certain functions/classes to their own `.py` files, but I don't know the file structure well enough to be sure of this. (Some of the files are getting rather long anyway, so it's probably worth doing in its own right.) If you can solve it in this PR that would obviously be preferable, but it's not the end of the world if it's too big of a task. In that case I would open an issue to do it later though.

virtual_rainforest/core/constants_loader.py
davidorme: > Exceptions should be used to deal with unexpected circumstances. Not having constants is not an unexpected circumstance, but a perfectly reasonable use case, so an Exception is not the right control structure here.

> Exceptions are about exceptional circumstances and things that should not be happening, normally.

I'm struggling with this. 

There are questions of style - I think this implementation is easier to read and maintain than the multiple clause `if` statement - but those don't really matter. But I *do* want to know if I'm missing something subtle about this approach - it seems really useful in some cases. I don't want to be misapplying it and your advice is clearly that I am.

Even in the standard libraries, `try`-ing something that will routinely fail in an expected way and then trapping those failures with `except` seems widespread. An example is `os.exists`, where a file not existing will happen all the time.

https://github.com/python/cpython/blob/1de9406f9136e3952b849487f0151be3c669a3ea/Lib/genericpath.py#L16-L22

What am I missing?

virtual_rainforest/core/constants_class.py
davidorme: Oh god yes, but all of the docstrings and docs need rewriting too. Once the implementation has settled down, I'll do that.

virtual_rainforest/core/registry.py
davidorme: Can't merge those last two tests - the last clause doesn't just raise that exception but otherwise 👍

virtual_rainforest/core/registry.py
davidorme: No. I think we've already standardised  (e.g.) "plants_model_schema.json" to "model_schema.json". I don't want to have `core/model_schema.json` because it isn't a model schema. We could standardise everything to `module_schema.json` (or just `schema.json` 😄 ).

@jacobcook1995 ?

virtual_rainforest/core/schema.py
davidorme: Seems it isn't anymore.

virtual_rainforest/core/constants_loader.py
dalonsoa: There might be cases where checking if something works or not is not straight forward, might require way too much code to fully check if it works or the multiple cases in which it might fail. Or simply there is no other way to do the checks (like in the example you mention - if you just check if the file actually exists, you will get the wrong result as the file exist in the case of broken symbolic links; they are just not connected to anything). In those cases, it is way simpler to just use `try...except...`. The conciseness and simplicity of the code clearly outweighs any benefit related to explicitness of the checks. I'm happy to use Exceptions in those cases.

But I don't think that having that possibility is an excuse when a simple conditional statement can do the job perfectly fine. People abuse of the EAFP idea massively and applies it everywhere, diminishing the actual role of Exceptions which is dealing with expectational circumstances. And they also apply it wrong, as it is your case, as there are two possible sources or errors, the dictionary and the call to `from_config`, with the issues I explained in my previous message.

virtual_rainforest/main.py
davidorme: But we use the missing names list in the logging output.

davidorme:
> I'm slightly suspicious about the circular import problem. I _suspect_ it could be resolved by parcelling out certain functions/classes to their own `.py` files, but I don't know the file structure well enough to be sure of this. (Some of the files are getting rather long anyway, so it's probably worth doing in its own right.) If you can solve it in this PR that would obviously be preferable, but it's not the end of the world if it's too big of a task. In that case I would open an issue to do it later though.

I'm _deeply_ suspicious of this circular import problem. I haven't spotted the fix yet, but the problem is that `BaseModel` uses all sorts of high level stuff (`Config`, `Data` etc.) and we want to _describe_ the typing registry well before any of that is loaded.

If that manky `Any` typing substitute is all that is holding us back on the implementation, I'm tempted to park that issue.

virtual_rainforest/core/constants_loader.py
davidorme: On a different note - on balance I'm going to stick with the `# type: ignore` here. Using `constants_class()` _clearly_ should not throw an error and it _is_ callable. I'd rather ignore it and leave an indication that `mypy` is being a bother than do something unnecessary just to keep `mypy` quiet.

virtual_rainforest/core/schema.py
davidorme: No - wait. The intention here is to turn it into a **generator**, right? Not just a tuple. 

That doesn't work because as the `property_paths` generator yields paths in the following loop,  the underlying `comb_schema` is modified, which changes the dictionary size,  and breaks the `search` iteration.

davidorme:
> I think the addition of the `ConstantsDataclass` _should_ fix the problem I was having with upgrading `mypy`, which is nice.

@jacobcook1995 I've just tried `precommit run --all` with `mypy 1.5.1` and all seemed well.

virtual_rainforest/core/registry.py
davidorme: I'll go ahead and standardise on `module_schema.json`

davidorme:
## Round 3

OK - I am not going to try and resolve the circular import problem with `BaseModel` here, which means leaving `core.registry.ModuleInfo.model` typed as `Any`. I've raised #328 to kick this down the track.

That aside, I think the draft implementation is probably stabilised. I'll ask for another round of reviews.

virtual_rainforest/core/schema.py
```diff
@@ -0,0 +1,151 @@
...
+def merge_schemas(schemas: dict[str, dict[str, Any]]) -> dict[str, Any]:
+    """Merge the validation schemas for desired modules.
+
+    The method merges a set of schemas for a set of desired modules into a single
+    integrated schema that can then be used to validate a merged configuration for those
+    modules. The merge also updates the resulting schema to enforce that only properties
+    explicity listed in the schema can be included.
+
+    Args:
+        schema: A dictionary of schemas keyed by module name
```
jacobcook1995, line 123: this name is different to the argument (weird as I would have thought the pre-commit hook would catch it)

jacobcook1995 (approved):
LGTM! I just had some minor comments

virtual_rainforest/core/constants.py
```diff
@@ -1,121 +1,15 @@
...
-        check_valid_constant_names(config, model_name, class_name)
-        # If an error isn't raised then generate the dataclass
-        return constants_class(**config[model_name]["constants"][class_name])
+@dataclass(frozen=True)
+class CoreConsts(ConstantsDataclass):
+    """Core constants for use across the :mod:`virtual_rainforest` modules."""
 
-    # If no constants are supplied then the defaults should be used
-    return constants_class()
+    k_to_c: float = 273.15
```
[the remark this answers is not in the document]
alexdewar: Agreed ^^

virtual_rainforest/core/constants_class.py
alexdewar: It does also help with reviewing if you can provide a docstring, even if it's just one line. Adding dummy docstrings like this just seems like an exercise in making `flake8` shut up

virtual_rainforest/core/registry.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # Try and import the module from the name to get a reference to the module
+    try:
+        module = import_module(module_name)
+    except ModuleNotFoundError as excep:
+        LOGGER.critical(f"Unknown module - registration failed: {module_name}")
+        raise excep
+
+    # Extract the last component of the module name to act as unique short name
+    module_name_short = module_name.split(".")[-1]
```
alexdewar, line 74: Alternatively:

```suggestion
    _, _, module_name_short = module_name.rpartition(".")
```

virtual_rainforest/core/schema.py
```diff
@@ -0,0 +1,151 @@
...
+    # Check that relevant keys are included - deliberately re-raising as ValueError here
+    # as the KeyError has built in quoting and is expecting to provide simple missing
+    # key messages.
+    try:
+        json_schema["properties"][module_name]
+        json_schema["required"]
+    except KeyError as excep:
+        to_raise = ValueError(f"Missing key in module schema {module_name}: {excep}")
+        LOGGER.error(to_raise)
+        raise to_raise
```
alexdewar, line 109: While I agree that in general, sometimes it can make for more readable code to use a `try..except` block, I don't think this is one of those cases.

virtual_rainforest/core/constants.py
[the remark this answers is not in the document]
davidorme: I see what you mean but do we really want multiple constants frameworks? And really, the whole simulation is one big red button 😄 . I'm not convinced we should bend over backwards to stop someone who is running a massive ecological simulation from redefining gravity or whatever (and secretly, I kinda want to see what happens).

Having said that - maybe we can update this to allow _some_ constants to be class attributes, not instance attributes, which I think will make them unconfigurable.

virtual_rainforest/core/constants_loader.py
```diff
@@ -0,0 +1,54 @@
+"""The :mod:`~virtual_rainforest.core.constants` module is used to store constants that
+are used across the Virtual Rainforest. This includes universal constants, such as the
+strength of gravity, but also model constants that are shared between multiple models,
+such as the depth of the biogeochemically active soil layer.
+
+At the moment, no constants are actually stored in this module. It currently only
```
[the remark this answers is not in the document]
davidorme: Yeah - working on that 😄

virtual_rainforest/core/schema.py
davidorme: Likewise - that's not good. Seems like something is borked in `pre-commit`? Missing a plugin maybe? @alexdewar @dalonsoa

virtual_rainforest/core/schema.py
alexdewar: So it does. My bad.

alexdewar (approved):
LGTM, modulo some tidyups and missing unit tests.

virtual_rainforest/core/constants.py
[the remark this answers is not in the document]
jacobcook1995: I mean I'm fine with changing things like the strength of gravity (though that would be tricky as things like flow rates will assume gravity's strength implicitly), I just feel like allowing people to change the number of millimetres in a meter would be bizarre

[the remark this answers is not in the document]
jacobcook1995: If we ever did a "what is the most sensitive parameter" style sensitivity analysis sweep we would want some way to exclude universal constants + unitary conversions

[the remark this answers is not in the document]
alexdewar: But when would a user ever want to redefine the Celsius-to-Kelvin conversion factor? That is just odd. You don't need a framework for these constants -- just define them somewhere.

For physical units, I think you should also try to use `scipy.constants` wherever possible.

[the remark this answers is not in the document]
davidorme: The main motivation here is that constants are organised so that they come through one route, rather than having to remember which bits come from where. I agree that `scipy.constants` is great, but it is predominantly physics and quite limited for constants we need for the VR.

We can explicitly set constants as being non-configurable using class variables though. Does this seem sensible:

```python
from typing import ClassVar

from scipy.constants import zero_Celsius

from virtual_rainforest.core.constants_class import ConstantsDataclass

@dataclass(frozen=True)
class Test(ConstantsDataclass):

    zero_Celsius: ClassVar[float] = zero_Celsius
    b: int = 3
```

And hence:

```python
In [288]: Test(zero_Celsius=263.15)
TypeError                                 Traceback (most recent call last)
Cell In[288], line 1
----> 1 Test(zero_Celsius=263.15)

TypeError: Test.__init__() got an unexpected keyword argument 'zero_Celsius'
```

We could update `ConstantsDataclass.from_config` to report attempts to configure things that should never change.

virtual_rainforest/core/registry.py
davidorme: Sure, marginally cleaner 😄

virtual_rainforest/core/constants.py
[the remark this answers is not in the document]
jacobcook1995: @davidorme definitely agree that that is a good solution in general. However, probably should mention that `zero_Celsius` is no longer going to be used in the model as I've switched to using `scipy.constants.convert_temperature()` to switch between temperature scales rather than writing out my temperature conversions myself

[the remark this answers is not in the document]
davidorme: That's fine - it can sit there as a demo for now. We're going to have to review all of this.

davidorme:
## Round 4

OK - so this now implements a revised mode of operation, following discussion, that explicitly separates the registration process from module import. The `MODULE_REGISTRY` is populated using `register_module` - that does import the module to be registered in order to discover components, but the module is _not_ registered in the `__init__.py`. 

The test suite is now updated to work with this new approach and all tests pass. That has included the addition of a global `autouse=True` fixture that ensures that the `MODULE_REGISTRY` is cleared before _all_ tests. This is because the registry contents persist between tests and we want to ensure that all tests can stand alone, with the correct module registration, rather than simply passing by accident due to earlier registration.

The docs are update too now. There is one odd `sphinx` warning that I can't figure out right now, which is something to do with the `id_of` property of JSONSchema validators.

```sh
/home/runner/.cache/pypoetry/virtualenvs/virtual-rainforest-Csde_K7l-py3.9/lib/python3.9/site-packages/jsonschema/validators.py:docstring of jsonschema.validators.create.<locals>.Validator:1:<autosummary>:1: WARNING: py:obj reference target not found: virtual_rainforest.core.schema.ValidatorWithDefaults.ID_OF
```

I've turned this from WIP into a full PR. I know it's a lot of changes but the key things that need attention are really just the latest `register_module` function in `core/registry.py`. If that seems solid then the test changes are all pretty clear. I have updated some test structures to make them easier to follow and to use the `Config` process corretly.

jacobcook1995 (approved):
LGTM!

tests/core/test_cli_integration.py
```diff
@@ -31,8 +32,8 @@ def test_vr_run(capsys):
                 str(example_dir),
                 "--outpath",
                 str(example_dir),
-                "--logfile",
-                str(example_dir / "vr_example.log"),
+                # "--logfile",
```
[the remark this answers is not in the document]
davidorme: No - but there's a whole bunch of fixing over in #331 which will replace much of this.

tests/models/litter/test_litter_model.py
```diff
@@ -77,180 +48,139 @@ def test_litter_model_initialization(caplog, dummy_litter_data):
...
-            DEBUG,
-            "litter model: required var 'lignin_below_structural' checked",
-        ),
-        (
-            ERROR,
-            "Lignin proportions not between 0 and 1 found in: ",
-        ),
-    )
+    from virtual_rainforest.models.litter.constants import LitterConsts
+    from virtual_rainforest.models.litter.litter_model import LitterModel
```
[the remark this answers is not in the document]
davidorme: If I understood @dalonsoa correctly, I think the idea is that as much as possible of the environment used by each test should be isolated within it? With some `vr` things - notably custom `Exceptions` - we have to import them at the module level so they can be used in parameterisation, but otherwise tests should be self contained? Is that right?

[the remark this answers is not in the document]
dalonsoa: That's correct. As much as possible, setting up the tests (as oppose to running them) should not depend on the package under test itself, and that means having all the imports - even if repeated - within the tests. This avoids side effects (keeping appropriate isolation) as well as preventing the whole test suit to fail at setup time just because a failure of the package under test. 

Having said that, as David points out, there are specific use cases where that is not possible, but it should be the general approach.

PR #327, opened 2023-10-04
Feature/adjust surface evaporation

vgro:
Surface evaporation was considered as bare soil evaporation, which was an overestimation under a forest canopy. This PR introduces a shading parameter that reduces evaporation.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/hydrology/constants.py
```diff
@@ -117,3 +117,6 @@ class HydroConsts:
     :cite:t:`cosby_statistical_1984`; see documentation for
     :attr:`air_entry_water_potential` for further details.
     """
+
+    shading_parameter: float = 0.5
+    """Parameter that reduced potential soil evaporation to shaded soil evaporation."""
```
jacobcook1995, line 122: It might be worth adding more details here as we are (eventually) planning to use these details to document the model constants. At the moment it isn't clear how exactly changing this parameter changes the strength of shading. I guess you could either explain this in words or link to the function

jacobcook1995 (approved):
LGTM! 

Had a minor point about the constants docstring, but otherwise it looks good to go.

PR #336, opened 2023-10-18
Configurable system for setting model execution sequence

### LINKED ISSUE #334: The `select_models` function is repetitive

The `Config.build_schema` method is now where module registration occurs and - in the process of running this method - the model information for the configuration is loaded and validated. The `vr_run.select_models` function then separately looks up requested models in the `MODULE_REGISTRY` and traps unknown models.

This is repetitive: `build_schema` has already seen this list of requested models, so  it makes more sense for `Config.build_schema` (possibly under a new name) to populates a `Config.models` attribute with a dictionary of the requested models, which can then just be used in `vr_run`.

Delete `select_models` and replace with functionality in `Config`.

### LINKED ISSUE #335: Models need a configurable system for setting running sequence

The methods for models can require specific execution order to make sure that variables are populated in the correct sequence. An example is that the `plants` model `__init__` method populates the canopy structure, which is required to calculate abiotic profiles. The execution order is not necessarily the same for the different `BaseModel` methods.

 * A configuration section that can be used to describe which models a given model should take priority over when applying a particular method.
 * Functionality to take a set of models and resolve the configured execution order for a method. This should detect configurations that contain cycles - the execution order must define an acyclic set of execution priorities.
 * Extend `vr_run` to use the configure execution order.

This _could_ be implemented using properties of the `BaseModel` subclasses, but hard-coding priorities is less flexible and means that model definitions themselves would need to change if new models are introduced.

Ultimately, this probably should be resolved at the _variable_ level, by producing a set of the variables in a simulation and resolving the order in which those should be initialised and updated, but an initial implementation is needed to allow control of execution order.

davidorme:
# Description

This PR:

* Implements a new model configuration schema that allows the execution order of different model methods to be set:

```toml
[core]
modules=['soil','abiotic_simple']
[soil.priority]
init=['abiotic_simple]
```
This is an instruction that the soil model `__init__` method should be run before the `abiotic_simple` `__init__` method.

* Implements a `get_model_sequence` function in `main.py` that sorts a dictionary of models according to the configured priorities for a given method.
* Only currently implements `init` and `update` but can be easily extended to `setup`, `spinup` etc if needed.
* Deletes the `select_models` function and instead adds `Config.model_classes` to get a simple dictionary of the configured model classes more directly (#334).
* Updates the tests and adds specific tests of the functionality.
* A couple of JSONSchema files had CRLF endings for some reason, so I've standardised back to LF.

One thing I'm not sure about - at the moment the code objects if the priorities for a configured model include model names that are not in the config. That basically makes defaults for different models impossible, because it relies on all the model priorities being included. It may make more sense to simple discard 'extra' model names from a priority with a warning.

Fixes #335 and fixes #334

## Type of change

## Key checklist

## Further checks


virtual_rainforest/main.py
```diff
@@ -181,6 +138,80 @@ def extract_timing_details(
...
+    config: Config, models: dict[str, Any], method: str
+) -> dict[str, Any]:  # FIXME dict[str, Any]-> dict[str, Type[BaseModel]]
+    """Get a tuple of the model execution sequence for a given model stage.
+
+    This function uses the ``priority`` sections in model configurations to establish an
+    execution sequence for a given model method (currently, one of ``init`` or
+    ``update``). For example, the configuration:
+
+    .. code-block:: toml
+        [plants.priorities]
```
dalonsoa, line 151: Based on the code below, shouldn't this be `priority` and not `priorities`?

virtual_rainforest/main.py
```diff
@@ -181,6 +138,80 @@ def extract_timing_details(
...
+        init = ['abiotic']
+        update = ['abiotic']
+
+    would set that the ``plants`` model should be both initialised and updated before
+    the ``abiotic`` model.
```
dalonsoa, line 156: I was a bit confused when checking the code until I saw that you used `reversed` in the static order. Now it makes sense.

Having said that, do you think it works better to use `priorities` (model X has higher priority than model Y and therefore should run first) or `dependencies` (model Y depends on model X, so X should run first)? The question is more philosophical than practical - either way works fine. In my mind, you need to define an order because a model depends on another, not because it has a higher priority.

virtual_rainforest/main.py
```diff
@@ -181,6 +138,80 @@ def extract_timing_details(
...
+
+        # Check the priorities are all models included in the configuration.
+        unknown_priorities = set(config_priorities).difference(models.keys())
+        if unknown_priorities:
+            to_raise = (
+                f"Model {method} priorities for {model_name} includes "
+                f"unconfigured models: {','.join(unknown_priorities)}"
+            )
+            LOGGER.critical(to_raise)
+            raise ConfigurationError(to_raise)
```
dalonsoa, line 194: I don't understand why you need this. As long as the `priorities` dictionary includes all the models, the sorter will work, regardless of each of the models having a complete list of priorities or not. So, the following works totally fine:

```python
priorities = {"A": ["B", "C"], "B": [], "C": [], "D": ["A"]}
sorter = TopologicalSorter(priorities )
list(reversed(list(sorter.static_order())))
```
resulting in:
```output
['D', 'A', 'C', 'B']
```
`C` and `D` have no priorities, so they are just allocated at some point when it is suitable to fulfil other model's priorities. Am I missing anything?

dalonsoa (approved):
Looks good to me! A tricky business the ordering one, but well solved. I've just left a couple of minor comments.

virtual_rainforest/main.py
davidorme: This catches the case where - for example - `soil` has the priorities `['abiotic_simple', 'logging']` but  the `logging` model is not in the model configuration.

That's my last point in the PR description - I now think this step a mistake actually - we'd want a default priority list for models that includes all known priorities and these just get removed if a given model is not being used. That way we don't have to continually mess around with configuration - `soil` gets default priority over `logging` when `logging` is included, but otherwise the priorities get reduced to the configured models. @jacobcook1995? 

I wasn't expecting the code below to work, but we certainly don't want it happening:

```python
In [131]: priorities = {"A": ["B", "C", "not_a_configured_model"], "B": [], "C": [], "D": ["A"]}
     ...: sorter = TopologicalSorter(priorities )
     ...: list(reversed(list(sorter.static_order())))
Out[131]: ['D', 'A', 'not_a_configured_model', 'C', 'B']
```

jacobcook1995: Yeah I was going to suggest adding default priorities (or dependancies) to the `JSON` schema to minimise the amount of messing around with config (which is probably unavoidable in some cases, but shouldn't be necessary to run our core set of models). I do think having a default priority list that can be overridden using the configuration is probably the better way to go though

virtual_rainforest/models/soil/module_schema.json
```diff
@@ -1,28 +1,48 @@
...
+                  "SoilConsts"
+               ]
+            },
+            "priority": {
+               "type": "object",
+               "default": {},
+               "properties": {
+                  "init": {
+                     "type": "array",
+                     "default": [],
```
jacobcook1995, line 26: Was going to suggest adding default priorities here, but maybe this isn't the best place for them

jacobcook1995 (approved):
LGTM! My main feeling is that a default priority (or dependancy) order needs to be set up somehow

davidorme:
I have updated the PR to:
* Switch to `depends` as that gives a clearer sense of what the system does.
* Updated the sequencer to check for unconfigured dependencies and then remove them with a warning. The simulation may well crash here, but this allows defaults to be set and left alone without having to constantly tinker with `depends` for the set of models.
* Added a short description in the docs.

PR #337, opened 2023-10-19
Alter schema structure to use a cleaner model set definition

### LINKED ISSUE #333: Alter schema structure to use a cleaner model set definition.

At the moment, the config `core.modules` property is used to specify which models are to be used and to provide a default set of models. There are some issues:

* It doesn't seem sane to have a default set of models - we want people to be explicit about what models they want used.
* It duplicates information - if a model is being configured then it will have a top-level section of its own:
 
```toml
[core]
modules = ['plants']
[plants]
a_parameter = 1
```

* It can also be internally inconsistent.  In the example below, it is clear that the `soil` model is wanted, but model registration is driven by `core.modules`

```toml
[core]
modules = ['plants']
[soil]
a_parameter = 1
```

The configuration files should provide an explicit top level set of the models to be used. Those can validly be empty to adopt the default configuration for any or all models:

```toml
[core]
[soil]
[abiotic]
[plants]
[animals]
```

That is then a clear statement about what models to use without redundancy or internal conflict and each section can be expanded as needed.

Note that this does _not_ resolve the wider problem of which modules are compatible or mandatory for a complete `vr_run` but that is another question!

davidorme:
# Description

This implements the suggested change in #333, that moves a configuration having to explicitly list the modules as `[core] modules = ['soil', ...]` with defaults to an explicit requirement to put a (possibly minimal) configuration entry for each required module:

```toml
[core]
[soil]
```

Fixes #333 - see that issue for the rationale for this PR.

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

Only question I have is whether we should define a default set of models to be run, which would probably be (`plants`, `animals`, `simple_abiotic`, `hydrology`, `litter`, `soil`)? I guess module choice is probably high enough level that writing it out for each set of config files isn't too much hassle though

virtual_rainforest/core/config.py
```diff
@@ -421,32 +421,35 @@ def build_config(self) -> None:
...
+        generate a single validation schema for model configuration.
         """
 
-        # Register the core module components and then extract the modules requested in
-        # the configuration, falling back to the schema defaults.
+        # Extract the requested modules, which are the top-level config keys.
+        requested_modules: list[str] = list(self.keys())
+
+        # Must include core
+        if "core" not in requested_modules:
```
dalonsoa, line 435: As `core` must always be there, why raising an error if not present? Just use it with the default options in that case.

dalonsoa (approved):
It looks good, although I don't completely follow the need of including explicitly core, just to remove it afterwards.

davidorme:
@jacobcook1995 I'm not convinced by a default set of models - I think the minimum configuration should be an explicit list of the modules to be used (which is a very short TOML doc/string). Even within testing, I can't really see many circumstances where a default set of modules, as opposed to a minimal set for testing particular features, is useful.

virtual_rainforest/core/config.py
davidorme: Yeah - my thinking here was that the simplest possible TOML config lists all of the modules wanted, so for consistency should include `[core]` so that users have to keep that configuration element obvious. All the sections are there to be expanded right from the most basic permitted config.

As a user though, I'd definitely be irritated if my config failed on that step.

dalonsoa: I see the beauty of that symmetry, but asking the user to include it when it is not optional and might be used with the default configuration, seems, as you say, annoying.

davidorme: Bah. Form over function,  I say 😄

davidorme: OK - reverted to `[core]` being optional, and logging a message to that effect.

PR #340, opened 2023-10-25
Feature/abiotic tools

vgro:
This PR adds general abiotic tools in the new (mechanistic) version of the abiotic model which I build on a separate branch `feature/new_abiotic_model` before merging in develop.
This code has been reviewed in the past, I only made minor changes.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -0,0 +1,96 @@
...
+        standard_mole: Moles of ideal gas in 1 m^3 air at standard atmosphere
+        standard_pressure: Standard atmospheric pressure, [kPa]
+        celsius_to_kelvin: Factor to convert temperature in Celsius to absolute
+            temperature in Kelvin
+
+    Returns:
+        molar density of air, [mol m-3]
+    """
+
+    temperature_kelvin = temperature + celsius_to_kelvin
```
alexdewar, line 34: I don't think you need to pass in `celsius_to_kelvin` as an argument as it'll never change.

Scipy has a function for doing this directly: `from scipy.constants.constants import C2K`

alexdewar: The same *might* apply for some of these other constants.

virtual_rainforest/models/abiotic/abiotic_tools.py
```diff
@@ -0,0 +1,96 @@
...
+    return (
+        specific_heat_equ_factor_1 * temperature**2
+        + specific_heat_equ_factor_2 * temperature
+        + molar_heat_capacity_air
+    )
+
+
+def calculate_latent_heat_vaporisation(
+    temperature: NDArray[np.float32],
+    celsius_to_kelvin: float,
```
alexdewar, line 71: Ditto above.

virtual_rainforest/models/abiotic/constants.py
```diff
@@ -12,3 +12,29 @@
 @dataclass(frozen=True)
 class AbioticConsts(ConstantsDataclass):
     """Dataclass to store all constants for the `abiotic` model."""
+
+    celsius_to_kelvin: float = 273.15
+    """Factor to convert temperature in Celsius to absolute temperature in Kelvin."""
```
alexdewar, line 17: See comment above.

virtual_rainforest/models/abiotic/constants.py
```diff
@@ -12,3 +12,29 @@
...
+    standard_pressure: float = 101.3
+    """Standard atmospheric pressure, [kPa]"""
+    standard_mole: float = 44.6
+    """Moles of ideal gas in 1 m^3 air at standard atmosphere."""
+    molar_heat_capacity_air: float = 29.19
+    """Molar heat capacity of air, [J mol-1 K-1]."""
+    gravity: float = 9.81
+    """Gravitational acceleration constant, [m s-1]."""
+    stefan_boltzmann_constant: float = 5.67e-8
+    """Stefan-Boltzmann constant, [W m-2 K-4]"""
```
alexdewar, line 27: These two constants (and maybe more) are defined in `scipy.constants`, so I'd use those values.

To be clear, I think it's fine to define `gravity` etc. here along with your other constants, but I would define it with `scipy.constants.g` instead of `9.81`.

alexdewar (changes requested):
Looks good!

I've got a couple of suggestions re constants:

1. Define your constants using the values in `scipy.constants` if they exist
2. Use `scipy.constants.C2K()` for converting from Celsius to Kelvin and drop the `celsius_to_kelvin` constant

davidorme (approved):
I don't have anything to add to @alexdewar's comments. I guess some of these may go into core constants. Since we're importing scipy for constants, we may as well use `CtoK`?

virtual_rainforest/models/abiotic/abiotic_tools.py
vgro: I agree, this makes sense.

vgro:
> I don't have anything to add to @alexdewar's comments. I guess some of these may go into core constants. Since we're importing scipy for constants, we may as well use `C2K`?
> 
> ETA - I see `C2K` and `K2C` have been deprecated in favour of the general `convert_temperatures`. It seems like heavy lifting over simply adding a constant?

I agree, I will use the `zero_Celsius` for now, I guess we will have another discussion about the constants :-)

virtual_rainforest/models/abiotic/abiotic_tools.py
vgro: I moved this to the `CoreConsts` now, how else could I call it or where would I import scipy?

alexdewar: I guess what I meant was that you probably didn't need to make the physical constants arguments. It just makes it a bit harder to read and `standard_pressure` etc. aren't going to change (right?).

That said, if it makes things easier to test, then it's fine.

alexdewar (approved):
I'm still not convinced of the need to pass in constants like `celsius_to_kelvin` as arguments to your functions, but it's up to you if you'd rather leave it as is.

vgro:
> I'm still not convinced of the need to pass in constants like `celsius_to_kelvin` as arguments to your functions, but it's up to you if you'd rather leave it as is.

I understand what you mean. The good news is that we realized that we all convert the temperatures from kelvin to Celsius to do our calculations and the convert it back (because we thought that others needed it in Celsius). So as soon as @jacobcook1995  has merged his enzyme PR, he will create an issue to move everything to Kelvin. I'll make the adjustment here and then merge 👍

PR #341, opened 2023-11-20
Add soil enzymes to model

jacobcook1995:
# Description

This pull request changes the soil carbon model to make it an enzyme explicit model, i.e. one where carbon flows are controlled by enzymes pools which are included in the model. This is unfortunately quite a large pull request, as I couldn't find a sensible place to stop midway.

The largest structural change I've made is moving the calculations of environmental factors impact on biogeochemical rates into a separate submodule. I did this as `soil.carbon` felt like it was becoming too large.

I've also started to make use of core constants as well as model constants. So feedback on how I've done this would be useful.

Generally looking for feedback on the structure, but if @vgro is able to check that the leaching function I've added is sensible that would be great.

## Type of change

## Key checklist

## Further checks


dalonsoa:
Working on this, but it might take me a couple of days.

virtual_rainforest/models/litter/litter_pools.py
```diff
@@ -39,7 +40,8 @@ def calculate_change_in_litter_variables(
     decomposed_excrement: NDArray[np.float32],
     decomposed_carcasses: NDArray[np.float32],
     update_interval: float,
-    constants: LitterConsts,
+    model_constants: LitterConsts,
+    core_constants: CoreConsts,
```
dalonsoa, line 44: And the same here.

dalonsoa (changes requested):
It looks good! I've paid special attention to the constants part and left a single comment, but that has ramifications if implemented. Feel free to ignore or postpone.

virtual_rainforest/models/litter/litter_model.py
```diff
@@ -106,7 +107,8 @@ def __init__(
         update_interval: Quantity,
         soil_layers: list[float],
         canopy_layers: int,
-        constants: LitterConsts,
+        model_constants: LitterConsts,
+        core_constants: CoreConsts,
```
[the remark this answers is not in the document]
jacobcook1995: Ahh yeah when I was writing it I did find having to track both to be pretty clumsy. So, a different approach definitely makes sense. I guess that the difficultly lies in setting it up in such a way so that the value of constants can still be changed using the configuration.

I guess one option would be for the constants classes to take a configured `CoreConsts` object as an argument, e.g. 

```python
class LitterConsts(core_constants: CoreConsts)
    ...
    depth_of_active_soil_layer: float = core_constants.depth_of_active_soil_layer
```

Then the configuration process could configure `CoreConsts` first (which could change the value of `depth_of_active_soil_layer`), and use this to generate model specific constants data classes.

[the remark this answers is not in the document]
dalonsoa: The problem with that is that `LitterConst` is a `dataclass` so its inputs are already the constant values it contains. however, it might be a step in the right direction if combined with `properties`:

```python
class LitterConsts:
    ...
    core_constants: CoreConsts = CoreConsts()

    @property
    def depth_of_active_soil_layer(self) -> float:
        return core_constants.depth_of_active_soil_layer
```

So `core_constants` becomes just another attribute of the `LitterConst` dataclass, with a default value of `CoreConst()` but that can get as input a modified `CoreConst` object, if present. Then, the constants of interest are accessed with properties.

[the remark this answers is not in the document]
jacobcook1995: Okay makes sense, should I make an issue for this then?

tests/conftest.py
```diff
@@ -172,12 +172,37 @@ def dummy_carbon_data(layer_roles_fixture):
...
     data["pH"] = DataArray([3.0, 7.5, 9.0, 5.7], dims=["cell_id"])
     data["bulk_density"] = DataArray([1350.0, 1800.0, 1000.0, 1500.0], dims=["cell_id"])
-    data["percent_clay"] = DataArray([80.0, 30.0, 10.0, 90.0], dims=["cell_id"])
+    data["clay_fraction"] = DataArray([0.8, 0.3, 0.1, 0.9], dims=["cell_id"])
     data["litter_C_mineralisation_rate"] = DataArray(
         [0.00212106, 0.00106053, 0.00049000, 0.0055], dims=["cell_id"]
     )
+    # Data for combined vertical flow (for entire timestep)
+    data["vertical_flow"] = xr.concat(
+        [
```
vgro, line 191: The vertical flow variable that is returned from the hydrology model is only one-dimensional (`cell_id` only) and describes the total flow though one grid cell per time step, so you don't need the layer selection here. If you wanted to look at multiple layers in the future, I could return the flow through each layer if needed.

tests/conftest.py
```diff
@@ -200,7 +225,6 @@ def dummy_carbon_data(layer_roles_fixture):
             "cell_id": data.grid.cell_id,
         }
     )
-    # TODO - Eventually this should replace the dummy soil moisture entirely
     data["matric_potential"] = xr.concat(
```
vgro, line 228: Maybe we can chat about this, if you don't need soil moisture we can focus on the matric potential because this is also how P-Hydro uses

virtual_rainforest/core/constants.py
```diff
@@ -21,5 +21,11 @@ class CoreConsts(ConstantsDataclass):
     zero_Celsius: ClassVar[float] = constants.zero_Celsius
     """Conversion constant from Kelvin to Celsius (°)."""
 
-    placeholder: float = 123.4
-    """A placeholder configurable constant."""
+    depth_of_active_soil_layer: float = 0.25
+    """Depth of the biogeochemically active soil layer [m].
```
vgro, line 25: Based on previous discussions, we set the top soil layer to 50cm. Would it be worth changing that to get a better representation of the active zone?

virtual_rainforest/data_variables.toml
```diff
@@ -477,14 +477,6 @@ initialised_by = "external"
 updated_by = "soil"
 used_by = [ "abiotic", "soil"]
 
-[[variable]]
-name = "percent_clay"
-description = "Clay as a percentage of total soil"
-unit = "%"
-initialised_by = "external"
-updated_by = ""
-used_by = [ "abiotic", "soil"]
```
vgro, line 486: At the moment this is only indirectly used by the `hydrology` model but I guess we could make it more explicit in the future. Thus, would it be worth already adding the hydrology here? If we do it as implemented, we should probably remove the `abiotic` one here for now.

virtual_rainforest/core/constants.py
vgro: I also wonder if that is actually a core constant? It is very soil microbe specific

virtual_rainforest/models/soil/carbon.py
```diff
@@ -1,34 +1,60 @@
...
+
 import numpy as np
 from numpy.typing import NDArray
-from scipy.constants import convert_temperature, gas_constant
 
-from virtual_rainforest.core.logger import LOGGER
+from virtual_rainforest.core.constants import CoreConsts
 from virtual_rainforest.models.soil.constants import SoilConsts
+from virtual_rainforest.models.soil.env_factors import (
+    calculate_environmental_factors,
```
vgro, line 15: The name of the function feels a bit potentially misleading, it's fine for now but we might want to state more explicitly that is is about the effects, not the environment itself

virtual_rainforest/models/soil/env_factors.py
```diff
@@ -0,0 +1,306 @@
...
+    """Impact of soil clay fraction on necromass decay destination [unitless]."""
+
+
+def calculate_environmental_factors(
+    soil_water_potential: NDArray[np.float32],
+    pH: NDArray[np.float32],
+    clay_fraction: NDArray[np.float32],
+    constants: SoilConsts,
+) -> EnvironmentalFactors:
+    """Calculate the impact of the environment has on relevant biogeochemical rates.
```
vgro, line 36:
```suggestion
    """Calculate the impact of the environment on relevant biogeochemical rates.
```

virtual_rainforest/models/soil/env_factors.py
```diff
@@ -0,0 +1,306 @@
...
+        activation_energy: Energy of activation [J mol^-1]
+        soil_temperature: The reference temperature of the Arrhenius equation [C]
+
+    Returns:
+        A multiplicative factor capturing the effect of temperature on microbial rates
+    """
+
+    # Convert the temperatures to Kelvin
+    soil_temp_in_kelvin = convert_temperature(
+        soil_temperature, old_scale="Celsius", new_scale="Kelvin"
```
vgro, line 112: This is something we need to agree on in the context of constants. I currently just add the conversion factor to the temperature in Celsius, I am not sure if it needs a function for that simple operation.

virtual_rainforest/models/soil/env_factors.py
```diff
@@ -0,0 +1,306 @@
...
+
+    This function only returns valid output for soil water potentials that are less than
+    the optimal water potential.
+
+    Args:
+        water_potential: Soil water potential [kPa]
+        water_potential_halt: Water potential at which all microbial activity stops
+            [kPa]
+        water_potential_opt: Optimal water potential for microbial activity [kPa]
+        moisture_response_curvature: Parameter controlling the curvature of the moisture
```
vgro, line 140: This is also a parameter that we both use, so we should coordinate where this is stored

virtual_rainforest/models/soil/env_factors.py
```diff
@@ -0,0 +1,306 @@
...
+            [unitless]
+        soil_layer_thickness: Thickness of the biogeochemically active soil layer [m]
+
+    Returns:
+        The rate at which the solute in question is leached [kg solute m^-3 day^-1]
+    """
+
+    # Vertical flow rate has to be converted into m day^-1
+    vert_flow_meters = vertical_flow_rate / 1e3
+
```
vgro, line 300: We should have a chat about this, especially the units. I included a lot of conversions to go from m/s in a soil column to mm per month per area, which might all be unnecessary sources of error if you don't need that.

virtual_rainforest/models/litter/litter_model.py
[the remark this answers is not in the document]
davidorme: I've backed away from this slightly - see commments on #342. I think having a `coreConsts` as an input to `BaseModel` is much more direct than expecting each model to merge in core constants as properties of a model specific class. Both work, though, so I don't think there's too much to worry about here.

davidorme (approved):
Nothing to add really to the existing comments

tests/conftest.py
jacobcook1995: Oh great, that is actually what my model requires!

virtual_rainforest/core/constants.py
jacobcook1995: The reason this is in core constants is because both the `soil` and `litter` models use it, but agree that it is confusing when the abiotic model also defined a soil layer height

virtual_rainforest/data_variables.toml
jacobcook1995: Oh right I had not realised that, shall we chat about this tomorrow?

virtual_rainforest/models/soil/carbon.py
jacobcook1995: effect sounds good, originally called this `environmental_impact_factors` but felt like that sounded too much like it was impacts **on** the environment

virtual_rainforest/models/soil/env_factors.py
vgro: ok, I see that the C2K is deprecated, we can chat about this tomorrow

virtual_rainforest/models/soil/carbon.py
```diff
@@ -1,34 +1,60 @@
...
+    maom_enzyme_production: NDArray[np.float32]
+    """Rate at which MAOM degrading enzymes are produced [kg C m^-3 day^-1]."""
+    necromass_decay_to_lmwc: NDArray[np.float32]
+    """Rate at which biomass is lost to the LMWC pool [kg C m^-3 day^-1]."""
+    necromass_decay_to_pom: NDArray[np.float32]
+    """Rate at which biomass is lost to the POM pool [kg C m^-3 day^-1]."""
+
+
+# TODO - This function should probably be shortened, leaving as is for the moment as a
+# sensible split will probably be more obvious once more is added to this function.
```
alexdewar, line 39: This function is indeed rather a beast. It is fine to leave for now (though maybe open an issue to fix it later?), but there's an awful lot going on and the number of arguments is also too high.

I don't really understand enough to give a better design for this, but in general I would say:

1. If lots of the arguments to this function are just being passed through to one or two other functions, you could just pass in the *results* of these functions instead and have the caller calculate them
2. If some of the arguments would naturally group together, consider bundling them together in a class/dataclass (provided that this function won't be the only user of it)

virtual_rainforest/models/soil/env_factors.py
alexdewar: It's a pity it was deprecated, because it's a lot terser! I suppose you could write your own helper function though?

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -204,6 +218,7 @@ def integrate(self) -> dict[str, DataArray]:
                 self.data[str(name)].to_numpy()
                 for name in self.data.data.keys()
                 if str(name).startswith("soil_c_pool_")
+                or str(name).startswith("soil_enzyme_")
```
alexdewar, line 221: You're only using `name` in stringified form here... maybe this will work instead?

```suggestion
                self.data[name].to_numpy()
                for name in map(str, self.data.data.keys())
                if name.startswith("soil_c_pool_")
                or name.startswith("soil_enzyme_")
```

virtual_rainforest/models/soil/soil_model.py
```diff
@@ -212,19 +227,28 @@ def integrate(self) -> dict[str, DataArray]:
             str(name): np.array([])
             for name in self.data.data.keys()
             if str(name).startswith("soil_c_pool_")
+            or str(name).startswith("soil_enzyme_")
```
alexdewar, line 230: Ditto.

```suggestion
            name: np.array([])
            for name in map(str, self.data.data.keys())
            if name.startswith("soil_c_pool_")
            or name.startswith("soil_enzyme_")
```

alexdewar (approved):
I've made a few minor comments, but otherwise LGTM.

virtual_rainforest/data_variables.toml
jacobcook1995: Switched to using `clay_fraction` from using `clay_percentage` as it felt like a less confusing unit

virtual_rainforest/models/soil/env_factors.py
jacobcook1995: I'll rename this parameter as it _is_ a different parameter from the one you use, the names just overlap a bit making it easy to confuse them

virtual_rainforest/models/soil/soil_model.py
jacobcook1995: Ahh yes that works (i.e. `mypy` doesn't complain), and reads a lot more cleanly

virtual_rainforest/models/soil/carbon.py
jacobcook1995: Oh yes, I should make an issue!

jacobcook1995: Have made issue #344 for this problem

tests/conftest.py
jacobcook1995: Just made #345 to reflect our in person discussion of this

virtual_rainforest/models/soil/env_factors.py
jacobcook1995: @vgro per our discussion the other day I've made an issue about switching to using Kelvin as the default unit #346. That should minimise the number of conversions that have to be made throughout the model

jacobcook1995:
Thanks for all the feedback, I made a few changes in response to it, but mainly I made a lot of issues as a lot of the problems raised didn't make sense to tackle in this PR. The new issues are as follows: #342, #344, #345, #346, #347, #348

dalonsoa (approved):
It looks good! And it makes total sense to open issues in response to comments when not directly relevant for this PR. You just need to not forget tackling them at some point!

virtual_rainforest/core/constants.py
vgro: The depth of the soil layers is defined in the core schemer and initialized by the plant model. We could add the 0.25m as an additional layer or replace it; for the `abiotic` model it does not matter what the values are.

tests/models/soil/test_carbon.py
```diff
@@ -81,17 +33,23 @@ def test_calculate_soil_carbon_updates(dummy_carbon_data, top_soil_layer_index):
...
         bulk_density=dummy_carbon_data["bulk_density"],
-        soil_moisture=np.array([0.5, 0.7, 0.6, 0.2]),
+        soil_moisture=dummy_carbon_data["soil_moisture"][
+            top_soil_layer_index
+        ].to_numpy(),
         soil_water_potential=dummy_carbon_data["matric_potential"][
             top_soil_layer_index
         ].to_numpy(),
+        vertical_flow_rate=dummy_carbon_data["vertical_flow"] / 30.0,
         soil_temp=dummy_carbon_data["soil_temperature"][top_soil_layer_index],
```
vgro, line 47: Add a TODO here to remove the `/ 30.0` once the vertical flow is changed to mean flow as discussed here #347

tests/models/soil/test_carbon.py
```diff
@@ -81,17 +33,23 @@ def test_calculate_soil_carbon_updates(dummy_carbon_data, top_soil_layer_index):
         soil_c_pool_maom=dummy_carbon_data["soil_c_pool_maom"].to_numpy(),
         soil_c_pool_microbe=dummy_carbon_data["soil_c_pool_microbe"].to_numpy(),
         soil_c_pool_pom=dummy_carbon_data["soil_c_pool_pom"].to_numpy(),
+        soil_enzyme_pom=dummy_carbon_data["soil_enzyme_pom"].to_numpy(),
+        soil_enzyme_maom=dummy_carbon_data["soil_enzyme_maom"].to_numpy(),
         pH=dummy_carbon_data["pH"],
         bulk_density=dummy_carbon_data["bulk_density"],
-        soil_moisture=np.array([0.5, 0.7, 0.6, 0.2]),
+        soil_moisture=dummy_carbon_data["soil_moisture"][
+            top_soil_layer_index
```
vgro, line 41: I thought you had replaced all `soil_moisture` with `matric_potential`, should we keep the soil moisture in for now? See #345

vgro (approved):
This looks good to me, a few minor comments

Issue #342, opened 2023-11-21
Include core constants as an attribute of each model constants class 


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
It's fiddly to have to track both model constants and core constants data-classes separately. So it would be simpler for each model constants data class to contain every relevant constant. By including the core constants data class as an attribute we can also retain the advantage of having a single location where all core constants are defined.

**Describe the solution you'd like**
Core constants could be set as an attribute and relevant constants could be returned using properties, e.g. 

```python
class LitterConsts:
    ...
    core_constants: CoreConsts = CoreConsts()

    @property
    def depth_of_active_soil_layer(self) -> float:
        return core_constants.depth_of_active_soil_layer
```

Then when constants data classes are created from the `Config`, the `CoreConsts` would have to be created first, and used to created the model specific constants data classes.

**Describe alternatives you've considered**
I don't see a clear alternative aside just putting up with each model having a separate `core_constants` and a `model_constants` dataclass.

**Additional context**
[This thread](https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/341#discussion_r1400393462) contains discussion of the issue. Also this issue has overlap with the broader issue #208.

davidorme:
Closing this. Over in #364, we've found a clean solution to implement this, but at the moment have chosen to keep separate `CoreConsts` and `ModelConsts` namespaces, probably with the addition of making a single `CoreConsts` instance at startup to use across all models.

PR #343, opened 2023-11-24
Animal Module constants rework

TaranRallings:
# Description

This is the rework of constants handling in the animal module. Everything seems to be working but do let me know if I've misunderstood something about implementing constants in this way. 

Fixes # (issue)

## Type of change

## Key checklist

- [x All tests pass: `$ poetry run pytest`

## Further checks


davidorme (approved):
LGTM. You may need to add another constants class to the model if you end up using core constants (and this may just become part of the BaseModel signature) but see also #342 - this is still up in the air, so just to be aware of.

Issue #349, opened 2023-12-05
Clarify role of required init vars


vgro:
The question came up what exactly needs to go in the `required_init_vars` of a model to ensure that the dummy simulation runs. Is it all the input variables needed to run the model or only the ones that come from somewhere external, like ERA5 data?

davidorme:
I think we probably need to discriminate between these two things. I think this is going to be at the heart of any attempt to freeze modules - some data has to be provided to start the model running, but others are generated from those. It's linked to the idea of what variables are _provided_ by which module.

vgro:
Ok, so I include all the inputs that the model expects to run

davidorme:
That it needs to `__init__`. I think in the near future we may end up defining all of these variables in their own register and provide a resolution process of which variables are populated by which models and which need to be provided etc.

davidorme:
@dalonsoa - this is linked to #371. We might want to split `required_init_vars` into `required_config_data` and `required_init_vars`? It would be useful in the variable docs to be very clear about what data _must_ be provided as an input and at the moment cannot be seen from `required_init_vars`. I guess we could work it out from `required_init_vars` and a new `vars_initialised` property across models, but it seems cleaner to be explicit!

vgro:
resolved with #370

PR #350, opened 2023-12-05
Feature/wind profiles

vgro:
This PR adds the wind submodule to the new `abiotic` model. It is not very complex, basically just a long list of functions that in the end are called in the `AbioticModel`.

@TaranRallings did you make some commits to this branch or did this come from `git merge develop`? It looks like I missed an update, please ignore the animal files in the review, we can sort this out when I eventually merge in develop.

The question came up about what the `required_init_vars` should contain regarding the dummy data set, please comment on #349 .

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/abiotic/abiotic_model.py
```diff
@@ -122,5 +136,63 @@ def update(self, time_index: int, **kwargs: Any) -> None:
...
+            sensible_heat_flux_topofcanopy=(
+                self.data["sensible_heat_flux_topofcanopy"].to_numpy()
+            ),
+            wind_speed_ref=(
+                self.data["wind_speed_ref"].isel(time_index=time_index).to_numpy()
+            ),
+            wind_reference_height=(self.data["canopy_height"] + 10).to_numpy(),
+            turbulence_sign=True,
+            abiotic_constants=AbioticConsts(),
+            core_constants=CoreConsts(),
```
jacobcook1995, line 165: Doesn't this make it impossible for the user to change the constants, because default values will always be used?

virtual_rainforest/models/abiotic/wind.py
```diff
@@ -0,0 +1,642 @@
...
+
+    where :math:`u_z` is wind speed at height :math:`z` within the canopy, :math:`u_h`
+    is wind speed at the top of the canopy at height :math:`h`, and :math:`a` is a wind
+    attenuation coefficient given by :math:`a = 2 l_m i_w`, where :math:`c_d` is a drag
+    coefficient that varies with leaf inclination and shape, :math:`i_w` is a
+    coefficient describing relative turbulence intensity and :math:`l_m` is the mean
+    mixing length, equivalent to the free space between the leaves and stems. For
+    details, see :cite:t:`maclean_microclimc_2021`.
+
+    Args:
```
jacobcook1995, line 507: You forgot to fill these out!

jacobcook1995 (changes requested):
Noticed a couple of problems, but otherwise looks good to me!

tests/models/abiotic/test_wind.py
```diff
@@ -0,0 +1,364 @@
...
+        roughness_length_momentum=np.array([0.017, 1.4533, 0.9591]),
+        mixing_length_factor=AbioticConsts.mixing_length_factor,
+    )
+
+    np.testing.assert_allclose(
+        result, np.array([1.35804, 1.401984, 0.925228]), rtol=1e-4, atol=1e-4
+    )
+
+
+def test_generate_relative_turbulence_intensity(dummy_climate_data):
```
jacobcook1995, line 119: You probably should also test the case where `increasing_with_height=False`

virtual_rainforest/models/abiotic/abiotic_model.py
vgro: oh yes, you're right. I copied the arguments and missed changing that.

virtual_rainforest/models/abiotic/wind.py
vgro: Well spotted 👍

jacobcook1995 (approved):
LGTM!

Issue #307, opened 2023-09-07
Switch away from `pint` time units


davidorme:
**Describe the bug**

The `pint` time units are not calendar aware and use _average_ lengths for months etc. The critical `update_interval` is stored as a `pint.Quantity`. We need calendar aware time calculations, which are probably most easily achieved using `np.datetime64`and `np.timedelta64`. There doesn't seem to be a simple conversion between `pint.Quantity` and `np` without writing some kind of interpreter.

**Expected behavior**

Months should not be 30.4375 days.

I think `pint` is _only_ currently used for the `update_interval`. It seems easiest to replace it with using `np.timedelta64` and a config like `[1, 'month']` , with a slim wrapper to convert a subset of time intervals to e.g. `np.timedelta64(1, 'M')`.

Issue #348, opened 2023-12-04
Define depth of first soil layer used by the abiotic (etc) models based on soil model constant


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
Currently the `soil` model defines a constant (`CoreConsts.depth_of_active_soil_layer`) which sets the depth of soil which is considered to be biogeochemically active, i.e. the depth of topsoil simulated in the `soil` model. The `abiotic`/`hydrology` models then also define soil layers separately (using config options).

@vgro pointed out that this is strange as we really want the first soil layer to correspond to the topsoil simulated by the `soil` model. At present the default options for both are set to 0.25m, but this still would allow users to select different values for the different models.

**Describe the solution you'd like**
I'm not 100% sure if we want to do this, but we could populate the topsoil layer height based on the core `depth_of_active_soil_layer` constant. This would have the advantage that the abiotic environmental information calculated for the 1st soil layer would always correspond to the layer that the `soil` model simulates. The drawback would be that this makes the system more inflexible.

vgro:
The depth of the soil layers is defined in the core schemer and initialized by the plant model. We could add the 0.25m as an additional layer or replace it; for the `abiotic` model it does not matter what the values are.

jacobcook1995:
I think `CoreConsts.depth_of_active_soil_layer` should be renamed `CoreConsts.depth_of_active_soil` as it isn't a layer. Then a function can be written to extract reasonable averages for soil temp and moisture for the biologically active soil.

jacobcook1995:
This has been addressed by the introduction of a layer structure. I'm closing this issue in favour of #532 and #533, which specifically set out the cleaning up that I need to do to move the `litter` and `soil` models to use the new system

Issue #256, opened 2023-07-12
Fix time_index in dummy climate data


vgro:
Climate data is accessed via `isel(time_index=time_index)`. The dummy climate data currently has the coordinated `time` and `time_index`, but only one dimension which is `time_index`. This was created by renaming the `time` dimension, which is a bit of a hack. This is not of high priority,  I will find a solution when I have some time to test.

davidorme:
I've just run into this when integrating the plant model.

davidorme:
The specific issue here was addressed in #339

PR #352, opened 2023-12-07
Plants model integration into full `vr_run`

davidorme:
# Description

This PR handles the remaining issues from adding the plants model into the `vr_run` model stack. The main things of note are:

* The above canopy offset, surface height and subcanopy heights are now all configurable.
* The `plants` model now uses a `LayerStructure` dataclass to bundle those values with `n_canopy_layers` and `soil_layers` into a single structure used to define the vertical structure. One of those can be generated from `Config` instance using a factory method.
* Evapotranspiration is now populated by `PlantsModel.estimate_gpp`, which is the likely point at which this should be properly calculated as the model develops.
* Notably **evapotranspiration is no longer an required init var for `hydrology`**. It was only ever used in update. This reinforces my feeling that we need to revisit the variable system and make it more structured, but not now 😄 
* The existing dummy plant data for `evapotranspiration` is removed.
* The model dependencies in the example data configuration are set to get a passing (but likely flawed!) resolution of variable calculation.

## Type of change

## Key checklist

## Further checks


tests/models/plants/conftest.py
```diff
@@ -26,6 +26,13 @@ def plants_config():
         out_initial_file_name = "model_at_start.nc"
         out_final_file_name = "model_at_end.nc"
 
+        [core.layers]
+        canopy_layers = 10
+        soil_layers = [-0.5, -1.0]
```
jacobcook1995, line 31: Recently switched the default topsoil depth to 25cm, so might be good to change this to `[-0.25, -1.0]`.

More broadly we should think about whether this topsoil layer should defined based on the global constant (see #348), or alternatively whether that constant should be removed in favour of just using these layer heights. But that's out of the scope of this PR.

jacobcook1995 (commented):
Looks sensible to me (though I don't have much grasp of the plant-abiotic interface stuff). I just had a few queries looking through the PR.

tests/models/plants/conftest.py
```diff
@@ -112,7 +129,8 @@ def plants_data():
         data=np.full((4, 12), fill_value=1000),
         coords={
             "cell_id": np.arange(4),
-            "time": np.arange("2000-01", "2001-01", dtype="datetime64[M]"),
+            # "time": np.arange("2000-01", "2001-01", dtype="datetime64[M]"),
```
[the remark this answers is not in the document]
davidorme: I had left it as kind of a reminder that we need to fix #307 and #256 but that's what issues are for!

virtual_rainforest/core/module_schema.json
```diff
@@ -205,12 +205,33 @@
...
+                  "above_canopy_height_offset": {
+                     "description": "The height offset relative to the canopy top for climatic reference variables.",
+                     "type": "number",
+                     "exclusiveMinimum": 0,
+                     "default": 2.0
+                  },
+                  "surface_layer_height": {
+                     "description": "The height used to calculate ground surface microclimate conditions.",
+                     "type": "number",
+                     "exclusiveMinimum": 0,
```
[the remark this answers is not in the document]
vgro: yes, this value should be above zero

virtual_rainforest/models/plants/canopy.py
```diff
@@ -7,17 +7,60 @@
...
+    ``core.layers`` configuration section.
+    """
+
+    canopy_layers: int
+    """The maximum number of canopy layers."""
+    soil_layers: list[float]
+    """A list of the depths of soil layer boundaries."""
+    above_canopy_height_offset: float
+    """The height above the canopy at which reference climate variables are measured."""
+    surface_layer_height: float
```
vgro, line 38: This is a bit of a tricky one. Most climate data provide the 2m data above ground; if it is not measured they interpolate the data or model output to that. So technically, all the input from ERA5 for example is 2m above ground. For the model, we want the conditions above the canopy as an input. In reality, the downscaling will most likely introduce much more uncertainty than the difference between top of canopy and below canopy, we just need to make this very clear. 
Therefore, it might be better to say `The height above the canopy for which reference climate variables are provided.`

virtual_rainforest/models/plants/plants_model.py
```diff
@@ -93,9 +94,7 @@ class PlantsModel(BaseModel):
         "layer_fapar",
         "layer_leaf_mass",  # NOTE - placeholder resource for herbivory
         "layer_absorbed_irradiation",
-        "herbivory",
-        "transpiration",
-        "canopy_evaporation",
+        "evapotranspiration",
     )
```
vgro, line 98: Did we not decided that `plants` only updates the transpiration?

virtual_rainforest/models/plants/plants_model.py
```diff
@@ -154,28 +151,25 @@ def from_config(
...
         # Try and create the instance - safeguard against exceptions from __init__
         try:
             inst = cls(
                 data=data,
                 update_interval=update_interval,
                 flora=flora,
                 constants=constants,
-                canopy_layers=canopy_layers,
-                soil_layers=soil_layers,
+                layer_structure=layer_structure,
```
vgro, line 172: Is this new structure going to be rolled out to all models?

vgro (approved):
LGTM, only a few minor comments

virtual_rainforest/models/plants/plants_model.py
davidorme: See #353! 😄

virtual_rainforest/models/plants/plants_model.py
vgro: I see. This will need some adjustment from my side. The `evapotranspiration` is used in the hydrology model to estimate how much water is taken from the soil by the roots. So it should really only be transpiration, but at that time I thought that the plant model does calculate both but not return the components individually. If the canopy evaporation is calculated in the hydrology model, and the transpiration in the plants, we will only need the transpiration as input to the `setup_hydrology_input_current_timestep`. You can change it here if you want or I'll do it in a future PR

davidorme: I think we push this to a future PR 😄

Issue #354, opened 2023-12-14
`xarray.open_mfdataset` can cause problems in `vr_run`


jacobcook1995:
**Describe the bug**
Sometimes when `vr_run` is called it stalls at the point it uses `xarray.open_mfdataset()` to combine the files for individual time points into a single file. This sometimes happens on CI meaning the tests don't pass (because they don't finish). This appears to be a [known bug in `xarray`](https://github.com/pydata/xarray/issues/3961).

**To Reproduce**
By its nature it is hard to reproduce this bug but when I run `vr_run` locally (2017 MacBook Pro) it stalls ~1 in 3 times. This is more frequent than happens on the CI so I suspect the stall is somewhat spec driven.

**Approach to fix**
The bug thread suggests setting `lock=False` as a work around but suggests that this can cause errors as `hdf5` is not thread safe. I feel like stopping CI from randomly breaking is probably the top priority for now, so I think we should go ahead and set this option. If setting this option proves to be too error prone for our use case then we will need to think of an alternative way of merging data, e.g. possibly a dedicated tool for merging data

jacobcook1995:
The actual halt is now resolved by #355, but the option may cause occasional errors. If we start seeing `HDF5` errors with any regularity then we need to revisit our solution to this issue

davidorme:
This cropped up again and I've raised PR #457 to explicitly turn off any attempt to parallelise the file handling in `dask`. I think that is as much as we can do here - it's a problem all the way down the stack to the NetCDF code, I think.

PR #355, opened 2023-12-14
Changed `open_mfdataset` call to use `lock=False` to fix `vr_run` halt

jacobcook1995:
# Description

Where `open_mfdataset` is used it is now called with the `lock=false` option selected. This should prevent `vr_run` from halting during CI.

I couldn't think of any way to prove that this has fixed the issue, but I did run `vr_run` locally 10 times without a stall. Previously this would have caused ~3 stalls on average.

Partially addresses #354, think the issue should be kept open so we have a way of tracking if this fix works sufficiently well.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/core/data.py
+++ b/virtual_rainforest/core/data.py
@@ -497,7 +497,7 @@ def merge_continuous_data_files(
     check_outfile(out_path)
 
     # Open all files as a single dataset
-    with open_mfdataset(continuous_data_files) as all_data:
+    with open_mfdataset(continuous_data_files, lock=False) as all_data:
         # Specify type of the layer roles object to allow for quicker saving by dask
         all_data["layer_roles"] = all_data["layer_roles"].astype("S9")
 


alexdewar:
Hmm... how annoying!

Is the bug restricted to a particular OS or can it occur with any of them? In principle I think this is fine if it works, however in the upstream issue it mentions that this may cause the hdf5 library to throw an error as it isn't thread safe. Have you come across this?

If you're only seeing the problem in the CI system, would it be worth only setting this option when running under CI (e.g. by using an environment variable)?

davidorme:
I've seen it occur on Mac and Ubuntu on CI and Jacob is seeing it locally on a Mac.

alexdewar (approved):
If it's causing problems locally and not just on CI, then I guess you don't have much choice.

I would maybe add a comment and/or open an issue describing that this is a hack with potential problems though, in case it trips someone up in future.

jacobcook1995:
Yep I'll edit the issue to mention that this is a bit of a hack, and needs to be kept an eye on

PR #356, opened 2024-01-03
Remove Python 3.9

### LINKED ISSUE #324: Consider dropping Python 3.9

There are a bunch of new typing related syntaxes that are available from 3.10. The snakemake helper is moving towards 3.10+ and it might make sense to also update `vr` itself.

Before we start making releases (soon!), we probably want to push the codebase to support only the most recent Python versions, rather than tying us to legacy versions that were only used while the project was in `alpha` development.

* Drop Python 3.9
* Update syntax to use newer features (see comment: https://github.com/ImperialCollegeLondon/virtual_rainforest_snakemake_template/issues/21#issuecomment-1742693120)

Keep Python 3.9

davidorme:
# Description

This PR is intended to remove Python 3.9. It adds `pyupgrade` to the `pre-commit` hooks with a minimum Python version of 3.10 and that then made a bunch of changes automatically. These are mostly:

- Replacing e.g. `set([1,2,3])` with the set literal `{1,2,3}`
- Replacing e.g. `Optional[int]` with `int | None`

Fixes #324

But at the moment, there are some issues about the documentation of `vr.core.grid.GRID_STRUCTURE_SIG`. I've updated the way that type alias is created, but `sphinx` is throwing errors. I _think_ it believes the type alias (`tuple[list[int], list[Polygon]]` is an actual tuple and is trying to document the `__repr__`, `count` and `index` methods?

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.10"
       - uses: pre-commit/action@v3.0.0
 
   test:
@@ -19,7 +19,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        python-version: [ "3.9" , "3.10", "3.11" ]
+        python-version: [ "3.10", "3.11" ]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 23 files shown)


jacobcook1995 (approved):
LGTM!

Does it make sense to update the `poetry.lock` as part of this PR? The minimum supported python version has changed, which means quite a lot of dependancies should now be less version constrained.

I've also found `intersphinx` to be really annoying, and have no guess for the problem in this case.

davidorme:
> Does it make sense to update the `poetry.lock` as part of this PR? The minimum supported python version has changed, which means quite a lot of dependancies should now be less version constrained.

Yes it really does. Good catch. I've updated `pyproject.toml` and then specifically updated the minimum version on couple of packages (`sphinx` and `myst-nb`) that were also holding us back from more recent packages.

PR #362, opened 2024-01-16
Feature/initialise for first timestep

vgro:
This PR initializes the variables required for the first time step of the energy balance. This includes atmospheric variables, canopy temperature, and a bunch of conductivities.
I also updated the typing syntax in the wind submodule.

## Type of change

## Key checklist

## Further checks


virtual_rainforest/models/abiotic/abiotic_model.py
```diff
@@ -31,7 +31,7 @@ class as a child of the :class:`~virtual_rainforest.core.base_model.BaseModel` c
 from virtual_rainforest.core.data import Data
 from virtual_rainforest.core.logger import LOGGER
 from virtual_rainforest.core.utils import set_layer_roles
-from virtual_rainforest.models.abiotic import wind
+from virtual_rainforest.models.abiotic import energy_balance, wind
 from virtual_rainforest.models.abiotic.constants import AbioticConsts
```
vgro, line 35: I think this might be unnecessary here as the constants are imported via load_config?

virtual_rainforest/models/abiotic/abiotic_model.py
```diff
@@ -154,6 +142,41 @@ def setup(self) -> None:
             "vapour_pressure_deficit_ref"
         )
 
+        # Generate initial profiles of air temperature [C], relative humidity [-],
+        # vapour pressure deficit [kPa], soil temperature [C], atmospheric pressure
+        # [kPa], and atmospheric :math:`\ce{CO2}` [ppm]
+        initial_atmosphere = microclimate.run_microclimate(
```
vgro, line 148: rename `initial_microclimate`

virtual_rainforest/models/abiotic/abiotic_model.py
```diff
@@ -154,6 +142,41 @@ def setup(self) -> None:
...
         )
 
+        # Generate initial profiles of air temperature [C], relative humidity [-],
+        # vapour pressure deficit [kPa], soil temperature [C], atmospheric pressure
+        # [kPa], and atmospheric :math:`\ce{CO2}` [ppm]
+        initial_microclimate = microclimate.run_microclimate(
+            data=self.data,
+            layer_roles=self.layer_roles,
+            time_index=0,
+            constants=AbioticSimpleConsts(),
```
jacobcook1995, line 152: Do you supply `AbioticSimpleConsts()` instead of `self.constants` to shut `mypy` up?

virtual_rainforest/models/abiotic/constants.py
```diff
@@ -163,3 +163,64 @@ class AbioticConsts(ConstantsDataclass):
...
+    :cite:p:`aston_heat_1985`.
+    """
+
+    leaf_heat_transfer_coefficient: float = 50.0
+    """Leaf heat transfer coefficient, [s^1/2 m^-1/2],
+    :cite:p:`linacre_determinations_1964`, default = 50.
+    """
+
+    stomatal_resistance: float = 200.0
+    """Stomatal resistance, default = 200."""
```
jacobcook1995, line 196: I don't think you strictly need to add default values here, as we _should_ be able to document the defaults automatically. Though, if you find adding them here helpful to you then that's fine

jacobcook1995 (approved):
Couple of minor comments/queries, but otherwise looks good to me!

virtual_rainforest/models/abiotic/abiotic_model.py
vgro: That was the issue we talked about yesterday. The model constants  `self.constants` are `AbioticConsts` but I also need `AbioticSimpleConsts`, so I wasn't sure how to load the best

PR #365, opened 2024-01-19
Refactor BaseModel

### LINKED ISSUE #363: Refactor BaseModel class
### Discussed in https://github.com/ImperialCollegeLondon/virtual_rainforest/discussions/215

<div type='discussions-op-text'>

<sup>Originally posted by **dalonsoa** April 28, 2023</sup>
Don't panic! It is much simpler than it looks :) 

At the moment, to create a new model, you need to derive from the `BaseModel` class, ensure that there are a few attributes present (up to 4 after #211 ), validate that those attributes exist in the child class and that they fulfil certain conditions. It all works and makes sense.

However, it has always looked me as a bit convoluted, in particular the part about requiring the presence of the attributes and the check that they are indeed properties. I'm here to suggest an, I think, cleaner approach exploiting once more the `__init_subclass__` method.

The idea would be to require the presence of certain keyword arguments directly when subclassing, and validate those before assigning them to the class attributes. This would be a simple example:

```python
from abc import ABC

class Base(ABC):
    a: str
    b: int

    @classmethod
    def __init_subclass__(cls, a: str, b: int):
        try:
            cls.a = validate_a(a)
            cls.b = validate_b(b)
        except:
            pass

class Child(Base, a="a", b=25):
    pass

if __name__ == "__main__":
    something = Child()
    print(something.a)
    print(something.b)
```

This enforces the `Child` class to be created with appropriate values for `a` and `b` without having to define any properties in the subclass. You just need to check if the values passed are valid - eg. if they are within certain bounds or something like that - but their presence is guarantee by design of the `__init_subclass__` definition.

In summary, checks like these become unnecessary: https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/5a821952bdb25a044b135532f9f7ad7c161ffebf/virtual_rainforest/core/base_model.py#L304

And if you really, really want them to be `properties`, as you have now, you just make it so in the `BaseModel` class directly, saving code everywhere else:

```python
class Base(ABC):
    _a: str
    _b: int

    @property 
    def a(self):
        return self._a

    @property 
    def b(self):
        return self._b

    @classmethod
    def __init_subclass__(cls, a: str, b: int):
        try:
            cls._a = validate_a(a)
            cls._b = validate_b(b)
        except:
            pass

```

Anyway, just an idea, and certainly not urgent. 
</div>

davidorme:
# Description

This implements @dalonsoa's suggested refactor of `BaseModel` from #363. We thought this seemed like a good idea at the Github meeting  so I've worked it through.

Briefly, that changes how we set model class attributes from:

```python
class NewModel(BaseModel):
     model_name="base_model"
     ...
```

to 

```python
class NewModel(
     BaseModel,
     model_name="base_model"
):
     ...
```

**Another change** that fell out of this update is that it looked much cleaner to have a single `model_update_bounds` attribute set like `('1 day', '1 month')` rather than separating those as `lower_bound_on_time_scale` and `upper_bound_on_time_scale`. It means `_check_time_bounds_units` is replaced by `_check_model_update_bounds`, which is only called once and also moves the test that `1 day < 1 month` inside the check function.

### Pros on the switch

* **The BaseModel code becomes much simpler.** Before we had a bunch of properties that might or might not be set on a new model and had to test whether they'd been set or not and then what they were. Now they are explicitly set when `__init__subclass__` runs.
* **Model definition becomes more robust.** Now these values are arguments to the class creation signature, users get all the usual warnings about forgotten arguments and typing.
* **The testing code becomes much cleaner.**  Before, we were using `exec()` to run chunks of text that described model definitions in order to set different attributes for testing. We can now just pass those attributes in to tests as `kwargs` to the model creation call.

### Cons on the switch

* We now don't have an obvious place to put model specific docstrings on the attribute. For example, I had a docstring on `PlantsModel.required_init_vars` that gave an overview. I've moved that into a specific section in the model docstring instead.  (Honestly I'm not sure this is really a con  - that seems like a more logical place for people to read about model requirements).

Fixes #363 (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/api/core/base_model.md
+++ b/docs/source/api/core/base_model.md
@@ -22,5 +22,5 @@ kernelspec:
     :members:
     :special-members: __init_subclass__, __repr__, __str__, __init__
     :private-members: _check_required_init_vars, _check_model_name,
-      _check_time_bounds_units
+      _check_model_update_bounds, _check_vars_updated
 ```
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -162,7 +162,7 @@ from virtual_rainforest.models.freshwater.streamflow import calculate_streamflow
 
 Now create a new class, that derives from the
 {mod}`~virtual_rainforest.core.base_model.BaseModel`. To begin with, choose a class name
-for the model and define the following four class attributes.
+for the model and define the following class attributes.
 
 The {attr}`~virtual_rainforest.core.base_model.BaseModel.model_name` attribute
 : This is a string providing the name that is used to refer to this model class in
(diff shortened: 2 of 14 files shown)


dalonsoa (approved):
What a great idea XDDD

Jokes aside, the code now looks much cleaner and I feel it will be more robust. Great work!

jacobcook1995 (approved):
LGTM!

vgro (approved):
LGTM :-)

PR #368, opened 2024-01-21
Introduce core constants as a core element

### LINKED ISSUE #366: Create a single instance of CoreConst

Off the back of the decisions in #364, we are for the moment sticking with models taking `CoreConst` and `ModelConst` objects. We can see how that plays out and have a path in #364 for an alternative approach to a single object.

Given that, we should extend the BaseModel definition so that it includes that `CoreConst` argument and then create a single instance `vr_run` to be passed off when models are initiatialised.

 Some models may not use the `CoreConst` - although I'd expect that to get rare as we develop the contents of that class. We could simply stick with the current approach of having models create their own `CoreConst` instance from `Config` if they need it.

davidorme:
# Description

This PR fixes #366 - but I don't think we should accept it because this feels horribly around the houses and clunky. I'm creating a PR to make it easy to see those issues and discuss a wider change in the BaseModel.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/development/defining_new_models.md
+++ b/docs/source/development/defining_new_models.md
@@ -229,7 +229,12 @@ things.
    also known as the superclass:
 
    ```python
-   super().__init__(data, update_interval, **kwargs)
+   super().__init__(
+        data=data, 
+        update_interval=update_interval, 
+        core_constants=core_constants, 
+        **kwargs
+    )
    ```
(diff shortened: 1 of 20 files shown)


davidorme:
Closing as superseded by issue #369 and PRs #374 and #375

PR #373, opened 2024-01-23
Adds support for static models

dalonsoa:
# Description

This PR setups the models such that they can be _frozen_ or _static_, i.e. they are setup once and they are used the same, without updating, for the rest of the simulation. This is controlled by a `static` flag included in the configuration.

There are two options for static models:

- A frozen model: updated once in the first iteration and then use those values for any future update. In this case, **NONE** of the variables setup during the update can be provided.
- A pre-set model: pulling data from a dataset that contains the relevant variables for the whole time range of the simulations. In this case **ALL** of the data variables that the model is supposed to update - or provide during the setup process - need to be provided.

Documentation about using the flag has been added to the `config.md` file, but I'm not sure if there's a better place.

Most modifications to the test have consisted in bypassing the static checks because, as the tests were designed, they were all failing. In all cases, only some of the variables required for init or for update were present in the data object, typically provided as a fixture, which were making the checks to decide on the _staticity_ of the models fail. Options are to leave the patches used as they are, or to add all the missing variables. However, that might be a lot of work and I'm not entirely sure if it is worth the effort.

Fixes # N/A

## Type of change

## Key checklist

## Further checks


virtual_rainforest/core/base_model.py
```diff
@@ -192,13 +199,61 @@ def setup(self) -> None:
...
+
+    def _update_static(self, time_index: int, **kwargs: Any) -> None:
+        """Function to update the model in the static case.
+
+        Args:
+            time_index: The index representing the current time step in the data object.
+        """
+        if self._static_data is None:
+            self._update(time_index, **kwargs)
+            self._build_static_data(time_index)
```
jacobcook1995, line 230: Not sure that I follow this bit, does the update happen and then new static data gets built for the next time step?

jacobcook1995 (approved):
I got slightly lost in one bit of the pseudocode, but the general approach seems sensible and flexible.

virtual_rainforest/core/base_model.py
dalonsoa: Exactly. This is run only once, the first time that we are required to run the update and, thereafter the updated values are used for updating the data object again and again. We need to run update at least once in this case since, otherwise, we do not know what to update the data with. 

If pre-set data is provided as an external file/s, then the `static_data` object is not none and this step is not needed. 

Keeping aside changes in initialisation/validation, the only change to existing models would be to replace `update` by `_update`, since they are overwritting only one of the two options for updating the model. The `_update_static` option is not model specific, and therefore is coded directly in the base model.

vgro (approved):
I am not so familiar with `self._`, but the approach looks generally good to me.

davidorme:
This makes a lot sense to me - the flow of the logic for the "update once and then repeat" is very clean.

I wonder if the `static_data` could be `use_static_data` and the model class itself is then responsible for locating the update_vars in `Data` and checking all is well. That static data assembly could be outside of the `BaseModel` but it seems like a more natural home? One thing this should support is the data providing only a single time step, which then just gets re-used. Users could always just repeat the same values across all time steps, but it would be good to avoid that! It was behind my thinking on the original `constant` option in the Data configuration, which might be worth reviving - if a grid is loaded as a constant, use that as a constant, but it has to be explicitly flagged as constant.

There are no end of repeating patterns and sequences we _might_ want to push in (La Niña, El Niño), but that seems like an evolution 😄

davidorme (approved):

dalonsoa:
> I wonder if the `static_data` could be `use_static_data` and the model class itself is then responsible for locating the update_vars in `Data` and checking all is well. That static data assembly could be outside of the `BaseModel` but it seems like a more natural home?

I'm not entirely sure what you mean here. Do you mean having a method called `use_static_data` that pulls the static data itself from somewhere else - outside of the model class - and put it in the relevant place in the Data object? Isn't that the intention of the `_update_with_static_data` method I sketched? Maybe I'm misunderstanding something. 

> One thing this should support is the data providing only a single time step, which then just gets re-used. Users could always just repeat the same values across all time steps, but it would be good to avoid that! It was behind my thinking on the original `constant` option in the Data configuration, which might be worth reviving - if a grid is loaded as a constant, use that as a constant, but it has to be explicitly flagged as constant.

I guess that the model can hold a single copy of that single step data, but then in the Data object it will need to update timestep after timestep to create the time series (when relevant), so other models can use it. I think that models that use that data should not need to worry about it being static or not, but just take the appropriate time from the Data object and go on with it, right?

davidorme:
> I'm not entirely sure what you mean here.

Yeah - wasn't a model of clarity was it. So:

* If users are providing frozen data (rather than looping the first update) then the most obvious current way to do that is for them to add those variables in the configuration. Those variable names _normally_ wouldn't be included in the data configuration because they are calculated internally, but it is a simple existing route to setting up the model.
* If the `Data` instance then contains the variables, we don't need to pass in an `xr.Dataset` to the frozen model. We can replace:
  ```python
  static_data: Optional[xr.Dataset] = None,
  ```
  with
  ```python
  use_static_data: bool = False,
  ```
  and then the `_build_static_data` method is expecting to be able to find the right variables in `self.Data`.

davidorme:
> I guess that the model can hold a single copy of that single step data, but then in the Data object it will need to update timestep after timestep to create the time series (when relevant), so other models can use it. I think that models that use that data should not need to worry about it being static or not, but just take the appropriate time from the Data object and go on with it, right?

Yup - agreed.

dalonsoa:
Next week, this one. If you have any update or comment on what was said above, please let me know.

TaranRallings (approved):
Makes sense to me.

dalonsoa:
I've been trying to progress on this before going on leave but I've hit a blocker - precisely because of the brand new variables system.

For the case of the frozen model - updates once and then repeat for future iterations - things are clear as the inputs are the same as a normal model. I have not implemented it but I do not foresee any problem there: it will make a local copy of the relevant data (the `vars_updated`) after that first iteration and just keep using that down the line.

Now, for the pre-set option things have gotten more complicated after implementing the variables system. As discussed above, that preset data will need to be passed as part of the data object, but that would mean that the model has different vars required for update when it is static or when it is not: in the first case, the pre-set data will need to be present, but in the normal case it will not, in general (or at least, not necessarily). As these variables are set before the model instances are initialised, when collecting the models, there is no way to modify them at runtime based on the model input settings (the static flag).

There are a couple of options that come to my mind, none of them satisfactory:
- Create static model subclasses with the right variables, eg `PlantStaticModel`, but that defeats the purpose of having these simple flags to switch between model types.
- Postpone the validation step to after the models have been initialised, so the flags can be used to define one or other set of variables depending on the model being preset or not, but that means risking initialising models that are incompatible, in addition not not allowing the variable system to define the model execution order (future work).
- Pass the pre-set data directly to the model, without using the data object. It will be a question of updating the `from_config` methods of the different models to load that data if the appropriate flags and paths to the relevant file are provided in the config. This will create a local `static_data: Data` object that can be used to update the main data object each iteration. This might be the less problematic option if we use the same tooling of the data system to load and validate the preset data.

It will be good to have your feedback on this, so I can work on it as soon as I'm back from leave.

davidorme:
Yeah - that's a pain.  First thoughts:

* I don't like the static subclasses or postponed validation options - admittedly a knee jerk response though.
* The last option feels cleaner. 

Having said that, I haven't got my head around how we work with avoiding running code in static mode models. 

* If we insist that a static model can either run in frozen mode or in preset, but you have to preset _all_ variables required by the model, then I can see a route for the static data object for preset variables. If we have a `static_data` attribute for `BaseModel`, then the `super.__init__` and `super.update` methods could include an automatic  check if _all_ the required variables for that method are present in `static_data` and then simply copy them across to `data` and set a flag that allows the subclass to shortcut the actual method code. Something like:

```python
def __init__(self,  ...):

    super().__init__(data=data, static_data=static_data, core_components=core_components, **kwargs)
    
    if self.entirely_static_init:
        return
```

* However that can't work if a user doesn't provide _all_ the required variables, because the actual model code needs to run to fill in the missing variables for the method. At that point, if we are to avoid a whole ton of unpleasant `if var not in self.static_data` statements, we would have to run the actual model and then overwrite the calculations from `static_data` at the end of the method.

* If we're accepting that computational expense - and I'm not sure we should - then another way we could do this might be to have a "static" flag in the data setup. That could then be added as an attribute to the variable in `Data` and the code to set data could then simply refuse to set "static" variables. We've gone to the expense of calculating it, but the model isn't allowed to set it. This doesn't raise an error, just logs that it is refusing to overwrite a static variable. That removes the need for code at the end of the method overwriting with the preset data.

This reminds me a little of https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/188!

Thinking about it - that "static" flag might allow us to avoid the extra `static_data` object. 

* A model might start and immediately run the `super.__init__`, 
* That runs a `BaseModel._check_entirely_static` method and that looks at `Data` noting if all the required variables are present and flagged and static.
* If they are all there and static, we can set a flag to shortcut the whole method.
* If only some are there and static, then we have to calculate them at each time step but refuse to actually set them and hold the static values.
* And if any are there but not static - we throw an error? Attempt to provide an internally calculated variable.

dalonsoa:
I've been working on this and after a few trials and errors I think there are only 3 scenarios involving the `static` flag that make sense:

1. **The flag is provided and ALL the variables that would have been created/updated in `init` and/or `update` are provided as inputs**: In this case, we just bypass `init` and/or `update`. This applies to `init` and `update` independently.
2. **The flag is provided and NOT ALL of those variables are provided, only some**: We rise an error. Updating some and not others risk inconsistencies between the variables updated by the model.  This applies to `init` and `update` independently.
3. **The flag is provided and NONE of those variables are provided**: In this case we run the methods once and then bypass them in any future call. This only applies to `update`, since `init` is run only once, anyway. For those variables that are a time series, we might need to update the Data object with the previous value - which is always the same - but I'm not sure.

I will implement a prototype of how that _bypassing_ can be done with minimal fuss.

dalonsoa:
Ok, so here you have another take at this. The code seems verbose because I tried to be thorough with the error messages, but actual lines changed are very few. 

This implements two methods:
- One that checks if the setup (the new `init`, see below) needs to be bypassed based on the presence of the `static` flags and the variables available in the data object. 
- Another that checks if the update method should be run once when if the model is `static` and the relevant variables are not present in the data object.

For this to work, the following changes will need to be done to all the models:
- [x] Everything in `init` after the `super().__init__(...)` call should be moved to a new abstract method called `_setup`. I guess we can use `setup` if you don't have any use for that one anymore. This method should take as input everything that is not already passed to `super().__init__(...)`, as that will be already available.
- [x] The subclasses `init` method should be removed. So no custom `init` anymore, just custom `_setup`. If it is present, there will be an exception.
- [x] Subclasses will need to override `_update` and NOT `update`, as it is now. If they override `update`, there will be an error, as with `__init__`.
- [x] In the `from_config` method, the models will need to read a `static` flag from the configuration and pass it to the class constructor. If the flag is not present, the default is not being static.
- [ ] Optionally, the input files will need to contain the data that will be used by the static model when it is operating in a preset model - i.e. all data provided externally.

OK, so give it some thoughts and we can move from here. The pending things/open questions are:
- Do we need to update the `data` object even when the model is static, coping the previous entry in the case of time series?
- Do we need to do anything to load the data files related to preset data?
- I guess that we will need to update the schemas to allow for a `static` flag, right?

jacobcook1995:
> * Do we need to update the `data` object even when the model is static, coping the previous entry in the case of time series?

The `data` object currently stores the present state of the model variables as well as time series data for inputs, i.e. regional climate projections over time. The variables that represent the current state get output at every time step (I think we currently output all of them, but it could be a user defined subset). There's nothing within the `data` object that records the previous state of the model itself. I feel like this means the `data` object doesn't need to be updated for static models as the present state is by definition unchanging.

I guess we might want to change how the `data` object is used so that previous values are stored (for at least some variables) at some point in the future. But I feel like this would be a fairly big refactor which would also inevitably change some of the static model logic so probably isn't worth hedging against

dalonsoa:
@davidorme , did you have a chance to look into the new version of this?

davidorme:
@dalonsoa I hadn't had a chance to review this new version yet.  I'll try and have a look this week - for a mix of school inset days and thesis vivas to juggle.

dalonsoa:
No rush. I was mentioning just in reference to #541

vgro:
> > * Do we need to update the `data` object even when the model is static, coping the previous entry in the case of time series?
> 
> The `data` object currently stores the present state of the model variables as well as time series data for inputs, i.e. regional climate projections over time. The variables that represent the current state get output at every time step (I think we currently output all of them, but it could be a user defined subset). There's nothing within the `data` object that records the previous state of the model itself. I feel like this means the `data` object doesn't need to be updated for static models as the present state is by definition unchanging.
> 
> I guess we might want to change how the `data` object is used so that previous values are stored (for at least some variables) at some point in the future. But I feel like this would be a fairly big refactor which would also inevitably change some of the static model logic so probably isn't worth hedging against

I think it should be easy to store time series of variables in the data object as long as they have a time index dimension , like the climate input data, and this time index is selected in the model when static

dalonsoa:
> I think it should be easy to store time series of variables in the data object as long as they have a time index dimension , like the climate input data, and this time index is selected in the model when static

Ok, but that is not how the data object is used at the moment, right? I think that if that is to be implemented in the future, then this static model thing will need to be edited, but implementing that functionality now when it is not needed seems unnecessary.

davidorme:
@dalonsoa Sorry for the really slow response on this one. I agree that this is the way to go - it is a pretty clean path and has the added benefit of trapping superfluous variables in the initial data. I don't think we currently do that and that is important.

I think we could probably avoid the inner `_update` method if we rely on users including `super.update(...)`. That feels like a simpler namespace but I guess does rely on that call being included, so is more brittle. Might just have talked myself out of it!

davidorme:
I've maybe got a mild preference for not using 'private' style class method names here for the things that model developers would have to provide. 

Actually - no - I'm wrong. I think my argument was going to be that it feels more obscure, but the private name is used to indicate the important information that these functions should _never_ be called directly as it would strip the execution of the context of the static method flags.

dalonsoa:
It seems you're having a nice discussion with yourself :) .

Yeah, I think asking people to override `_whatever` instead of `whatever`, is not a bad practice and give us the freedom to introduce any actions that we want in the middle. It also saves the developer from having to call `super()...`, which often is unclear if it should be call at the beginning of the overriding method, at the end or not called at all. We will do that for them.

davidorme:
> It seems you're having a nice discussion with yourself :) .
> 
> Yeah, I think asking people to override `_whatever` instead of `whatever`, is not a bad practice and give us the freedom to introduce any actions that we want in the middle. It also saves the developer from having to call `super()...`, which often is unclear if it should be call at the beginning of the overriding method, at the end or not called at all. We will do that for them.

Yeah - two successive comments where I was going to suggest a change for superficially good reasons and then wanted to record why they are only _superficially_ good reasons.

alexdewar:
@dalonsoa Sorry it's taken me so long to get to this. I'm going through now

virtual_ecosystem/core/base_model.py
```diff
@@ -240,12 +241,110 @@ def __init__(
...
 
         # Check the required init variables
         self.check_init_data()
         # Check the configured update interval is within model bounds
         self._check_update_speed()
 
+        if not self._bypass_setup_due_to_static_configuration():
+            self._setup(**kwargs)
+
+    def _bypass_setup_due_to_static_configuration(self) -> bool:
```
alexdewar, line 259: I'm a bit confused as to why setup needs to be bypassed (I'm sure it's something I'm missing). Why can't static models be set up in the same way as dynamic ones? Presumably the setup of static models happens before `__init__` in some way?

virtual_ecosystem/core/base_model.py
```diff
@@ -254,10 +353,28 @@ def setup(self) -> None:
...
+
+        Args:
+            time_index: The index representing the current time step in the data object.
+            **kwargs: Further arguments to the update method.
+        """
+        if self._static:
+            if not self._run_update:
+                return
+            else:
+                self._run_update = False
```
alexdewar: This feels like a bit of a hack, but I'm not sure of a cleaner way to do things :shrug:.

Is there a way we could combine the initialisation and initial update steps for static models? Do we do anything with models while they're in an initialised-but-not-yet-updated state? Just a thought.

alexdewar (approved):
I think I've lost track of the inner workings of the code a bit by now, but this seems broadly sensible! I've made a couple of comments about the architecture, but they're largely borne of ignorance :stuck_out_tongue:

virtual_ecosystem/core/base_model.py
dalonsoa: Setup and init are in practice the same thing and indeed #541 is mean to get rid of the setup. But, whatever we call it, the thing is that static models do not need to run through the setup - indeed, you do not want to - since that step calculates some initial variables to be used by other models. Static models have those initial values defined upfront.

virtual_ecosystem/core/base_model.py
dalonsoa: We initialise the other models, which might provide variables that the update methods require.

Why is hacky? I thought I was being clever 😆 . Jokes aside, I don't see another way of running the inner `_update` just once (or none) for the static models. Amy suggestion is most welcome.

dalonsoa:
Ok, I'm going to crack this today, so I'll need to update a few models here and there, probably.

dalonsoa:
Ok, I've implemented the main changes and now I'm navigating through the update of the tests, which is a very painful experience. If you could have a look at the non-test files, to make sure I'm not doing anything stupid, that would be great.

virtual_ecosystem/core/base_model.py
davidorme: I think the only issue I have here is the name `_run_update`, which could be a little more explicit? `_run_initial_static_update` or something?

davidorme:
I've only really looked at the plants code but the structure of those changes looks as expected.

dalonsoa:
@davidorme @vgro @jacobcook1995 @TaranRallings @alexdewar , this is complete and ready for a full review.

dalonsoa:
Well, docs are failing, but I've no idea why. Let's see...

davidorme:
The ve_run example page in the docs is failing. I've grabbed the error from the action artefacts and it says: `Configuration failed for models: soil,litter`

dalonsoa:
Yeah, I suspected as much. Both two models - by default static - are being initialised with some variables that they are supposed to initialise - either in setup or in update - present in the Data object. As the actual error is lost, it is difficult to figure out what are the offending variables, but I'll see what I can find.

dalonsoa:
OK, I got the errors:

For "soil":

`Non-static model soil requires none of the variables in vars_populated_by_first_update or vars_updated to be present in the data object. Present variables: ['soil_c_pool_microbe', 'soil_c_pool_necromass', 'soil_c_pool_lmwc', 'soil_enzyme_maom', 'soil_enzyme_pom', 'soil_c_pool_pom', 'soil_c_pool_maom']`

For "litter:

`Non-static model litter requires none of the variables in vars_populated_by_first_update or vars_updated to be present in the data object. Present variables: ['c_p_ratio_below_structural', 'c_n_ratio_above_structural', 'litter_pool_above_structural', 'litter_pool_above_metabolic', 'c_p_ratio_above_structural', 'litter_pool_below_structural', 'lignin_below_structural', 'lignin_above_structural', 'c_p_ratio_below_metabolic', 'c_p_ratio_above_metabolic', 'c_n_ratio_below_metabolic', 'c_n_ratio_below_structural', 'litter_pool_below_metabolic', 'c_n_ratio_woody', 'c_p_ratio_woody', 'litter_pool_woody', 'c_n_ratio_above_metabolic', 'lignin_woody']`

So, the solution would be to get rid of those from the data object in the example, but chances are that will break a lot of things.

davidorme:
Or @jacobcook1995 is this simply that you've now implemented the calculation of things we had to feed in by hand before, and we just need to update the data? That seems quite likely - there's been a lot of movement on those models.

jacobcook1995:
This error seems to be related to the content of `vars_updated` (rather than `vars_populated_by_first_update`). They are all things that the `soil` and `litter` models track but no other model uses. I'm not sure if I have misunderstood `vars_updated` but I assumed it was about the output from the models rather than the inputs to the models?

jacobcook1995:
> This error seems to be related to the content of `vars_updated` (rather than `vars_populated_by_first_update`). They are all things that the `soil` and `litter` models track but no other model uses. I'm not sure if I have misunderstood `vars_updated` but I assumed it was about the output from the models rather than the inputs to the models?

Actually this isn't right sorry, the animal model needs to know about the size of the litter pools so there is a dependance there

jacobcook1995:
All of those variables do need to stay in the example data though as they are required for model initialisation. Maybe it's that they are all in `vars_required_for_init`, `vars_required_for_update` and `vars_updated` which might not be the correct approach

virtual_ecosystem/core/base_model.py
```diff
@@ -240,24 +241,142 @@ def __init__(
...
+        required = set(self.vars_populated_by_first_update + self.vars_updated)
+        present = [var for var in required if var in self.data]
+        found = len(present)
+        expected = len(required)
+
+        if not self._static and present:
+            raise ConfigurationError(
+                f"Non-static model {self.model_name} requires none of the variables in "
+                f"vars_populated_by_first_update or vars_updated to be present in the "
+                f"data object. Present variables: {present}"
```
jacobcook1995: The message here seems different from the way the error is described in `Raises:` section of the docstring (I might be missing something). This message suggests the error is because none of the variables in "vars_populated_by_first_update or vars_updated" are required to be in the data object whereas the docstring says that non-static models should only error if "some of the variables in vars_populated_by_init are not present in the data object"

jacobcook1995 (commented):
In general looks good to me, just had a query about the error message that was causing problems above

virtual_ecosystem/core/base_model.py
dalonsoa: Yep, the docstring is wrong. The result of copy/paste. Sorry.

jacobcook1995: Ahh that makes more sense! So is it the case that my models (litter and soil) shouldn't add things to `vars_updated` that are also included in `vars_required_for_init`?

dalonsoa: No. In `vars_updated` you only need to include things that are updated by the model `update` method. If the model is not static (which is the default), then these variables should NOT be in the data object when setting up the model - i.e. they should not be provided as input to a simulation - which is what this error is complaining about.

There are two separate checks, one in relation to the `vars_required_for_init` a few lines above and another one related to the `vars_populated_by_first_update` and `vars_updated` (this check). Both are run when running `__init__`.

jacobcook1995: Ahh right! So for example, if `"soil_c_pool_microbe"` is needed for init (to set the initial condition) and then gets continually updated by update method. Would it need to be included in both `vars_required_for_init` and `vars_updated` (which is what I'm currently doing), or just be included in `vars_required_for_init`?

dalonsoa: In that case, it should appear in both groups, as it fits both conditions: it is required for init and, as well, it is updated every iteration.

dalonsoa: Now, this flags a problem: you need the variable to be in the data object, so the model can be initialised, but at the same time you do not want it here otherwise it will cause an error - as it is happening. 

I guess the solution might be to include in this check only the variables initialised by the first update and not those vars_updated, since these might be required to initialise the model. Or exclude the variables that are both in required by init and vars_updated.

jacobcook1995: I guess either approach would work to fix the problem (don't have a strong opinion on which is preferable)

dalonsoa: And now all works!

@jacobcook1995 , thanks for flagging this - it was actually the solution to the docs problem as well.

docs/source/using_the_ve/configuration/config.md
```diff
@@ -191,3 +191,22 @@ formatted the configuration process will critically fail.
...
+
+- For the initialisation, if the variables that are populated are:
+  - **All present**: The model initialisation process is bypassed.
+  - **None present**: The model initialisation process runs normally.
+- For the update, if the variables that are populated are:
+  - **All present**: Then the update process is bypassed.
+  - **None present**: The update *is run just once*, keeping the same values for those
+  variables throughout the simulation.
+
+Providing some but not all of the variables case will result in an error in all cases.
```
alexdewar: Not sure what this sentence means...?

virtual_ecosystem/core/base_model.py
```diff
@@ -240,24 +241,144 @@ def __init__(
...
+
+        In particular, it checks that the appropriate variables populated by init are
+        present or not in the data object. Based on this, an exception is raised is
+        there is a problem or a decission is made on whether or not bypass the setup.
+
+        Raises:
+            ConfigurationError: If the model is static and some but not all the
+            variables or None in vars_populated_by_init are present in the data object
+            or if the model is not static and some of the variables in
+            vars_populated_by_init are not present in the data object.
```
alexdewar: I assume sphinx will parse this correctly, but I'm just flagging it in case the text after the line with `ConfigurationError:` on it needs indenting a bit more. Do you happen to know @davidorme?

virtual_ecosystem/core/base_model.py
```diff
@@ -240,24 +241,144 @@ def __init__(
...
+        """
+        present = [var for var in self.vars_populated_by_init if var in self.data]
+        found = len(present)
+        expected = len(self.vars_populated_by_init)
+
+        if not self._static and present:
+            raise ConfigurationError(
+                f"Non-static model {self.model_name} requires none of the variables in "
+                f"vars_populated_by_init to be present in the data object. "
+                f"Present variables: {present}"
```
alexdewar:
```suggestion
                f"Non-static model {self.model_name} requires none of the variables in "
                "vars_populated_by_init to be present in the data object. "
                f"Present variables: {', '.join(present)}"
```

virtual_ecosystem/core/base_model.py
```diff
@@ -240,24 +241,144 @@ def __init__(
...
+            )
+
+        elif self._static:
+            if 0 < found < expected:
+                raise ConfigurationError(
+                    f"Static model {self.model_name} requires to either all variables "
+                    "in vars_populated_by_init to be present in the data object or "
+                    f"all to be absent. {found} out of {expected} found: {present}."
```
alexdewar:
```suggestion
                    f"all to be absent. {found} out of {expected} found: {', '.join(present)}."
```

virtual_ecosystem/core/base_model.py
```diff
@@ -240,24 +241,144 @@ def __init__(
...
+        )
+        present = [var for var in required if var in self.data]
+        found = len(present)
+        expected = len(required)
+
+        if not self._static and present:
+            raise ConfigurationError(
+                f"Non-static model {self.model_name} requires none of the variables in "
+                f"vars_populated_by_first_update or vars_updated to be present in the "
+                f"data object. Present variables: {present}"
```
alexdewar: As above:

```suggestion
                f"Non-static model {self.model_name} requires none of the variables in "
                "vars_populated_by_first_update or vars_updated to be present in the "
                f"data object. Present variables: {', '.join(present)}"
```

virtual_ecosystem/core/base_model.py
```diff
@@ -240,24 +241,144 @@ def __init__(
...
+            if found == 0:
+                return True
+            elif found == expected:
+                return False
+            else:
+                raise ConfigurationError(
+                    f"Static model {self.model_name} requires to either all variables "
+                    "in vars_populated_by_first_update and vars_updated to be present "
+                    f"in the data object or all to be absent. {found} out of {expected}"
+                    f" found: {present}."
```
alexdewar:
```suggestion
                    f" found: {', '.join(present)}."
```

virtual_ecosystem/core/base_model.py
```diff
@@ -475,6 +596,13 @@ class ExampleModel(
                 defined
             TypeError: If model_name is not a string
         """
+        if cls.__init__ != BaseModel.__init__:
+            raise NotImplementedError("Model subclasses cannot override __init__.")
+
+        if cls.update != BaseModel.update:
+            raise NotImplementedError(
+                "Model subclasses cannot override the update method."
+            )
```
alexdewar, line 606: Because we're comparing identities, I wonder if this is more correct:

```suggestion
        if cls.__init__ is not BaseModel.__init__:
            raise NotImplementedError("Model subclasses cannot override __init__.")

        if cls.update is not BaseModel.update:
            raise NotImplementedError(
                "Model subclasses cannot override the update method."
            )
```

virtual_ecosystem/models/abiotic/abiotic_model.py
```diff
@@ -197,15 +183,26 @@ def setup(self) -> None:
...
         all time steps of the simulation. All variables are added directly to the
         self.data object.
+
+        Args:
+            model_constants: Set of constants for the abiotic model.
+            **kwargs: Further arguments to the setup method.
         """
 
+        self.model_constants: AbioticConsts = model_constants
+        """Set of constants for the abiotic model."""
```
alexdewar, line 202: Presumably this will never be changed after assignment? Just checking because we don't want the default arg above to be reused otherwise

tests/core/test_base_model.py
```diff
@@ -557,3 +557,256 @@ def from_config(
...
+        @classmethod
+        def from_config(
+            cls, data: Data, core_components: CoreComponents, config: Config
+        ) -> BaseModel:
+            return super().from_config(
+                data=data, core_components=core_components, config=config
+            )
+
+    for var in data_vars.keys():
+        fixture_data[var] = fixture_data["existing_var"].copy()
```
alexdewar, line 800: Just checking this doesn't need to be a `deepcopy`....?

tests/models/abiotic/test_abiotic_model.py
```diff
@@ -245,12 +260,17 @@ def test_generate_abiotic_model_bounds_error(
...
-    with raises:
-        _ = AbioticModel.from_config(
-            data=dummy_climate_data,
-            core_components=core_components,
-            config=config,
-        )
+    with (
+        patch_run_update("abiotic"),
+        patch_bypass_setup("abiotic") as mock_bypass_setup,
+    ):
```
alexdewar, line 266: TIL this syntax exists! I've been nesting all my `with` blocks till now, which looks awful :see_no_evil:

alexdewar (approved):
LGTM! I've put some small comments and suggestions.

docs/source/using_the_ve/configuration/config.md
dalonsoa: Neither do I 😆 . Will re-write it into something meaningful.

dalonsoa:
```suggestion
Providing some but not all of the variables will result in an error in all cases.
```

virtual_ecosystem/core/base_model.py
davidorme: I think it might need indenting. But I'm not sure!

virtual_ecosystem/core/base_model.py
dalonsoa: Not entirely sure, to be honest. The `!=` works - I had to change a few models because of that - so I'd leave it as it is.

tests/core/test_base_model.py
dalonsoa: It defaults to deepcopy: https://docs.xarray.dev/en/stable/generated/[xarray.DataArray.copy](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.copy.html).html

virtual_ecosystem/models/abiotic/abiotic_model.py
dalonsoa: All the constant classes are frozen dataclasses, so they cannot be changed.

dalonsoa:
I'm planning to work on #581 from next week, but for that I really need this PR agreed and merged. Is there any barrier for that to happen?

vgro (approved):
Apologies for the delayed response, I was pretty much out of office for the last 6 weeks. I do not see any reason why you shouldn't merge this branch, in particular if it stops you from proceeding.

dalonsoa:
Yes, please. We really need to merge this - or ditch it altogether. I've already had to fix merge conflicts several times because - obviously - you keep working and adding stuff incompatible with the changes here, and today I've wasted 3h trying to incorporate the latest changes to the animal model without success because in the data object there's some data that should not be there.

@TaranRallings , I get this error:

```
virtual_ecosystem.core.exceptions.ConfigurationError: Non-static model animal requires none of the variables in vars_populated_by_first_update or vars_updated to be present in the data object. Present variables: total_animal_respiration
```

Either `total_animal_respiration` should not be in the `dummy_animal_data` object used by the `prepared_animal_model_instance` fixture or `total_animal_respiration` should not be one of the `vars_populated_by_first_update` neither `vars_updated`. I cannot fix that myself. 

So, keeping aside this being a barrier or not, once @TaranRallings has fixed the above issue, please merge or close this PR without merging, but do not leave it lingering around for longer.

Issue #369, opened 2024-01-22
Simplifying the model creation setup.


davidorme:
This is a proposal for a refactor of the BaseModel to make a easier to update interface 
between the core and specific models.

We are in the process of adding new core content into models, centralising 
a bunch of stuff. I've done that for core_constants in a WIP pull request (#368), 
but it was a surprising pain in the arse, because:

* a `core_constants` instance first needs to be created in `vr_run` before models 
  are initialised.
* the argument then needs to be added to the `BaseModel.__init__` signature...
* ... and every model's `__init__` signature...
* ... and every model's `from_config` signature.

There is a bunch of core stuff on its way:

* core constants
* vertical layer structure
* axis definitions
* temporal definitions

It makes a lot of sense to add this all to model objects as attributes using `BaseModel.__init__()`, so
that the required `super().__init__()` call within a model populates all that stuff. The simplest thing
to avoid complex signatures, is simply to pass it a `Config` object and use that to
populate those base attributes

But - now we have a model `__init__` signature that contains `Config`, so what is the point
in the `from_config method`. I think we had a split between programmatic creation of a
model instance - where a detailed signature might help people understand model
differences - and then the runtime creation in simulation, where we just supply the
config.

I'm  not sure that keeping that model specific signature overrides the complexity needed
to add new attributes. We could just get rid of from_config approach and let all models 
be defined from simple core inputs. Models could then basically be created using:

```python
model <- ModelClass(data=data, config=config)
```
The `__init__` method - or possibly an encapsulated `configure` abstract method then
checks if the model can be configured from the inputs.

As a simplified sketch:

```python
from abc import ABC

def build_widely_used_thing(thing: str) -> str:
    """Build a widely used thing from core config contents."""
    return thing

class DemoConfig(dict):
    """Demo config."""
    pass

class DemoBase(ABC):
    """A base class that sets a bunch of widely used stuff directly from the config."""
    def __init__(self, config: DemoConfig) -> None:
        self.config = config
        self.widely_used_thing = build_widely_used_thing(config["core_thing"])

class DemoModel(DemoBase):
    """A daughter of the base class that adds a new model_thing attribute."""
    def __init__(self, config: DemoConfig):
        super().__init__(config)
        self.model_thing = self.config["model_thing"]
        print(self.__class__)

cfg = DemoConfig(
    {
        "layers": [1, 2, 3, 4],
        "core_thing": "core_thing",
        "model_thing": "model_thing",
[...]

PR #374, opened 2024-01-24
Defining the CoreComponents, LayerStructure and ModelTiming classes

davidorme:
# Description

Following the discussion in #369, we've aligned on a solution that tries to strike a balance between the original completely identical `BaseModel` subclass signatures and something that retains some of the model to model differences. See 
https://github.com/ImperialCollegeLondon/virtual_rainforest/issues/369#issuecomment-1908099087 for a sketch of this solution.

That requires a dataclass of core details that can be created once and then passed to all model initialisations. Models will then share a genuine single copy of those details - not individual identical copies - and model signatures can avoid the more opaque `Config` objects while still inheriting core components.

This PR implements that by creating `CoreComponents` which includes nested classes for `LayerStructure` and `ModelTiming`. Those borrow heavily from existing code for generating `layer_roles` and the `extract_timing_details` function. Once this is accepted, those functions will disappear and an instance of `CoreComponents` can be used in their place.

Issue #369 is not completed by this PR, but it would be good to agree on this structure and I can then adopt this structure in a second PR onto this, so we can break up the review a little bit!

Edited to add: That second WIP PR (#375) is now in progress - I wanted to see if implementing the changes here threw up anything unexpected. They did not. 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (commented):
Looks sensible to me, just one small query.

Also I can't work out why `codecov` is claiming that your changes have 0% test coverage, despite the fact you've written quite a few unit tests?

virtual_rainforest/core/core_components.py
```diff
@@ -0,0 +1,278 @@
...
+    reconciled_run_length: np.timedelta64 = field(init=False)
+    """The difference between start and calculated end time."""
+    run_length: np.timedelta64 = field(init=False)
+    """The configured run length."""
+    run_length_quantity: Quantity = field(init=False)
+    """The configured run length as a pint Quantity."""
+    update_interval: np.timedelta64 = field(init=False)
+    """The configured update interval."""
+    update_interval_quantity: Quantity = field(init=False)
+    """The configured update interval as a pint Quantity."""
```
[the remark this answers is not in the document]
davidorme: Yeah - I wasn't sure how all of those bits got used, so I've been inclusive and they will get winnowed down as this gets worked through. Also, this will presumably change when we get around to thinking about how to move away from `pint` and a month being 30.75 days and towards something calendar aware.

virtual_rainforest/core/core_components.py
```diff
@@ -0,0 +1,278 @@
...
+            and not (isinstance(canopy_layers, float) and canopy_layers.is_integer())
+            and canopy_layers < 1
+        ):
+            to_raise = ConfigurationError(
+                "The number of canopy layers is not a positive integer."
+            )
+            LOGGER.error(to_raise)
+            raise to_raise
+
+        self.canopy_layers = canopy_layers
```
dalonsoa, line 228: I would make this a separate function, for clarity of the code and easy testing:

```python
def _validate_canopy_layers(canopy_layers) -> int:
    ...

self.canopy_layers = _validate_canopy_layers(lyr_config["canopy_layers"])
```

virtual_rainforest/core/core_components.py
```diff
@@ -0,0 +1,278 @@
...
+
+        np_soil_layer = np.array(soil_layers)
+        if not (np.all(np_soil_layer < 0) and np.all(np.diff(np_soil_layer) < 0)):
+            to_raise = ConfigurationError(
+                "Soil layer depths must be strictly decreasing and negative."
+            )
+            LOGGER.error(to_raise)
+            raise to_raise
+
+        self.soil_layers = soil_layers
```
dalonsoa, line 253: Likewise, make this a separate function:

```python
def _validate_soil_layers(soil_layers) -> list[int | float]:
    ...

self.soil_layers = _validate_soil_layers(lyr_config["soil_layers"])
```

virtual_rainforest/core/core_components.py
```diff
@@ -0,0 +1,278 @@
...
+            ("subcanopy_layer_height", lyr_config["subcanopy_layer_height"]),
+        ):
+            if not isinstance(value, (float, int)) or value < 0:
+                to_raise = ConfigurationError(
+                    f"The {attr} value must be a positive float."
+                )
+                LOGGER.error(to_raise)
+                raise to_raise
+            else:
+                setattr(self, attr, value)
```
dalonsoa, line 268: And the same here:

```python
def _validate_positive(value) -> int | float:
    ...

setattr(self, attr, _validate_positive(value))

**Review by dalonsoa — CHANGES_REQUESTED** [review]
Looks good, but a couple of things can be made more modular. 

Otherwise, I think this will make the rest of the code much easier!

File: virtual_rainforest/core/core_components.py (hunk 4 above)
Comment (davidorme) [reply virtual_rainforest/core/core_components.py:268]: Things like this might be a useful wider `util` - we should probably do a code base scan at some point looking for duplicated patterns like this.

Comment (davidorme) [reply virtual_rainforest/core/core_components.py:268]: @dalonsoa Turns out `np.infty` and `np.nan` both sail straight through `isinstance(value, (float, int))`, so we now have `_validate_positive_finite_numeric` 😄

**Review by vgro — APPROVED** [review]
LGTM. My only concern is that we maybe overdo it with the wrapping and lose transparency.

File: virtual_rainforest/core/core_components.py [hunk 5]
```diff
@@ -0,0 +1,294 @@
...
+        """Populate the ``ModelTiming`` instance.
+
+        This method populates the ``ModelTiming`` attributes from the provided
+        :class:`~virtual_rainforest.core.config.Config` instance.
+
+        Args:
+            config: A Config instance.
+        """
+
+        timing = config["core"]["timing"]
```
dalonsoa, line 93: By the time this is called, the config has already validated to ensure there's a `core` and a `timing`, right? I think so, but just cross checking.

dalonsoa (approved):
Looks good!

TaranRallings (approved):

davidorme:
I've merged #375 into this branch, ready to merge the whole thing down into `develop`.

PR #385, opened 2024-02-06
Update ci.yml to use CODECOV_TOKEN

dalonsoa:
# Description

It seems you need a token now to upload coverage reports to Codecov, so here it is. 

Fixes # (issue)

## Type of change

## Key checklist
N/A

## Further checks
N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,9 @@ jobs:
 
     - name: Upload coverage to Codecov
       if: (matrix.os == 'ubuntu-latest' && matrix.python-version == '3.10')
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
+      env:
+        CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
 
   docs_build:
     needs: qa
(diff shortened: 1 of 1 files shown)


davidorme:
Does that differ from this commit to #384 : https://github.com/ImperialCollegeLondon/virtual_rainforest/pull/384/commits/1ad61b031392a6bcae827b21497b56976db01264

Also, it looks like the `codecov` upload doesn't complete, even though the action doesn't fail:
https://github.com/ImperialCollegeLondon/virtual_rainforest/actions/runs/7801415615/job/21276472911?pr=385#step:7:60

Where is that `secrets.CODECOV_TOKEN` value actually set?

davidorme:
> Anyway, yes, the main difference is that I've added the secret to the repo. I'm not sure why you need me, though. Any repo admin should be able to do it.

Honestly, it wasn't clear how to generate the token and whether that is per repo or per user - in which case it would be an `ImperialCollegeLondon` wide token?

dalonsoa:
@davidorme , sorted! It seems it did not like using environment variables, but the token needed to be provided directly. I've also added a couple more flags to the coverage upload. 

Feel free to ignore this PR and implement these changes in #384 directly

davidorme:
Fantastic! Just for the record,  how did you generate the token to put in the secrets settings?

dalonsoa:
In the settings tab in Codecov. Maybe that's where I'm needed... 

Screenshot from 2024-02-07 08-13-42

davidorme:
Exactly where you're needed, I think 😄 I have a feeling that secrets can be set at the Org level, so that this wouldn't have to be enabled on a per repo basis.

davidorme:
Handled along with other changes over in #384

PR #388, opened 2024-02-15
Add variables setup

dalonsoa:
# Description

Supersedes #371 

This PR adds a more full featured version of the variables infrastructure. It adds two registries, one containing all the known variables based on the modules that are being used in the simulation and another one for runtime variables that are actually being used by the requested models.

Each variable can be initialised by one single model and can be updated (with a warning) and used by more than one model. Who initialises, updates and use each variable is added to the relevant attributes of each variable. 

The availability of the axis each variable requires is also validated.

### What is known to be wrong or missing

At the moment, we use the models `required_init_vars` to decide which variable is initialised by each model, but that is not what `required_init_vars` represents, but rather what variable needs to be in the `data` object to be able to initialise the model. So, in reality, they are more like `vars_used`, a new attribute that needs to be populated for each model. Likewise, the `vars_updated` might actually be more closely related to what the model is initialising, although maybe not in all cases.

Keeping this confusion aside, all models (including the core ones) will need to define their own `variable.py` module with the relevant variables. An example for one variable used by the `plants` model is included to show how that needs to be done. To implement that, it might be easier if each of you create such a variable module for each of your models and open a PR against this one. 

Needless to say, tests and docs need to be implemented. That will be done once we are happy with the implementation. Until all the above is sorted, test will fail, naturally.

Fixes # (issue) - partly addresses #371 

## Type of change

- [x] Breaking change

## Key checklist

## Further checks


davidorme (commented):
Definitely looks like what we're after. Some comments below. On this part of your PR message:

> At the moment, we use the models required_init_vars to decide which variable is initialised by each model, but that is not what required_init_vars represents, but rather what variable needs to be in the data object to be able to initialise the model. So, in reality, they are more like vars_used, a new attribute that needs to be populated for each model. Likewise, the vars_updated might actually be more closely related to what the model is initialising, although maybe not in all cases.

Assuming we're only dealing with the `init` and `update` methods:

* `required_init_vars` are variables that must be either provided through the run configuration or created by a previously initialised model. (I am _vaguely_ wondering here whether there's mileage in having a `Variable.initialise(self, data: Data)` method that allows a variable initialisation to be importable across modules. That would save code duplication).
* We probably need `required_update_vars` - variables that are required for the `update` method.
* We have `vars_updated` but need `vars_initialised`.
* require

virtual_rainforest/core/config.py
```diff
@@ -17,6 +17,7 @@
 import tomli_w
 from jsonschema import FormatChecker
 
+import virtual_rainforest.core.variables as variables
 from virtual_rainforest.core.exceptions import ConfigurationError
```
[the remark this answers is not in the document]
dalonsoa: It is not identical. The method I used does not import the contents of the module until `setup_variables` is used deep in the code. This avoids circular import errors since by the time it is necessary to really load the contents of `variable` all the dependencies are already loaded. 

And it keep the namespace clean 😃

virtual_rainforest/core/variables.py
```diff
@@ -0,0 +1,190 @@
...
+        ValueError: If a variable is already in the known variables registry.
+    """
+    try:
+        variables_submodule = import_module(f"{module_name}.variables")
+    except ModuleNotFoundError:
+        LOGGER.info("No variables registered for %s.", module_name)
+        return
+
+    list(getmembers(variables_submodule))
+
```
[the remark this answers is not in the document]
dalonsoa: Yeah, I could not figure out a more elegant way of loading the contents of a module so the variables are registered. Any suggestion is most welcomed.

virtual_rainforest/core/variables.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+
+        KNOWN_VARIABLES[self.name] = self
+
+
+RUN_VARIABLES_REGISTRY: dict[str, Variable] = {}
+"""The global registry of variables used in a run."""
+
+KNOWN_VARIABLES: dict[str, Variable] = {}
+"""The global known variable registry."""
```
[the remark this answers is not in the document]
dalonsoa: Ok, I get it. It should have contained all possible variables, but clearly I put the registration in the wrong place. 

I will look into this.

virtual_rainforest/core/variables.py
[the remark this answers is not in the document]
dalonsoa: This has now been ditched.

dalonsoa:
Ok @davidorme , I've done a few changes following your comments:

- **ALL** existing variables within the `models` folder are now registered in the `KNOWN_VARIABLES` registry. This is populated before any configuration takes place.
- A function that outputs this registry to a file has also been included. Where/when to run that is up to you. I'm not sure if you want to do it manually from time to time or automatically when `docs/source/conf.py` is run.
- A function has been added that discovers ALL models and pulls from them the different usages of the variables. This **MUST NOT** be run at runtime since it defeats the purpose of having a way of identifying unfeasible configurations. At the moment is run only when outputting the registry.

There's still the issue of identifying the different types of vars (linked to #349), but it will be good to have your feedback on this.

dalonsoa:
Ok, I've added a few things. Now the models need to define 4 groups of variables:
- Required to be initialised
- The ones they initialise
- Required to be updated
- The ones they update

There is no distinction on where the variables required for initialisation are coming from.

These are all pull into the information for each variable, populating the relevant fields. Additionally, the variables available in the data object just after loading the data are also pulled into the registry before a verification of the consistency of the models based on the variables they require and provide is run.

There might be a few loose ends and I'm sure that some things will fail, but this is just a start.

davidorme:
OK - so we're now at the point where the `ve_run` tests are failing because the variable registries aren't being populated because the models haven't set up their `variables.py` files.

So, I think the implementation looks sound and  the next steps are:

- [ ] Update each model to create and populate `variables.py`
- [ ] Adapt the model variable attributes to the new system.
- [ ] Add `output_known_variables` to the Sphinx `conf.py` and get it rendered in the docs.
- [ ] Tests... could be a pain because the whole discovery system is so interlinked.

We can't merge until those are done, right?

virtual_ecosystem/core/variables.py
```diff
@@ -0,0 +1,290 @@
...
+    for mod in pkgutil.iter_modules(models.__path__):
+        if not mod.ispkg:
+            continue
+
+        try:
+            import_module(f"{models.__name__}.{mod.name}.variables")
+        except ImportError:
+            LOGGER.debug(
+                f"No 'variables' module found for model {models.__name__}.{mod.name}."
+            )
```
davidorme, line 78: Actually - looking at the way `ve_run`  works at the moment, it loops over _all_ defined models, whether they are used or not in a run. That:

* sort of doesn't matter, because all we're doing is populating a registry with more information than is needed,  
* means we can share the same function between `ve_run` and `output_known_variables`

But it does also mean that logs will contain messages about models that aren't being used. I think it might make more sense to include lines 74-78 in `core.registry.register_module`. That gets run in the next line in `ve_run` - as each model in the config files is added to the configuration, the `MODULE_REGISTRY` gets populated with `ModuleInfo` instances. We don't want to add information by module, but the same function could be used to populate the separate cross module `RUN_VARIABLES_REGISTRY`?

The configuration build is explicitly used to only register the modules included in the config, so that variable registry then only contains the supposed set of required variables, which is probably better for spotting bad model setups or missing data?

Somewhere after this - finish adding the `ModuleInfo` and then populate the variables?

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/89169fbb1e3e1e7cdb5d963c0bd141bc122f7df3/virtual_ecosystem/core/registry.py#L156

I think `_discover_all_variables_usage` and `register_all_variables` probably then collapse into `output_known_variables`, although we could keep it modular?

dalonsoa: I think there is value on keeping things modular and concise, doing only one thing. Having said that, I'll check the best way of reducing redundancy when registering stuff. You suggestion seems sensible.

dalonsoa:
That's correct - we cannot merge this until all the models are adapted to, at least, _populate_ the variables registry. _Using_ the registry can be done afterwards. I'd suggest each of you take care of writing the variables for their own models to distribute the effort.

I can tackle the last two. I might modify a bit the functions to make them more testable. As you said, dealing with the global state can be a pain.

virtual_ecosystem/core/variables.py
dalonsoa: I'm checking your suggestions in detail and I think that only the requested modules are registered, and not all of the available ones. The `requested modules` scanned in the `Config` and used to build the registry are only those included in the configuration files, in turn passed as `config_paths` or `config_strings`, which are provided by the user. Am I missing something?

davidorme: Yes, but `register_all_variables` scans _all_ of the `model.variable` modules. We already have duplicate parallel models - and its something we want to encourage - so those `variables.py` are going to contain duplicates and not all are needed for the specific run, so loading all modules seem unnecessary and possible a problem.

I think there's a bigger question  - I haven't quite got my head around _where_ we put variables. The `models.variables` module seems like a natural place, but where do we put variables always loaded from data? They don't really belong in any one model - some may only be used by one model but others will overlap, so how do model designers know which to include where. The same issue occurs with parallel models - we have repetition across files and multiple places to edit if we change definitions.

It seems more DRY and easy to manage if we just have one variable data source - an integrated home outside the models where all definitions get loaded from. That could even be the existing `data_variables.toml` file, but stripped of the model usage, which this system can then populate for the configured models.  That will break when we come to consider how users can plug in models - we probably want to be able to specify a model directory to add local models, but that could provide another `data_variables.toml` source file that _must_ not repeat.

davidorme: I'm not sure loading all known variables is a problem - it's just a lookup table for possible values - but the overlap, duplication and uncertainty about which module a variable should go in seems undesireable.

virtual_ecosystem/models/plants/variables.py
```diff
@@ -0,0 +1,12 @@
+"""Variables associated to the plans modudle."""
```
vgro, line 1:
```suggestion
"""Variables associated to the plans module."""
```

vgro:
Apologies for the very late reply.
As far as I can tell, this all makes sense and it does exactly what we asked for.
Could you please give an example for what adding a variable to `variable.py` should look like, then I will contribute the variables from the `abiotic`, `abiotc_simple`, and `hydrology` model.

vgro:
> Apologies for the very late reply. As far as I can tell, this all makes sense and it does exactly what we asked for. I will contribute the variables from the `abiotic`, `abiotic_simple`, and `hydrology` model soon.

@dalonsoa can I just double check which variables to include here, the ones required for `init` and `update` as well as the ones that the model initialises and updates?

virtual_ecosystem/core/variables.py
dalonsoa: I'm not sure about variables loaded form data, to be honest, but model variables should live in their own module within the appropriate folder. For parallel models, designers just need to check if all the variables they need have already been defined in another model of the same kind - or anywhere else - and define only those missing. That's were the documentation comes into play.

dalonsoa:
> > Apologies for the very late reply. As far as I can tell, this all makes sense and it does exactly what we asked for. I will contribute the variables from the `abiotic`, `abiotic_simple`, and `hydrology` model soon.
> 
> @dalonsoa can I just double check which variables to include here, the ones required for `init` and `update` as well as the ones that the model initialises and updates?

Only those the model initialises. Those that the model updates should have been defined somewhere else...

virtual_ecosystem/core/variables.py
davidorme: I think there's a case for defining variables in one location - a root variables module maybe? I haven't entirely thought this through but it seems wrong to have parallel models having to contain the same variable definitions. Each model might have slightly different sets of variables, but it doesn't seem DRY to have that duplication.

dalonsoa: As far as I can tell, there are two options:
- A root `variables.py` module that contains ALL variables and where new variables are added. Easy to maintain and to monitor for changes. Will make the code simpler and solves the issue of where to put the input data variables. I don't see a major drawback except, maybe, that you have variables for all models all mixed up together, which somewhat reduces modularity (a model is no longer a self-contained directory).
- A _per model_ `variables.py` module, BUT not duplicating variables, i.e. if the `adiabatic` model already defines variables A, B and C, `adiabatic_simple` does not need to defined them again - indeed, it would result in an error, as the code is now. This groups variables that are conceptually related together, but might make it harder to trace where they are actually defined, and doesn't solve the issue of input data variables. As variable definition is not duplicated, modularity is also not that great.

Duplicating the variable definition I don't think is a solution, so I would not go down that route even if it makes the code more modular and self'contained.

My preference would be the first option.

davidorme: I've just realised that my last message repeats stuff I've already said...

dalonsoa: Ok, I'll review the code based on this new interpretation this week (possibly tomorrow or on Thursday). I won't enter into the "providing external models" idea as that might require further thought, given that it might have more ramifications. 

@vgro don't add any variable, yet, as the implementation is still fluid.

davidorme: @dalonsoa The external models is definitely for another PR. I just wanted to flag that we will definitely need to do this, but I don't think the implementation will be super hard if we take the new route.

jacobcook1995 (approved):
This looks like a sensible approach to me!

virtual_ecosystem/core/variables.py
jacobcook1995: I also agree that option 1 seems the cleanest. And while it does make the system slightly less modular, I kind of doubt that the main limit on modularity will come from the code side anyway (I think it'll mainly come from the theory/science/concept definition side)

davidorme: @dalonsoa Just to add that we all chatted about this in a catch up today and we all agree that Option 1 looks good.

dalonsoa:
OK, so now the variables are defined all together in a `toml` file in the root of `virtual_ecosystem` and the `KNOWN_VARIABLES` registry is taken from there. The rest is the same. 

Following @davidorme list above, the next steps would be:

- [ ] Update each model definition with the appropriate variables (`required_init_vars`, `vars_updated`, `required_update_vars` and `vars_initialised`.
- [ ] Adapt the model variable attributes to use the new system.
- [ ] Add `output_known_variables` to the Sphinx `conf.py` and get it rendered in the docs.
- [ ] Tests... could be a pain because the whole discovery system is so interlinked.

dalonsoa:
OK, so a brief update on this:
- I've added the tests - some might need tweaking later on.
- Updated the list of variables currently needed by the existing models, but these will need your input/update to provide the right description, units and axis.
- Pull the known variables into a file and render it as a table. I think a tabular format makes sense considering the nature of the list, the the rendering is pretty bad as the table is too wide and one neds to scroll sidewise. Any suggestion on this would be most welcomed - as well as to avoid creating the intermediate CSV file. 

I must also mention that the instructions to build the documentation are not that robust (they don't work for me in Windows). I'd suggest updating them to:

```bash
poetry run sphinx-build docs/source docs/build
```

davidorme:
I've added this branch to the RTD build list so we can see how it renders:

https://virtual-ecosystem.readthedocs.io/en/variables_1/

davidorme:
Oh god, we really need to figure out how to simplify the auto expanding signatures: 
https://virtual-ecosystem.readthedocs.io/en/variables_1/api/core/variables.html#virtual_ecosystem.core.variables.KNOWN_VARIABLES

dalonsoa:
Wow! how does it know the contents of those dictionaries???

davidorme:
I think this is because `sphinx.ext.autodoc` is dynamic - it actively _has_ to import the `virtual_ecosystem` package. I've been idly looking at `autodoc2` because it supports MyST in docstrings and that uses a static approach. But I think there must be options.

We get the same kind of spam with the constants classes, so hopefully we can solve both with one setting.

dalonsoa:
I've implemented an alternative way of displaying the variables, just below the tabular data:

https://virtual-ecosystem.readthedocs.io/en/variables_1/virtual_ecosystem/variables.html#aerodynamic-resistance-surface

Ignore the code to do that - it is quick and dirty - and just check if this form seems more readable and useful.

davidorme:
I don't think we worry too much about the rendering now. The table is a pain because of the scroll bar, so I think that new layout is easier to search. It would be good to add sorting and filtering to allow people to see what variables come with which model, but that's a whole HTML and probably JS interface, but that's a new PR.

I think we could also add a PR to put a new entry point providing a tool that takes a list of models and spits out a data config TOML template with the required input variables for those models and path placeholders to be filled in.

dalonsoa:
Ok, so I'm going to tidy up the code and then make a list of specific things to do by all of you, contributing to this branch, so we have the models defined in such a way that the the variables registries are populated properly.

dalonsoa:
The documentation building is failing because of something happening in [this block](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/13ef5321346e72c6dfdd1071b76e8e8d386ebedc/docs/source/development/documentation/api_generation.md?plain=1#L70) (that I have not touched) but other than that, this is ready to be contributed to. 

Next steps:
- Branch off the branch of this PR, `variables 1`.
- @davidorme , @vgro , @jacobcook1995 , @TaranRallings : Review the [`data_variables.toml`](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/variables_1/virtual_ecosystem/data_variables.toml) file and complete the information related to the variables of your models. This include adding the right units, meaningful description, the type of the variable AND the name of the axis they use. If you spot a variable that is no longer relevant, delete it. 
- Then, review your models and update the contents of `required_init_vars`, `vars_updated`, `required_update_vars` and `vars_initialised`. They should all be just simple lists and the name of the variables, obviously, should match those of the `data_variables.toml`. 
- With those changes, create a PR against this branch. 

Once the above is done for all models, I'll strip some unnecessary validation code and, potentially, adapt tests. By that point, VE would not be _using_ the variables registry, yet, but having it would not break the existing functionality, so this would be ready for merging.

After that, is up to you to exploit that functionality.

vgro:
> The documentation building is failing because of something happening in [this block](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/13ef5321346e72c6dfdd1071b76e8e8d386ebedc/docs/source/development/documentation/api_generation.md?plain=1#L70) (that I have not touched) but other than that, this is ready to be contributed to.
> 
> Next steps:
> 
>     * Branch off the branch of this PR, `variables 1`.
> 
>     * @davidorme , @vgro , @jacobcook1995 , @TaranRallings : Review the [`data_variables.toml`](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/variables_1/virtual_ecosystem/data_variables.toml) file and complete the information related to the variables of your models. This include adding the right units, meaningful description, the type of the variable AND the name of the axis they use. If you spot a variable that is no longer relevant, delete it.
> 
>     * Then, review your models and update the contents of `required_init_vars`, `vars_updated`, `required_update_vars` and `vars_initialised`. They should all be just simple lists and the name of the variables, obviously, should match those of the `data_variables.toml`.
> 
>     * With those changes, create a PR against this branch.
> 
> 
> Once the above is done for all models, I'll strip some unnecessary validation code and, potentially, adapt tests. By that point, VE would not be _using_ the variables registry, yet, but having it would not break the existing functionality, so this would be ready for merging.
> 
> After that, is up to you to exploit that functionality.

Thanks for the instructions on next steps, just a quick question: by type, to you mean float, string, etc?

dalonsoa:
Yes. In principle, that is [one of the required parameters](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/11a19a68494d353e201eea8e2a95105165948eee/virtual_ecosystem/core/variables.py#L75) that we were thinking to include. It is not used for computation, so maybe it should be just human-readable type information to guide the user, and not an actual type. Or we can ditch it altogether.

vgro:
> Yes. In principle, that is [one of the required parameters](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/11a19a68494d353e201eea8e2a95105165948eee/virtual_ecosystem/core/variables.py#L75) that we were thinking to include. It is not used for computation, so maybe it should be just human-readable type information to guide the user, and not an actual type. Or we can ditch it altogether.

oh ok, I'm asking because I thought all variables in the data object were xr.DataArray

davidorme:
@vgro - Yup, they are but I think it is useful to specify the type of the array contents. I suspect the only time it will be actually important is if the variable should be an integer and not a float (string types are very rare?) and most variables can be generically numeric (i.e. float, but doesn't really matter if the inputs are stored as int?).

davidorme:
> @dalonsoa At the moment, this retains the original definition of `required_init_vars` with the format `(("var_name", ("axis_name"))`, but in the convo above I think the intention is to make this a tuple of strings like the other matching attributes?
> 
> I think that makes sense because the axis validation can be offloaded into the variable setup process rather than being repeated across models.
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/e612fd510698c875414c090d8146dd83b6d2c0ad/virtual_ecosystem/core/base_model.py#L152

OK - wait, that is a logical step but does break the compatibility of this PR with the current state of `develop`. We should probably restore the original `required_init_vars` setup and then create new PRs to adopt/shift that part of the variables system. Rather than trying to fix everything here.

dalonsoa:
> OK - wait, that is a logical step but does break the compatibility of this PR with the current state of `develop`, which is not the strategy you outlined above for adopting this PR.
> 
> Although the updated docstrings indicate the intention to reduce `required_init_vars` to a tuple of names and not the more complex structure, we should probably restore the original `required_init_vars` setup and then create new PRs to adopt/shift that part of the variables system. Rather than trying to fix everything here.

Implementing this PR will do things differently, but as this PR is not going to be merged until all models have been updated to setup the variables in this new way, offloading the validation to the variables system, there's no breaking of anything.

Restoring the previous `required_init_vars` will break the current state of this branch, implementing the variables and their validation outside of the models, so I really advise against it. As I said, nothing is to be broken in `develop` because by the time this is merged, the models would be adapted.

davidorme:
@dalonsoa OK - I was thinking that we would merge this PR in close to its present state (variables all described and registered but models still implement data validation) with then subsequent PRs to hook up the wider utility of the system. In that case, I'd have restored the `required_init_vars` format because it's a breaking change. 

But IIUC, your intention is to hook all that up once we've got the variables definitions sorted out and merged in here. That does make more sense - in my defence, I'm trying to fix my tendency to balloon PRs 😄

davidorme:
Just to note that I'm expecting those `pre-commit` fails - the incoming changes from `variables_1_vgro` will update the model structures to apply the new simpler list of strings in `BaseModel.required_init_vars`.

davidorme:
@dalonsoa We've just talked through the variable process and we think we definitely need to be able to have variables that only appear at `update`. These are typically internal state variables, that use the `Data` object as memory. It doesn't really make sense to force all variables to be present after the `__init__` stages complete. 

Can we allow an explicit `populated_by_update` variables attribute? It might be worth moving away from the `initialised_by` terminology to avoid confusion? Maybe something like:`populated_by_data`, `populated_by_init`, `required_by_init`, `populated_by_update` and `required_by_update`?

dalonsoa:
@davidorme , sounds good. I'll allocate time to work on this and leave it ready for merging next Tuesday 18th.

davidorme:
Another thing we might want is a very small  set of core variables (at the moment _one_)  that are initialised from the configuration before models initalise.

The motivating case is `layer_heights`, which is initialised by the Plants model at the moment. That makes some sense, since most of it is canopy height, but the actual initial state is _all_ from config and the core components, but it is used by all sorts of models as a central reference for vertical layer heights. 

So, we could introduce those core variables - possible as part of the initialisation process of `CoreComponents` - and maybe flag them as core somehow. I don't think this is part of the variables PR, just something that I'm thinking about right now and wanted to log.

vgro:
@dalonsoa I currently initialize a few variables in `setup`, but we agreed that we want to retire the `setup` from the base model and move all the functionality to `init`. Would it make sense for me (or you) to move everything that sits in `setup` to `init` first? I can list the variables that are meant to be in `populated_by_init` and `populated_by_update`.

dalonsoa:
Happy to do the change myself, and also in relation to the variables need moving.

vgro:
> Happy to do the change myself, and also in relation to the variables need moving.

That would be really nice. I am currently updating the lists and will commit to this branch for you to review

virtual_ecosystem/models/hydrology/hydrology_model.py
```diff
@@ -82,6 +81,37 @@ class HydrologyModel(
...
+        "wind_speed",
+        "leaf_area_index",
+        "layer_heights",
+        "soil_moisture",
+        "evapotranspiration",
+        "surface_runoff_accumulated",
+        "subsurface_flow_accumulated",
+    ),
+    populated_by_update_vars=(),
 ):
```
vgro, line 115: `air_temperature`, `relative_humididty` and `atmospheric_pressure` are here initilised in a very simple manner under the assumption that the hydrology runs before the abiotic model (which than initializes an actual profile for these variables). Is this still our assumed model order?

virtual_ecosystem/models/abiotic/abiotic_model.py
```diff
@@ -98,21 +98,28 @@ class AbioticModel(
...
         "soil_temperature",
         "vapour_pressure_ref",
         "vapour_pressure_deficit_ref",
         "air_temperature",
         "relative_humidity",
-        "vapour_pressure",
         "vapour_pressure_deficit",
         "atmospheric_pressure",
         "atmospheric_co2",
+        "canopy_absorption",  # DAVID This is assuming that abiotic runs before plants
```
vgro, line 112: @davidorme can you please comment on this?

davidorme: Yeah  - this should be a `required_init_vars` instead and `plants` will have it as `populated_by_init_vars`

dalonsoa:
I'm trying to make this work and I'm facing the following problem:

The `aerodynamic_resistance_surface` is not in the `daily_lists` dictionary, so things fail in [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/76680c884837cc7c4a7f27361542e180872a75e0/virtual_ecosystem/models/hydrology/hydrology_model.py#L692). Adding it is straight forward - I just made it one of the `vars_updated` when creating the model - but that does not populate the dictionary entry with any data, so the line above still fails. @vgro how should the `aerodynamic_resistance_surface` variable be updated in the `HydrologyModel.update` method?

This is making two tests fail, but otherwise, once that is sorted, this should be ready to go.

@vgro , @jacobcook1995 , I've added `aerodynamic_resistance_surface` and `soil_c_pool_necromass` to the variables registry. Please, check that the information there makes sense.

dalonsoa:
Ups... should those files under `jupyter_execute` be there? I'll git ignore the folder and remove them if they should not be there.

davidorme:
@dalonsoa I have some unpushed updates on this too - I think this was one of the issues. Will have a look now.

docs/source/variables.rst
```diff
@@ -0,0 +1,1663 @@
...
+name                 soil_c_pool_necromass
+description          Necrotic organic matter pool
+unit                 kg C m-3
+variable_type        float
+axis                 ['space']
+populated_by_init    []
+populated_by_update  []
+required_by_init     []
+updated_by           ['soil']
+required_by_update   []
```
jacobcook1995, line 1230: This should have the same properties as the other `soil_c_pool_*` so should `required_by_init` and `required_by_update` should both be 'soil'

vgro:
> I'm trying to make this work and I'm facing the following problem:
> 
> The `aerodynamic_resistance_surface` is not in the `daily_lists` dictionary, so things fail in [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/76680c884837cc7c4a7f27361542e180872a75e0/virtual_ecosystem/models/hydrology/hydrology_model.py#L692). Adding it is straight forward - I just made it one of the `vars_updated` when creating the model - but that does not populate the dictionary entry with any data, so the line above still fails. @vgro how should the `aerodynamic_resistance_surface` variable be updated in the `HydrologyModel.update` method?
> 
> This is making two tests fail, but otherwise, once that is sorted, this should be ready to go.
> 
> @vgro , @jacobcook1995 , I've added `aerodynamic_resistance_surface` and `soil_c_pool_necromass` to the variables registry. Please, check that the information there makes sense.

Thanks @dalonsoa . The `aerodynamic_resistance_surface` is not returned by the function that calculates it [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/76680c884837cc7c4a7f27361542e180872a75e0/virtual_ecosystem/models/hydrology/above_ground.py#L106)

could you add a line `output["aerodynamic_resistance_surface"] = aerodynamic_resistance` below, please?

The daily lists are created from `vars_updated` [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/76680c884837cc7c4a7f27361542e180872a75e0/virtual_ecosystem/models/hydrology/hydrology_model.py#L426),  so if you add the `aerodynamic_resistance_surface` to `vars_updated` it will hopefully work.

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -46,13 +46,24 @@ class SoilModel(
...
-        ("soil_c_pool_microbe", ("spatial",)),
-        ("soil_c_pool_pom", ("spatial",)),
-        ("pH", ("spatial",)),
-        ("bulk_density", ("spatial",)),
-        ("clay_fraction", ("spatial",)),
+        "soil_c_pool_maom",
+        "soil_c_pool_lmwc",
+        "soil_c_pool_microbe",
+        "soil_c_pool_pom",
+        "soil_enzyme_pom",
```
jacobcook1995, line 53: "soil_c_pool_necromass" should be here (I think I forgot to add it sorry)

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -46,13 +46,24 @@ class SoilModel(
...
+        "pH",
+        "bulk_density",
+        "clay_fraction",
+    ),
+    populated_by_init_vars=(),
+    required_update_vars=(
+        "soil_c_pool_maom",
+        "soil_c_pool_lmwc",
+        "soil_c_pool_microbe",
+        "soil_c_pool_pom",
```
jacobcook1995, line 64: "soil_c_pool_necromass" should also be here

jacobcook1995 (approved):
LGTM! I've just added feedback on how "soil_c_pool_necromass" should be specified (which I think I specified wrong when I added it)

docs/source/variables.rst
dalonsoa: This file is generated automatically when building the documentation - actually, we might want to git-ignore it to avoid confusion. The file to review is only the TOML one in the root directory.

virtual_ecosystem/data_variables.toml
```diff
@@ -1,550 +1,727 @@
...
-updated_by = "soil"
-used_by = ["soil"]
+variable_type = "float"
 
 [[variable]]
+axis = ["space"]
+description = "Necrotic organic matter pool"
+name = "soil_c_pool_necromass"
+unit = "kg C m-3"
+variable_type = "float"
```
jacobcook1995, line 573: This info is all correct!

virtual_ecosystem/data_variables.toml
```diff
@@ -1,550 +1,727 @@
...
-name = "stream_flow"
-description = "Estimated stream flow"
-unit = "mm per time step"
-initialised_by = "hydrology"
-updated_by = "hydrology"
-used_by = [ "hydrology"]
+axis = ["space"]
+description = "Aerodynamic resistance surface"
+name = "aerodynamic_resistance_surface"
+unit = "m2"
```
vgro, line 334:
```suggestion
unit = "kg m2 s–3"
```
The unit is Ohm or kg m2 s–3

dalonsoa:
I hate Sphinx. Everything seems to be working, but the build of the documentation appears to be failing. However, looking at the logs, there's nothing indicating a failure. might it be because of the warnings at the beginning, @davidorme ?

dalonsoa:
OK, I managed to find the error. When running [this notebook](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/variables_1/docs/source/using_the_ve/virtual_ecosystem_in_use.md), we get: `ValueError: Variable surface_runoff initialised by hydrology already in registry as initialised by data.`. This suggest that the example data used by the notebook contains data that was supposed to be initialised by the hydrology model, instead, and not provided as input data. 

So, @vgro , @davidorme , what do you suggest to fix this? I don't know the ramifications of just removing the `surface_runoff ` from the example data.

davidorme:
OK, so removing `surface_runoff` from the example data doesn't seem to be an issue, but that reveals more issues with that specific notebook. 

1. The first issue is the three variables in the `hydrology` model `populated_by_init_vars` that are also in the `abiotic` model. This is a known issue - the overlap is noted in comments - but the variables system can't run until this is resolved, right @dalonsoa ?
 
https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/fd0df0e55bf13bffe09bcd0bfffec785cd81c52f/virtual_ecosystem/models/hydrology/hydrology_model.py#L98-L101

2. If I comment those three lines out, it then runs but halts at another error:

```python
  File "/Users/dorme/Research/Virtual_Rainforest/virtual_ecosystem/virtual_ecosystem/core/variables.py", line 312, in _collect_updated_by_vars
----- stderr -----
    raise ValueError(
----- stderr -----
ValueError: Variable aerodynamic_resistance_surface required by hydrology is not initialised by any model.
```

@vgro - I think this is right. `aerodynamic_resistance_surface` is first populated by `HydrologyModel.update` but is never initialised. Not quite sure what is going on here.

davidorme:
It's really late in the day to be moaning about this, but I find I constantly stumble over the meanings of the model variable attributes. Does anyone else find this - if so could we switch up the names?

```text
    required_init_vars -> vars_required_for_init
    vars_updated -> vars_updated
    required_update_vars -> vars_required_for_update
    populated_by_init_vars -> vars_populated_by_init
    populated_by_update_vars -> vars_populated_by_first_update
```

davidorme:
OK -  this has basically been me smacking it with stick. I don't think I've done anything mad but @vgro and @TaranRallings please have a look at the changes to `hydrology` and `animal` model files.

1. Removed `surface_runoff` from initial input data.
2. Added a bunch of variables that are populated by the first update but were missing from `populated_by_update_vars`.
3. Correct `space` to `spatial` in variable axis definitions, added a placeholder `time` axis validator, moved some more outlandish hypothetical axes into comments in `data_variables.toml`.
4. Update the axis testing to expect the new `time` axis name.

I haven't got my head around why `test_ve_run` passed, yet basically running the same thing in a notebook failed.

vgro:
Good morning, you were busy last night!

- it is ok to remove the `surface_runoff` from the input data now, however, it will probably be needed once we introduce some sort of spinup. The idea is that the surface runoff needs to accumulate over a spinup period to read a steady state. I can add it back when we get there.
- the `aerodynamic_resistance_surface` is indeed initialized by the first update
- the abiotic variables that are populated in the hydrology [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/fd0df0e55bf13bffe09bcd0bfffec785cd81c52f/virtual_ecosystem/models/hydrology/hydrology_model.py#L98-L101) are somewhat optional; it depends on the order of models. If abiotic runs first that is fine, but I thought we wanted to run the hydrology before abiotic?

vgro:
This conversation definitely highlights the value of your contribution @dalonsoa  , thank you :-)

dalonsoa:
> It's really late in the day to be moaning about this, but I find I constantly stumble over the meanings of the model variable attributes. Does anyone else find this - if so could we switch up the names?
> 
> ```
>     required_init_vars -> vars_required_for_init
>     vars_updated -> vars_updated
>     required_update_vars -> vars_required_for_update
>     populated_by_init_vars -> vars_populated_by_init
>     populated_by_update_vars -> vars_populated_by_first_update
> ```

No problem changing the names - just tell me what you want, and I'll change it :)

> The first issue is the three variables in the hydrology model populated_by_init_vars that are also in the abiotic model. This is a known issue - the overlap is noted in comments - but the variables system can't run until this is resolved, right @dalonsoa ?

That's right. That's the point of the validation process done in the variables system, to avoid having two models initialising the same variables. It was mentioned in the comments that the adiabatic models were also initialising those, but there was no suggestion of a solution (at lease not one I can implement without knowing the science). The adiabatic models and the hydrology models are not parallel models (like `adiabatic` and `adiabatic_simple` that are mutually exclusive), so if they are to coexist, they need to have compatible variable initialisation processes.

vgro:
> > It's really late in the day to be moaning about this, but I find I constantly stumble over the meanings of the model variable attributes. Does anyone else find this - if so could we switch up the names?
> > ```
> >     required_init_vars -> vars_required_for_init
> >     vars_updated -> vars_updated
> >     required_update_vars -> vars_required_for_update
> >     populated_by_init_vars -> vars_populated_by_init
> >     populated_by_update_vars -> vars_populated_by_first_update
> > ```
> 
> No problem changing the names - just tell me what you want, and I'll change it :)
> 
> > The first issue is the three variables in the hydrology model populated_by_init_vars that are also in the abiotic model. This is a known issue - the overlap is noted in comments - but the variables system can't run until this is resolved, right @dalonsoa ?
> 
> That's right. That's the point of the validation process done in the variables system, to avoid having two models initialising the same variables. It was mentioned in the comments that the adiabatic models were also initialising those, but there was no suggestion of a solution (at lease not one I can implement without knowing the science). The adiabatic models and the hydrology models are not parallel models (like `adiabatic` and `adiabatic_simple` that are mutually exclusive), so if they are to coexist, they need to have compatible variable initialisation processes.

A solution would be to make sure that the `abiotic` model (whichever is selected) is initialized BEFORE the `hydrology`, then we can just delete these variables from the initialization of the `hydrology` model

dalonsoa:
Could we indicate in the config file that `hydrology` depends on `adiabatic`? At the moment, it is the other way around in here, for example:
https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/68129075e9bd03f77d612eb5b6b2447d49a59aef/virtual_ecosystem/example_data/config/ve_run.toml#L2-L4

vgro:
> Could we indicate in the config file that `hydrology` depends on `adiabatic`? At the moment, it is the other way around in here, for example:
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/68129075e9bd03f77d612eb5b6b2447d49a59aef/virtual_ecosystem/example_data/config/ve_run.toml#L2-L4

ah, this is what I was looking for. I _think_ we want

init in this order: `plants`, `abiotic_simple`, `hydrology`, others
update in this order: `hydrology`, `abiotic_simple`, `plants`, others

which should look like this:
```
[hydrology.depends] 
init = ['plants', 'abiotic_simple'] 

[abiotic_simple.depends]
init=['plants']
update=['hydrology']

[plants.depends]
update=['abiotic_simple', 'hydrology']
```

Also, I just noticed that the infamous `aerodynamic_resistance_surface` is a `required_update_var` for the `abiotic` model. So it would be best to update the `hydrology` first.

can you confirm this @davidorme ?can you confirm this @davidorme ?

dalonsoa:
Actually, we don't need to worry about the order. If we indicate correctly what variables are initialised by what model and when, then we can use the information stored in the variables registry to come up with the right sequence for the init and the update. We don't need to indicate that in the config file - and we can completely ditch that part. At least when it comes to the initialisation.

davidorme:
> Actually, we don't need to worry about the order. If we indicate correctly what variables are initialised by what model and when, then we can use the information stored in the variables registry to come up with the right sequence for the init and the update. We don't need to indicate that in the config file - and we can completely ditch that part. At least when it comes to the initialisation.

Yeah - this is one of the main advantages of the variables system, I think. It can automatically establish if there is a feasible variable setup sequence for any given suite of models and give meaningful errors about why no sequence can be established.

davidorme:
> At least when it comes to the initialisation.

I think the same should be true for the update? Going into `update`, we know which variables have been loaded from data and populated by the model `__init__` methods. So for each model, we can know which variables are missing and then which models populate that variable when they update. That ought to give us the update order.

davidorme:
> > It's really late in the day to be moaning about this, but I find I constantly stumble over the meanings of the model variable attributes. Does anyone else find this - if so could we switch up the names?
> > ```
> >     required_init_vars -> vars_required_for_init
> >     vars_updated -> vars_updated
> >     required_update_vars -> vars_required_for_update
> >     populated_by_init_vars -> vars_populated_by_init
> >     populated_by_update_vars -> vars_populated_by_first_update
> > ```
> 
> No problem changing the names - just tell me what you want, and I'll change it :)

@jacobcook1995 gave this the thumbs up, so unless anyone else is wildly bothered, I say we go for those switches. It probably makes sense to update the `_check_...` functions in `core.variables` to match.

davidorme:
> * it is ok to remove the `surface_runoff` from the input data now, however, it will probably be needed once we introduce some sort of spinup. The idea is that the surface runoff needs to accumulate over a spinup period to reach a steady state. I can add it back when we get there.

I've a nasty feeling that we'll need to add to variables system then to track spinup variables and the order of execution of `spinup` methods... Although if spinup is limited to a very few variables and models, we could just provide external code to estimate the initial values to put in, rather than having to embed it all in the VE code.

dalonsoa:
> > > It's really late in the day to be moaning about this, but I find I constantly stumble over the meanings of the model variable attributes. Does anyone else find this - if so could we switch up the names?
> > > ```
> > >     required_init_vars -> vars_required_for_init
> > >     vars_updated -> vars_updated
> > >     required_update_vars -> vars_required_for_update
> > >     populated_by_init_vars -> vars_populated_by_init
> > >     populated_by_update_vars -> vars_populated_by_first_update
> > > ```
> > 
> > 
> > No problem changing the names - just tell me what you want, and I'll change it :)
> 
> @jacobcook1995 gave this the thumbs up, so unless anyone else is wildly bothered, I say we go for those switches. It probably makes sense to update the `_check_...` functions in `core.variables` to match.

I'm going to make this change in this PR, right now. Let's see if I don't break anything else...

davidorme:
> Do we want to do this in this issue? 

No. Nope. Nix. **Hard** pass.

> It's becoming pretty large and I feel it would be best to consolidate what we have before starting to use the new functionality to improve the workflow.

☝️ 👆 This.

davidorme (approved):
Looks good to me! Time to bank?

dalonsoa:
Yes, please!!! 🙏

vgro:
yes, go for it 👍

vgro:
oh you already did, perfect :-) Happy weekend!

Issue #370, opened 2024-01-23
Variables registry


dalonsoa:
All models in the Virtual Rainforest interact through variables stored in the data object. Dependencies of one model on another model essentially boil down to needing certain variables to exist in the data object, i.e. initialized by the model in question. We’ve slightly formalised this by introducing a configuration system to order init and update methods across modules, but in the main it remains quite loosely defined. 

What would be useful would be a system that registers all variables, along with information about which module initialises them, which modules use them, and which modules updates them. This could then be used to constrain what sets of models can be run together and provide information of what variables need to be added to get a set of models to play nicely with each other. This becomes even more relevant when we have different versions of the same components, for example a simple and a mechanistic version of the abiotic model. This should also help automate the  documentation of variables used in the models, which currently documented manually. 

As a sketch: 

- Variable descriptions are created from a new ABC that defines a unique variable name along with properties like dimensionality, a description and the expected units. 
- Individual models might then specify `required_init_vars`, `updated_vars` and `required_update_vars`. A particular set of models then defines which variables need to be provided in the input data, and potentially defines solutions for the model init  and update order. 
- Creating a new Variable instance automatically adds that variable to a registry, which checks for pre-existing usage of a name. 
- It might be better to have a separate variables module, so that models then import the variable definitions that they need. That makes it easier to avoid clashes.  However, in the long-term, we want users to be able to add their own models and so these probably should be able to declare their own variables. Maybe we just need to define `Variable.__eq__` so that we can check if a variable declaration is identical?

davidorme:
Closed by #388

PR #393, opened 2024-02-26
Fixing broken tests

### LINKED ISSUE #390: Test failures in Ubuntu with Python 3.10
> Having said that, and totally underelated, the tests with Ubuntu and Python 3.10 are a bit flaky, right? They fail for no apparent reason and pass if re-run. Any idea of why that might be?

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/virtual_rainforest/issues/384#issuecomment-1932064501_
        
They sure are flaky - initially this looked like some kind of transient issue but over on #384 it is failing consistently. It isn't clear what is going wrong - partly because the error reporting within `vr_run` is too monolithic and partly because we can't access the log from the Github Action runner at the moment.

davidorme:
# Description

Hopefully this fixes #390.

I think the issue is that the `hydrology` model needs to initialise _after_ `plants`, because that populates the `leaf_area_index` data. That isn't explicitly set in the configuration, so then I think we are getting platform specific variation in the way `TopologicalSorter` is returning 'tied' models, such that `plants` and `hydrology` can vary in their execution order.

More generally - this kinda thing needs to break in a much more informative way, but the variables stuff is targeted at this (#370, #388).

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_rainforest/example_data/config/vr_run.toml
+++ b/virtual_rainforest/example_data/config/vr_run.toml
@@ -1,5 +1,6 @@
 [core]
 [hydrology.depends]
+init = ['plants']
 update = ['plants', 'abiotic_simple']
 [abiotic_simple]
 [animals]


vgro (approved):
Fine with me if that fixes the problem.

PR #400, opened 2024-03-01
Adding explicit ipykernel install step to RTD

### LINKED ISSUE #399: The `vr_python3` kernel is not found on RTD
**Describe the bug**

Following on from the fixes in #398, RTD builds are _now_ failing with:

```sh
[WARNING] - kernelspec - get_kernel_spec(284) - Kernelspec name vr_python3 cannot be found!
[ERROR] - manager - wrapper(95) - No such kernel named vr_python3
```

See the notes in [`docs/source/development/documentation/jupyter_notebooks.md`](
https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/e1ac1ad1960f63f27c0ec73a4659cc893236439b/docs/source/development/documentation/jupyter_notebooks.md) for the rationale for using a specific named kernel.

I think what has happened is that `poetry` environment handling has changed, so the RTD config now explicitly sets the install environment using `READTHEDOCS_VIRTUALENV_PATH` but that is no longer registered as an `ipykernel` under the `vr_python3` name. So I _think_ we just need to include the following line from our `jupyter_notebooks` guide in the RTD build.

```zsh
poetry run python -m ipykernel install --user --name=vr_python3
```

davidorme:
# Description

This hopefully fixes #399 by adding an explicit `ipykernel install` step to `.readthedocs.yaml` build sequence.

@dalonsoa - this is a bit of an informed guess. I guess we just try it and see what happens, since we can only test by seeing if RTD builds!

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -22,6 +22,7 @@ build:
       # VIRTUAL_ENV needs to be set manually for now.
       # See https://github.com/readthedocs/readthedocs.org/pull/11152/
       - VIRTUAL_ENV=$READTHEDOCS_VIRTUALENV_PATH poetry install --with docs
+      - VIRTUAL_ENV=$READTHEDOCS_VIRTUALENV_PATH poetry run python -m ipykernel install --user --name=vr_python3
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:


davidorme:
I've force merged this to see if I'm on the right track and try and get the rename done.

PR #407, opened 2024-03-07
Feature/remove unnecessary required vars

vgro:
This PR removes unnecessary variables from `required_vars` and moves the bounds for variables to the `constants` file.

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/abiotic_simple/constants.py
+++ b/virtual_ecosystem/models/abiotic_simple/constants.py
@@ -13,24 +13,12 @@ from virtual_ecosystem.core.constants_class import ConstantsDataclass
 class AbioticSimpleConsts(ConstantsDataclass):
     """Dataclass to store all constants for the `abiotic_simple` model."""
 
-    air_temperature_gradient: float = -1.27
-    """Gradient for linear regression to calculate air temperature as a function of
-    leaf area index from :cite:t:`hardwick_relationship_2015`"""
-
-    relative_humidity_gradient: float = 5.4
-    """Gradient for linear regression to calculate relative humidity as a function of
-    leaf area index from :cite:t:`hardwick_relationship_2015`"""
-
-    vapour_pressure_deficit_gradient: float = -252.24
-    """Gradient for linear regression to calculate vapour pressure deficit as a function
-    of leaf area index from :cite:t:`hardwick_relationship_2015`"""
-
     saturation_vapour_pressure_factor1: float = 0.61078
-    """factor 1 for saturation vapour pressure calculation."""
+    """Factor 1 for saturation vapour pressure calculation."""
     saturation_vapour_pressure_factor2: float = 7.5
-    """factor 2 for saturation vapour pressure calculation."""
+    """Factor 2 for saturation vapour pressure calculation."""
     saturation_vapour_pressure_factor3: float = 237.3
-    """factor 3 for saturation vapour pressure calculation."""
+    """Factor 3 for saturation vapour pressure calculation."""
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


virtual_ecosystem/models/abiotic_simple/constants.py
```diff
@@ -31,3 +31,37 @@ class AbioticSimpleConsts(ConstantsDataclass):
...
+
+
+@dataclass(frozen=True)
+class AbioticSimpleBounds(ConstantsDataclass):
+    """Upper and lower bounds for abiotic variables.
+
+    When a values falls outside these bounds, it is set to the bound value. Note that
+    this approach does not conserve energy and matter in the system. This will be
+    implemented at a later stage.
+    """
```
davidorme, line 43: I think this would be cleaner as tuples:

```python

@dataclass(frozen=True)
class AbioticSimpleBounds(ConstantsDataclass):
    air_temperature: tuple[float, float] = (-20.0, 80.0)
```

It's more concise and makes it easier to unpack the data when you need it.

virtual_ecosystem/models/abiotic_simple/microclimate.py
```diff
@@ -107,15 +93,18 @@ def run_microclimate(
 
     # interpolate atmospheric profiles
     for var in ["air_temperature", "relative_humidity", "vapour_pressure_deficit"]:
+        upper_bound_attr = f"{var}_max"
+        lower_bound_attr = f"{var}_min"
```
davidorme, line 97: If you using bounds as tuples this can be dropped and replaced with:

```python
lower, upper = getattr(bounds, var)
```

virtual_ecosystem/models/abiotic_simple/microclimate.py
```diff
@@ -107,15 +93,18 @@ def run_microclimate(
...
             data=data,
             reference_data=data[var + "_ref"].isel(time_index=time_index),
             leaf_area_index_sum=leaf_area_index_sum,
             layer_roles=layer_roles,
             layer_heights=data["layer_heights"],
-            upper_bound=Bounds[var + "_max"],
-            lower_bound=Bounds[var + "_min"],
-            gradient=getattr(constants, var + "_gradient"),
+            upper_bound=getattr(bounds, upper_bound_attr),
+            lower_bound=getattr(bounds, lower_bound_attr),
```
davidorme, line 106: And hence:
```suggestion
            upper_bound=upper,
            lower_bound=lower,
```

virtual_ecosystem/models/abiotic_simple/constants.py
```diff
@@ -31,3 +31,37 @@ class AbioticSimpleConsts(ConstantsDataclass):
...
+
+    air_temperature_min: float = -20.0
+    """Minimum air tempertature, [C]."""
```
davidorme, line 46:
```suggestion
    """Minimum air temperature, [C]."""
```

davidorme (approved):
This looks good to me. I think it would be cleaner to use tuples of bounds - as in the comments. It might also be cleaner to use bounds as tuples in function arguments (rather than separate `upper_bound` and `lower_bound` arguments) - that keeps function signatures shorter and packages two logically inseperable values neatly.

Having said that - this is fine and I'm not sure whether the changes to handle mass and energy conservation are basically going to replace this. Your call!

virtual_ecosystem/models/abiotic_simple/constants.py
vgro: This is a good suggestion. I added the gradients to the tuples, too, so it's a nice package for each variable.

davidorme: I did wonder about including the gradients too - not everything has a gradient, so I kept it to bounds, but I think that works too.

PR #410, opened 2024-03-15
Feature/animal module - The Madingley Foraging Update

TaranRallings:
# Description

This PR is to get the VE in line with the Madingley Model's approach to foraging dynamics. It is a lot more complicated than what I had done previously and now I understand why foraging took up most of their computing resources! Foraging is completely reworked and now involves a lot of hopping back and forth between methods of `AnimalCohort` and `scaling_functions`. I attempted to impose some consistent logic on which methods went where but I am confident it can be improved. 

All feedback is welcome but I am particularly looking for errors/improvements to the organisational structure and how things are woven together. If you notice things that could speed up run-time, that would be useful as well. 

This work is both too large and the product of new-dad brain, mea culpa. I am pretty sure some obvious stuff slipped through the cracks but I've been staring at this for too long to see it!

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+
+        for plant in plant_list:
+            # Calculate the mass to be consumed from this plant
+            consumed_mass = self.calculate_consumed_mass_herbivory(plant_list, plant)
+            # Update the plant resource's state based on consumed mass
+            actual_consumed_mass = plant.get_eaten(consumed_mass, self, excrement_pool)
+            # Update total mass gained by the herbivore
+            total_consumed_mass += actual_consumed_mass
+
+        return total_consumed_mass
```
dalonsoa, line 505: There's no excrement generation here. Is this correct?

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+            The float value of theta.
+        """
+        A_cell = 1.0  # temporary
+        theta = 0.0
 
-        Currently, all this does is call the food's get_eaten method and pass the
-        returned mass value to the consumer.
+        for cohort in animal_list:
+            if self.mass_current == cohort.mass_current:
+                theta += cohort.individuals / A_cell
```
dalonsoa, line 574: As above, this _might_ be faster:
```suggestion
        return sum(
            cohort.individuals / A_cell
            for cohort in animal_list
            if self.mass_current == cohort.mass_current
        )
```

virtual_ecosystem/models/animals/animal_communities.py
```diff
@@ -215,20 +227,20 @@ def forage_community(self) -> None:
         plant_list = [plant_community]
 
         for consumer_cohort in self.all_animal_cohorts:
-            if (
-                consumer_cohort.individuals == 0
-            ):  # temporary while finalizing die_cohort placements
-                continue
-
+            # Prepare the prey list for the consumer cohort
             prey_list = self.collect_prey(consumer_cohort)
```
dalonsoa, line 231: Should there be a similar function for the plants? Surely not all animals can eat all the plants, right?

virtual_ecosystem/models/animals/animal_model.py
```diff
@@ -178,7 +178,13 @@ def spinup(self) -> None:
     def update(self, time_index: int, **kwargs: Any) -> None:
         """Function to step the animal model through time.
 
-        Currently this is a toy implementation.
+        Temporary solution.
```
dalonsoa, line 181: Depending on how much interdependence there is between the different events, we might want to use [coroutines ](https://docs.python.org/3.11/library/asyncio-task.html)here.

virtual_ecosystem/models/animals/scaling_functions.py
```diff
@@ -264,3 +266,188 @@ def natural_mortality_scaling(mass: float, terms: tuple) -> float:
...
+
+    return alpha_0_herb * mass
+
+
+def k_i_k(alpha_i_k: float, phi_herb_t: float, B_k_t: float, A_cell: float) -> float:
+    """The potential biomass (g) of plant k eaten by cohort i, per day.
+
+    TODO: Finish docstring
+
+    Madingley
```
dalonsoa, line 297: Yep :)

dalonsoa (changes requested):
This looks very good - and very interesting! I've made a couple of comments/questions, and you already have a few ToDos (eg, improve docstrings), but in general I've spotted very few things. 

Once we have a full size simulation up and running, we will need to run this and profile it to spot which parts are slow and need improving.

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+            ) + sf.H_i_k(
+                self.constants.h_herb_0,
+                self.constants.M_herb_ref,
+                self.mass_current,
+                self.constants.b_herb,
+            )
+        return total_handling_t
+
+    def F_i_k(self, plant_list: Sequence[Resource], target_plant: Resource) -> float:
+        """Refactored method to determine instantaneous herbivory rate on plant k.
```
davidorme, line 322: The function name is a obscure. Also the reader of the docstrings does not care it is refactored 😄

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+        """Calculate the search rate of the predator for preying."""
+        return sf.alpha_i_j(self.constants.alpha_0_pred, self.mass_current, w_bar)
+
+    def calculate_potential_prey_consumed(
+        self, alpha: float, theta_i_j: float
+    ) -> float:
+        """Calculate the potential number of prey consumed."""
+        N_i = self.individuals
+        A_cell = 1.0  # temporary
+        return sf.k_i_j(alpha, N_i, A_cell, theta_i_j)
```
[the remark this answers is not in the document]
davidorme: @dalonsoa I was wondering about this up above with the line:

```python
individual_mass = (
            self.mass_current
        ) 
```

That value is then used several times, so extracting it once saves several `getattr` calls within the function. Is this a balance or is it generally better just to use the attribute directly?

virtual_ecosystem/models/animals/animal_communities.py
```diff
@@ -133,6 +142,9 @@ def migrate_community(self) -> None:
     def die_cohort(self, cohort: AnimalCohort) -> None:
         """The function to change the cohort status from alive to dead.
 
+        This method is a computation question, not a scientific one. The science is in
+        the various methods that call die_cohort after a mortality inflicting event.
+
```
davidorme, line 147: This is a bit obscure. Also generally avoid "The function to" or "A method that" etc in docstring titles.

```suggestion
        """Remove a dead cohort  from a community.
        
        This method is used to mark a cohort as dead and remove it a community by 
        any of the methods that can inflict mortality.
```

I guess we could rename it `remove_dead_cohort` for further clarity?

virtual_ecosystem/models/animals/animal_model.py
davidorme: Oh no... I mean that sounds appropriate but `asyncio` is scary! 😱

virtual_ecosystem/models/animals/scaling_functions.py
```diff
@@ -264,3 +266,188 @@ def natural_mortality_scaling(mass: float, terms: tuple) -> float:
...
 
     return daily_mortality_prob
+
+
+def alpha_i_k(alpha_0_herb: float, mass: float) -> float:
+    """Effective rate at which an individual herbivore searches its environment.
+
+    This is linear scaling of herbivore search times with current body mass.
+
+    Madingley
```
davidorme, line 276: I get why you're using these function names, but it is harder to follow:
```suggestion
def herbivore_search_rate(alpha_0_herb: float, mass: float) -> float:
    """Effective rate at which an individual herbivore searches its environment.

    This is a linear scaling of herbivore search times with current body mass. It is the :math:`\alpha_{i,k}` term from the Madingley model.

```

davidorme (changes requested):
This is looking really good - hard to keep all the flow in mind, so it would be good to create some kind of a flow diagram for docs at some point - but the functions and commenting are all very clear. I've basically got three main issues:

1. In tests, it would be good to standardise on using the pytest `mocker` fixture rather than importing `mock`.
2. Again in tests, I've a strong preference for using `pytest.param` and `id=` to get meaningful failure messages (and it makes it easier to pick a single test to run for troubleshooting if we can pick from ids).
3. I think it would be better to use more readable method and function names than the Madingley notation.

But that's all minor.

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+        # Calculate the mass represented by the individuals actually eaten
+        actual_killed_mass = actual_individuals_eaten * individual_mass
+
+        # Calculate the actual amount of mass consumed by the predator
+        actual_consumed_mass = min(actual_killed_mass, potential_consumed_mass)
+
+        # Calculate the amount of mass that goes into carcass pool
+        carcass_mass = (actual_killed_mass - actual_consumed_mass) + (
+            actual_consumed_mass * (1 - predator.functional_group.mechanical_efficiency)
+        )
```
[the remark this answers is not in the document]
TaranRallings: Ah, yeah I am accidentality double-counting here.

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+
+        Args:
+            target_cohort: The prey cohort from which mass will be consumed.
+
+        Returns:
+            The mass to be consumed from the target cohort by the predator (in kg).
+        """
+        F = self.F_i_j_individual(
+            [target_cohort], target_cohort
+        )  # Note: Adjusting this call may be necessary
```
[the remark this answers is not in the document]
TaranRallings: I think this was something temporary that slipped through the cracks!

virtual_ecosystem/models/animals/animal_cohorts.py
TaranRallings: That's right. Currently, I am putting excrement generation inside of get_eaten methods. I think its an inelegant place for them but its going to make things easier when we try to build in stoichiometry and seed dispersal. My hope is to clean it up after seed dispersal is implemented so I can weight the pros and cons properly.

virtual_ecosystem/models/animals/animal_communities.py
TaranRallings: They can eat all of the plants... for now. The problem of which animals can eat which plants is unfortunately complex and there is no similar first-pass allometric relationship that captures 70% of the variance. Once the plant module has herbivory-facing trait diversity I will be building something parallel out for managing the plants end of things.

virtual_ecosystem/models/animals/animal_model.py
TaranRallings: It would be interesting to test it out and see how it changes the outcomes.

virtual_ecosystem/models/animals/scaling_functions.py
TaranRallings: I should have mentioned it in my original post but these technical method names are temporary until we get into unit-testing but I could probably shorten that timing to the other side of a proper flow diagram. It is... challenging to do the cross-referencing with madingley to make sure it's all assembled properly. If I started changing the names right now it will be extremely difficult.

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
-        AnimalCommunity.forage_community and performs a simple random uniform selection.
-        After this, excrete is called to pass excess waste to the excrement pool.
-        Later this function will involve more complex weightings of prey options.
+        This is Madingley's delta_assimilation_mass_predation
 
-        TODO: Fix the occasional division by zero bug in eat and then remove Optional
+        TODO: Replace delta_t values with actual reference
+        TODO: add epsilon conversion efficiency
 
         Args:
```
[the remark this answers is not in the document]
dalonsoa: Still missing stuff in docstrings...

dalonsoa (approved):
LGTM!!

davidorme (approved):
LGTM too - there are three unresolved comments from my previous review. Could you resolve those?

TaranRallings:
> LGTM too - there are three unresolved comments from my previous review. Could you resolve those?

Bah! I missed the "expand comments" bit they were hiding in!

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -195,145 +188,418 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+            self.mass_current
+        )  # Mass of an average individual in the cohort
+        max_individuals_eaten = ceil(potential_consumed_mass / individual_mass)
+        actual_individuals_eaten = min(max_individuals_eaten, self.individuals)
+
+        # Calculate the mass represented by the individuals actually eaten
+        actual_killed_mass = actual_individuals_eaten * individual_mass
+
+        # Calculate the actual amount of mass consumed by the predator
+        actual_consumed_mass = min(actual_killed_mass, potential_consumed_mass)
```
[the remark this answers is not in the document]
TaranRallings: I don't think it is the case that `actual_killed_mass <= potential_consumed_mass` is always true, unless I am missing something (unfortunately always possible).

Let's assume:
prey_individuals = 1
prey_mass = 1.0
potential_consumed_mass = 0.5 (this is a possible result from `calculate_consumed_mass_predation` under these conditions)

(I've updated my language a bit here for clarity)
`       # Mass of an average individual in the cohort
        individual_mass = self.mass_current

        max_individuals_killed = ceil(potential_consumed_mass / individual_mass)
        actual_individuals_killed = min(max_individuals_killed, self.individuals)

        # Calculate the mass represented by the individuals actually killed
        actual_mass_killed = actual_individuals_killed * individual_mass

        # Calculate the actual amount of mass consumed by the predator
        actual_mass_consumed = min(actual_mass_killed, potential_consumed_mass)`
        
max_individuals_killed = ceil(0.5 / 1) = 1
actual_individuals_killed = min(1, 1) = 1

actual_mass_killed = 1 * 1.0 = 1.0
actual_mass_consumed = min(1.0, 0.5) = 0.5

In which case, actual_killed_mass is 1.0 and potential_consumed_mass is 0.5. Describing the ecological situation where a predator has killed more than they can actually eat. 

Does that make sense or have I fudged something up?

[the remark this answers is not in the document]
davidorme: As we discussed on Teams, this makes sense. I'm copying in a comment from that because although it isn't going into this PR, there was some stuff we might want to implement as part of a model review:

> OK - I think the logic makes sense. I think those variable name changes make it clearer.  Another thing I found tricky was the switch in perspectives from predator to prey in those two lines in delta_mass_predation and the naming of the functions.
 
```python
# Calculate the mass to be consumed from this cohort
consumed_mass = self.calculate_consumed_mass_predation(cohort)
# Call get_eaten on the prey cohort to update its mass and individuals
actual_consumed_mass = cohort.get_eaten(consumed_mass, self, carcass_pool)
```
 
> That first value is actually target_mass , right? Nothing has been consumed at all until get_eaten runs. I don't think that target_mass is needed elsewhere and I wonder if having a single function would be more efficient and show the logic more cleanly? I think the same structure is used on plants too?
 
```python
def consume_prey(
    self, target_cohort: AnimalCohort, carcass_pool: DecayPool
) -> float:
    """Consumes prey from a target cohort.
    This method utilizes the F_i_j_individual method to determine the rate at which
    the target cohort is consumed, and then calculates the number of individuals
    killed and the actual mass to be consumed based on this rate and other model
    parameters. The target cohort and carcass pool are also adjusted to reflect the
    predation.
    Args:
        target_cohort: The prey cohort from which mass will be consumed.
        carcass_pool: The pool to which remains of eaten individuals are delivered.
    Returns:
        The actual mass consumed by the predator,(in kg).
    """
    F = self.F_i_j_individual(
        [target_cohort], target_cohort
    )  # Note: Adjusting this call may be necessary
    delta_t = 30.0  # days, TODO: Replace with actual reference or model parameter

    # Calculate the potential consumed mass based on Mad. formula for
    # delta_mass_predation
    potential_consumed_mass = (
        target_cohort.mass_current
        * target_cohort.individuals
        * (
            1
            - exp(-(F * delta_t * self.constants.tau_f * self.constants.sigma_f_t))
        )
    )

    # Calculate the number of individuals killed
    max_individuals_killed = ceil(
        potential_consumed_mass / target_cohort.mass_current
    )
    actual_individuals_killed = min(
        max_individuals_killed, target_cohort.individuals
    )

    # Calculate the mass represented by the individuals killed
    actual_killed_mass = actual_individuals_killed * target_cohort.mass_current

    # Calculate the actual amount of mass consumed by the predator, which can be
    # lower that the killed mass when there is overkill
    actual_consumed_mass = min(actual_killed_mass, potential_consumed_mass)

    # Update the number of individuals in the prey cohort
    target_cohort.individuals -= actual_individuals_killed

    # Calculate the amount of mass that goes into carcass pool and update the pool
    carcass_mass = (actual_killed_mass - actual_consumed_mass) + (
        actual_consumed_mass * (1 - self.functional_group.mechanical_efficiency)
    )

    # Update the carcass pool with carcass mass
    target_cohort.update_carcass_pool(carcass_mass, self, carcass_pool)

    return actual_consumed_mass
```

PR #419, opened 2024-04-26
Feature/juvenile dispersal - Adding juvenile dispersal to the Animal Module

TaranRallings:
# Description

Following Madingley, this builds on previous starvation-induced dispersal to add a dispersal check immediately after a cohort is born. When migrate_community is called during the normal flow of model execution there is a secondary check to see if the cohort is newly created. If the cohort is new, it creates a probability of dispersal based on allometric movement speed and grid size. 

Currently, this only works for 'rook move' dispersal. The logic can be expanded for diagonal movement and multi-grid occupancy but we can leave that for a little later. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -711,3 +705,33 @@ def inflict_natural_mortality(
...
+        Assuming cohort individuals are homogenously distributed within a grid cell and
+        that the move is non-diagonal, the probability is then equal to the ratio of
+        dispersal speed to the side-length of a grid cell.
+
+        Returns:
+            The probability of diffusive natal dispersal to a neighboring grid cell.
+
+        """
+
+        A_cell = 1.0  #
```
dalonsoa, line 727: Maybe add another `TODO` to remember to update this to the right number.

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -711,3 +705,33 @@ def inflict_natural_mortality(
...
+        grid_side = sqrt(A_cell)
+        velocity = sf.juvenile_dispersal_speed(
+            self.mass_current,
+            self.constants.V_disp,
+            self.constants.M_disp_ref,
+            self.constants.o_disp,
+        )
+        probability_of_dispersal = velocity / grid_side
```
dalonsoa, line 735: It will be useful to do a dimensional discussion here (an explanation is ok) to confirm this is unitless - as any probability should be - as well as a check to ensure that it is always <=1 .

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -711,3 +705,33 @@ def inflict_natural_mortality(
         )
 
         self.die_individual(number_of_deaths, carcass_pool)
+
+    def migrate_juvenile_probability(self) -> float:
```
dalonsoa, line 709: The output of this function is not really a probability, but rather a flag if it should migrate or not based on the calculate probability. Both the docstring explaining the function, the return type (I'm surprised `mypy` didn't picked this) and the function name itself should be updated to reflect this. Maybe calling it `juvenile_should_migrate` or something like that to reflect that is actually a boolean check.

Or return the actual probability and then use `random.random()` in the calling function, which I think makes more sense.

virtual_ecosystem/models/animals/animal_communities.py
```diff
@@ -131,14 +132,25 @@ def migrate(self, migrant: AnimalCohort, destination: AnimalCommunity) -> None:
...
             if cohort.is_below_mass_threshold(self.constants.dispersal_mass_threshold):
                 # Random walk destination from the neighbouring keys
                 destination_key = choice(self.neighbouring_keys)
                 destination = self.get_destination(destination_key)
                 self.migrate(cohort, destination)
 
+            elif cohort.age == 0.0 and cohort.migrate_juvenile_probability():
+                destination_key = choice(self.neighbouring_keys)
+                destination = self.get_destination(destination_key)
+                self.migrate(cohort, destination)
```
dalonsoa, line 152: There's a lot of code duplication here. Unless you're planning t follow a different migration approach for each branch, I' would just put together all the conditions for migration:
```suggestion
        migrate = (
            cohort.is_below_mass_threshold(self.constants.dispersal_mass_threshold)
            or (cohort.age == 0.0 and cohort.migrate_juvenile_probability())
        )
        
        if not migrate:
            return
            
        destination_key = choice(self.neighbouring_keys)
        destination = self.get_destination(destination_key)
        self.migrate(cohort, destination)
```

dalonsoa (changes requested):
I've a couple of comments related to the naming or interpretation of the probability. But otherwise looks good.

virtual_ecosystem/models/animals/animal_cohorts.py
TaranRallings: Oh good point! I'm surprised that didn't get picked up.

virtual_ecosystem/models/animals/animal_communities.py
TaranRallings: That is certainly much more straightforward, thanks!

dalonsoa (approved):
A couple of suggestions for the probability part, but looks good otherwise.

PR #420, opened 2024-05-01
Soil moisture units to mm in data object

### LINKED ISSUE #345: Soil moisture isn't actually needed in the Virtual Rainforest simulation
At the moment the `hydrology` model calculates soil moisture, but none of the other models actually need to use it. At the moment, the soil model uses it to calculate amount of water in the topsoil layer, something the `hydrology` model could return instead.

The `hydrology` model could get rid of the soil moisture calculation, and just return the quantity the `soil` model needs instead. It might be worth retaining the soil moisture calculation function, as this is a nice output to have though.

vgro:
This PR changes the soil moisture unit to mm in the data object. For some calculations in the hydrology and abiotic model, it is still needed in volumetric moisture content, but I think I reduced the number of conversions.
I also tidied the code based on what we have learned, but there is still a big outstanding issue with accessing indices etc.

@jacobcook1995 could you add the changes needed in the soil model to this branch please?

Fixes #345 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -289,7 +289,7 @@ def dummy_carbon_data(fixture_core_components):
             # At present the soil model only uses the top soil layer, so this is the
             # only one with real test values in
             DataArray(
-                [[0.9304620050, 0.787549327, 0.504263188, 0.302527807]],
+                [[232.61550125, 196.88733175, 126.065797, 75.63195175]],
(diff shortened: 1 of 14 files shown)


jacobcook1995 (approved):
LGTM, I'll change the `soil` and `litter` models to match the new setup

PR #422, opened 2024-05-17
Adding non-predation mortality and semelparity to the Animal Module.

TaranRallings:
# Description

This adds a new trait to functional groups, reproductive type, which is broken into two variations, iteroparity (reproducing multiple times) and semelparity (reproducing once and then dying). Iteroparous organisms reproduce as before, semelparous organisms also use some of their non-reproductive mass in the reproductive event and then the cohort dies. 

This PR also expands non-predation mortality to include a fixed background mortality, senescence mortality, and starvation mortality. There were small errors in the Madingley paper for the latter two mortalities so you may notice an excess of diagnostics in the testing from when I was trying to figure out if I had gone mad. I left them in, in case they might be useful later.  Let me know if you think I should remove them. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/animals/animal_cohorts.py
+++ b/virtual_ecosystem/models/animals/animal_cohorts.py
@@ -766,10 +766,10 @@ class AnimalCohort:
 
         return min(1.0, probability_of_dispersal)
 
-    def total_non_predation_mortality(
+    def inflict_non_predation_mortality(
         self, dt: float, carcass_pool: CarcassPool
     ) -> None:
-        """Combine background, senescence, and starvation mortalities."""
+        """Inflict combined background, senescence, and starvation mortalities."""
 
         pop_size = self.individuals
         mass_current = self.mass_current
--- a/virtual_ecosystem/models/animals/animal_cohorts.py
+++ b/virtual_ecosystem/models/animals/animal_cohorts.py
@@ -766,7 +766,9 @@ class AnimalCohort:
 
         return min(1.0, probability_of_dispersal)
 
-    def total_non_predation_mortality(self, dt: float) -> None:
+    def total_non_predation_mortality(
+        self, dt: float, carcass_pool: CarcassPool
+    ) -> None:
         """Combine background, senescence, and starvation mortalities."""
 
         pop_size = self.individuals
@@ -802,4 +804,4 @@ class AnimalCohort:
         number_dead = ceil(pop_size * (1 - exp(-u_t * dt)))
 
         # Remove the dead individuals from the cohort
-        self.individuals -= number_dead
+        self.die_individual(number_dead, carcass_pool)
--- a/virtual_ecosystem/models/animals/animal_cohorts.py
+++ b/virtual_ecosystem/models/animals/animal_cohorts.py
@@ -769,7 +769,13 @@ class AnimalCohort:
     def inflict_non_predation_mortality(
         self, dt: float, carcass_pool: CarcassPool
     ) -> None:
-        """Inflict combined background, senescence, and starvation mortalities."""
+        """Inflict combined background, senescence, and starvation mortalities.
+
+        Args:
+            dt: The time passed in the timestep (days).
+            carcass_pool: The local carcass pool to which dead individuals go.
+
+        """
 
         pop_size = self.individuals
         mass_current = self.mass_current
--- a/virtual_ecosystem/models/animals/animal_cohorts.py
+++ b/virtual_ecosystem/models/animals/animal_cohorts.py
@@ -46,7 +46,6 @@ class AnimalCohort:
         """The functional group of the animal cohort which holds constants."""
         self.name = functional_group.name
         """The functional type name of the animal cohort."""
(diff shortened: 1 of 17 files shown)


virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -748,3 +765,49 @@ def migrate_juvenile_probability(self) -> float:
...
+            self.constants.u_bg
+        )  # constant background mortality
+
+        if self.is_mature is True:
+            # senescence mortality is only experienced by mature adults.
+            u_se = sf.senescence_mortality(
+                self.constants.lambda_se, t_to_maturity, t_since_maturity
+            )  # senesence mortality
+        elif self.is_mature is False:
+            u_se = 0.0
```
dalonsoa, line 797: This can be slightly re-organised to save a conditional:
```suggestion
        u_se = 0.0
        if self.is_mature:
            # senescence mortality is only experienced by mature adults.
            u_se = sf.senescence_mortality(
                self.constants.lambda_se, t_to_maturity, t_since_maturity
            )  # senesence mortality
```

virtual_ecosystem/models/animals/animal_cohorts.py
```diff
@@ -748,3 +765,49 @@ def migrate_juvenile_probability(self) -> float:
...
+            self.constants.J_st,
+            self.constants.zeta_st,
+            mass_current,
+            mass_max,
+        )  # starvation mortality
+        u_t = u_bg + u_se + u_st
+
+        # Calculate the total number of dead individuals
+        # TODO: the use of ceil here might have unintended outcomes, keep an eye on it
+        number_dead = ceil(pop_size * (1 - exp(-u_t * dt)))
```
dalonsoa, line 810: I'm curious: what are the unexpected consequences?

`pop_size * (1 - exp(-u_t * dt))` can be, at most, `pop_size`, so there's no risk of going above the population size when using `ceil`. And I guess that, at the lower end, it makes sense that a fraction of individual dead is a full individual dead, so sort a reasonable behaviour also there, right?

virtual_ecosystem/models/animals/animal_communities.py
```diff
@@ -183,17 +183,32 @@ def birth(self, parent_cohort: AnimalCohort) -> None:
...
+        if parent_cohort.functional_group.reproductive_type == "semelparous":
+            non_reproductive_mass_loss = (
+                parent_cohort.mass_current
+                * parent_cohort.constants.semelparity_mass_loss
+            )
+            parent_cohort.mass_current -= non_reproductive_mass_loss
+            # kill the semelparous parent cohort
+            parent_cohort.is_alive = False
+        else:
+            non_reproductive_mass_loss = 0.0
```
dalonsoa, line 204: As above:
```suggestion
        non_reproductive_mass_loss = 0.0
        if parent_cohort.functional_group.reproductive_type == "semelparous":
            non_reproductive_mass_loss = (
                parent_cohort.mass_current
                * parent_cohort.constants.semelparity_mass_loss
            )
            parent_cohort.mass_current -= non_reproductive_mass_loss
            # kill the semelparous parent cohort
            parent_cohort.is_alive = False    
```

virtual_ecosystem/models/animals/scaling_functions.py
```diff
@@ -268,6 +268,86 @@ def natural_mortality_scaling(mass: float, terms: tuple) -> float:
...
+
+    return u_se
+
+
+def starvation_mortality(
+    lambda_max: float, J_st: float, zeta_st: float, mass_current: float, mass_max: float
+) -> float:
+    """Mortality from body-mass loss.
+
+    There is a error in the madingley paper that does not follow their source code. The
```
dalonsoa, line 325: The paper seems to be full of errors 😄

dalonsoa (changes requested):
I've added a few comments here and there, bur it looks good to me. To be honest, it's a fascinating topic!

virtual_ecosystem/models/animals/scaling_functions.py
TaranRallings: There have been some unpleasant surprises! When I get to doing the final parametrization, I'm going to comb through all of their actual C# code to make sure.

virtual_ecosystem/models/animals/animal_cohorts.py
dalonsoa: I see... 

You're already using probabilistic events in other parts of the code, so I guess it might make sense. In that case, you will have a certain number of dead individuals for sure (might be none) and 1 more that might or might not be dead. This would be rather irrelevant for ants but might make a whole difference for elephants, as you say.

virtual_ecosystem/models/animals/scaling_functions.py
dalonsoa: It should be them publishing an addendum...

dalonsoa (approved):
All good!

PR #424, opened 2024-05-29
Feature/test ragged arrays

vgro:
This PR adds a new dummy climate data fixture to test that the `abiotic_simple` model works with different numbers of canopy layers (the mechanistic abiotic model tests are not updated here). I kept the old tests for now, could be removed or merged in the future.

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM (with the caveat that I don't 100% get how ragged arrays work)

vgro:
> LGTM (with the caveat that I don't 100% get how ragged arrays work)

Thank you for the quick review, I'll wait what David thinks before I merge because it will affect him, too :-)

tests/conftest.py
```diff
@@ -729,3 +747,315 @@ def dummy_climate_data(fixture_core_components):
...
+
+# dummy climate data with different number of canopy layers
+@pytest.fixture
+def dummy_climate_data_ragged(fixture_core_components, fixture_empty_array):
+    """Creates a dummy climate data object for use in tests."""
+
+    from virtual_ecosystem.core.data import Data
+    from virtual_ecosystem.core.grid import Grid
+
+    # Setup the data object with four cells.
```
davidorme, line 760:
```suggestion
    # Setup the data object with three cells.
```

tests/conftest.py
```diff
@@ -729,3 +747,315 @@ def dummy_climate_data(fixture_core_components):
     )
 
     return data
+
+
+# dummy climate data with different number of canopy layers
+@pytest.fixture
+def dummy_climate_data_ragged(fixture_core_components, fixture_empty_array):
```
davidorme, line 754: I wouldn't really describe this as using ragged arrays. Because we are padding with `np.nan` to a fixed number of canopy layers, the data are always proper square arrays. It would be truly ragged if we stored the canopy for each cell as a list containing arrays with data for only the 'true' canopy layers. 

This is picky - but I think it _could_ confuse code review if people expect properly ragged structures across cells (which we most definitely get with the community data!).

```suggestion
def dummy_climate_data_canopy_variable(fixture_core_components, fixture_empty_array):
```

And then similar changes in the other names.

davidorme (approved):
This looks great - I've complained about using 'ragged' here, which would be better to change for clarity, but the actual functionality etc. is fine.

tests/conftest.py
vgro: Ok, I changed it to `varying_canopy` because I think using the word `variable` here might be confusing.

PR #430, opened 2024-06-05
Feature/fixes for variables pr

vgro:
This PR adds missing variables to the data object and updates some units in preparation for #388 

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

## Key checklist

- [x] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [x] All tests pass: `$ poetry run pytest`

## Further checks

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works

PR #431, opened 2024-06-05
Variables 1 vgro

vgro:
I reviewed the variables in `data_variables.toml` that are related to the abiotic models and the hydrology model. I also updated the contents of `required_init_vars`, `vars_updated`, `required_update_vars` and `vars_initialised`. I hope I captured everything.

I have not updated the tests because I was not quite sure if that was sensible at this stage.


virtual_ecosystem/models/abiotic/abiotic_model.py
```diff
@@ -67,12 +69,63 @@ class AbioticModel(
...
+    required_update_vars=(
+        "air_temperature_ref",
+        "relative_humidity_ref",
+        "vapour_pressure_deficit_ref",
+        "atmospheric_pressure_ref",
+        "atmospheric_co2_ref",
+        "wind_speed_ref",
+        "leaf_area_index",
+        "layer_heights",
+        "topofcanopy_radiation",
```
vgro, line 99:
```suggestion
        "topofcanopy_radiation",
        "stomatal_conductance",
```

virtual_ecosystem/data_variables.toml
```diff
@@ -1,31 +1,23 @@
+[[variable]]
+name = "air_heat_conductivity"
+description = "Air heat conductivity between layers"
+unit = "mol m-2 s-1"
+variable_type = "DataArray(float)"
+axis = [space]
```
dalonsoa, line 6: All of these `space` or `time` throughout this file need to be in quotes, as they are strings:
```suggestion
axis = ["space"]
```

virtual_ecosystem/data_variables.toml
```diff
@@ -1,31 +1,23 @@
...
+
 [[variable]]
 name = "air_temperature"
 description = "Air temperature profile"
 unit = "C"
-variable_type = ""
-axis = []
+variable_type = "DataArray(float)"
```
dalonsoa, line 12: I'm not sure if we want to indicate that this is a `DataArray` of floats or just indicate `floats`, taking for granted that they are all going to be `DataArray`s. Totally up to you and the use case you have in mind.

virtual_ecosystem/models/abiotic_simple/abiotic_simple_model.py
dalonsoa, line 1: The `required_init_vars` of the `AbioticSimpleModel` also need updating.

virtual_ecosystem/models/hydrology/hydrology_model.py
dalonsoa, line 1: The `AbioticSimpleModel` of the `HydrologyModel` also need updating.

virtual_ecosystem/models/abiotic/abiotic_model.py
```diff
@@ -54,6 +54,8 @@ class AbioticModel(
         ("air_temperature_ref", ("spatial",)),
         ("relative_humidity_ref", ("spatial",)),
         ("topofcanopy_radiation", ("spatial",)),
+        ("leaf_area_index", ("spatial",)),
+        ("layer_heights", ("spatial",)),
```
dalonsoa, line 58: In the final version, the `required_init_vars` are just a tuple, like `vars_updated` and the others since the axis information is contained in the variable description in the toml file.
```suggestion
        "air_temperature_ref",
        "relative_humidity_ref",
        "topofcanopy_radiation",
        "leaf_area_index",
        "layer_heights",
```
This change will make the validator fail, as it is expecting a different structure, but we will deal with that once this is merged, as it will be necessary to do some work harmonizing things.

dalonsoa (changes requested):
A step in the right direction, but there are a few things to change. You have A LOT of variables!!!

virtual_ecosystem/data_variables.toml
vgro: Could you please remind me what exactly this `variable_type` is used for or validated against? If all entries are DataArrays, then I can easily change in to floats. If we have some very different types, like the functional groups or so, I would prefer keeping a more complete description. Thoughts @davidorme @jacobcook1995 ?

davidorme: I think just `float`. It's more compact and it is also easier to use the value directly in checking, if we wanted to use the variables system to type check inputs.

I don't think we'd be loading functional groups by this method - if we do implement a `read_csv` for dataframes, then the fields should be loaded into `Data` as individual `DataArrays`. Anything dict/toml like would come in by a bespoke route or config.

virtual_ecosystem/models/abiotic/abiotic_model.py
davidorme: Yup - that sounds right. We're centralising the axis requirements, rather than allowing them to potentially clash between models, so all we need do in the model is provide the names of the variables needed.

vgro:
> A step in the right direction, but there are a few things to change. You have A LOT of variables!!!

Thanks for the feedback.
I agree that I have a lot of variables, however, they do cover three different models ;-)
I had not delete the ones that are in other models from this list as I thought this will all be merged in the end. I did that now that for clarification.
I think a lot of the variables I listed will not be interesting for anyone other than me to actually look at in the end. It is very convenient to store them in the data object to make them accessible for different models and functions, but I am happy to adopt an alternative approach that takes the weight off the data object and variable system.

@davidorme  I am using "topofcanopy_radiation" as input which I think you call "shortwave_in"? And I use "canopy_absorption" for what you call "layer_absorbed_irradiation". Do you have any reasons for picking one or the other, like pyrealm dependencies? Then I'm happy to rename it in my models. I expect "stomatal_conductance" as an input from the plant model, is that what your outputting?

davidorme:
@vgro - I think it's easier if you don't delete the ones that aren't yours. Sorry! Otherwise we'll be in a position of having multiple files to remerge - that would be fine if we didn't share variables but we'll get dupes!

davidorme:
I'm happy with those name changes!

vgro:
> @vgro - I think it's easier if you don't delete the ones that aren't yours. Sorry! Otherwise we'll be in a position of having multiple files to remerge - that would be fine if we didn't share variables but we'll get dupes!

haha ok, no problem, I'll add them back in

davidorme:
Super - thanks. I might work directly on this branch, if that's okay with you - easier to see what has been fixed up already!

vgro:
@davidorme all the variables that you added are already in the toml file but the name is different.
`air_conductivity` is called `air_heat_conductivity`, I fixed that in #430 but it is probably not merged in this branch.
The accumulated fluxes are call `surface_flux_accumulated` and `subsurface_flow_accumulated`, did you take the names from the docstring? Might be misleading so maybe it is better to change the docstring.

davidorme:
@vgro - they all got picked up from running them through the system to discover all variable usage by models. I think they're all just instances where the model variable lines need deleting or changing?

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/86656f7a5a32490ff33000bb563801ed896891ad/virtual_ecosystem/models/abiotic/abiotic_model.py#L66

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/86656f7a5a32490ff33000bb563801ed896891ad/virtual_ecosystem/models/hydrology/hydrology_model.py#L112-L113

Should we just update the names in those lines and then delete the variables I've added to `data_variables.toml`.

vgro:
> @vgro - they all got picked up from running them through the system to discover all variable usage by models. I think they're all just instances where the model variable lines need deleting or changing?
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/86656f7a5a32490ff33000bb563801ed896891ad/virtual_ecosystem/models/abiotic/abiotic_model.py#L66
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/86656f7a5a32490ff33000bb563801ed896891ad/virtual_ecosystem/models/hydrology/hydrology_model.py#L112-L113
> 
> Should we just update the names in those lines and then delete the variables I've added to `data_variables.toml`?

Like I said, I have already renamed the `air_conductivity` to `air_heat_conductivity`in #430, I don't know why this still shows the old name. For the accumulated fluxes, yes, name change is probably best. The names are correct in `vars_updated`. I can do that later or you can do it if you are still making changes, I'm on the stats at the moment

virtual_ecosystem/core/base_model.py
```diff
@@ -285,70 +289,47 @@ def _check_model_name(cls, model_name: str) -> str:
...
+        Creating an instance of the BaseModel class requires that several variable
+        attributes are set. Each of these provides a list of variable names that are
+        required or updated by the model at various points. This method is used to
+        validate the structure of the new instance and ensure the resulting model
+        structure is consistent.
 
-        TODO: Remove. Validation is performed at runtime in the variables side.
+        TODO: Diego - this is possibly paranoid. It no longer checks anything about the
+            variables (which is the job of the variables system) but is basically a
+            check on creating the model correctly.
```
dalonsoa, line 308: A little bit, but it's harmless and is only run once per model, so totally fine. 

```suggestion
```

dalonsoa (approved):
I think all looks good to me. This includes the changes to all models?

davidorme:
> I think all looks good to me. This includes the changes to all models?

It does include all models. Tests all now pass - I have updated all the code and tests to handle the simplified `required_init_vars` structure and added a couple of small bits. 

The docs fail for relevant reasons - the docs now include a walk through of the model running on the example data. That is using a wider set of models than are included in `test_ve_run` and that has revealed some variables issues - the "initialised by update" variables that we've mentioned elsewhere. So - those need resolving still. 

I think we probably still merge this down to have one branch on this issue again and tackle that problem separately?

dalonsoa:
Yes, I think that's the best way forward.

PR #433, opened 2024-06-05
Update QA tool chain and add Python 3.12

### LINKED ISSUE #432: Update QA tool chain and add Python 3.12

We really should be running this on Python 3.12, as the most recent active version. We currently can't because `pytest-flake8` is not compatible, but having `flake8` and `mypy` in `pytest` is overkill anyway - we enforce that with `pre-commit`.
* Generally, moving to `ruff` also greatly simplifies our QA chain and is miles faster.

A grab bag of QA improvements:

* Add Python 3.12 to `pyproject.toml`
* Simplify the `pytest` config to ditch `flake8` and `mypy`
* Simplify the `pre-commit` config to replace `isort`, `flake8`, `black`, `pyupgrade` etc. with `ruff`.

We could leave things as they are but this is going to be faster and more robust.

davidorme:
# Description

This PR:

* Simplifies the `pytest` config to remove `mypy` and `flake8`: runs faster, less cruft.
* Replaces `flake8`, `isort`, `pyupgrade`, `black` in `pre-commit` with `ruff`: runs faster, less cruft
* Removes `setup.cfg` - the new QA chain can all be configured using `pyproject.toml`.
* Updates VSCode extension recommendations and settings to match
* Fixes a bunch of minor formatting and linting issues that `ruff` spots but `flake8` doesn't.

This PR is initially about checking it all works across platforms and rolling out adoption of the new QA chain. Note that the improved runtime applies to local work _and_ the runner speeds for GA workflows.

This PR **does not**:

* Update the documentation to match this practice - there is a separate plan to completely overhaul that, which would also tackle these changes. 

Fixes #432

## Type of change

## Key checklist

## Further checks


davidorme:
@alexdewar and @dalonsoa Can you take an initial look at this and see if it fits with your experience of best practice in migrating away from `flake8` etc to `ruff`.

pyproject.toml
```diff
@@ -84,3 +79,46 @@ sphinxcontrib-mermaid = "^0.9.2"
...
+# Enable all `pydocstyle` rules, limiting to those that adhere to the
+# Google convention via `convention = "google"`, below.
+select = [
+  # "B", # flake8-bugbear
+  # "SIM", # flake8-simplify
+  "E", # pycodestyle
+  "F", # pyflakes
+  "D", # pydocstyle
+  "I", # isort
+  "UP", # pyupgrade
```
alexdewar, line 107: I also quite like the `"RUF"` checks, which are a few `ruff`-specific ones.

pyproject.toml
```diff
@@ -84,3 +79,46 @@ sphinxcontrib-mermaid = "^0.9.2"
...
+]
+
+# On top of the Google convention, disable:
+ignore = [
+  "D202", # Blank line after docstring is ok
+  "D107", # Location of __init__ docstring in class not __init__"
+  "F401", # module imported but unused
```
alexdewar, line 114: Are you sure you want to ignore this warning globally? I find it quite convenient to have `ruff` point out when I have unused imports (and automatically remove them). I'd just suppress it in the few places where you're doing it on purpose, if I were you.

.pre-commit-config.yaml
```diff
@@ -4,20 +4,11 @@ repos:
...
-  - repo: https://github.com/psf/black
-    rev: "24.4.2"
-    hooks:
-      - id: black
-  - repo: https://github.com/PyCQA/flake8
-    rev: 7.0.0
-    hooks:
-      - id: flake8
-        additional_dependencies: [flake8-docstrings]
+      - id: ruff        #  Run the linter.
```
alexdewar, line 10: Do you want `ruff` to try to auto-fix problems? If so:

```suggestion
      - id: ruff        #  Run the linter.
        args: [--fix, --exit-non-zero-on-fix]
```

virtual_ecosystem/models/litter/litter_model.py
```diff
@@ -189,6 +189,7 @@ def update(self, time_index: int, **kwargs: Any) -> None:
 
         Args:
             time_index: The index representing the current time step in the data object.
+            kwargs: Further arguments to the update method.
```
alexdewar, line 192: I think I've had situations where putting `kwargs` and not `**kwargs` in a docstring has broken things, but that may have been with `mkdocs` rather than `sphinx`.

alexdewar (approved):
I've made a few suggestions, but LGTM.

I've found using `ruff` to be quite a pleasant experience. We've updated all of our active projects to using it, away from `flake8` et al.

pyproject.toml
davidorme:
```
Found 97 errors.
```

But they all seem sensible, so will adopt it.

pyproject.toml
davidorme: I think that's a 'legacy' setting.

PR #435, opened 2024-06-06
434 soil temperature not returned correctly

### LINKED ISSUE #434: Soil temperature not returned correctly
The soil temperature is not returned as expected by the microclimate submodule. I missed testing for it explicitly and the data in the dummy data replaced the value unexpectedly.

This should be easy to fix with correct assignment of indices.

vgro:
The soil layer heights were missing in the dummy data set with variable layer heights. Therefore, all soil temperatures were nan.

Fixes #434 

## Type of change

## Key checklist

## Further checks


davidorme (approved):
LGTM - great to simplify it into a loop too.

Issue #376, opened 2024-01-28
Move `Grid` instance from `Data` into `CoreComponents`


davidorme:
**Is your feature request related to a problem? Please describe.**
It's a much more natural home for `Grid`, and can then be added cleanly to all models via the `BaseModel`. Passing `CoreComponents` and possibly `Config` to `Data` is then likely part of the plan for fixing old problems like #208.

**Describe the solution you'd like**
Add `Grid` to `CoreComponents` and populate `BaseModel.grid` as a core shared attribute, rather than it being oddly accessed through `Data`.

Issue #437, opened 2024-06-10
Add statistics to abiotic variables


vgro:
As discussed in #425, we need the `abiotic`, `abiotic_simple`, and `hydrology` model to return statistics of microclimate variables to represent the variability over a monthly time step.

We decided to introduce an additional axis `climate_stats` to add mean, min, and max to the requested variables (air temperature, soil temperature, vpd, soil mositure).

We considered alternatively to add variables for each statistic, but that would blow up the code unnecessarily. With the setup of numpy arrays, we can hopefully run all functions smoothly for the additional dimension without extra looping or similar

vgro:
Paused until #441  and #388  are sorted and merged in develop.

PR #441, opened 2024-06-12
Amending vertical layer structure

### LINKED ISSUE #421: Additional information in LayerStructure
There are a number of static properties of our 3D structure that we use to create arrays, select indices etc, for example the `topsoil_layer_index`, number of layers above ground, and so on.

I suggest that we add these static properties to the `LayerStructure` so they are easier accessible and the same across all models.

davidorme:
# Description

This is a work in progress PR implementing the removal of the explicit subcanopy layer from the vertical layer structure.

The current progress is

- [X] The LayerStructure no longer has the `subcanopy` layer. The `layer_roles` attribute is now shorter by 1.
- [X] The `LayerStructure.from_template(name)` method is added, returning a fresh blank named DataArray with `layer_roles` and `cell_id` dimensions.
- [X] There are now two dictionaries `LayerStructure.role_indices` and `LayerStructure.role_indices_bool`, depending on whether you want integer indices or boolean indices onto the layer role dimension. Those are keyed by `above`, `canopy`, `surface`, `soil`. 
- [X]  Implement the specific `topsoil` indexing option - the soil layer roles are now explicitly split into a single `topsoil` role and then additional `subsoil` layers but an additional `all_soil` index is provided to refer to all soil layers collectively.
- [x] Add indices to identify the subset of soil layers that are biologically active along with the thicknesses of the layers and the biologically active thickness of each soil layer.

## New location for Grid

- [X] As a side effect, the `Grid` object now also exists within `CoreComponents` and is inherited by models automatically as `Model.grid` via the `BaseModel.__init__` mechanism. As noted in #376, this is saner place to put it now that `CoreComponents` exists, and the `LayerStructure` component needs some minimal information from the grid.

    This PR **will not remove `Data.grid`** so we will have two copies of that for a while, but this will be the long term home and it is easier to work with two copies than try and do this all in one go. At some point a separate PR will fix the other half of #376 and remove the dupe from `Data`.

## Things to do:

Fix everything that this breaks:

- [ ]  example data and tests - mainly the removal of the subcanopy layer.
- [ ] adoption of the new API across model code. This doesn't strictly have to be done here, but it would be good to pick off as much as possible.

These fixes can be by individual PR **onto this branch** to pick off issues until we have a passing PR to merge down.

Fixes #421 (issue)

## Type of change

- [X] Optimization (back-end change that speeds up the code)

## Key checklist

## Further checks


davidorme:
First review of WIP

@jacobcook1995  and @vgro - Do you need any of this implementation changing or extending?
@alexdewar Anything obviously mad?

jacobcook1995:
Yeah looks sensible to me, I'm not planning to continue using `topsoil_index` in the long run (I don't know if @vgro is though), so from my perspective I'm happy with any implementation that doesn't break the current tests

tests/conftest.py
```diff
@@ -270,177 +211,86 @@ def dummy_carbon_data(fixture_core_components):
...
-    grid = Grid(
-        grid_type="square",
-        cell_nx=3,
-        cell_ny=1,
-        cell_area=3,
-        xoff=0,
-        yoff=0,
-    )
-    data = Data(grid)
+    data = Data(fixture_core_components.grid)
```
vgro, line 288: Just to double-check, did you change the number of cell_ids for all my tests from 3 to 4?

tests/conftest.py
```diff
@@ -455,280 +305,125 @@ def dummy_climate_data(fixture_core_components):
...
-    )
-    data["attenuation_coefficient"] = DataArray(
-        np.broadcast_to(attenuation_coefficient, (3, 15)).T,
-        dims=["layers", "cell_id"],
-        name="attenuation_coefficient",
-        coords=full_coordinates,
-    )
-    data["mean_mixing_length"] = DataArray(np.repeat(1.3, 3), dims="cell_id")
+    data["absorbed_radiation"] = from_template()
+    data["absorbed_radiation"][[1, 2, 3]] = 10.0
```
vgro, line 376: I think this variable is no longer used, we have the `canopy_absorption` instead

vgro:
Looks sensible to me, also good job tidying up the conftest :-)
One concern is that I have the parallel branch for including stats which and I don't know in which order we want to merge this in develop to make the least mess ?

tests/conftest.py
davidorme: I did. 

I'm not happy about it and I'm still trying to decide if it is the right thing to do. 

Currently we have models that use 4 cells and models that use 3. The underlying issue is that we have a big pile of fixtures that are used more or less as a pick and mix stall. That's sorta ok, but we end up hand coding indices all over the place (3 v. 4 for example) to make the selection work together. With the more integrated core components, that's going to lead to more inconsistencies in tests and stuff basically not working because half the test uses one setup and  the other half uses something else.

Ultimately, if everything uses `fixture_core_components` and a small number of extra data frames, then updating tests etc. becomes much cleaner. 

It's a massive pain in the arse: I _think_ it is the right thing to do, but this PR went down a rabbit hole on Friday and I just need to figure out what to do.

vgro: I see and I understand. It's become a bit of a mess with the fixtures and conftest file in general.
I totally agree that this is something that needs sorting, but in the interest of keeping PRs manageable I suggest to focus on what we discussed, add more features to the layer structure and get rid of the subcanopy layer, make that work, merge in develop and branch off again to do the next step, like a PR that just focuses on tidying up the test fixtures. Would you agree?

vgro: This also links to the overall comment I made about developing other features on parallel branches, which is becoming an absolute nightmare if we make too many changes here at once.

I will pause the stats axis for now #437 (it's not required for min viable product), it just makes it more confusing with the layer structure and the variables that Diego is working on. Once these things are in place it should be much easier to add an axis and update the relevant locations.

davidorme: > I suggest to focus on what we discussed, add more features to the layer structure and get rid of the subcanopy layer, make that work, merge in develop and branch off again to do the next step

I think that's probably right - let me run with this for today - I've got a bunch of fixes to get the tests working that I'll probably put into a sub PR.

davidorme: > get rid of the subcanopy layer

Just to note - this is the tricky bit. The tests all need updating to remove this, so cleaning them to make this process cleaner is kinda built in.

vgro: I agree, the same thing happened to me with the climate_stats, which is why I think it makes more sense for me to  be here with you and respond where needed or contribute rather than making different changes on a different branch.

alexdewar (commented):
I think I've more or less lost track of the inner workings of the code at this point, so I don't think I can advise whether or not this is a good idea, though it sounds sensible on the face of it. I've gone through and made some fairly surface-level suggestions.

This PR is a little on the long side, as @vgro says. Not sure how easy it would have been to break this work up, but in general I think little PRs often is better than big PRs occasionally, especially when the work touches a lot of different areas of code.

davidorme:
> This PR is a little on the long side, as @vgro says. Not sure how easy it would have been to break this work up, but in general I think little PRs often is better than big PRs occasionally, especially when the work touches a lot of different areas of code.

@alexdewar.  I completely agree - the problem here is that one of the core changes we _want_ to make in this PR changes the vertical structure of the simulation so that the dimensions of the data change. That's not something that we can adopt incrementally, but we are trying to isolate the bits for different models in sub-branches. Thanks for the comments!

virtual_ecosystem/core/core_components.py
```diff
@@ -188,21 +230,33 @@ class LayerStructure:
 
     canopy_layers: int = field(init=False)
     """The maximum number of canopy layers."""
-    soil_layers: list[float] = field(init=False)
+    soil_layers: NDArray[np.float_] = field(init=False)
```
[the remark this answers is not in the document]
davidorme: So:

* `mypy` objects to `NDArray[float]` with the message 'Type argument "float" of "NDArray" must be a subtype of "generic".' 
* We probably could go with `NDArray[np.float32]` but we've been kinda inconsistent about typing our float arrays so I went with a generic float 😄. I suspect we could usefully have a systematic clean up of array typing!

virtual_ecosystem/core/core_components.py
```diff
@@ -218,28 +272,118 @@ def __post_init__(self, config: Config) -> None:
...
+            a_min=0,
+            a_max=np.inf,
+        )
+
+        self.role_indices["active_soil"] = self.role_indices["all_soil"][
+            self.soil_layer_active_thickness > 0
+        ]
+        self.role_indices_bool["active_soil"] = np.array(
+            [
+                True if v in self.role_indices["active_soil"] else False
```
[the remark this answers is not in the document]
davidorme: <looks ashamed>

tests/models/litter/conftest.py
```diff
@@ -2,17 +2,18 @@
...
 @pytest.fixture
-def fixture_litter_model(dummy_litter_data):
+def fixture_litter_model(dummy_litter_data, fixture_core_components):
     """Create a litter model fixture based on the dummy litter data."""
 
     from virtual_ecosystem.core.config import Config
     from virtual_ecosystem.core.core_components import CoreComponents
     from virtual_ecosystem.models.litter.litter_model import LitterModel
 
+    fixture_core_components
```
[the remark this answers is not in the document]
davidorme: Yes.

tests/core/conftest.py
```diff
@@ -0,0 +1,95 @@
...
+    data = Data(fixture_square_grid_simple)
+
+    # Create an existing variable to test replacement
+    data["existing_var"] = DataArray([1, 2, 3, 4], dims=("cell_id",))
+
+    return data
+
+
+@pytest.fixture
+def new_axis_validators():
```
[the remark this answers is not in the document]
davidorme: Yeah - invoking this fixture registers these two axis validators, so they become available in tests that use the fixture. It's a bit around the houses.

virtual_ecosystem/core/core_components.py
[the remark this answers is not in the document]
alexdewar: Makes sense. And I think generic `float` is actually probably best here anyway.

davidorme:
@dalonsoa Any idea on the `qa` failure here? The `pre-commit.ci` is running the same thing with the same configuration - wait. Might be python version?

The only other thing I looked at is the `mypy` configuration in `pyproject.toml` specifies the `numpy.testing.mypy-plugin`, but that shouldn't affect the `pre-commit` configuration and that is identical between the two actions.

virtual_ecosystem/core/core_components.py
```diff
@@ -218,33 +272,120 @@ def __post_init__(self, config: Config) -> None:
...
+            for layer_role in ("above", "canopy", "surface", "topsoil", "subsoil")
+        }
+        self.role_indices = {
+            ky: np.nonzero(vl)[0] for ky, vl in self.role_indices_bool.items()
+        }
+
+        # Add the `all_soil` and `atmospheric` indices
+        self.role_indices["all_soil"] = np.concatenate(
+            [self.role_indices["topsoil"], self.role_indices["subsoil"]]
+        )
```
dalonsoa, line 309: Maybe, for consistency with the atmosphere - and because I think it looks like a nicer syntax - I would move this below and do:
```suggestion
        self.role_indices["all_soil"] = np.where(
            self.role_indices_bool["all_soil"]
        )[0]
```

virtual_ecosystem/core/core_components.py
```diff
@@ -218,33 +272,120 @@ def __post_init__(self, config: Config) -> None:
...
+        # actual template.
+        self._array_template: DataArray = DataArray(
+            np.full((self.n_layers, self.n_cells), np.nan),
+            dims=("layers", "cell_id"),
+            coords={
+                "layers": self.layer_indices,
+                "layer_roles": ("layers", self.layer_roles),
+                "cell_id": np.arange(self.n_cells),
+            },
+        )
```
dalonsoa, line 361: Given that you need to deepcopy this array, I don't see much point on keeping a private copy of it. I guess there might be some performance improvement but given that all of these arrays will be created when setting up the simulation and not when running it (I think), the overall gains will be minimal - if any. I would just create this in the `from_template` below, which also avoids risking careless users modifying the template at runtime (that's the paranoid's vision).

virtual_ecosystem/core/core_components.py
```diff
@@ -188,21 +230,33 @@ class LayerStructure:
...
+    """A dictionary providing integer indices of each roles within the vertical layer
+    structure."""
     config: InitVar[Config]
     """A validated model configuration."""
+    n_cells: int
+    """The number of grid cells in the simulation."""
+    max_depth_of_microbial_activity: float
+    """The maximum depth of microbial activity (m)."""
 
     def __post_init__(self, config: Config) -> None:
```
dalonsoa, line 261: This post init method is pretty long and doing a lot of stuff. I suggest breaking it down in smaller chunks for:
- validation
- setting roles and indices
- set microbial information

virtual_ecosystem/core/core_components.py
```diff
@@ -218,33 +272,120 @@ def __post_init__(self, config: Config) -> None:
...
+        self.layer_indices = np.arange(0, self.n_layers)
+
+        # Set up the various indices onto those roles
+        self.role_indices_bool = {
+            layer_role: self.layer_roles == layer_role
+            for layer_role in ("above", "canopy", "surface", "topsoil", "subsoil")
+        }
+        self.role_indices = {
+            ky: np.nonzero(vl)[0] for ky, vl in self.role_indices_bool.items()
+        }
```
dalonsoa, line 304: If you want to go a step further with the syntactic sugar, you could create properties for all of these. It is more code here, but will reduce the verbosity in other parts of the codebase as well as save them from typos as people will not need to use a string to access the `role_indices` dictionaries:

```python
@property 
def above_index(self):
    return self.role_indices["above"]

@property 
def above_bool(self):
    return self.role_indices_bool["above"]
...
```
As I said, this is just syntactic sugar, but I think it will have a positive impact in readability.

dalonsoa (changes requested):
I think it makes sense to make the layer structure part of the core functionality, as well as the `Grid`, as they both underpin much of the rest of the functionality. In the end, they define the 3D physical space where the different models of the simulation will run, so the change makes total sense. 

As the PR is pretty long, I've focused my review in the `core_components`, leaving the ramifications of those changes up to you. I have a couple of suggestions for simplifying things and readability.

virtual_ecosystem/core/core_components.py
davidorme: That's an interesting point, @dalonsoa and something I had parked in the back of my head. I think the different models differ in whether they update an existing DataArray or simply replace it with an updated one. I'm sure we have `from_template()` being used in model `update` methods, so it will be used when running. That _might_ be something we want to avoid doing, or it might be that the `deepcopy` is worth hanging on to.

I think we might want to start leaving a few comments in the code base like `# PERFORMANCE - does deepcopy of the template here provide any real benefit over on the fly creation".

virtual_ecosystem/core/core_components.py
davidorme: I did consider that and I agree this would be more readable. It doesn't work so cleanly with the user-requested aggregate indices over in #452.

tests/core/test_base_model.py
```diff
@@ -274,7 +274,7 @@ class UM(
         assert str(err.value) == exp_msg
 
 
-def test_check_failure_on_missing_methods(data_instance, fixture_core_components):
+def test_check_failure_on_missing_methods(dummy_climate_data, fixture_core_components):
```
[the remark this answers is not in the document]
davidorme: It does still exist - it's a local fixture in the file and presumably locally overrode the `data_instance` that existed in the root `conftest.py`. But, I think its better to keep fixtures _out_ of test files, so will move it and probably rename it.

virtual_ecosystem/core/core_components.py
davidorme: Updated in #451

virtual_ecosystem/core/core_components.py
davidorme: That's true, but there are a lot of attributes, which should be declared and documented in `__init__` rather than other methods. I could simply declare and type them all and then populate them in sub methods, but most of them are populated in one liners. It makes clear sense to break out initialising the indices though.

virtual_ecosystem/core/core_components.py
davidorme: @dalonsoa Do you have any thoughts on whether integer or boolean indexing is likely to be more efficient? We'll be doing it a lot. We don't _need_ both boolean and integer indices, I think, so if one is clearly better we can simplify.

vgro (approved):
LGMT (only checked changes since my last review)

davidorme:
@dalonsoa I think we need you to approve this too after your requested changes. We know it's a halfway house, but we want to be able to move on with other stuff in `develop` while the changes in #452 are adopted.

dalonsoa (approved):
Done!

PR #445, opened 2024-06-18
Fixing Hydrology model to work with LayerStructure

### LINKED ISSUE #444: Fixing Hydrology for #421
This issue is for a PR to align the hydrology model with the changes requested in #421 and being implemented in #441

davidorme:
This is a WIP PR to adapt the hydrology model and tests to use the revised LayerStructure.

@vgro - I haven't merged in #441 yet, so this was pushed *without verify* and there are a bunch of `mypy` complaints. My thinking here was that I wanted you to see _just_ the changes I've made to get a feel for them, before they get a bit hidden when the LayerStructure implementation changes come in.

When I run these changes locally with the changes in #441, nearly all the test pass, but `test_hydrology_model/test_setup` fails and so does `test_hydrology_tools.py::test_setup_hydrology_input_current_timestep`

Fixes #444 (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/hydrology/hydrology_model.py
```diff
@@ -387,8 +379,8 @@ def update(self, time_index: int, **kwargs: Any) -> None:
             time_index=time_index,
             days=days,
             seed=seed,
-            layer_roles=self.layer_structure.layer_roles,
-            soil_layer_thickness=self.soil_layer_thickness,
+            layer_structure=self.layer_structure,
+            soil_layer_thickness=self.soil_layer_thickness_mm,
```
davidorme, line 383: @vgro - the `LayerStructure` provides soil layer thickness in metres, so I've renamed and converted here. It probably makes sense for `LayerStructure.soil_layer_thickness` to switch over to mm. @jacobcook1995 doesn't have strong feelings.

vgro: I think it is fine in meters and convert here, I use it in the soil energy balance in meters, so one of them must be converted and to be consistent with the layer heights, meters works better in the LayerStructure I think

virtual_ecosystem/models/hydrology/hydrology_tools.py
```diff
@@ -107,15 +107,16 @@ def setup_hydrology_input_current_timestep(
...
-    for out_var, in_var in (
-        ("subcanopy_temperature", "air_temperature"),
-        ("subcanopy_humidity", "relative_humidity"),
-        ("subcanopy_wind_speed", "wind_speed"),
-        ("subcanopy_pressure", "atmospheric_pressure"),
-    ):
-        output[out_var] = (
-            data[in_var].isel(layers=layer_roles.index("subcanopy")).to_numpy()
-        )
+    # VIVI: these to go?
```
vgro, line 110: We still need these values, but I will rename them to 'surface' for now until we have sorted the interpolation to 1.5m

tests/models/hydrology/test_above_ground.py
```diff
@@ -60,21 +60,25 @@ def test_calculate_soil_evaporation(dens_air, latvap):
     )
 
 
-def test_find_lowest_neighbour(dummy_climate_data):
+def test_find_lowest_neighbour(fixture_core_components, dummy_climate_data):
     """Test finding lowest neighbours."""
 
-    from math import sqrt
+    # FIXME: At the moment this is being tested on a 2x2 grid with these elevations and
```
vgro, line 66: Sorry, I don't understand this commend, what would you suggest to change?

tests/models/hydrology/test_hydrology_model.py
```diff
@@ -99,7 +100,8 @@ def test_hydrology_model_initialization(
         assert repr(model) == "HydrologyModel(update_interval=1209600 seconds)"
         assert model.initial_soil_moisture == ini_soil_moisture
         assert model.initial_groundwater_saturation == ini_groundwater_sat
-        assert model.drainage_map == {0: [], 1: [0], 2: [1, 2]}
+        # VIVI - not sure about the value below.
```
vgro, line 103: This will depend on the elevation input

tests/models/hydrology/test_above_ground.py
davidorme: Yeah - bit obscure 😄.

This doesn't need fixing now. I'm not convinced that the function is doing what we want right now - but might be wrong. So at some point (possibly when this gets ported to `Grid`?) we should review these test values and the implementation.

tests/models/hydrology/test_hydrology_model.py
davidorme: Yup. Which is that grid of `[[200,100],[10,10]]`. We probably just leave this - it passes tests - and then generate some more expansive tests of the drainage mapping system elsewhere. If that changes the implementation, we just come back and fix this value.

virtual_ecosystem/models/hydrology/hydrology_tools.py
```diff
@@ -126,14 +127,15 @@ def setup_hydrology_input_current_timestep(
     # Select soil variables
     # FIXME - there's an implicit axis order built into these calculations (vertical
     #         profile is axis 0) that needs fixing.
+    # VIVI: I think this is implicit order is now baked in?
     output["top_soil_moisture_capacity_mm"] = (
```
vgro, line 131: I think it is, we are not going to revert the order of layers?

davidorme: I can't see why we would want to! As long as these functions are always dealing with vertical layer x grid cells, then assuming the axis order seems easier and is probably faster.

virtual_ecosystem/models/hydrology/hydrology_model.py
```diff
@@ -135,7 +132,7 @@ def __init__(
         """Set of constants for the hydrology model"""
         self.core_constants = core_components.core_constants
         """Set of core constants for the hydrology model"""
-        self.data.grid.set_neighbours(distance=sqrt(self.data.grid.cell_area))
+        self.grid.set_neighbours(distance=sqrt(self.grid.cell_area))
```
davidorme, line 135: Was going to do this in another PR, but ahead of the curve 😄

vgro: upsi, I hope it won't break anything

davidorme:
@vgro With the last commit,  all of the Hydrology tests now pass. You can see the specific alterations here (since your changes and excluding the merge up of the base branch #421):

https://github.com/ImperialCollegeLondon/virtual_ecosystem/compare/0abf176...444-fixing-hydrology-for-421

Main things:

* The test of which variables are returned in `test_setup_hydrology_input_current_timestep` was not working as expected. I've updated the test so it checks the returned and expected variable names match, but had to edit the expected list. Are those correct?

* In `hydrology_tools.setup_hydrology_input_current_timestep`, the new code was returning 2D arrays for the surface layers, which was inconsistent with some of the other surface layer variables. The example I found was in the section below where the two variables being averaged had different dimensions - the `river_discharge_rate` entries were `(n_cells,)` but the `aerodynamic_resistance_surface` entries came via `wind_speed` and the surface variables and were `(1, n_cells)`.

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/15268a5c8aaa66b36135602dcf6c7a47aeeaed1a/virtual_ecosystem/models/hydrology/hydrology_model.py#L660-L665

* I've explicitly reshaped `HydrologyModel.soil_layer_thickness` to be `(n_soil_layers, n_cells)` - I think all of the use cases needed it to be at least rotated to a vertical array and it seemed sensible just to make it explicitly 2D and broadcast to cells. One outcome of that is that I've simplified the calculation of `soil_moisture` to only handle float values for `init_soil_moisture` - at present the model `__init__` can only take a float value here from config, so that seemed sane. We may want to make `initial_soil_moisture` a data variable at some point, and then things change.

## Problems!

The the test values in `test_hydrology_model.py::test_setup` have changed. I've substituted the reported values into the tests to check the testing _mechanics_ all work but don't have insight into why these values have changed. 

I've kept the old values as comments from comparison, if you can review these values and see where I've wrecked it, that would be great. Then we're done with this branch, I think!

tests/conftest.py
```diff
@@ -190,10 +104,9 @@ def fixture_config():
 
         [core.layers]
         canopy_layers = 10
-        soil_layers = [-0.25, -1.0]
+        soil_layers = [-0.5, -1.0]
         above_canopy_height_offset = 2.0
```
vgro, line 108: That looks right, as long as it does not break any other tests.

tests/models/hydrology/test_hydrology_model.py
```diff
@@ -196,181 +197,113 @@ def test_generate_hydrology_model(
...
+            config=fixture_config,
         )
 
         model.setup()
 
+        # Test soil moisture
+
+        exp_soilm_setup = fixture_core_components.layer_structure.from_template()
+        soil_indices = fixture_core_components.layer_structure.role_indices["all_soil"]
+        exp_soilm_setup[soil_indices] = np.array([[250], [250]])  # [[125], [375]]
```
vgro, line 243: Hmm, something is not quite right here. The values should be different between topsoil and subsoil. I'll see if I can find out where this goes wrong.

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
+        # "subcanopy_humidity",
+        # "subcanopy_pressure",
+        # "subcanopy_wind_speed",
+        "surface_temperature",
+        "surface_humidity",
+        "surface_pressure",
+        "surface_wind_speed",
+        "leaf_area_index_sum",
+        "current_evapotranspiration",
+        # "soil_layer_heights",
```
vgro, line 46:
```suggestion

```

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
+        # "soil_layer_thickness",
```
vgro, line 47:
```suggestion

```

tests/models/hydrology/test_hydrology_tools.py
vgro: can be deleted

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
+        "top_soil_moisture_capacity_mm",
+        "top_soil_moisture_residual_mm",
+        # "soil_moisture_true",
```
vgro, line 50:
```suggestion

```

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
+        "previous_accumulated_runoff",
+        "previous_subsurface_flow_accumulated",
         "groundwater_storage",
+        "soil_moisture_mm",
```
vgro, line 54: This variable should be returned as 'soil_moisture', not soil_moisture_mm'

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
         np.sum(result["current_precipitation"], axis=1),
         (dummy_climate_data["precipitation"].isel(time_index=0)).to_numpy(),
     )
     np.testing.assert_allclose(
-        result["subcanopy_temperature"], dummy_climate_data["air_temperature"][11]
+        result["surface_temperature"],
+        dummy_climate_data["air_temperature"]
+        .isel(layers=fixture_core_components.layer_structure.role_indices["surface"])
+        .squeeze(),
     )
```
vgro, line 71: This seems a bit odd to me, is there no more direct way to select a layer?

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
+        .isel(time_index=[0])
+        .squeeze()
+        .to_numpy()
+        .T,
     )
     np.testing.assert_allclose(
-        result["soil_moisture_mm"], DataArray(np.full((2, 3), 50.0))
+        result["soil_moisture_mm"],
+        DataArray(np.tile([[5], [500]], fixture_core_components.grid.n_cells)),
     )
```
vgro, line 91: Again, this should not be returned, could this have to do with writing directly in the output dict without renaming the variable?

tests/models/hydrology/test_hydrology_tools.py
```diff
@@ -31,80 +16,107 @@ def test_setup_hydrology_input_current_timestep(
...
+            layer_structure.soil_layer_thickness[:, None] * 1000,
+            fixture_core_components.grid.n_cells,
+        ),
+        initial_soil_moisture=init_soilm,
+    )
+    # The fixture is configured with soil layers [-0.25, -1.0]
+    exp_result = DataArray(np.broadcast_to(expected, (2, 4)))
+    np.testing.assert_allclose(
+        result[layer_structure.role_indices["all_soil"]], exp_result
     )
```
vgro, line 122: This is not matching up the dimensions correctly, the values should differ. Why can we not do `soil_layer_thickness * 1000`?

tests/models/hydrology/test_hydrology_model.py
vgro: Ah no, silly me, I was looking at the layer heights instead of the layer thickness. 1-0.5=0.5, so the values should be the same, and should be 250.
You can delete the values in comments

virtual_ecosystem/models/hydrology/hydrology_tools.py
```diff
@@ -107,14 +89,18 @@ def setup_hydrology_input_current_timestep(
         seed=seed,
     )
 
+    # named 'surface_...' for now TODO needs to be replaced with 2m above ground
+    # VIVI - these variables are explicitly 1 dimensional for a single layer. They come
+    # out of the subset as 2D (1, n_cells) and so I've squeezed them.
+    # squeezed them
     for out_var, in_var in (
```
vgro, line 96: That's fine with me, I think I had misunderstood how exactly the new layer_structure works

virtual_ecosystem/models/hydrology/hydrology_tools.py
```diff
@@ -107,14 +89,18 @@ def setup_hydrology_input_current_timestep(
         seed=seed,
     )
 
+    # named 'surface_...' for now TODO needs to be replaced with 2m above ground
+    # VIVI - these variables are explicitly 1 dimensional for a single layer. They come
```
vgro, line 93:
```suggestion
    # These variables are explicitly 1 dimensional for a single layer. They come
```

vgro: I think it is still useful to have this info, but does not need my name :-)

vgro:
> @vgro With the last commit, all of the Hydrology tests now pass. You can see the specific alterations here (since your changes and excluding the merge up of the base branch #421):
> 
> [0abf176...444-fixing-hydrology-for-421](https://github.com/ImperialCollegeLondon/virtual_ecosystem/compare/0abf176...444-fixing-hydrology-for-421)
> 
> Main things:
> 
>     * The test of which variables are returned in `test_setup_hydrology_input_current_timestep` was not working as expected. I've updated the test so it checks the returned and expected variable names match, but had to edit the expected list. Are those correct?
> 
>     * In `hydrology_tools.setup_hydrology_input_current_timestep`, the new code was returning 2D arrays for the surface layers, which was inconsistent with some of the other surface layer variables. The example I found was in the section below where the two variables being averaged had different dimensions - the `river_discharge_rate` entries were `(n_cells,)` but the `aerodynamic_resistance_surface` entries came via `wind_speed` and the surface variables and were `(1, n_cells)`. I've now squeezed these variables so they are `(n_cells,)`.
> 
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/15268a5c8aaa66b36135602dcf6c7a47aeeaed1a/virtual_ecosystem/models/hydrology/hydrology_model.py#L660-L665
> 
>     * I've explicitly reshaped `HydrologyModel.soil_layer_thickness_mm` to be `(n_soil_layers, n_cells)` - I think all of the use cases needed it to be at least rotated to a vertical array and it seemed sensible just to make it explicitly 2D and broadcast to cells. One outcome of that is that I've simplified the calculation of `soil_moisture` to only handle float values for `init_soil_moisture` - at present the model `__init__` can only take a float value here from config, so that seemed sane. We may want to make `initial_soil_moisture` a data variable at some point, and then things change.
> 
> 
> ## ~Problems!~
> 
> ~The the test values in `test_hydrology_model.py::test_setup` have changed. I've substituted the reported values into the tests to check the testing _mechanics_ all work but don't have insight into why these values have changed.~
> 
> ~I've kept the old values as comments from comparison, if you can review these values and see where I've wrecked it, that would be great. Then we're done with this branch, I think!~
> 
> The 'problem' is that the `fixture_core_components.layer_structure.soil_layers` is `[-0.25, -1. ]` is inconsistent with the depths set in `dummy_climate_data` at `[-0.5, -1.0]`. If I match those up, it works with the original values with minor tweaks for the new 4th cell.

LGTM, few minor points, then you can merge it:
- the variable names are not all correct, it should be `soil_moisture` not `soil_moisture_mm`, see inline comment. The layer related variables can be deleted.
- The 2D arrays for surface values are meant to be 1D and I'm happy with the squeeze option if that is the best way to deal with it.
- The idea behind arrays for `initial_soil_moisture` was that different grid cells might have different initial values, for example if we put in an external dataset. But this is not happening anytime soon, and like you said the it's not setup, so limiting this to float is fine.

- New returned values in `test_hydrology_model.py::test_setup` are correct, you can delete the values in comments.

tests/conftest.py
davidorme: It doesn't seem to and I think it's shared across your models.

tests/models/hydrology/test_hydrology_model.py
davidorme: Phew - thought I'd radically misunderstood the code 😄

tests/models/hydrology/test_hydrology_tools.py
davidorme: Yup - good point.

tests/models/hydrology/test_hydrology_tools.py
davidorme: I don't think we can - this is checking the soil moisture.

virtual_ecosystem/models/hydrology/hydrology_tools.py
davidorme: All of the indices are numpy arrays, which leads to 2D slices. That's not obvious and needs documenting better. This can be done more cleanly by converting the index to a scalar where the index is know to be to a single layer.

tests/models/hydrology/test_hydrology_tools.py
davidorme: Is this right? The `setup_hydrology_input_current_timestep` function is called in `HydrologyModel.update` as:

```python
hydro_input = hydrology_tools.setup_hydrology_input_current_timestep(
```

and there are then lots of references to `hydro_input['soil_moisture_mm']` in `hydrology_model.py`.

vgro: yes, this is not the soil moisture that goes in the data object, it is the daily soil moisture. The naming comes from when we stored volumetric soil moisture in the data object and I had to convert it in this step. I could rename it to daily, but that only works if we stick with daily timesteps

vgro: The initial soil moisture in the configuration is still in relative water content which is adjusted to each soil layer thickness, so there is a function 'initialise_soil_moisture_mm` which is meaningful

vgro: I can stick that on my todo list

davidorme:
Jeepers - had the base for the merge set to `develop` and nearly bypassed protections to merge it into that.

Right. Anyway,  all the tests for the `hydrology` model now pass locally, so this branch has done its job of fixing this up for #421 and I'll merge it down.

PR #448, opened 2024-06-18
LayerStructure updates for abiotic_simple

### LINKED ISSUE #446: Fixing abiotic_simple for #441
This issue is for a PR to align the abiotic_simple model with the changes requested in #421 and being implemented in #441

davidorme:
# Description

This PR introduces changes to make the `abiotic_simple` model work with the new LayerStructure in #421. 

@vgro - if you can have a look at the changes to see if anything looks mad (but the tests pass, so hopefully not) then I can merge it down into #421.

The main changes to the actual code are passing `LayerStructure` down the chain rather than `layer_roles` and then using that functionality.

Fixes #446

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -140,13 +128,12 @@ def setup(self) -> None:
...
-        self.data["vapour_pressure_deficit_ref"] = (
-            vapour_pressure_and_deficit["vapour_pressure_deficit"]
-        ).rename("vapour_pressure_deficit_ref")
-
-        self.data["vapour_pressure_ref"] = (
-            vapour_pressure_and_deficit["vapour_pressure"]
-        ).rename("vapour_pressure_ref")
+        self.data["vapour_pressure_deficit_ref"] = vapour_pressure_and_deficit[
+            "vapour_pressure_deficit"
+        ]
```
davidorme, line 130: DataArrays don't need to be named when inserted into a Dataset.

virtual_ecosystem/models/abiotic_simple/microclimate.py
```diff
@@ -301,34 +271,28 @@ def interpolate_soil_temperature(
...
 
-    # set upper and lower bounds and return soil and surface layers, further layers are
-    # added in the 'run' function
-    return DataArray(
-        np.clip(layer_values, lower_bound, upper_bound),
-        coords=interpolation_heights.coords,
-    ).drop_isel(layers=0)
+    # return
+    return_xarray = layer_structure.from_template()
+    return_xarray[layer_structure.role_indices["all_soil"]] = layer_values[1:]
```
davidorme, line 296: It was cleaner to work internally with `numpy` arrays and then insert into the DataArray at the end.

virtual_ecosystem/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -118,19 +115,7 @@ def setup(self) -> None:
...
-            ),
-            dims=["layers", "cell_id"],
-            coords={
-                "layers": np.arange(0, self.layer_structure.n_layers),
-                "layer_roles": ("layers", self.layer_structure.layer_roles),
-                "cell_id": self.data.grid.cell_id,
-            },
-            name="soil_temperature",
-        )
+        self.data["soil_temperature"] = self.layer_structure.from_template()
```
vgro, line 118: wow, this is so elegant :-)

virtual_ecosystem/models/abiotic_simple/microclimate.py
```diff
@@ -101,62 +100,47 @@ def run_microclimate(
...
-        .isel(time_index=time_index)
-        .where(output["air_temperature"].coords["layer_roles"] != "soil")
-        .rename("atmospheric_pressure")
-        .T
+    # VIVI: the previous code implied this should only be filled for atmospheric roles,
+    # but that didn't happen.
+    output["atmospheric_pressure"] = layer_structure.from_template()
+    output["atmospheric_pressure"][:] = data["atmospheric_pressure_ref"].isel(
+        time_index=time_index
     )
```
vgro, line 117: That was always a bit of a hack, I just filled all the non-soil layers with the reference data, even the canopy layers that are nan. So it would be ideal if we could assign the values only to the true above ground layers, do we have that somewhere in the layer structure? Or would it make sense for it to be where layer heights is not nan? Same for atmospheric CO2.

virtual_ecosystem/models/abiotic_simple/microclimate.py
```diff
@@ -101,62 +100,47 @@ def run_microclimate(
...
-    output["atmospheric_co2"] = (
-        data["atmospheric_co2_ref"]
-        .isel(time_index=0)
-        .where(output["air_temperature"].coords["layer_roles"] != "soil")
-        .rename("atmospheric_co2")
-        .T
-    )
+    # VIVI: This had a non varying time index
+    output["atmospheric_co2"] = layer_structure.from_template()
+    output["atmospheric_co2"] = data["atmospheric_co2_ref"].isel(time_index=time_index)
```
vgro, line 122: yeah, I think I assumed constant CO2 when I set this up but you are right, that should be a time series, too.

virtual_ecosystem/models/abiotic_simple/microclimate.py
```diff
@@ -301,34 +271,28 @@ def interpolate_soil_temperature(
...
-    interpolation_heights = xr.concat(
-        [
-            surface_layer,
-            layer_heights[layer_heights.coords["layer_roles"] == "soil"] * -1
-            + surface_layer.values,
-        ],
-        dim="layers",
+    # select surface layer (atmosphere) and generate interpolation heights
+    surface_layer = layer_heights[layer_structure.role_indices["surface"]].to_numpy()
+    soil_depths = layer_heights[layer_structure.role_indices["all_soil"]].to_numpy()
```
vgro, line 276: I think we could use the `layer_structure.soil_layer_thickness` here?

virtual_ecosystem/models/abiotic_simple/microclimate.py
vgro: That's a good point, I noticed that when I tried to set up the climate_stats axis, too 👍

vgro:
@davidorme  thanks for your help with this, it looks really good, a few minor tweaks. I'll finish the hydrology stuff that I started yesterday and then I'll make the changes here, I'll let you know when I need you to review again.

tests/models/abiotic_simple/test_microclimate.py
```diff
@@ -5,12 +5,10 @@
...
 
 
-def test_log_interpolation(
-    dummy_climate_data, fixture_core_components, fixture_empty_array
-):
+def test_log_interpolation(dummy_climate_data, fixture_core_components):
     """Test interpolation for temperature and humidity non-negative."""
 
-    from virtual_ecosystem.models.abiotic_simple.microclimate import log_interpolation
+    from virtual_ecosystem.models.abiotic_simple.microclimate_2 import log_interpolation
```
vgro, line 11: Why microclimate_2?

virtual_ecosystem/models/abiotic_simple/abiotic_simple_model.py
davidorme: It's what you asked for 😄

virtual_ecosystem/models/abiotic_simple/microclimate.py
davidorme: Yeah - was thinking about this. Longer term, I suspect there are more elegant ways to handle the canopy layers, but for the moment I think it makes sense to provide a `LayerStructure.set_filled_canopy()` method, that the `plants` module and testing can use to add a new index layer. That could _technically_ be viewed as Data, but it makes sense to manage that specific bit of information in the context of the other layer indices. 

I was also thinking we could then support the generation of aggregate indices: `LayerStructure.get_indices(['AB', 'FC', 'SF'])` might return `[0,1,2,3,11]`.

tests/models/abiotic_simple/test_microclimate.py
davidorme: Bugger - was testing the new implementation in parallel because I kept screwing it up, and forgot to revert that switch once I'd aligned them!

virtual_ecosystem/models/abiotic_simple/microclimate.py
vgro: I think this would be useful and agree that for now we can just fill the whole array (above ground), that should not break anything. Do you want me to do that or are you already on it?

virtual_ecosystem/models/abiotic_simple/microclimate.py
davidorme: I think this is equally direct.

davidorme:
The `abiotic_simple` tests all pass locally, which is the point of this subbranch, so merging it down into #421.

PR #449, opened 2024-06-18
Initial draft of abiotic model changes for new LayerStructure

### LINKED ISSUE #447: Fixing Abiotic for #421
This issue is for a PR to align the abiotic model with the changes requested in https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/421 and being implemented in https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/441

davidorme:
# Description

These are the updates needed to get the `abiotic` model working with #421.  This is very WIP - I haven't merged in #421 yet,  but these are the current changes. Many tests pass but not all. 

The initial push was made with `--no-verify` to isolate the model and test changes in a single commit, rather than blurring them with the downstream changes.

Fixes #447 (issue)

## Type of change

## Key checklist

## Further checks


tests/models/abiotic/test_conductivities.py
```diff
@@ -23,46 +23,20 @@ def test_initialise_conductivities(dummy_climate_data, fixture_core_components):
...
-    coords = {
-        "layers": np.arange(15),
-        "layer_roles": (
-            "layers",
-            fixture_core_components.layer_structure.layer_roles,
-        ),
-        "cell_id": [0, 1, 2],
-    }
+    exp_air_cond = fixture_core_components.layer_structure.from_template()
+    # VIVI - these test values have changed from:
```
vgro, line 27: yes, that results from a different number of layer heights

vgro: I think the function assumes that the atmosphere layers all have the same depth which is not true, I'll add a TODO to the function.

virtual_ecosystem/models/abiotic/conductivities.py
```diff
@@ -69,6 +71,7 @@ def initialise_conductivities(
     conductivity between the ground and the lowest canopy node.
```
vgro, line 72:
```suggestion
TODO account for variable layer depths
```

tests/models/abiotic/test_soil_energy_balance.py
```diff
@@ -135,38 +137,16 @@ def test_calculate_soil_heat_balance(dummy_climate_data):
...
-        "longwave_emission_soil",
-        "sensible_heat_flux_soil",
-        "latent_heat_flux_soil",
-        "ground_heat_flux",
-    ]
-    variables = [var for var in result if var not in var_list]
-    assert variables
+    # VIVI - I can't get these to work. I think there is a bug in the function, that was
+    # getting the total canopy absorption across all cells, not the per cell sum across
+    # layers, so not sure what the right answer is here.
```
vgro, line 142: You are most likely correct, this might have been overlooked by having the same number of canopy layers and grid cells; the 3x4 is a much better test configuration. What values do you get?
I also have a big todo for the soil energy balance, it is still a bit crappy and inconsistent

tests/models/abiotic/test_wind.py
```diff
@@ -295,113 +283,97 @@ def test_calculate_wind_above_canopy():
...
 
 
-def test_calculate_wind_canopy(dummy_climate_data):
+def test_calculate_wind_canopy(dummy_climate_data_varying_canopy):
     """Test below canopy wind profile."""
 
     from virtual_ecosystem.models.abiotic.wind import calculate_wind_canopy
 
+    # VIVI - different from attenuation coefficient in the data object? Can we switch to
+    # using the fixture?
```
vgro, line 309: yes, I think that is left over from when I sporadically tested the different number of canopy layers, before having a fixture for that

davidorme:
This branch now passes all `abiotic` models tests except the one that has been skipped. 

@vgro has looked over the suggested changes and thinks they all look sane and we both agree that - given the `abiotic` model is still in early development - it makes sense for now simply to update the model code to the new LayerStructure to make it easier to pick up in the future and then move on. So I'm going to merge this down to #441 and iron out the remaining top level (CLI, `ve_run`) implementation issues there.

PR #452, opened 2024-06-21
Initial draft of abiotic model changes for new LayerStructure

### LINKED ISSUE #451: Add dynamic filled canopy index to LayerStructure.

The extended `LayerStructure` object being implemented in #421 and sub-branches #444, #446 and #447 provides indexing across all of the vertical layers in the simulation. However, it only has access to the constant information set up in the model configuration and many models make use of indexing of the **occupied** canopy layers, not just the complete configured set of canopy layers.

It is possible that many of these methods _could_ be rewritten to handle `np.nan` in the input data, but at the moment we use a lot of indexing and insertion to only update layers where the canopy in at least one cell is not `np.nan`. This information needs to be set dynamically from within the `plants` model, but that update process should be exposed as a method so that it can be set statically in testing.

This information _could_ go into `Data`, but adding it to `LayerStructure` keeps the vertical indexing all in one place, and allows the `LayerStructure` code to provide aggregate indices  e.g. above canopy + filled canopy + surface. This should simplify the code and make the layer structure selection much more transparent.

A new `LayerStructure.set_filled_canopy` method, that takes a boolean index of filled canopy layers, and updates all the internal canopy structures.

davidorme:
# Description

This PR implements the updates to the `LayerStructure` object described in #451 and a few more things.

* The class is refactored from a dataclass to a standard class. As more methods were added, it got easier to define attributes within `__init__`.
* The `set_filled_canopy` method now takes an array of canopy heights and updates the `filled_canopy` role indexing and the `lowest_canopy_filled` attribute.
* The use of the `role_indices` attribute has switched over to using the  new`get_indices` method. This allows users to request aggregate indices (`['above', 'filled_canopy', 'surface']`) and get a single array of integer or boolean layer indices. Aggregate indices are cached on request and updated by `set_filled_canopy` so they don't get outdated.

  This will break the current use of `layer_structure.role_indices` in #421 but that should be a fairly straight swap and results in a much clearer interface: the code explicitly states what roles are needed.

Fixes #451

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/core/test_core_components.py
+++ b/tests/core/test_core_components.py
@@ -50,8 +50,8 @@ ALTERNATE_CANOPY = np.array(
         pytest.param(
             "[core]",
             {
-                "canopy_layers": 10,
-                "soil_layers": np.array([-0.25, -1.0]),
+                "n_canopy_layers": 10,
+                "soil_layer_depth": np.array([-0.25, -1.0]),
                 "above_canopy_height_offset": 2.0,
                 "surface_layer_height": 0.1,
                 "n_layers": 14,
@@ -83,8 +83,8 @@ ALTERNATE_CANOPY = np.array(
             max_depth_of_microbial_activity = 0.8
             """,
             {
-                "canopy_layers": 3,
-                "soil_layers": np.array([-0.1, -0.5, -0.9]),
+                "n_canopy_layers": 3,
+                "soil_layer_depth": np.array([-0.1, -0.5, -0.9]),
                 "above_canopy_height_offset": 1.5,
                 "surface_layer_height": 0.2,
                 "n_layers": 8,
@@ -153,6 +153,7 @@ def test_CoreComponents(config, expected_layers, expected_timing, expected_const
                     "all_soil": np.array([12, 13]),
                     "active_soil": np.array([12]),
                     "atmosphere": np.arange(0, 12),
+                    "filled_canopy": np.array([]),
                 },
                 soil_thickness=np.array([0.25, 0.75]),
                 soil_active=np.array([0.25, 0]),
@@ -184,6 +185,7 @@ def test_CoreComponents(config, expected_layers, expected_timing, expected_const
                     "all_soil": np.array([5, 6, 7]),
                     "active_soil": np.array([5, 6]),
                     "atmosphere": np.arange(0, 5),
+                    "filled_canopy": np.array([]),
                 },
                 soil_thickness=np.array([0.1, 0.4, 0.4]),
                 soil_active=np.array([0.1, 0.15, 0]),
@@ -217,6 +219,7 @@ def test_CoreComponents(config, expected_layers, expected_timing, expected_const
                     "all_soil": np.arange(5, 14),
                     "active_soil": np.array([5, 6, 7, 8, 9]),
                     "atmosphere": np.arange(0, 5),
+                    "filled_canopy": np.array([]),
                 },
                 soil_thickness=np.repeat(0.1, 9),
                 soil_active=np.array([0.1, 0.1, 0.1, 0.1, 0.05, 0, 0, 0, 0]),
@@ -297,6 +300,9 @@ def test_LayerStructure_init(
             layer_structure.soil_layer_active_thickness, expected_values["soil_active"]
         )
 
+        assert np.all(
+            np.equal(np.isnan(layer_structure.lowest_canopy_filled), np.repeat(True, 9))
+        )
         # Check the index dictionaries - convert expected keys to frozendicts
         exp_indices = {
             frozenset([ky]): vl for ky, vl in expected_values["layer_indices"].items()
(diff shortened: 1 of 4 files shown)


davidorme:
@alexdewar A bit difficult to review this stuff because it's all such a big jigsaw of things being assembled, but all of the changes we're working on are based around the upgraded `LayerStructure` class in `virtual_ecosystem.core.core_components`. That is a centralised tool, shared among all models for communicating and coordinating information about the vertical layering of the model. 

It's been a bit of a two stage process - the class was updated and then we've identified how best to build out the class further in rolling out those updates. I think this is where the functionality stops for now, though, and since its so central, it would be really useful to get a focussed review of that element as part of this PR.

tests/core/test_core_components.py
```diff
@@ -288,36 +293,42 @@ def test_LayerStructure(
...
-        )
-        assert np.allclose(
-            layer_structure.soil_layer_active_thickness, expected_values["soil_active"]
-        )
 
         # Check the from_template data array
         template = layer_structure.from_template("a_variable")
         assert isinstance(template, DataArray)
-        assert template.shape == (layer_structure.n_layers, layer_structure.n_cells)
+        assert template.shape == (layer_structure.n_layers, layer_structure._n_cells)
```
vgro, line 331: Why did you choose this syntax for n_cells and not other variables?

tests/core/test_core_components.py
```diff
@@ -327,9 +338,127 @@ def test_LayerStructure(
...
+        layer_structure.get_indices("filled_canopy", as_bool=False), np.array([])
+    )
+
+    # Set an aggregate index
+    agg_roles = ["above", "filled_canopy", "surface"]
+    agg_index = frozenset(agg_roles)
+    agg_values = layer_structure.get_indices(agg_roles, as_bool=False)
+
+    assert agg_index in layer_structure._role_indices_int
+    assert np.all(np.equal(agg_values, np.array([0, 11])))
```
vgro, line 433: Why does this not include any canopy indices?

tests/core/test_core_components.py
```diff
@@ -327,9 +338,127 @@ def test_LayerStructure(
...
+    canopy_heights[0:8, 0:8] = np.where(np.flipud(np.tri(8)), 1, np.nan)
+
+    layer_structure.set_filled_canopy(canopy_heights=canopy_heights)
+
+    # Check everything has been set
+    assert np.allclose(
+        layer_structure.lowest_canopy_filled,
+        np.concatenate([np.arange(8, 0, -1), [np.nan]]),
+        equal_nan=True,
+    )
```
vgro, line 448: This test is a bit unclear to me

tests/core/test_core_components.py
davidorme: It's run _before_ `set_filled_canopy` is called and the default is that there is no canopy, so those layers are all empty. The intention is that this information can be set up before any models run, so we don't have to rely on any one model to populate `layer_heights` and then models can pop the the canopy into place.

virtual_ecosystem/core/core_components.py
```diff
@@ -160,23 +160,28 @@ def __post_init__(self, config: Config) -> None:
...
+    configuration settings from the ``[core.layers]`` section.
+
+    * ``above_canopy_height_offset``: the height above the canopy top of the first layer
+      role ``above``, which is used as the measurement height of reference climate data.
+    * ``canopy_layers``: a fixed number of layers with the ``canopy`` role. Not all of
+      these necessarily contain canopy during a simulation as the canopy structure
+      within these layers is dynamic.
+    * ``surface_layer_heights``: the height above ground level of the ground surface
+      atmospheric layer.
+    * ``soil_layers``: the number and depths of soil layers.
```
vgro, line 179: Should we add a comment that this is different from the layer thickness?

virtual_ecosystem/core/core_components.py
```diff
@@ -188,138 +193,155 @@ class LayerStructure:
...
+        4. The ``filled_canopy`` role indicates canopy layers that contain any canopy
+           across all of the grid cells. Canopy layers below these layers do not contain
+           any canopy, and this is initialised to show no filled canopy layers. The
+           ``lowest_canopy_filled`` attribute contains an array showing the lowest
+           filled canopy layer in each grid cell. It contains ``np.nan`` when there is
+           no canopy in a grid cell and is initalised as an array of ``np.nan`` values.
+
+           Both of these are updated by the :meth:`~LayerStructure.set_filled_canopy`
+           method, which is used to dynamically update these layer components during a
+           simulation.
```
vgro, line 228: At what point in the simulation is this updated? Will be relevant for plants/atmosphere

vgro (approved):
LGTM!

tests/core/test_core_components.py
davidorme: Yeah - it does a tonne of work. I've extended the docstring to make it clearer what the test stages are.

virtual_ecosystem/core/core_components.py
davidorme: I've updated the description throughout the docstring to make it clear that ``soil_layers`` are the depths of horizons relative to the surface and thicknesses are individual layer... thicknesses!

virtual_ecosystem/core/core_components.py
davidorme: That's to be determined, but I think I'd drop it into ``plants`` in the ``__init__`` to set the canopy heights from the initial plant community data  and then just re-run it in update as they grow. It won't be any different from the current flow.

dalonsoa (approved):
Ok, I review the wrong PR despite @davidorme being explicit about which one I should review...

Anyway, these changes make total sense if aggregate roles are really a thing. Nevertheless, there are a couple of suggestions in the other PR that you might want to consider.

davidorme:
@dalonsoa Aggregate roles are very much a thing but we _could_ just try and define all of them as named roles (like `all_soil`). I just wanted to avoid future models having to tinker with LayerStructure to work. I did consider adding a `requires_aggregate_index` attribute to the model to set them all up at the start, but didn't want to further complicate the base model structure.

virtual_ecosystem/core/core_components.py
```diff
@@ -340,31 +368,153 @@ def __post_init__(self, config: Config) -> None:
...
+
+        # Get the requested index type
+        if as_bool:
+            idx_dict: dict = self._role_indices_bool
+        else:
+            idx_dict = self._role_indices_int
+
+        # Check if the key is already present
+        if idx_key in idx_dict:
+            return idx_dict[idx_key]
```
dalonsoa, line 467: I comes to my mind that you could do something clever with caching here:

```python
from functools import lru_cache
...

@lru_cache
def get_indices(self, *args: tuple[str], as_bool: bool = True) -> NDArray:
    if len(args) == 0:
        raise ...

    idx_dict = self._role_indices_bool if as_bool else self._role_indices_int
    idx_key = frozenset(args)

    ...
```
What I think is nice about this approach is that:
1. The user does not need to pass explicitly a tuple when more than one role, just writing them one after the other, eg. `get_index("soil", "surface")` (this can also be implemented in your approach).
2. If the function has already been called with exactly the same inputs, the result is cached, so the body is not executed. 
3. You only need to manually store the base indices, as the others would be available via automatic caching - instead of manual caching, which is what you do now within the `_set_aggregate_index`.

Using the builtin caching is up to you, but using `*args` will make your code simpler.

dalonsoa:
> @dalonsoa Aggregate roles are very much a thing but we _could_ just try and define all of them as named roles (like `all_soil`). I just wanted to avoid future models having to tinker with LayerStructure to work. I did consider adding a `requires_aggregate_index` attribute to the model to set them all up at the start, but didn't want to further complicate the base model structure.

No, I think this is the right place, and the way you have implemented it is very neat and flexible. I've just added a small suggestion, but totally optional.

virtual_ecosystem/core/core_components.py
```diff
@@ -534,6 +511,61 @@ def from_template(self, array_name: str | None = None) -> DataArray:
 
         return template_copy
 
+    @property
+    def index_above(self):
```
dalonsoa, line 515: Don't they need type annotation for the output?

dalonsoa (approved):
This looks much cleaner 👍

virtual_ecosystem/core/core_components.py
davidorme: Huh - nothing locally complained.. Good point!

virtual_ecosystem/core/core_components.py
davidorme: That is clever, but we've moved away from the aggregate indices (thanks to your questioning of whether we _really_ needed it 😁 ).

PR #457, opened 2024-06-27
Turning on single-threaded handling in dask

### LINKED ISSUE #354: `xarray.open_mfdataset` can cause problems in `vr_run`

davidorme:
# Description

This PR is to fix segfaults in testing and docs building when `xarray.open_mfdataset` is used. This is due to underlying issues with the netCDF code not being thread safe. We previously turned on `lock=False` in that command, following advice here:

https://github.com/pydata/xarray/issues/3961

And that worked, but the issue has started occurring again. Following the advice in this issue thread:

https://github.com/pydata/xarray/issues/7079

this PR now explicitly sets the `dask` scheduler to use single-threading.

Fixes #354

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/core/data.py
+++ b/virtual_ecosystem/core/data.py
@@ -123,6 +123,7 @@ file.
 from pathlib import Path
 from typing import Any
 
+import dask
 import numpy as np
 from xarray import DataArray, Dataset, open_mfdataset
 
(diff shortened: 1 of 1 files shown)


dalonsoa (approved):
If it works, it looks good!

jacobcook1995 (approved):
LGTM!

PR #459, opened 2024-06-27
Adopt extended LayerStructure functionality

### LINKED ISSUE #451: Add dynamic filled canopy index to LayerStructure.

### LINKED ISSUE #458: Adopt extended LayerStructure functionality

The PR #452 adds more functionality to the `LayerStructure` object and will close #451. That PR has been reviewed and approved but we need to update the code in develop to use that proposed structure.

This issue is to raise a branch to make those changes. We could just merge `develop` directly into #452, but I want to keep the definition of the new API clear of the adoption, in case we need to update the API implementation in response to adopting it. So:

* #452 is being used to define _what_ is being adopted.
* This branch and PR is being used to adopt it.

davidorme:
# Description

This PR is taking the code in `develop` and applying the new API for the LayerStructure from #452, which has been merged down into this.

The main changes are:

* In the models, replacing locally calculated vertical indices with the core `LayerStructure` indices.
* Replacing the `LayerStructure.role_indices['canopy']` approach from the first pass in #441, which has been deprecated in favour of explicit per role attributes (`LayerStructure.index_canopy`).
* In the tests, updating tests to accept the central `fixture_core_components` and then systematically using the `from_template` and index attributes to generate expected and test values, replacing hard coded integer layer indices with layer specific indices from the `fixture_core_components.LayerStructure` instance.

Fixes #451 and #458 

## Type of change

## Key checklist

## Further checks


tests/conftest.py
```diff
@@ -261,43 +271,24 @@ def dummy_carbon_data(fixture_core_components):
...
-    data["soil_temperature"] = fixture_core_components.layer_structure.from_template()
-    data["soil_temperature"].loc[
-        {"layers": fixture_core_components.layer_structure.role_indices["topsoil"]}
-    ] = [35.0, 37.5, 40.0, 25.0]
-    data["soil_temperature"].loc[
-        {"layers": fixture_core_components.layer_structure.role_indices["subsoil"]}
-    ] = [22.5, 22.5, 22.5, 22.5]
-
-    return data
+    lyr_str = fixture_core_components.layer_structure
```
jacobcook1995, line 274: This is a minor thing but could this be `lyr_struct` or something instead? I'm primed to read "str" as "string", so end up spending time wondering what the layer string is used for

virtual_ecosystem/core/core_components.py
```diff
@@ -188,148 +196,214 @@ class LayerStructure:
...
+    n_cells: InitVar[int]
+    """The number of grid cells in the simulation."""
+    max_depth_of_microbial_activity: float
+    """The maximum soil depth of significant microbial activity."""
+
+    # Attributes populated by __post_init__
+    n_canopy_layers: int = field(init=False)
     """The maximum number of canopy layers."""
-    soil_layers: NDArray[np.float32] = field(init=False)
+    soil_layer_depth: NDArray[np.float32] = field(init=False)
```
jacobcook1995, line 291: Would this be better named as `depths` because `depth` makes me think that every layer is of equal thickness

jacobcook1995 (approved):
LGTM (though I only looked at the soil related stuff)!

tests/conftest.py
vgro: why not `layer_structure`?

tests/models/abiotic/test_conductivities.py
```diff
@@ -52,43 +58,42 @@ def test_interpolate_along_heights(dummy_climate_data, fixture_core_components):
...
+    exp_result = exp_result[lyr_str.index_filled_atmosphere]
 
     np.testing.assert_allclose(result, exp_result, rtol=1e-04, atol=1e-04)
 
 
 def test_interpolate_along_heights_arrays(fixture_core_components, dummy_climate_data):
     """Test linear interpolation along heights with arrays of boundary values."""
 
+    # TODO - I don't think this differs from the test above.
+
```
vgro, line 87: The difference is that in the first test, the boundaries are floats, in the second one they are arrays. Could probably drop the first one

tests/models/abiotic/test_energy_balance.py
```diff
@@ -142,47 +138,54 @@ def test_calculate_leaf_and_air_temperature(
...
+    exp_gv = lyr_str.from_template()
+    exp_gv[lyr_str.index_filled_canopy] = np.array([0.203513, 0.202959, 0.202009])[
+        :, None
+    ]
 
-    exp_gv = layer_structure.from_template()
-    gv_vals = [0.203513, 0.202959, 0.202009]
-    exp_gv.T[..., [1, 2, 3]] = gv_vals
+    # TODO - flux layer index does not include above but these tests do - what is best.
+    flux_index = np.logical_or(lyr_str.index_flux_layers, lyr_str.index_above)
```
vgro, line 178: I think the flux above could probably be set to nan in the function, just need to check that this does not cause any errors. So will become irrelevant here, I think the solution is ok for now

tests/models/plants/conftest.py
```diff
@@ -90,35 +90,37 @@ def fixture_canopy_layer_data():
...
+            lyr_str.index_filled_canopy,
         ),
         "layer_fapar": (
             "layer_fapar",
             np.array([0.4, 0.2, 0.1]),
-            np.array([1, 2, 3]),
+            lyr_str.index_filled_canopy,
         ),
         "layer_absorbed_irradiation": (
             "layer_absorbed_irradiation",
```
vgro, line 117: Do you consider the absorption by the topsoil?

tests/models/plants/test_plants_model.py
```diff
@@ -91,14 +91,13 @@ def test_PlantsModel_update_canopy_layers(
...
-        expected = fixture_core_components.layer_structure.from_template()
+        expected = from_template()
         expected[layer_indices] = layer_vals[:, None]
         xarray.testing.assert_allclose(fxt_plants_model.data[layer_name], expected)
 
     # Check layer_absorbed_irradiation is indeed still empty
     xarray.testing.assert_allclose(
-        fxt_plants_model.data["layer_absorbed_irradiation"],
-        fixture_core_components.layer_structure.from_template(),
+        fxt_plants_model.data["layer_absorbed_irradiation"], from_template()
```
vgro, line 100: Did we not rename this variable in the context of the variables PR?

virtual_ecosystem/core/core_components.py
```diff
@@ -162,21 +162,29 @@ def __post_init__(self, config: Config) -> None:
...
+    ordered from above the canopy to the bottom of the soil, that perform different
+    roles in the simulation. The layers are defined using the following five
+    configuration settings from the ``[core.layers]`` section.
+
+    * ``above_canopy_height_offset``: the height above the canopy top of the first layer
+      role ``above``, which is used as the measurement height of reference climate data.
+    * ``canopy_layers``: a fixed number of layers with the ``canopy`` role. Not all of
+      these necessarily contain canopy during a simulation as the canopy structure
+      within these layers is dynamic.
+    * ``surface_layer_heights``: the height above ground level of the ground surface
```
vgro, line 178:
```suggestion
    * ``surface_layer_height``: the height above ground level of the ground surface
```

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -538,67 +532,42 @@ def calculate_leaf_and_air_temperature(
...
-            data["layer_heights"]
-            .where(data["layer_heights"]["layer_roles"] == "surface")
-            .dropna(dim="layers"),
-        ]
-    )
+
+    # TODO - This only uses the index of the _last_ filled layer, which works with the
+    #        current test where the canopy layers are consistent across cells, but will
+    #        break with uneven canopy layers.
+
```
vgro, line 544: This is why I used the function to find the lowest values for each grid cell at some point, that should make it work with varying number of canopy layers. But does not need to be addressed here

vgro: There is probably a better way to do that now that we no longer have the subcanopy layer

vgro (approved):
LGTM, a few comments that are probably for another PR.

tests/conftest.py
davidorme: Honestly, just to keep the line wrapping sane, but I'll have a look 😄

vgro: fair enough, that makes a lot of sense

tests/models/abiotic/test_conductivities.py
davidorme: OK - no. I see the difference.

tests/models/plants/conftest.py
davidorme: It should do, will look.

tests/models/plants/test_plants_model.py
davidorme: We did but they're parallel branches! I'm going to merge and then update the `variables_1` branch.

virtual_ecosystem/models/abiotic/energy_balance.py
davidorme: Yeah - the LayerStructure will be able to provide specific coordinates to extract the target values using `lowest_filled_layer`, but may be able to do it with numpy array commands:

```python
In [128]: y = np.arange(25).reshape((5,5))

In [129]: y
Out[129]: 
array([[ 0,  1,  2,  3,  4],
       [ 5,  6,  7,  8,  9],
       [10, 11, 12, 13, 14],
       [15, 16, 17, 18, 19],
       [20, 21, 22, 23, 24]])

In [130]: y[tuple([np.arange(5),np.arange(4, -1, -1)])]
Out[130]: array([ 4,  8, 12, 16, 20])
```

tests/models/plants/conftest.py
davidorme: OK - as discussed, since this is specifically the fraction of PAR absorbed, it doesn't really make sense to assign a fAPAR value to the topsoil layer and we need to revisit the light penetration through the canopy to include SW downwelling radiation more widely.

PR #475, opened 2024-06-28
Updated TODOs across animal module code.

TaranRallings:
# Description

It was suggested that I should be turning my TODOs into issues and that I should be making branches out of those issues. To begin this process I needed to clean up my TODOs and that is what I have done here.

The TODOs are now much better organized and they have been made into issues. There were a couple little docstring updates I did along the way when I realized something was out of date. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


davidorme (approved):
LGTM!

dalonsoa (approved):
Fantastic! 

Rather than creating issues with an "Animal - ..." title, I think it will be useful to use [labels](https://github.com/ImperialCollegeLondon/virtual_ecosystem/labels) to classify the issues by categories or themes. That way is also easier to filter or quickly visualise what the issues you have are about. So, I would create a label for `Animal`, another for `Plats`, etc. and use those.

PR #476, opened 2024-07-01
abiotic and hydrology models overview updated

vgro:
This PR updates the overview model descriptions of the abiotic models and hydrology model. Is this the style we want?

- I moved the model to the top of the list because they describes the environment for the other models
- I set a few links to the API docs
- I mostly focused on checking that descriptions are up to date, not on being particularly nicely written (priority on filling gaps)

### DIFF
--- a/docs/source/virtual_ecosystem/module_overview.md
+++ b/docs/source/virtual_ecosystem/module_overview.md
@@ -126,16 +126,20 @@ Ecosystem workflow.
 Some of the features described here are not yet implemented.
 ```
 
-The first part of the hydrology model determines the water balance within each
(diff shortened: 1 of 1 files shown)


davidorme (approved):
LGTM - I think the style is right. In the long term, we might break these out into separate documents and maybe add some dynamic code to illustrate model behaviour but this is great.

jacobcook1995 (approved):
Yep, I also agree that it's the right level for an overview!

davidorme:
I guess one thing is that the links to the API docs are quite a big jump in technical content - which is where those dynamic content bits might come in -  but I think that's something to work on later.

Issue #344, opened 2023-11-30
simplify calculate_soil_carbon_updates() function


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
The function to calculate the updates for the soil carbon pool (`calculate_soil_carbon_updates`) is much too complex, it's 150 lines long and has 17 arguments.

**Describe the solution you'd like**
I should remember to simplify the function. There should hopefully be an obvious moment to do this at some point when I'm adding more stuff to the soil model.

@alexdewar also suggested the following as good generic approaches to this sort of problem:

> 1. If lots of the arguments to this function are just being passed through to one or two other functions, you could just pass in the _results_ of these functions instead and have the caller calculate them
> 2. If some of the arguments would naturally group together, consider bundling them together in a class/dataclass (provided that this function won't be the only user of it)

jacobcook1995:
@davidorme had some a helpful suggestions [here](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/477#pullrequestreview-2152973231), which I should bear in mind for the next time I need to simplify this function

PR #477, opened 2024-07-01
Adding a microbial necromass pool

jacobcook1995:
# Description

This pull request extends the soil carbon model to include a pool for microbial necromass (dead cells, denatured enzymes, etc). As part of this I have changed some of the previous flows between carbon pools.

I've also tried to break the main function used by the soil model integration up a bit (see #344). This is far from complete, but hopefully it's a bit more readable now.

I haven't added any documentation changes to the PR as I'm planning to overhaul the soil model documentation in the near future anyway.

## Type of change

## Key checklist

## Further checks


davidorme (approved):
This all looks good to me.

On the re-org, I was wondering if it might make sense to package the carbon and env factors as classes with methods (or just add methods on to the existing dataclasses). 

One advantage of that is that at the moment you have long lines passing many variables into functions - but most of those variables are shared widely across the functions. If those variables were attributes of the classes (possibly `__post_init__` ones) then the function signatures reduce to `xyz(self, actual_new_info)` and there is much less unpacking and packing of information.

I think that would simplify the internals a lot and then in `construct_full_soil_model` you'd end up using something like:

```python 
carbon_updates = SoilCarbon(...)
return carbon_updates.get_updates()
```


That kind of design was pretty much where I ended up with the `LayerStructure` update (https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/022cb50a8526fe3576ba484fbdba4c21d6a4b483/virtual_ecosystem/core/core_components.py)

virtual_ecosystem/models/soil/carbon.py
```diff
@@ -12,36 +12,52 @@
...
-    """Rate at which MAOM degrading enzymes are produced [kg C m^-3 day^-1]."""
-    necromass_decay_to_lmwc: NDArray[np.float32]
-    """Rate at which biomass is lost to the LMWC pool [kg C m^-3 day^-1]."""
-    necromass_decay_to_pom: NDArray[np.float32]
-    """Rate at which biomass is lost to the POM pool [kg C m^-3 day^-1]."""
-
-
-# TODO - This function should probably be shortened, leaving as is for the moment as a
-# sensible split will probably be more obvious once more is added to this function.
+    lmwc_uptake: NDArray[np.float32]
```
vgro, line 26: are you always talking about rates of change? At the moment, most of these names can be easily interpreted as stocks, for example the `microbe`

virtual_ecosystem/models/soil/carbon.py
```diff
@@ -12,36 +12,52 @@
...
+    necromass_generation: NDArray[np.float32]
+    """Rate at which necromass is being produced [kg C m^-3 day^-1]."""
+
+
+@dataclass
+class EnzymeMediatedRates:
+    """Rates of each enzyme mediated transfer between pools."""
+
+    pom_to_lmwc: NDArray[np.float32]
+    """Rate of POM decomposition to LMWC [kg C m^-3 day^-1]."""
```
vgro, line 47: I would spell out the names of the pools in the docstrings

vgro (approved):
LGTM

virtual_ecosystem/example_data/generation_scripts/soil_example_data.py
```diff
@@ -35,6 +35,10 @@
 # Generate a range of plausible values (0.1-1.0) for the POM pool [kg C m^-3].
 pom_values = 0.1 + 0.9 * (gradient) / (64)
 
+# Generate a range of plausible values (0.00015-0.0005) for the microbial necromass pool
+# [kg C m^-3].
+necromass_values = 0.00015 + 0.00035 * (gradient) / (64)
```
alexdewar, line 40: Not sure what the brackets are for:

```suggestion
necromass_values = 0.00015 + 0.00035 * gradient / 64.0
```

It's probs also best practice to make the 64 explicitly a floating-point number.

virtual_ecosystem/models/soil/carbon.py
```diff
@@ -96,153 +127,256 @@ def calculate_soil_carbon_updates(
...
         soil_temp: soil temperature for each soil grid cell [degrees C]
-        clay_factor_decay: A factor capturing the impact of soil clay fraction on
-            necromass decay destination [unitless]
+        env_factors: Data class containing the various factors through which the
+            environment effects soil cycling rates.
         constants: Set of constants for the soil model.
 
     Returns:
-        A dataclass containing all the losses from the microbial biomass pool.
+        A dataclass containing, the rate at which microbes uptake LMWC, the rate of
```
alexdewar, line 237: Stray comma:

```suggestion
        A dataclass containing the rate at which microbes uptake LMWC, the rate of
```

alexdewar (approved):
LGTM! I've made a couple of small comments

PR #509, opened 2024-07-10
Direct input from plant model to the litter model

jacobcook1995:
# Description

This PR changes the litter model to receive input from the plant model. This input then gets split between pools based on its lignin content and C:N ratio. Science wise it is relatively straightforward, but there's a couple of things to note from a code perspective:

1. @davidorme this PR alters plant model code, let me know if you are happy with the general structure
2. This PR adds 11 new variables to the data object, this has lead to some fairly huge functions signatures, if anyone has any suggestions for how best to handle this please let me know.

## Type of change

## Key checklist

## Further checks


pyproject.toml
```diff
@@ -92,6 +92,7 @@ addopts = """
   --cov=virtual_ecosystem
   --cov-report=html:htmlcov/coverage
   --doctest-modules --ignore=virtual_ecosystem/__main__.py
+  --import-mode importlib
```
dalonsoa, line 95: Out of curiosity, what is this for?

virtual_ecosystem/models/litter/input_partition.py
```diff
@@ -0,0 +1,225 @@
...
+    reproduct_turnover: NDArray[np.float32],
+    root_turnover: NDArray[np.float32],
+    leaf_turnover_lignin_proportion: NDArray[np.float32],
+    reproduct_turnover_lignin_proportion: NDArray[np.float32],
+    root_turnover_lignin_proportion: NDArray[np.float32],
+    leaf_turnover_c_n_ratio: NDArray[np.float32],
+    reproduct_turnover_c_n_ratio: NDArray[np.float32],
+    root_turnover_c_n_ratio: NDArray[np.float32],
+    constants: LitterConsts,
+) -> dict[str, NDArray[np.float32]]:
```
dalonsoa, line 31: Do you think this is a long signature? Have a look at [`pandas_read_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas-read-csv). _That_ is long. 

Seriously, I don't see a big problem with having all of these inputs. The _look_ more than they are because the names are long (and descriptive, no complain in there), but otherwise it looks totally fine.

virtual_ecosystem/models/litter/litter_pools.py
```diff
@@ -35,6 +40,17 @@ def calculate_change_in_litter_variables(
     lignin_above_structural: NDArray[np.float32],
     lignin_woody: NDArray[np.float32],
     lignin_below_structural: NDArray[np.float32],
+    deadwood_production: NDArray[np.float32],
```
dalonsoa, line 43: Ok, yes, this is on the long side. 

I wonder if you need this function at all, to start with. In the end, this is a top level function calling a bunch of other functions that do the heavy lifting, and that is all the `update` method above is doing. So you have a top level function (`update`) that just calls another high level function (`calculate_change_in_litter_variables`). I don't see why you cannot put all the contents of this function directly into `update` and save unnecessary, top level functions with long signatures.

dalonsoa: Another option is to make the litter pools a class, with a reference to data, and rewrite all of these as method that use that data.

```python
class LitterPool:
    def __init__(self, data):
        self.data = data

    def calculate_change_in_litter_variables(self) -> dict:
        ...

    # And other relevant methods
```
And then, in the `LitterModel` just use above:

```python
class LitterModel(...):
    self __init__(...):
        ...
        self.litter_pool = LitterPool(data)

    self.update(...):
        updated_vars = self.litter_pool.calculate_change_in_litter_variables()
        ...
```
Or something along these lines. This is just a sketch.

dalonsoa (approved):
The code itself looks good, so I'm approving. I just have a couple of suggestions to simplify a bit the signatures and the readability.

virtual_ecosystem/models/litter/litter_pools.py
jacobcook1995: Yeah I do wonder whether the separate function is really warranted. I think I've basically duplicated the structure I've used for the `soil` model, but there the function sets up an integrator, passes the lower level function into the integrator and then extracts the result, whereas here it just calls the lower level function

pyproject.toml
dalonsoa: Cool! I did not know that.

virtual_ecosystem/models/litter/litter_pools.py
jacobcook1995: I think I will just delete the lower level function and move everything into `LitterModel.update()` as that seems like the simplest approach. I'll bear the extra class approach in mind when thinking about the `soil` model case where a more complex structure is required

pyproject.toml
davidorme: The default is basic the legacy option, but this one is faster and better.

virtual_ecosystem/models/litter/env_factors.py
```diff
@@ -0,0 +1,129 @@
...
+
+from virtual_ecosystem.models.litter.constants import LitterConsts
+
+
+def calculate_environmental_factors(
+    surface_temp: NDArray[np.float32],
+    topsoil_temp: NDArray[np.float32],
+    water_potential: NDArray[np.float32],
+    constants: LitterConsts,
+) -> dict[str, NDArray[np.float32]]:
```
davidorme, line 17: I'm not convinced that nesting these three functions into a single wrapper is worthwhile. The three functions below are all very straightforward and there's essentially no overlap apart from the constants. If you drop this, you have three calls with smaller signatures and then need to add two items to the `decay_rates` signature. It's seems much cleaner.

davidorme (approved):
Looks good to me too. No problems with that placeholder in the plants model. 

I like @dalonsoa's idea of having an internal class -  having a class with methods does tend to avoid having a lot of data being passed around, which I think could be clearer here - but possibly something to park and come back to.

PR #524, opened 2024-07-17
Move setup step to init in `Abiotic` and `AbioticSimple` models

dalonsoa:
# Description

It was done already for the `Hydrology` model but I forgot to do the same for the other two. The main changes have more to do with fixing the tests than the actual change in the code.

Fixes # N/A

## Type of change

## Key checklist

## Further checks


davidorme:
@TaranRallings - The animal model also uses  a `setup` method, which is currently also called from within `__init__`, so runs twice.

vgro (approved):
LGTM

PR #528, opened 2024-07-24
493 hydrology add canopy evaporation

### LINKED ISSUE #493: Hydrology - Add canopy evaporation
Canopy evaporation is currently not implemented as it was not clear if that is handled in the plant model. It isn't, so it needs to be included in the hydrology model.

Implement equation that accounts for canopy evaporation from the intercept pool.

vgro:
This turns out to be more complex than anticipated. In order to calculate canopy evaporation, we need a range of variables that are provided by the `abiotic` model. These are currently not in the `abiotic_simple` model; the attempt to add them caused a crash in `ve_run`. This is a Draft PR to see error message.

I have tried to reproduce this step by step and it all works fine until the [this commit](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/528/commits/20c05a3c1090a8ed548cdab9d5747debd16ed3bf), unit tests are fine but the integration tests runs into nowhere when it tried to update the `self.data['wind_speed']`  in line 205 in `abiotic_simple_model.py`.  (After that something went funny with pushing and pulling and I had weird duplications of lines, I hope this is all sorted)

@dalonsoa and @davidorme  could you take a look into this, maybe you spot something? I suspect it has to do with the order of models, it work ok when the wind comes from the `hydrology` model, but it should work with the changes in variables?

Fixes # (issue)

## Type of change

- [] Optimization (back-end change that speeds up the code)

## Key checklist

## Further checks


vgro:
An update on this, I have tried to run the `ve_run` example locally, it also get's stuck but interestingly when it tries to update the `soil`model, see logfile attached. @davidorme do you have an idea how I could get to the bottom of this?
[logfile.log](https://github.com/user-attachments/files/16499784/logfile.log)

vgro:
We have identified the issue, see #536.
Will work on this and then put it up for review.

davidorme:
I'll de-request those outstanding reviews for now.

vgro:
Quick update (mostly for me to pick up after summer break):

The problem with the endless loop in `ve_run` is solved; I added a tentative check in the soil model, will need to be discussed for soil, see #536 with @jacobcook1995,  and more generally for integration tests

The problem was in the wind update, there seemed to be an issue with filling the right layers with zeros for the `sensible heat flux` variable. It was fixed with a hack in the `abiotic_simple` model to continue working, but came back with the `abiotic` model - the first canopy layer was Nan for all temperatures and fluxes. This points to the general issue with the inconsistencies in the `abiotic` model subcomponents. Hopefully, the conference will shed light in this.

Fixes to do (potentially in separate PR):
[ ] integrate model versions better (constants, wrapper functions, same order, same varaibles tec)
[ ] fix layer selection for sensible heat flux for both model versions
[ ] add wind to `abiotic` init and check why it causes first canopy layer to be nan
[ ] update tests for both models to capture the additional variables that are returned
[ ] add checks for Nan in key places to stop process early

To do's for evaporation:
[x] add aero resistance canopy
[ ] add potential evaporation (drafted)
[ ] add canopy evaporation function (drafted)

Issue #533, opened 2024-07-30
Soil model should average abiotic quantities over relevant soil layers


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
We now have a system where the soil layers expected to be biologically active is defined within the layer structure, but currently the soil model doesn't use this it just takes abiotic variables from the first soil layer.

**Describe the solution you'd like**
Instead the soil model should average the abiotic variables over the relevant layers. 

**Describe alternatives you've considered**
More complicated schemes of averaging could be used, i.e. weighting towards the surface layers as these are more active. But to begin with I think a simple average is enough

jacobcook1995:
This can be implemented now, it has the problem of a lack of a boundary condition (see #533), but that already exists in the litter model implementation

PR #534, opened 2024-07-31
532 litter model should average abiotic quantities over relevant soil layers

### LINKED ISSUE #532: Litter model should average abiotic quantities over relevant soil layers
We now have a system where the soil layers expected to be biologically active is defined within the layer structure, but currently the litter model doesn't use this it just takes abiotic variables from the first soil layer.

Instead the litter model should average the abiotic variables over the relevant layers. 

More complicated schemes of averaging could be used, i.e. weighting towards the surface layers as these are more active. But to begin with I think a simple average is enough

jacobcook1995:
# Description

This pull request makes use of the new and improved `LayerStructure` object to average temperature and matric potential over the microbially active depth, rather than just taking the value from the topsoil layer. This has been done for just the litter model as that's the one I'm currently working on (but soil should follow a similar process when I get round to it see #533).

This PR is generally pretty straightforward, the only slight weirdness being the testing. Basically the layer structure fixture has a really deep first soil layer, so it doesn't actually matter whether you average or just take the top layer as you get the exact same result. So I had to overwrite the fixture in some of the tests so that they actually test a case with averaging.

Fixes #532

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/litter/env_factors.py
```diff
@@ -5,6 +5,81 @@
...
+    surface_temp = air_temperatures[layer_structure.index_surface_scalar].to_numpy()
+    below_ground_temp = average_over_microbially_active_layers(
+        environmental_variable=soil_temperatures, layer_structure=layer_structure
+    )
+    water_potential = average_over_microbially_active_layers(
+        environmental_variable=water_potentials, layer_structure=layer_structure
+    )
+
+    # Calculate temperature factor for the above ground litter layers
+    temperature_factor_above = calculate_temperature_effect_on_litter_decomp(
```
vgro, line 57: you could probably combine above and below ground te,temperature calculation to make it slightly more concise, but I think it's also ok like this for readability

virtual_ecosystem/models/litter/env_factors.py
```diff
@@ -69,3 +144,33 @@ def calculate_soil_water_effect_on_litter_decomp(
...
+
+    Args:
+        environmental_variable: The environmental variable to be averaged
+        layer_structure: The LayerStructure instance for the simulation.
+
+    Returns:
+        The average of the environmental variable across the soil depth considered to be
+        microbially active.
+    """
+
```
vgro, line 165: I think for the below ground temperature it would make sense to include the surface temperature (which we might want to move closer to the surface, maybe only 2 cm or so) when calculating where you sit on the temperature profile, else you will not have a gradient but just the value of the topsoil layer of that is deeper than you microbial activity. Am I understanding this correctly?

vgro (changes requested):
Looks overall good to me, I like how the use of the LayerStructure components makes it much clearer. Also, I think the structure is becoming more and more clear as functions are grouped.
My main point to review is about how you calculate the temperature below ground, I think if you want a value along a gradient it would make sense to include the surface temperature (if the microbial activity depth is within the topsoil layer)

virtual_ecosystem/models/litter/env_factors.py
vgro: It is at the actual depth which is why we do the vertical averaging I thought

jacobcook1995: Oh right! Yes then I absolutely need to include the surface temperature in the calculation (and rework the calculation)

virtual_ecosystem/models/litter/env_factors.py
jacobcook1995: combine as in wrap them up into a single function, or like a list comprehension?

vgro: I thought of list comprehension (do write the same thing twice)

jacobcook1995: Ahh gotcha! List compressions still confuse/scare me, but that something I probably should get over 😅

vgro: I think it is good to do it in two steps, write it all out first and then when you realize that you've written the same function several times it's quite straight forward to loop over the variables :-)

virtual_ecosystem/models/litter/env_factors.py
vgro: I've created an issue to address this #535

virtual_ecosystem/models/litter/env_factors.py
```diff
@@ -5,6 +5,88 @@
...
+        below ground litter decay.
+    """
+
+    temperatures = {
+        "surface": air_temperatures[layer_structure.index_surface_scalar].to_numpy(),
+        "below_ground": average_temperature_over_microbially_active_layers(
+            soil_temperatures=soil_temperatures,
+            surface_temperature=air_temperatures[
+                layer_structure.index_surface_scalar
+            ].to_numpy(),
```
vgro, line 59: Would you not us the new shallower topsoil layer here? If not now, add a TODO?

virtual_ecosystem/models/litter/env_factors.py
```diff
@@ -5,6 +5,88 @@
...
+
+    Returns:
+        A dictionary containing three environmental factors, one for the effect of
+        temperature on above ground litter decay, one for the effect of temperature on
+        below ground litter decay, and one for the effect of soil water potential on
+        below ground litter decay.
+    """
+
+    temperatures = {
+        "surface": air_temperatures[layer_structure.index_surface_scalar].to_numpy(),
```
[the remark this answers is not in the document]
jacobcook1995: Yep this one is for the above ground litter, in this case I just use the surface temperature directly and don't interpolate (because that seems tricky without tracking litter depth etc)

virtual_ecosystem/models/litter/env_factors.py
jacobcook1995: I'm not sure actually. How far above the surface is the "surface" air temperature?

vgro: At the moment it is 10cm

jacobcook1995: Oh right, yes it probably makes sense to use the top layer then (once it's a thin layer). I'll add a TODO in

vgro (approved):
LGTM

PR #538, opened 2024-08-15
Animal - Feature/multi grid no community

TaranRallings:
# Description

Where to even begin! This is scandalously oversized rework of the entire animal module that has taken me over two months. I would never have chosen to do it this way but it happened organically in a manner that has taught me many lessons. 
Mea culpa. 

This rework combines two major changes:
1) the implementation of multi-grid occupancy
2) the removal of the `AnimalCommunity` class 

Multi-grid occupancy equips each animal cohort with a territory of grid cells (list(int)) that they occupy simultaneously. They forage and deposit wastes uniformly across these cells. Carcasses produced from predation are deposited at the intersection of predator and prey territories. Territories are produced using breadth-first search but this can be made more sophisticated later.   

The removal of the `AnimalCommunity` class is a major structural change. This class had been storing per-grid information on animal cohorts, waste pools, and the like as well as running cohort-specific methods in aggregate for those grid cells. Almost all of this functionality is now located within the `AnimalModel` class. Animal cohorts and waste pools are now stored as dictionaries in `AnimalModel`, tying the occupying lists to specific grid cell ids. The aggregating methods from `AnimalCommunity` have been modified to run over these dictionaries. Removing these community objects has, I hope, made the module significantly easier to understand. 

I abandoned the `AnimalTerritory` class I had been exploring in an earlier implementation of multi-grid.

There are some significant changes to the litter system here as well as it was designed while I was mid re-work. The version I have is a little structurally different than Jacob's original but works the same. 

There were some concerns with an earlier version about the amount of indexing going on in the foraging system. This is valid but I simply do not have the time to optimize the foraging system, especially with how behind I am after the rework. Hopefully the structural simplifications made here will simplify making such optimizations in the future, after the remaining required functionality is implemented. 

The important files here are animal_model and animal_cohort but unfortunately little tweaks had to be made just about everywhere. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -299,8 +381,17 @@ def get_eaten(
...
+            self.is_alive = False
+
+        # Find the intersection of prey and predator territories
+        intersection_carcass_pools = self.territory.find_intersecting_carcass_pools(
+            predator.territory, carcass_pools
+        )
+
         # Update the carcass pool with carcass mass
-        self.update_carcass_pool(carcass_mass, carcass_pool)
+        self.update_carcass_pool(carcass_mass, intersection_carcass_pools)
```
dalonsoa, line 394: Why do you need to update the carcass pool with the intersection of carcass pools instead of the normal carcass pool, as before? Indeed, why is the intersection of territories needed and why is it calculated based on the intersection of the carcass pools?

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -177,21 +238,32 @@ def defecate(
...
-            (1 - self.decay_fraction_excrement) * waste_energy * self.individuals
-        )
-        excrement_pool.decomposed_energy += (
-            self.decay_fraction_excrement * waste_energy * self.individuals
+        # Find total waste mass, the total amount of waste is found by the
+        # average cohort member * number individuals.
+        waste_energy = (
+            mass_consumed
+            * self.functional_group.conversion_efficiency
+            * self.individuals
```
dalonsoa, line 253: Maybe is just a question of definition, but it seems to me that the higher the conversion efficiency, the lower the waste should be, but here is the opposite so maybe I am not understanding what waste is.

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -177,21 +238,32 @@ def defecate(
...
         )
 
+        waste_energy_per_community = waste_energy / number_communities
+
+        for excrement_pool in excrement_pools:
```
dalonsoa, line 258: How many pools (i.e. number of communities) are we talking about? If they are many, having the `excrement_pools` as an array and somewhat vectorising this (or at least, using list comprehensions) might speed up things here compared with a plain loop.

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -239,41 +313,49 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
-
-        # Update the carcass pool with the remainder
-        carcass_pool.scavengeable_energy += (
-            1 - self.decay_fraction_carcasses
-        ) * carcass_mass
-        carcass_pool.decomposed_energy += self.decay_fraction_carcasses * carcass_mass
+        number_carcass_pools = len(carcass_pools)
+        carcass_mass_per_pool = carcass_mass / number_carcass_pools
+
+        for carcass_pool in carcass_pools:
```
dalonsoa, line 333: And the same here. Also, why you use a list here for these pools but a set for the `excrement_pools` above?

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -239,41 +313,49 @@ def die_individual(self, number_dead: int, carcass_pool: CarcassPool) -> None:
...
+                self.decay_fraction_carcasses * carcass_mass_per_pool
+            )
 
     def get_eaten(
         self,
         potential_consumed_mass: float,
-        predator: Consumer,
-        carcass_pool: DecayPool,
+        predator: AnimalCohort,
+        carcass_pools: dict[int, set[CarcassPool]],
```
dalonsoa, line 346: Indeed, the carcass pools here are a set, not a list.

virtual_ecosystem/models/animal/animal_territories.py
```diff
@@ -0,0 +1,248 @@
...
+                new_cell = nr * cell_nx + nc
+                # If the cell hasn't been visited, mark it as visited and add to the
+                # territory
+                if new_cell not in visited:
+                    visited.add(new_cell)
+                    territory_cells.append(new_cell)
+                    queue.append((nr, nc))
+                    # If we have reached the target number of cells, exit the loop
+                    if len(territory_cells) >= target_cell_number:
+                        break
```
dalonsoa, line 246: I'm a big fan of un-indenting stuff, so I would suggest the following to improve readability:
```suggestion
            if not(0 <= nr < cell_ny and 0 <= nc < cell_nx):
                continue
                
            new_cell = nr * cell_nx + nc
            # If the cell hasn't been visited, mark it as visited and add to the
            # territory
            if new_cell in visited:
                continue
                    
            visited.add(new_cell)
            territory_cells.append(new_cell)
            queue.append((nr, nc))
            
            # If we have reached the target number of cells, exit the loop
            if len(territory_cells) >= target_cell_number:
                break
```

virtual_ecosystem/models/animal/animal_territories.py
```diff
@@ -0,0 +1,248 @@
+"""The ''animal'' module provides animal module functionality."""  # noqa: #D205, D415
+
+from virtual_ecosystem.models.animal.animal_cohorts import AnimalCohort
+from virtual_ecosystem.models.animal.decay import CarcassPool, ExcrementPool
+from virtual_ecosystem.models.animal.plant_resources import PlantResources
+
+
+class AnimalTerritory:
```
dalonsoa, line 8: So, as a sort of summary, all the iteration between an animal cohort and its surroundings to thrive - or die - now happens here, right?

dalonsoa (commented):
The changes seem to make sense, but I have to admit I've trouble following the relationships between the different components. 

Let's see if I get it:

The Animal model manages... the animals! They take the form of communities (each of a different animal, like elephants, ants, etc). Each community is made of cohorts that are born, grow and die. Each cohort lives in a territory where they interact with other cohorts of other communities, as well as plants and the pools. The interactions between the cohorts and the environment is orchestrated at the highest level by the Animal model.

Is this picture correct?

Keeping this general understanding of the problem, I've just made a few suggestions here and there.

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -547,40 +642,46 @@ def calculate_consumed_mass_predation(
 
     def delta_mass_predation(
         self,
-        animal_list: Sequence[AnimalCohort],
-        excrement_pool: DecayPool,
-        carcass_pool: CarcassPool,
+        animal_list: list[AnimalCohort],
+        excrement_pools: list[ExcrementPool],
+        carcass_pools: dict[int, set[CarcassPool]],
```
davidorme, line 647: OK - I think I see why the cell_id keyed dicts here. I think the issue is that animals can only eat an animal and leave its carcass in overlapping territory cells, but defecation can occur anywhere. So you have to retain the `cell_id` on the carcass pools, so that `get_eaten` can drop corpses in the right locations.

Something that might make it simpler: 
* `AnimalCohort` has an attribute `excrement_pools` that is updated when territory changes. 
* `delta_mass_predation` works on a single prey item at a time. So something like:

```python
for prey in animal_list:
    self.delta_mass_predate(prey)

def delta_mass_predate(self, prey: AnimalCohort):
    intersection = self.territory & prey.territory  # assuming sets
    etc.
```

That seems cleaner than having to do a bunch of repeated indexing at multiple levels? I think we'd need to add the complete set of carcass pools  to each instance?

virtual_ecosystem/models/animal/animal_communities.py
```diff
@@ -1,32 +1,18 @@
...
 from virtual_ecosystem.models.animal.decay import CarcassPool, ExcrementPool
 from virtual_ecosystem.models.animal.functional_group import (
     FunctionalGroup,
-    get_functional_group_by_name,
 )
 from virtual_ecosystem.models.animal.plant_resources import PlantResources
-from virtual_ecosystem.models.animal.scaling_functions import damuths_law
 
 
 class AnimalCommunity:
```
davidorme, line 18: I don't fully understand why this class is required. The carcass, excrement and plant resources _feel_ like they are redundant, given the per_cell list within the Animal Model? And I don't understand why we want to group by the centroid?

davidorme (commented):
I do think this looks cleaner than #530. There are a few things that don't feel quite right but this is a draft and in the bigger picture I _think_ this is the way to go.

There are a couple of aspects I'm not quite getting - the main one is the indexing, which I know you said got ugly! It feels like too much indexing is going on - and I _think_ this is to do with handling the multiple prey items and needing to keep track of defecation and carcass locations.

I have a strong suspicion that this might get much cleaner if the `delta_mass_predation` function operates only on prey cohort at a time. When everything was in a single cell, looping over cohorts internally made sense, but now each prey cohort has a potentially different territory intersection, so calculating the cell ids of that intersection for each pair (which  we do anyway) and then indexing from the per-cell lists seems cleaner?

The other thing is the AnimalTerritory itself - I haven't followed through how it is used perfectly, but it seems redundant?

I've added a bunch of comments to #530 about our discussion and the indexing issue we're trying to solve. @dalonsoa seems intrigued by the problem 😄

virtual_ecosystem/models/animal/protocols.py
```diff
@@ -5,6 +5,7 @@
 
 from typing import Protocol
 
+# from virtual_ecosystem.models.animal.decay import ExcrementPool
```
jacobcook1995: Guess this line can just be deleted

virtual_ecosystem/models/litter/input_partition.py
jacobcook1995, line 1: You should delete this file, I renamed it recently and I think your merge is trying to add it back in under the old name

jacobcook1995 (approved):
The stuff that interfaces with the soil and litter is all fine by me

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -431,39 +503,47 @@ def calculate_litter_additions_from_herbivory(self) -> dict[str, DataArray]:
...
+            for excrement_pool in excrement_pools:
+                excrement_pool.decomposed_carbon = 0.0
+                excrement_pool.decomposed_nitrogen = 0.0
+                excrement_pool.decomposed_phosphorus = 0.0
+
+        for carcass_pools in self.carcass_pools.values():
+            for carcass_pool in carcass_pools:
+                carcass_pool.decomposed_carbon = 0.0
+                carcass_pool.decomposed_nitrogen = 0.0
+                carcass_pool.decomposed_phosphorus = 0.0
```
dalonsoa: I think it will make sense to implement a `reset` method for these pools, so you can do the following without having to know the internal detail:

```suggestion
        for excrement_pools in self.excrement_pools.values():
            for excrement_pool in excrement_pools:
                excrement_pool.reset()

        for carcass_pools in self.carcass_pools.values():
            for carcass_pool in carcass_pools:
                carcass_pool.reset()
```

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -148,45 +212,37 @@ def _setup_grid_neighbors(self) -> None:
...
+        # cohorts
+        self.communities = {cell_id: list() for cell_id in self.data.grid.cell_id}
+
+        # Iterate over each cell and functional group to create and populate cohorts
+        for cell_id in self.data.grid.cell_id:
+            for functional_group in functional_groups:
+                # Calculate the number of individuals using Damuth's Law
+                individuals = damuths_law(
+                    functional_group.adult_mass, functional_group.damuths_law_terms
+                )
```
dalonsoa, line 232: I'm not entirely sure, but it seems to me this law is used already within the the `AnimalCohort` constructor. Not sure if for the same purpose, but just flagging it in case this is an error.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -148,45 +212,37 @@ def _setup_grid_neighbors(self) -> None:
...
+                # Create a cohort of the functional group
+                cohort = AnimalCohort(
+                    functional_group=functional_group,
+                    mass=functional_group.adult_mass,
+                    age=0.0,
+                    individuals=individuals,
+                    centroid_key=cell_id,
+                    grid=self.data.grid,
+                    constants=self.model_constants,
+                )
```
dalonsoa, line 243: As far as I can tell, all cohorts in all grid cells are identical. Where do you set - if you do - the initial distribution of individuals across the simulation domain?

dalonsoa (approved):
The PR is massive, as we all know, and I've might have missed something, but it looks good overall.

I've left a couple of comments, but what really confuses me is the separation of concerns. What methods/information goes into the `AnimalCohort` and what goes into the `AnimalModel` (and before in the `AnimalCommunity`. I'm sure there's a pattern, but I just cannot see it.

virtual_ecosystem/models/animal/animal_model.py
TaranRallings: Good catch!

virtual_ecosystem/models/animal/animal_model.py
TaranRallings: There is not currently a setting for number of individuals across the entire grid. The number of individuals are determined through population densities and the spatial range of the simulation. I expect this format to be much more commonly used but we may need a "fixed number of individuals" input later, particularly for species of special interest and reintroduction programs.

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: @dalonsoa this is code that I added and Taran had to refactor to get to work with the new setup. I agree that this is a better way of doing things, I'll rewrite and push the changes to this branch

PR #539, opened 2024-08-28
Add phosphorus to litter model

### LINKED ISSUE #512: Litter phosphorus
Functionality has been developed to handle litter nitrogen, but it also needs to be developed for phosphorus.

Functionality should be developed to track litter phosphorus. The litter pool splits should also be partially determined by phosphorus content

jacobcook1995:
# Description

This pull requests adds tracking of litter phosphorus concentrations. This follows an identical structure to the nitrogen case. I did consider trying to make a generic functions rather than duplicating functions, but it didn't feel worth the effort as we have no current plans to add further nutrients (e.g. Potassium).

As there weren't any structural changes I didn't split this up into separate PRs, hopefully it is a manageable size as a single PR.

Fixes #512 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/example_data/config/data_config.toml
```diff
@@ -106,6 +106,21 @@ var_name = "c_n_ratio_below_metabolic"
 [[core.data.variable]]
 file = "../data/example_litter_data.nc"
 var_name = "c_n_ratio_below_structural"
+[[core.data.variable]]
+file = "../data/example_litter_data.nc"
```
dalonsoa, line 110: This is just a generic question and nothing to do with this PR in particular. For what I can see, running a full Virtual Ecosystem simulation requires A LOT of data to be available. Is that data typically available or easy to find? Are you planning to provide defaults for them - some, at least? If the data can be downloaded from online repositories (for free), would you be including hooks to download that data in the background?

dalonsoa (approved):
LGTM. IT follows a similar pattern than nitrogen and, to be honest, I've not spotted that many repetitions to be worrisome.

virtual_ecosystem/example_data/config/data_config.toml
jacobcook1995: Yeah this is a good question. In general, a lot of this data will not be particularly easy to get, partly because ecological data tends to get reported in fairly in consistent ways. But also because (at least in the soil and litter models) some variables are chosen because they make sense in the model context rather than being commonly measured empirically, e.g. structural litter isn't measured empirically, mass of dead leaves, twigs, branches etc is.

A few more (Malaysia based) postdocs have joined the team and their role will be getting the data needed for parameterisation, calibration and validation together. I'm not sure of the exact planned workflow there, but I believe the idea is to have all the scripts used to wrangle the data and the final datasets ready for input available online. So this probably make sense to be hooked in some way to the model itself, particularly for the non-site specific data.

PR #548, opened 2024-09-05
Execution order based on variables

### LINKED ISSUE #389: Setup Variables (umbrella)
Ok, so I'm booking myself Thursday and Friday next week to work on this. My plan would be:

### Step 1 (one PR)
- [x] Start from the simplified version created by @davidorme 
- [x] Add a step [somewhere in the `Config` class](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/76f19065815edc8eebdef6016b94859123697847/virtual_rainforest/core/config.py#L129) after the required models have been pulled out of the configuration to populate the `initialised_by`, `updated_by`, and `used_by` lists. The models might need updating.
- [x] Run some validation based initially on the following conditions, raising an error if they are broken:
    - [x] No variable is initialised by two models
    - [x] There is no variable used or updated that has not been initialised
    - [x] No two models update the same variable (or they do, I don't know if that is a use case)

### Step 2 (another PR, after the above is merged)
- [x] Update the [`get_model_sequence`](https://github.com/ImperialCollegeLondon/virtual_rainforest/blob/76f19065815edc8eebdef6016b94859123697847/virtual_rainforest/main.py#L62) function to use this information, as well as (or in place of?) the explicit dependencies already in place, to try to come up with the right sequence.
- [x] Update the models so they initialise the variables when they are initialised and the registry of initialised variables is populated.

### Step 3 (yet another PR)
- Update the models so they use the information from this registry -> TBC by each model at its own pace

I'm sure I'm missing something and I will need to improvise, but is there anything obvious I'm forgetting?

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/virtual_rainforest/issues/371#issuecomment-1934300630_

dalonsoa:
# Description

Replaces the current calculation of the model execution order in init and update - based on the input configuration - with another one based on the information on which variables each model initialises and which ones they require.

Using this method has identified also an inconsistency in the variables used by the `soil` model, which I guess it is a good thing!

The `depends` section of the example config files have been removed, and I had not spotted it being mentioned anywhere else, including the documentation, but let me know if I missed anything.

In principle, this is the last bit of the implementation of the variables system.

Fixes #389

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

vgro (approved):
LGTM

dalonsoa:
I'm in no rush, but if this is Ok, it will be best to merge it. The longer a PR is open, the more challenges it brings merging it for ongoing work. It is approved but I would prefer if you merge it, so it is done when it is a good time (eg. to avoid conflicts).

davidorme (approved):
LGTM

Issue #544, opened 2024-09-04
ve_run on example does not output initial_state.nc


hrlai:
When I try to following the [example ](https://virtual-ecosystem.readthedocs.io/en/latest/using_the_ve/virtual_ecosystem_in_use.html#create-the-model-configuration-and-initial-data)tutorial, `ve_run` finishes:
```
Starting Virtual Ecosystem simulation.
* Logging to: ve_example\out\logfile.log
* Loading configuration
* Saved compiled configuration: ve_example\out\vr_full_model_configuration.toml
* Built core model components
* Initial data loaded
* Models initialised: animals, plants, litter, hydrology, abiotic_simple, soil
* Starting simulation
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 24/24 [00:04<00:00,  5.34it/s] 
* Simulation completed
* Merged time series data
* Saved final model state
Virtual Ecosystem run complete.
```

But it did not save the `initial_state.nc` file to the `out` directory. My `out` folder only contains:
```
out\.gitignore
out\all_continuous_data.nc
out\final_state.nc
out\logfile.log
out\vr_full_model_configuration.toml
```
and as a result I wasn't able to continue the tutorial from [here ](https://virtual-ecosystem.readthedocs.io/en/latest/using_the_ve/virtual_ecosystem_in_use.html#looking-at-the-results)onwards...

jacobcook1995:
Whether or not the initial state of the model is set by the config option `core.data_output_options.save_initial_state`. This option seems to default to False, which is probably sane in general so that people don't get spammed with files they don't want. 

For the tutorial we need the initial file for the examples to actually work, so the config file for the example run needs to be changed so that this file actually gets outputted.

If you keep this issue open I'll make a pull request to fix this, (and the documentation problem you noticed in #543)

jacobcook1995:
@hrlai I'm having a bit of trouble replicating the bug on my machine, would you be able to send info on what version you are using? (found by calling `ve_run --version`)

Also could you look in your `vr_full_model_configuration.toml` file and look for a tag/heading called `[core.data_output_options]`. Does it exist in the file? If so can you post all of the config options under that heading?

Cheers!

hrlai:
[...]

PR #549, opened 2024-09-05
Release/0.1.1a5

jacobcook1995:
# Description

Release branch to ensure that the downloadable version of the code isn't so outdated that our usage tutorial doesn't actually work (see #544)


davidorme (approved):
As discussed, release to make newer functionality installable for test users.

PR #562, opened 2024-09-16
Improving how excrement and carcass decay is handled

### LINKED ISSUE #510: Animal input into litter model should be improved
In #509, a more consistent way of adding plant input to the litter model has been developed. By comparison the animal inputs are handled in quite a slapdash way.

I should change how the animal inputs are handled so that they follow the same broad procedure as the plant inputs.

### LINKED ISSUE #546: Improve how excrement and carcass decay is handled
At the moment excrement and carcass decay is handled by splitting generated carcasses and excrement into consumable and decayed pools. From a model timing perspective this is probably the least worst option, but it isn't very intuitive.

Also at the moment these pools flow into the metabolic litter pool, which means that the fraction defined as not being consumed by animals can actually be consumed by animals.

Also at the moment all these flows are carbon only when they should also track nitrogen and phosphorus

These flows can be moved to be animal to soil model flows to bypass the litter model entirely. The parameterisation of the split between animal consumption and decay can be generated using a mechanistic equation, which should improve the explainability of this part of the model.

jacobcook1995:
# Description

This PR does three things:
1) stops excrement and carcass biomass from flowing into the litter model, as this meant that we were basically requiring it to decay twice (once in the animal model and once in the litter model) before it entered the soil.
1) adds tracking of excrement and carcass nitrogen and phosphorus. These links are really placeholders at the moment, which should be filled out once animal model stoichiometry is implemented.
1) adds a function to calculate the split between the "scavengable" and "decayed" pools. This was previously handled by a fixed constant split. The change I made basically does the same thing with an extra step, but (hopefully) makes the logic of what we are doing clearer

The excrement and carcass decay flows from the animal model don't actually link to anything at the moment, I'll make an issue to track this, and then fix this when I'm back working on the `soil` model.

Fixes #510

## Type of change

## Key checklist

## Further checks


TaranRallings (approved):
Looks good to me! (weeping softly at my desk)

PR #577, opened 2024-10-03
Add waste from herbivore mechanical inefficiencies to the litter

jacobcook1995:
# Description

This PR adds a new `HebivoryWaste` pool which stores things that herbivores take from plants which don't reach their digestive tracks. At the moment herbivory is only implemented for leaves, so this also only works for leaves, but this should be readily extensible to other plant matter types.

## Type of change

## Key checklist

## Further checks


TaranRallings (approved):
This LGTM. We're going to have to rework it basically immediately anyway lol.

PR #583, opened 2024-10-10
Added handling of herbivory waste products to the litter model

### LINKED ISSUE #510: Animal input into litter model should be improved

jacobcook1995:
# Description

A previous PR (#577) added a flow from the animal model representing waste products from herbivory (e.g. an elephant pulls off a branch but doesn't eat it all and drops the remainder). This PR changes the litter model so that it actually does something with this type of flow (which at present only exists for leaves).

Getting this to work required restructuring of the litter model. Primarily by adding an `InputPartition` class, which collects all the steps needed to calculate the total input of each plant matter type as well as to calculate the split of each plant matter type between litter pools. This class returns an `InputDetails` dataclass which gets used by the rest of the `litter` model. Feedback on this restructure would be very helpful!

Fixes #510

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/litter/input_partition.py
```diff
@@ -1,32 +1,197 @@
...
+    input_below_structural: NDArray[np.float32]
+    """Total input to the below ground structural litter pool [kg C m^-2]"""
+
+
+class InputPartition:
+    """This class handles the partitioning of plant matter between litter pools.
+
+    This class contains methods to calculate the partition of input plant matter between
+    the different litter pools based on the contents of the `data` object.
+    """
```
davidorme: This feels structurally a bit odd - the `InputPartition` class really only exists as a factory method for `InputDetails` and as a thin wrapper around `Data`. I can think of two alternative structures:

* Make `Data` and `constants` as `InitVar` attributes to `InputDetails` and make all the rest of the attributes `field(init=False)`. Then all of the logic in `InputPartition` can be put into a `__post_init__` method for `InputDetails`.

* That does restrict you to only ever populating an instance of `InputDetails` using `Data`, which to be honest sounds like it is the only use case. But you could also have it as a factory class method on the current dataclass. You could then create an instance programatically from dicts of variables.

```python3
@dataclass
class InputDetails:
    ...
    
    @classmethod
    def create_from_data(cls, data: Data, constants: LitterConsts) -> InputDetails:
         ...
```
  You would also need to pass `Data` to `combine_input_sources` - that could then just be a standalone function.

davidorme (changes requested):
I don't love the name `InputDetails` - it's a bit generic. `LitterInputs`? Asides from that I think the science all looks fine, but as I've noted in comments, the `InputPartition`/`InputDetails` implementation feels like a single class thats fallen in half.

jacobcook1995:
@davidorme thanks for the feedback fully agree that the two separate classes were clunky so I have now merged them into one

davidorme (approved):
Yeah - that seems cleaner. I'd be tempted to rename the module though, since that class is now gone. Probably fine with `models/litter/inputs.py`?

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -0,0 +1,349 @@
...
+
+import numpy as np
+from numpy.typing import NDArray
+
+from virtual_ecosystem.core.data import Data
+from virtual_ecosystem.core.logger import LOGGER
+from virtual_ecosystem.models.litter.constants import LitterConsts
+
+
+@dataclass
```
dalonsoa: If the `LitterInputs` objects are not to be modified after creation, make it frozen with `@dataclass(frozen=True)` to avoid mistakes.

dalonsoa (approved):
This looks good, and pretty clean to put all related stuff in its own class. My only comment is that you change the name of the class to `LitterInputs`, but let the variables as `input_details`. I'd change this to `litter_inputs` to be more explicit about what inputs we are talking about.

PR #616, opened 2024-10-28
Adding nitrogen mineralisation

### LINKED ISSUE #605: Soil model needs to track organic nitrogen
The soil model does not currently track organic nitrogen, and I should add this.

jacobcook1995:
# Description

This pull request adds two new pools to track nitrogen in the soil "soil_n_pool_don" and "soil_n_pool_particulate". These pools both receive input from litter mineralisation, and the DON (dissolved organic nitrogen) pool also leaches. Both pools are organic nitrogen pools, so effectively track the nitrogen stoichiometry of their corresponding carbon pools (the low molecular weight carbon (LMWC) pool and the particulate organic matter (POM) pool). The breakdown rate of POM to LMWC therefore determines the rate at which nitrogen goes from the particulate to the dissolved pool.

I've also tried to restructure the code a bit. I've added a new `SoilPools` class, which has method to calculate updates for all soil pools. I still feel like the contents of this method is a bit messy, but I couldn't see an obvious way to group the function steps that would make it clearer. I'm leaving this as is for now, but intended to refactor as further pools are added

Fixes #605

## Type of change

## Key checklist

## Further checks


docs/source/_toc.yaml
```diff
@@ -196,8 +196,8 @@ subtrees:
           - file: api/models/soil
             title: The soil model
             entries:
-              - file: api/models/soil/carbon
-                title: The carbon submodule
+              - file: api/models/soil/all_pools
```
dalonsoa: Calling a python module `all_pools` sounds pretty odd. Why not just `pools`?

virtual_ecosystem/models/soil/all_pools.py
```diff
@@ -1,15 +1,16 @@
-"""The ``models.soil.carbon`` module  simulates the soil carbon cycle for the Virtual
-Ecosystem. At the moment five pools are modelled, these are low molecular weight carbon
+"""The ``models.soil.all_pools`` module simulates all soil pools for the Virtual
+Ecosystem. At the moment four carbon pools are modelled (low molecular weight carbon
 (LMWC), mineral associated organic matter (MAOM), microbial biomass, particulate organic
-matter (POM), and POM degrading enzymes.
+matter (POM)), as well as two enzyme pools (POM and MAOM) degrading enzymes, and two
+nitrogen pools (dissolved organic nitrogen and particulate organic nitrogen).
```
dalonsoa: Add their acronyms as well, for consistency with other pools.

```suggestion
nitrogen pools (dissolved organic nitrogen (DON) and particulate organic nitrogen).
```

virtual_ecosystem/models/soil/all_pools.py
```diff
@@ -64,145 +68,179 @@ class EnzymeMediatedRates:
...
+        """Pools which can change during the soil model update.
+        
+        These pools need to be added outside the data object otherwise the integrator
+        cannot update them and the integration will fail.
+        """
+        self.constants = constants
+
+    def calculate_all_pool_updates(
+        self,
+        delta_pools_ordered: dict[str, NDArray[np.float32]],
```
dalonsoa: As far as I can tell, this pool is an output of this function - new entries are assigned to it -, not an input, so why is it an input argument?

virtual_ecosystem/models/soil/all_pools.py
```diff
@@ -64,145 +68,179 @@ class EnzymeMediatedRates:
...
+        """The data object for the Virtual Ecosystem simulation."""
+        self.pools = pools
+        """Pools which can change during the soil model update.
+        
+        These pools need to be added outside the data object otherwise the integrator
+        cannot update them and the integration will fail.
+        """
+        self.constants = constants
+
+    def calculate_all_pool_updates(
```
dalonsoa: This is a big function, but I don't see a way - or a need - to make it shorter. It is a top level function that just call other functions, so it just need to be big.

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -208,15 +220,19 @@ def integrate(self) -> dict[str, DataArray]:
...
         )
 
         # Find and store order of pools
         delta_pools_ordered = {
             name: np.array([])
             for name in map(str, self.data.data.keys())
-            if name.startswith("soil_c_pool_") or name.startswith("soil_enzyme_")
+            if name.startswith("soil_c_pool_")
+            or name.startswith("soil_enzyme_")
+            or str(name).startswith("soil_n_pool_")
```
dalonsoa: This should not be needed, if it is not needed for the others. 
```suggestion
            if name.startswith("soil_c_pool_")
            or name.startswith("soil_enzyme_")
            or name.startswith("soil_n_pool_")
```

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -208,15 +220,19 @@ def integrate(self) -> dict[str, DataArray]:
             [
                 self.data[name].to_numpy()
                 for name in map(str, self.data.data.keys())
-                if name.startswith("soil_c_pool_") or name.startswith("soil_enzyme_")
+                if name.startswith("soil_c_pool_")
+                or name.startswith("soil_enzyme_")
+                or str(name).startswith("soil_n_pool_")
```
dalonsoa: This should not be needed, if it is not needed for the others. 
```suggestion
            if name.startswith("soil_c_pool_")
            or name.startswith("soil_enzyme_")
            or name.startswith("soil_n_pool_")
```

dalonsoa (changes requested):
I've made a couple of comments/suggestions, but it looks good, otherwise.

virtual_ecosystem/models/soil/soil_model.py
davidorme: I think there's an argument for having an explicit list of these variables somewhere. Partly that's to avoid having repeated string matches, but I also think it's clearer if the model declares the set of of things it cares about.

I think you could declare it as an additional class attribute (don't think the ABC forbids that?) or it could be a GLOBAL variable within the module. Sure, you'd have to maintain it, but that seems better?

virtual_ecosystem/models/soil/all_pools.py
jacobcook1995: This is because `scipy.solve_ivp` can only operate on a `numpy` array, but the data we are trying to update is stored as a dictionary. So something is needed to consistently relate the dictionary entries to specific rows in the`numpy` array. 

The solution I've used here is to create the dictionary that updated output is added to (`delta_pools_ordered`) in a higher level function so that the order of dictionary keys is always the same, which ensures that calling

`np.concatenate(list(delta_pools_ordered.values()))`

always results in a `numpy` array where a given dictionary key (e.g. "soil_n_pool_don") will always be used to fill in the same rows (e.g. rows 1-9).

There might be a smarter way to pass this kind of information about order around, but it didn't occur to me. If you have any suggestions for a cleaner way to approach this that would be useful!

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: Yeah this is a sensible suggestion, particularly as this is basically the list of variables that get altered by the integrator (everything else that gets used from the data object is fixed for any given integration run).

jacobcook1995: This list is actually identical to `vars_updated` (because everything updated by the soil model is updated based on integration) so it probably makes sense to just use that (if I can figure out how to pass it everywhere it needs to go)

virtual_ecosystem/models/soil/all_pools.py
dalonsoa: Ah, ok I get it. So the dictionary format - having keys and values - is a convenience thing. It is not really used except for ensuring the order of the data when transforming it into/from an array, right?

Well, it is pretty neat, to be honest. I cannot think on a better way of doing it. I would, however, explain this somewhere - maybe the docstring of the high level function - otherwise it is quite confusing and difficult to understand why it is the way it is.

jacobcook1995:
@dalonsoa, I've tried to address all your suggestions. Let me know whether the changes look okay. Cheers!

docs/source/_toc.yaml
```diff
@@ -196,8 +196,8 @@ subtrees:
           - file: api/models/soil
             title: The soil model
             entries:
-              - file: api/models/soil/carbon
-                title: The carbon submodule
+              - file: api/models/soil/pools
+                title: The all_pools submodule
```
dalonsoa: Maybe...
```suggestion
                title: The pools submodule
```

dalonsoa (approved):
All looks good!

PR #628, opened 2024-11-29
Added nitrogen limitation for microbial growth

jacobcook1995:
# Description

This PR adds in nitrogen flows between the full set of soil carbon pools and adds nitrogen as a factor limiting microbial growth. The nitrogen content of the microbial biomass (and enzymes) is not explicitly tracked as microbial biomass is assumed to have fixed stoichiometry (which means it is already tracked implicitly).

This pull request basically just adds a few new variables and new functions to update them, there are not any significant structural changes. Let me know if you notice any areas needing improvement!

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/soil/pools.py
```diff
@@ -755,3 +876,111 @@ def calculate_soil_nutrient_mineralisation(
...
+            growth, death enzyme production, etc
+        constants: Set of constants for the soil model.
+
+    Returns:
+        The rate at which nitrogen is added to the necromass pool [kg N m^-3 day^-1]
+    """
+
+    return np.divide(
+        microbial_changes.necromass_generation, constants.microbial_c_n_ratio
+    )
```
dalonsoa: Out of curiosity, why not just using `/` to perform the division? Aren't these normal arrays already?

virtual_ecosystem/models/soil/pools.py
```diff
@@ -755,3 +876,111 @@ def calculate_soil_nutrient_mineralisation(
...
+) -> NDArray[np.float32]:
+    """Calculate the rate at which nutrients flow into the necromass pool.
+
+    These flows comprise of the nitrogen and phosphorus content of the dead cells and
+    denatured enzymes that flow into the necromass pool.
+
+    TODO - A core assumption here is that the stochiometry of the enzymes are identical
+    to the microbial cells. This assumption works for now but will have to be revisited
+    when fungi are added (as they have different stochiometric ratios but will
+    contribute to the same enzyme pools)
```
dalonsoa, line 892: I guess you have a separate function for this and the chosen inputs as there might be other, more complex stuff going on in the future in this function, right? Otherwise, this functions seems a bit overkilling for a simple division.

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -124,6 +130,8 @@ def __init__(
             or np.any(data["soil_c_pool_necromass"] < 0.0)
             or np.any(data["soil_n_pool_don"] < 0.0)
             or np.any(data["soil_n_pool_particulate"] < 0.0)
+            or np.any(data["soil_n_pool_necromass"] < 0.0)
```
dalonsoa: Considering how big this condition is, I wonder if it would be better to make it a separate function. Just for readability purposes. Something like:

```python
def _all_pools_positive(data: Data) -> bool:
    """Checks if all pools have zero or positive values.
    return ...
```

So this becomes:

```python
if not _all_pools_positive(data):
    raise ...
```

dalonsoa (approved):
Just a couple of suggestions and questions, but looks good!

virtual_ecosystem/models/soil/pools.py
jacobcook1995: Not actually sure, I just noticed that I was using `np.divide()` elsewhere and assumed that it was generically best practice, so used it here. Is the advantage of using `np.divide()` that it allows you to convert lists of numbers to `numpy` arrays?

dalonsoa: Pretty much. You can use array-like objects like lists with `np.divide` while you can only use proper arrays with `/`. I don't think it is a better practice, but rather serves to different purposes.

virtual_ecosystem/models/soil/pools.py
jacobcook1995: Yep in future, the microbial pool will be separated into (at least) a bacterial pool and a fungal pool. These pools will have different stoichiometry which means the calculation will become more involved, so the function is basically a placeholder for when I will do this.

PR #636, opened 2024-12-10
Streamline getting started documentation

### LINKED ISSUE #631: Streamline "Getting Started" documentation
- Remove the "Running an example.." section from the [Installing VE page](https://virtual-ecosystem.readthedocs.io/en/latest/using_the_ve/getting_started.html#running-an-example-virtual-ecosystem-simulation) and clarify the [Using the Virtual Ecosystem](https://virtual-ecosystem.readthedocs.io/en/latest/using_the_ve/virtual_ecosystem_in_use.html) page.
- Other typos / corrections

sallymatson:
# Description

Updating the getting started documentation to remove redundancies between the Getting Started page and Using the VE page. For now, I have separated them out so that the Getting started is focused on installation/running the simulation for the first time, and the "Using the VE page" (which I renamed to "Exploring the VE outputs") is more to go through a few ways to interact with the outputs / visualize. 

Fixes #631 

## Key checklist

## Further checks


sallymatson:
Hi @hrlai , I've had a first go at it. Would love to hear your thoughts - if it would be easier to talk on Teams would be happy to set up a call.

davidorme:
@sallymatson The `precommit.ci` run has picked up a problem. It's really trivial - `markdownlint` wants to remove a trailing space - but it suggests that your development environment has an issue with `precommit` - either not installed or not configured. If you run `poetry run pre-commit run -a`, what happens?

sallymatson:
> @sallymatson The `precommit.ci` run has picked up a problem. It's really trivial - `markdownlint` wants to remove a trailing space - but it suggests that your development environment has an issue with `precommit` - either not installed or not configured. If you run `poetry run pre-commit run -a`, what happens?

Yes, that would be my fault for not running`pre-commit`. All set now :)

davidorme:
It should be running automatically when you `git commit`? Does it?

sallymatson:
> It should be running automatically when you `git commit`? Does it?

It does. The instructions on the doc were perfect ! Do you have any thoughts about the one build that failed?

davidorme:
> Do you have any thoughts about the one build that failed?

Yup - our CI workflow runs the pre-commit `qa` job and then does two things: runs the `test` suite on a bunch of OS and python version combos, but also tries to build the `sphinx` docs and it is that one that has failed. If you dive into the build details, there's a cross reference to a different file that fails - it's just a typo for `overview` but our QA testing is set up to be ultra picky.

https://github.com/ImperialCollegeLondon/virtual_ecosystem/actions/runs/12277162911/job/34255998318#step:6:401

docs/source/using_the_ve/getting_started.md
```diff
@@ -20,7 +20,7 @@ language_info:
   name: python
   nbconvert_exporter: python
   pygments_lexer: ipython3
-  version: 3.11.9
+  version: 3.12.0rc3
 ---
 
 # Getting started
```
hrlai, line 26: At the first mention of `pip` or a command, should we assume less coding experience of the users and say explicitly where to run it? For example, before `pip install` below, say something like "...open up the Terminal and type:"

docs/source/using_the_ve/getting_started.md
```diff
@@ -56,18 +56,36 @@ configuration and data files to run a model.
 ve_run --install-example /path/
 ```
 
-You can then run the model itself:
+You can then run the model itself. If you have already run the simulation you will need
+to delete or rename the output files, as previously generated output can prevent the
```
hrlai, line 60: Even though one could just remove the files using file explorer, should we give an example code to remove the files (e.g., using `rm`) for those who have little experience with the Terminal?

If so, probably write it after the `ve_run` chunk for a better flow...

docs/source/using_the_ve/getting_started.md
```diff
@@ -56,18 +56,36 @@ configuration and data files to run a model.
...
+## Simulation results
+
+The Virtual Ecosystem writes out a number of data files:
+
+* `initial_state.nc`: A single compiled file of the initial input data.
+* `all_continuous_data.nc`: An optional record of time series data of the variables
+  updated at each time step.
+* `final_state.nc`: The model data state at the end of the final step.
+
+These files are written to the standard NetCDF data file format.
```
hrlai, line 80: Maybe insert example code to remove them here.

docs/source/using_the_ve/getting_started.md
```diff
@@ -56,18 +56,36 @@ configuration and data files to run a model.
...
+
+## Next steps
+
+* To explore the simulation results further you can visit the [Visualising Virtual
+Ecosystem Output](virtual_ecosystem_in_use.md) tutorial, which walks you through basic
+graphs using model inputs and outputs.
+* The [Example Data](./example_data.md) pages provides a detailed description of the
+contents of the `ve_example` directory. Here you can dig into the strucutre of the
```
hrlai: Typo "strucutre" -> "structure".

docs/source/using_the_ve/getting_started.md
hrlai: In the other file I just realised that @jacobcook1995 had already written something, maybe ditto them here for completeness?

docs/source/using_the_ve/virtual_ecosystem_in_use.md
```diff
@@ -19,105 +19,68 @@ language_info:
...
-import matplotlib.dates as mdates
-import numpy as np
-import xarray
-```
-
-If you have previously attempted to run this example it is probably a good idea to
-delete the existing virtual ecosystem example directory, as previously generated files
-can prevent the example simulation from running successfully. That can be done as
-follows.
+The following commands allow you to run the simulation from a Jupyter Notebook.
```
hrlai, line 38: It is great to remind the users that they can run the python codes using Jupyter Notebook. However, I struggle to even comprehend where to type the following commands after launching Jupyter Notebook, sorry! Most ecologists or biologists would be more familiar with R than python, so it would be a shame to turn them away because we didn't ease them into python. 

That said, I struggle between providing too much details about python coding here (some users may already know how and find it boring) versus hand-carrying new users. Should we dedicate a new page just for using python codes in Jupyter, or at least point the users to an external tutorial page?

hrlai (changes requested):
Great start towards a clearer start-up tutorial! I have added my two cents, and the most major one is possibly to hand-carry the users more around how to run python codes using Jupyter Notebook...

docs/source/using_the_ve/virtual_ecosystem_in_use.md
```diff
@@ -19,105 +19,68 @@ language_info:
...
 ```{code-cell} ipython3
+---
+editable: true
+slideshow:
+  slide_type: ''
+tags: []
+---
 %%bash
 # Remove any existing VE data directory in the /tmp/ directory
 if [ -d /tmp/ve_example ]; then
```
hrlai: Can this be simplified as `rm -rf /tmp/ve_example` ?

davidorme: Huh - I don't think of the `-f` flag that way, but it does explicitly _not_ raise an error if the file is missing. We should note that is why we're using it, but it is less complex than the explicit test.

docs/source/using_the_ve/virtual_ecosystem_in_use.md
```diff
@@ -19,105 +19,68 @@ language_info:
...
+However, you can also run the script from the command line following the [Getting
+Started](getting_started.md) instructions.
 
 ```{code-cell} ipython3
+---
+editable: true
+slideshow:
+  slide_type: ''
+tags: []
+---
```
davidorme: I have _not_ figured out why Jupyter sometimes chucks some of these tags in or how to filter them out. I think we can do so with the myst filters in jupytext, and we definitely should because we do not want this!

docs/source/using_the_ve/virtual_ecosystem_in_use.md
```diff
@@ -129,12 +92,48 @@ The Virtual Ecosystem writes out a number of data files:
...
+
+```{code-cell} ipython3
+---
+editable: true
+slideshow:
+  slide_type: ''
+tags: []
+---
 # Load the generated data files
 initial_state = xarray.load_dataset("/tmp/ve_example/out/initial_state.nc")
```
hrlai, line 89: If a user install the develop branch using `pip` they probably won't have the initial state file by default settings, see https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/544 I can confirm that this is still the case today on my computer.

docs/source/using_the_ve/virtual_ecosystem_in_use.md
```diff
@@ -129,12 +92,48 @@ The Virtual Ecosystem writes out a number of data files:
...
+# Print the name of each variable in the final state
+for key in list(final_state.keys()):
+    print(key)
+```
+
++++ {"editable": true, "slideshow": {"slide_type": ""}, "tags": []}
+
 ### Initial state and input data
 
 The `initial_state.nc` file contains all of the data required to run the model. For some
```
hrlai, line 104: In the plot below, we reshape the vector output of elevation to be a matrix of dimension 9 x 9. Can we avoid hard-coding the row and column numbers from 9 to something from the netCDF output? I imagine someone to change the grid numbers eventually so this will help to automate the plotting a bit more.

davidorme: It's actually going to need to be a new function to reverse the internal representation of cells (which is a 1D array) back onto the grid. So that's going to be a general helper function (`convert_cell_data_to_grid` or similar).

But your point is a good one!

ETA - Actually that get's interesting. You need the grid configuration to go back to the 2D, so plotting spatial data will need a helper class (`SpatialPlotter`) that is created using the Grid config and then has a method (`to_spatial`) that maps the cell_id back onto the spatial layout.

hrlai: Awesome, something like `convert_cell_data_to_grid` would be a great addition indeed. Is the grid configuration (e.g., number of rows and columns) stored anywhere in the output files?

docs/source/using_the_ve/getting_started.md
sallymatson: I think it's generally not great practice to use bash scripts to delete directories, because you might end up deleting something by mistake. It's useful in that example because the files are being stored in a kind of weird location (in the user directory) which might be hard to find using Finder. But I personally would encourage users to keep the data files in the `virtual_ecosystem` directory, because they will likely need to access them while working, and it will all stay a lot more organized that way. That is why I used this method - and in addition, I just think it overcomplicates the task to run a bash script, when the files should be easy to find. 

Let me know what you think about that reasoning!!

docs/source/using_the_ve/getting_started.md
sallymatson: See above - I think it's better in this tutorial to not have a script for that. LMK if you feel strongly about including it!

docs/source/using_the_ve/getting_started.md
hrlai: All good. I agree that it is unnecessary :) The only thing left to decide if whether to list the filenames to be deleted, but this is not always the same (e.g., the initial states may not be saved depending on user setting.) When I first bumped into this I had to ask around which files to remove.

docs/source/using_the_ve/getting_started.md
hrlai: Sounds good to me, replied above :)

sallymatson:
Going to close this. I think the issue still stands, but I think from this conversation / other chats in meetings and over email, it seems like we need a bit of a deeper think about what is most useful for the introduction to VE docs. I still have this on my radar to work on soon, just going to wait a bit to make sure I do it right!

PR #641, opened 2024-12-16
Revert "Bump numpy from 2.1.3 to 2.2.0 in the production-dependencies group"

davidorme:
Reverts ImperialCollegeLondon/virtual_ecosystem#639

Numpy 2.2.0 changes some typing handling that raises issues - this failed `pre-commit.ci` as a result but that check wasn't required. So I'm backing this one off until it can pass.

I've made `pre-commit.ci` one of the required checks now - but because of the way it runs, we're getting that same failure. I'll have to force this push, but does this seem reasonable?


vgro (approved):
I'm not quite sure what's going on but I am in favor of reversing a change that broke everything ;-)

PR #653, opened 2025-01-06
Adding organic phosphorus flows

jacobcook1995:
# Description

This pull request adds organic flows of phosphorus to the soil model. This involved extending the functionality developed to track the organic nitrogen cycle to also track the organic phosphorus cycle. The scientific content is a bit dense here, but would be good to get feedback on the code structure/clarity.

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
The code itself looks OK - it pretty much follows what nitrogen is doing, as far as I can tell.

However, I've seen that now there are several functions returning dictionaries rather than tuples that can be unpacked, which is less readable. I wonder if it would be better to create dataclasses, as it is done for `MicrobialChanges`, to contain those outputs.

The same might apply to `pools`, which is a very large dict and that could be more more readable if it were a dataclass. 

None of this is a request for a change, just observations.

jacobcook1995:
@dalonsoa yes I did wonder whether returning dataclasses rather than dictionaries would be better practice. Good to know!

The next thing I'm working on will be adding inorganic nitrogen and phosphorus pools in, which will involve altering some of functions added/changed here. With that in mind I will merge this now and switch outputs to dataclasses when I'm working on the new code

PR #658, opened 2025-01-07
Add a data method to validate data frame variables

### LINKED ISSUE #657: Add a data method to check variables form a data frame
Both the plants and animal models need to handle cohort data provided as a data frame (see #656 for details). These particular use cases need to validate that a set of variables loaded into the central `Data` object do indeed form a data frame:

* They are all present.
* They are all one dimensional variables of equal length.

This should be shared functionality, so a new method of `Data`.

davidorme:
# Description

This PR adds a new method to `Data` that can be used to check that a named set of variables for a data frame: a set of one-dimensional arrays with the same length.

This is motivated by the need for both the plants and animals models to handle cohort data, which naturally forms a data frame.

Fixes #657 

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
This looks logical & thorough to me!

PR #661, opened 2025-01-08
Added soil model abbreviation glossary and `hoverxref` glossary tool tips 

### LINKED ISSUE #479: Glossary terms for abbreviations in documentation
The `soil` model (and I'm guessing the other models) uses a lot of acronyms, e.g. MAOM for mineral associated organic matter. Having to write out these acronyms every time in docstrings/documentation is tedious, but just writing out the acronyms leads to fairly unreadable documentation.

The `sphinx` extension [`hoverxref`](https://sphinx-hoverxref.readthedocs.io/en/latest/usage.html#tooltip-on-glossary-terms) allows us to define a glossary which is then used to generate tooltips explaining terms (and in this case acronyms). I will look into this for the soil model as it is pretty acronym heavy

jacobcook1995:
# Description

I've added a short glossary of soil model pool name abbreviations. I have also added the `sphinx.hoverxref` package, which I setup so that glossary item definitions can be viewed by tooltips. This is basically just a short demonstration to show the functionality. I plan on using this functionality as I extend/improve the soil model documentation.

I chose to add the glossary as a page within the soil model theory docs, but if others think that it's better to have a common glossary elsewhere I'm completely fine with doing that instead.

With `hoverxref` added we can basically add tooltips for any type of reference that we use. I did consider doing this for `intersphinx` references and for bibliographical references but decided against as I thought it could bog this PR down in side arguments.

I've built the docs and they can be viewed [here](https://virtual-ecosystem.readthedocs.io/en/479-glossary-terms-for-abbreviations-in-documentation/). The tooltip examples can be seen [here](https://virtual-ecosystem.readthedocs.io/en/479-glossary-terms-for-abbreviations-in-documentation/virtual_ecosystem/theory/soil/soil_theory.html#microbial-biomass) and [here](https://virtual-ecosystem.readthedocs.io/en/479-glossary-terms-for-abbreviations-in-documentation/api/models/soil/pools.html#virtual_ecosystem.models.soil.pools.calculate_enzyme_mediated_rates).

Fixes #479

## Type of change

## Key checklist

## Further checks


davidorme (approved):
This looks great - very clean and saves a whole bunch of repeated definitions. I think we should have a project wide glossary. It might get very long, but it's all crossrefed with links so I don't think that matters.

Watch for it being poached in `pyrealm` very soon!

jacobcook1995:
@davidorme, great! Yes if others also want to use this for their docs I agree that a project wide glossary makes most sense. Would the best place to put it be inside the "Reference" section?

sallymatson:
Nice !! @jacobcook1995 from my perspective a project-wide glossary under Reference seems like a great idea.

sallymatson (approved):

vgro (approved):
Looks cool, very good idea

PR #664, opened 2025-01-09
Csvxlsx file reader

### LINKED ISSUE #656: Add a CSV/XLSX file reader to core
Both the plants and animals models require the users to provide cohort data. For plants, this is providing tuples of data:

```
(cell id, plant functional type, number of individuals, individual size)
```

There can be multiple entries per cell id and different numbers of cohorts per cell. The easiest and sanest format for this data is a simple data frame of those tuples and the natural format for creating and maintaining that data is a CSV or XSLX file. Forcing users to convert this into NetCDF for input is not sensible.

So, we need to:

* Add a CSV/XLSX loader.
* This should use `pandas` as that is already a requirement of `xarray` and is designed explicitly to handle data frames, rather than using the standard library `csv` or any of the `numpy` structures.
* I think we will need to explicitly add `openxlsx` to `[tool.poetry.dependencies]` to support reading XLSX format.
* Test that it works!

It should go in `virtual_ecosystem.core.readers` and I think the signature will look like:

```python
@register_file_format_loader(file_types=(".csv", ".xlsx"))
def load_from_dataframe(file: Path, var_name: str) -> DataArray:
    """Loads a DataArray from a data frame format."""
```

The format registry should then automatically switch to using this loader for CSV and XLSX files.

There is some ugliness here in that the file is going to be opened multiple times to load each variable as we don't have persistent file handles, but the same is currently true for NetCDF. A better way to do this in future would be to open each _file_ within the data configuration _once_ to access a tuple of variables that are claimed to live in that file, rather than independently opening the file specified for each variable.

sallymatson:
# Description

Added a dataframe reader that uses pandas to handle csv and excel input. Wrote tests for the functionality of both. Added `openpyxl` to poetry as a dependency for reading excel files. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/core/readers.py
```diff
@@ -134,6 +136,59 @@ def load_netcdf(file: Path, var_name: str) -> DataArray:
...
+        LOGGER.critical(to_raise)
+        raise to_raise
+    except ParserError as err:
+        to_raise = ParserError(f"Could not load data from {file}: {err}.")
+        LOGGER.critical(to_raise)
+        raise to_raise
+    except Exception as err:
+        to_raise = Exception(f"Unidentified exception opening {file}: {err}")
+        LOGGER.critical(to_raise)
+        raise to_raise
```
davidorme, line 211: I think it's worth explicitly adding BadZipError - it's by far the most likely thing to happen and we can effectively report it as "Does not appear to be an Excel file". The general Exception can then handle anything else. I'd add a comment on that exception to say that `openpyxl` has a whole bunch of assorted failure modes for corrupt/non-excel Zip files and we're not going to try and track down an exhaustive list.

You can then add an actual zip file of (e.g.) a simple text file to test that failure mode.

davidorme (changes requested):
This all looks good. I've requested a minor change on the failure modes.

One thing - I'm not 100% sure about how the `try` stack is a mix of CSV failure modes and Excel failure modes without it being clear which one is which. That seems harder to maintain. We could split them, but then you'd duplicate the FileNotFoundError, unless the code specifically tests for the file using `Path().exists()` first - but that's bad practice because the file _could_ in theory vanish between the two calls.

Should we just have separate CSV and Excel reader functions? It's more verbose but feels cleaner. Thoughts?

virtual_ecosystem/core/readers.py
sallymatson: Will do!!

sallymatson:
> This all looks good. I've requested a minor change on the failure modes.
> 
> One thing - I'm not 100% sure about how the `try` stack is a mix of CSV failure modes and Excel failure modes without it being clear which one is which. That seems harder to maintain. We could split them, but then you'd duplicate the FileNotFoundError, unless the code specifically tests for the file using `Path().exists()` first - but that's bad practice because the file _could_ in theory vanish between the two calls.
> 
> Should we just have separate CSV and Excel reader functions? It's more verbose but feels cleaner. Thoughts?

Yeah, it already pained me a bit to have the duplication from `load_from_netcdf` because they are so similar (e.g. syntax for checking if the variable is in the dataset), but you're right that it would all be more explicit if they are separate. Especially since they `read_csv` and `read_excel` have relatively different behaviour. I can split them up :)

virtual_ecosystem/core/readers.py
sallymatson: I added a corrupted excel file which triggers that error - I think that should cover that failure case?

sallymatson:
@davidorme I pushed some changes to split up the tests and now the builds are failing. It seems like it is a problem with the poetry version in `ci.yml`, but that doesn't make sense because I didn't change any files related to that. I have a feeling there is something silly that I've done, does it set of any immediate lightbulbs for you?

davidorme:
@sallymatson No - that's not you. For reasons I don't understand, `dependabot` bumped the `poetry-action` version to `4.0.0` a few days ago (https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/652).

The only problem being `4.0.0` doesn't exist. I can't find any indication it ever did. No idea what happened.

 Can you push a change to to `ci.yml` in this branch setting:

```yaml
    - name: Install Poetry
      uses: abatilo/actions-poetry@v3.0.2
      with:
        poetry-version: 1.5.1
```

davidorme:
Aha! It did exist briefly, and the maintainer pulled it.  https://github.com/abatilo/actions-poetry/issues/88

davidorme:
@sallymatson Is this ready for re-review? If so, can you re-request review?

davidorme (approved):
LGTM. I guess I'd also add something to the `load_excel` that says at present the data is explicitly only loaded from the first sheet. We could fix that, but not now?

PR #669, opened 2025-01-10
Adding utils/fixture_mapper

### LINKED ISSUE #668: Test fixture usage tool
A while back, I wrote a short script to profile the fixtures in the `pytest` suite and then generate some simple stats and a graph representation of fixture usage by tests. The aim here is to allow us to identify potential duplicate fixture functionality and lower the maintenance load on the test suite.

We might never use it again, but it's sitting in my local clone at the moment and seems worth hanging on to in case we do want it again, even if it is just to count the tests...

davidorme:
# Description

This PR adds the `utils` folder and the `fixture_mapper` tool. This is a python script that calls `pytest` to discover all the fixtures in the test suite and then identify which tests use which features.

When it runs, it outputs this text and a couple of files. One is a simple list of tests with their fixtures, the other is a graph of tests and fixtures.

```
(virtual-ecosystem-py3.11) ➜  fixture_mapper git:(659-use-pyrealm-flora) ✗ python fixture_mapper.py
Running: pytest ../../tests --fixtures
Running: pytest ../../tests --setup-plan
Fixture report generated
Fixture graph generated:
Tests found: 342
Fixtures found: 74
Number of fixture usages: 989
```

It does require `networkx` but this is such a low usage tool that I've left it off the `poetry` requirements. There are no tests and documentation.

Fixes #668

## Type of change

## Key checklist

- [] All tests pass: `$ poetry run pytest`

## Further checks


utils/fixture_mapper/tests_fixtures.graphml
jacobcook1995, line 1: Should this file actually be committed? because surely it is immediately out of date as soon as tests are added, and I guess we don't want to be forced to maintain it

utils/fixture_mapper/tests_fixtures.txt
jacobcook1995, line 1: Same question as above

jacobcook1995 (approved):
LGTM, just had a question about whether we want to be committing the output files

davidorme:
That's a fair point, I guess they're quite small, text and will change infrequently, and it might be useful to look at an example to understand the script? We can always remove them later?

jacobcook1995:
Yeah I guess if the are just there as something for people to look at to get a broad picture, that gets updated from time to time that's fine.

It's just keeping them up to date that I am sceptical about as they will change (slightly) every time a test is added, renamed, or changed such that it uses a different set of fixtures (i.e. for the majority of PRs)

davidorme:
OK - on balance, I think you're right. We shouldn't keep outdated files around.

PR #673, opened 2025-01-13
Updating repo to be compatiable with `ruff``'s 2025 stable style

jacobcook1995:
# Description

`ruff` have released [their 2025 stable style](https://astral.sh/blog/ruff-v0.9.0), which every `ruff` version after `0.9` will now use. This changes the formatting rules slightly which was generating a weird back and forth for me, where my VScode `ruff` extension would change the formatting which the `ruff` version in pre-commit would then reject.

This PR fixes that issue by updating the `pre-commit` hooks to use a post `0.9` version of `ruff`. I have also reformatted all the code and test files to bring them up to date with the new format.

The most recent version of the `ruff` VSCode extension defaults to `>0.9.0`, so @davidorme, @TaranRallings, @vgro, @sallymatson  if you are using this extension you just need to make sure that it is up to date to avoid a clash between it and the `pre-commit` version

## Type of change

## Key checklist

## Further checks


davidorme (approved):
The `pyproject.yaml`  file has  already been updated by `dependabot` right? Other than that - looks great. I think I even like the changes being forced on us 😄

jacobcook1995:
@davidorme, `ruff` isn't currently included as a dependency in the `pyproject.toml` which feels like an oversight?

davidorme:
Yeah - since it _can_ be installed via pip we probably should do. We could also add it to the suggested VSCode extensions in `.vscode/extensions.json`

PR #678, opened 2025-01-15
Windows specific getting started instructions

jacobcook1995:
# Description

This PR adds Windows specific instructions for the "getting started page". This is done using tabs and can be seen [here](https://virtual-ecosystem.readthedocs.io/en/bugfix-add_windows_getting_started/using_the_ve/getting_started.html).

## Type of change

Documentation improvement


sallymatson (commented):
Looks good to me !! :)

davidorme:
The tabs are very pretty! This is Gitbash only though? Wouldn't windows paths need `C://` etc. Can we also extend to the installation process and checking paths work. I realise that is extending the scope of the PR!

jacobcook1995:
This is all windows powershell (which is default installed right). 

I can change the path to be `C://path/` rather than `path` if you think that's clearer (guess it's safe to assume that anyone saving to a drive other than C knows what they are doing)?

jacobcook1995:
The installation via `pip` is the same regardless of whether you use powershell or bash, obviously the installation of `python` differs but I was planning to just send the "How to install python on windows" instructions written by Python (which seems better than trying to write our own).

Where is the "checking paths work" section?

davidorme:
> Where is the "checking paths work" section?

There isn't one at present, but if a user installs Python and then `pip` fails or installs `VE` and then `ve_run` fails then they need to know how to fix it. Actually I don't know how to fix that latter one! `pip` should register the entry points but I don't know if the windows path then automatically finds them.

Completely agree on the follow the windows official instructions, but there are details on adding the path that are on that page that we should point at.

jacobcook1995:
> Completely agree on the follow the windows official instructions, but there are details on adding the path that are on that page that we should point at.

Wouldn't they ideally be installing python in a way that doesn't require them to add the path themselves though? (at least for current purposes)

jacobcook1995:
> > Where is the "checking paths work" section?
> 
> There isn't one at present, but if a user installs Python and then `pip` fails or installs `VE` and then `ve_run` fails then they need to know how to fix it. Actually I don't know how to fix that latter one! `pip` should register the entry points but I don't know if the windows path then automatically finds them.

Are there obvious common ways that `pip` could fail to install that we should highlight? I'm just concerned that if we tried to cover all the possibilities this page quickly becomes a "how to avoid common python setup errors" page

sallymatson:
I hope adding this about the required python version and rec of pyenv/pyenv-win doesn't start to make this too unweildly. @vgro  and I were just going through the installation while trying to run in static mode and found the python versioning a bit of a tricky one, especially for users unfamiliar with python, I think it's important to signpost them a bit.

sallymatson (approved):

Issue #666, opened 2025-01-10
Repair model attribute typing and docstrings


davidorme:
With the change to the static model mechanism (#373), model attributes are now largely being defined in the `_setup` method, which means that the actual model class documentation is missing the model attributes: they are only picked up from `__init__` or from class attributes.

We need to fix this - our docs are broken until we do. From @dalonsoa's comments on Teams, we have two clean choices. 

1) We start using `__init__` again in the models. It's banned at the moment - only the `BaseModel.__init__` gets run and it can't be redefined. This is because `__init__` shouldn't do anything but setup the static/non_static model mode and all actual stuff that needs to happen is isolated in `_setup`, so it can conditionally executed based on the model mode. But we can just agree that actual models `__init__` methods should be "logic-free" and are used simply to maintain the documentation and typing of model attributes. They _have_ to call `super().__init__()`.

2) We document and type all of the model attributes as if they were class attributes (below the `class ModelName` line and outside of any method). We then don't have to use `__init__`.

I prefer option (1). Although it is a bit more verbose, it follows the canonical Python practice of declaring **instance** attributes in `__init__`. These are attributes that differ between instances, which is true for all of these attributes. In option 2, we are pretending they are **class** attributes, which should be things that are fixed across all class instances. The only reason we can do this is because we don't actually want to assign values to the attributes at a class level, only at an instance level.

The code below shows the two options in a basic setup. The issue we are trying to fix is that attribute `eee` defined in `_setup` is not documented in `sphinx` (and I don't think _can_ be), and nearly all our model attributes are now in this state.

```python
from abc import ABC


class Base(ABC):
    """Base class docstring."""

    base_class_attr: int
    "A class attribute"

    def __init__(self) -> None:
        """Base init docstring."""
        self.aaa: int
        "An integer"
        self.bbb: str
        "A string"


class Actual(Base):
    """Actual class docstring."""

    def __init__(self) -> None:
        """Actual init docstring."""
        super().__init__()

        self.ccc: int
        "Another integer"

        self.ddd: str
        "Another string"

    def _setup(self) -> None:
        """Actual setup docstring."""

        self.eee: int
        "This attribute is not documented"


class Base2(ABC):
    """Base2 class docstring."""

    base_class_attr: int
    "A class attribute"

    def __init__(self) -> None:
        """Base2 init docstring."""
        self.aaa: int
        "An integer"
        self.bbb: str
        "A string"


class Actual2(Base):
[...]

PR #682, opened 2025-01-16
681 setup skipped improperly

### LINKED ISSUE #681: _setup skipped when static = True and the variables populated by init are not provided
**Describe the bug**

The intended behaviour of `_bypass_setup_due_to_static_configuration` is that it only returns `True` (i.e. bypasses the `_setup` function) if the model is set to static AND no `vars_populated_by_init` are present: 

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/913a228e4514e42e9641e161ab41f900aa24f41e/virtual_ecosystem/core/base_model.py#L272-L274

However, if static is true and `vars_populated_by_init` is 0 (meaning that `found = 0`), then the function returns `True` based on this logic:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/913a228e4514e42e9641e161ab41f900aa24f41e/virtual_ecosystem/core/base_model.py#L287-L296

I believe that in the case where `found=0` and `static=True` then it should return `False`, correct? If so I will fix this. Please let me know if I'm missing something here!!

sallymatson:
# Description

Fixed the bug where `_setup` skipped in static model when none of the init vars are provided. 

Fixes #681 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/core/test_base_model.py
+++ b/tests/core/test_base_model.py
@@ -575,7 +575,7 @@ def test_check_update_speed(
             True,
             ("var1", "var2"),
             {},
-            True,
+            False,
             does_not_raise(),
             None,
             id="static_no_vars_present",
--- a/virtual_ecosystem/core/base_model.py
+++ b/virtual_ecosystem/core/base_model.py
@@ -292,7 +292,11 @@ class BaseModel(ABC):
                     f"all to be absent. {found} out of {expected} found: "
                     f"{', '.join(present)}."
                 )
+            elif found == 0:
+                # The case when static is true and no init vars provided
+                return False
             else:
+                # The case when static is true and all init vars provideed
                 return True
 
         return False


vgro (approved):
Makes sense to me based on our discussion earlier, I give the creators the final word as I am not sure what the initial motivation was

dalonsoa (approved):
Yep, this particular case was an oversight from my side. The extra logic looks correct.

However, I think I was a bit naive when coding the bypassing rules of the `_setup` and we might need to deal a more complex logic. The `_setup` method not only populates the `data` object with some variables but, in general, can initialise some initial state for the model, only relevant internally. If we skip `_setup`, that initialisation of the state does not happen.

This is not relevant for the purpose of this pull request, which solves the issue it referrers to, but I feel we are going to hit a wall sooner or later...

davidorme:
> in general, can initialise some initial state for the model

@sallymatson and I were thinking about that yesterday. In the "update once and freeze" mode, you have to fully parameterise the model, `_setup` and `_update` have to run, so all the internals get populated correctly. In "use fixed data" mode, _nothing_ runs outside `__init__`. By design that is logic-free (even after the changes proposed in #666), so although attributes are not populated, they are guaranteed not to be used. We think it's ok?

dalonsoa:
@davidorme , @sallymatson  I think that in total we have 4 cases, with one of them ill-defined:

| Static | _setup | _update | Comments                                                         |
| ------ | ------ | ------- | ---------------------------------------------------------------- |
| No     | Yes    | Yes     | Normal run                                                       |
| Yes    | Yes    | Once    | Update once and freeze                                           |
| Yes    | No     | Once    | Ill defined. Setup is skipped but update is expected to run once |
| Yes    | No     | No      | Use fix data                                                     |

The third one is the problem. As the code is now, it is allowed: you just provide all the variables that need to be populated by init but none of those that need to be populated by update. As a result, the `_setup` method is not run and none of the internals that the `_update` method will need are there.

If this is not a use case, then we need to fix the code to prevent that from happening.

davidorme:
@dalonsoa Yeah - I think the conclusion we came to was that case 3 makes no sense and never will. The `update` _requires_ anything specified in the `init` variable listings (either required by or populated by), so it should be impossible to ever make that combination work.

Issue #676, opened 2025-01-14
Update "how to create a new model" documentation


davidorme:
We have #686 , which is user facing documentation, but our developer facing documentation on creating a new model is now wildly out of date, with the changes from the variables update, the static model and then the restructure in #666 of the required methods.

So we need to update this file:
https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/develop/docs/source/development/design/defining_new_models.md?plain=1

davidorme:
This will be fixed when #1193 is merged - that PR updates the flow used in `__init__` and I've overhauled the new model page.

PR #692, opened 2025-01-21
666 repair model attribute typing and docstrings

### LINKED ISSUE #666: Repair model attribute typing and docstrings

sallymatson:
# Description

Move attribute declarations out of `_setup` to subclass `__init__` functions. The attributes are now defined and documented in the proper place (e.g. see [updated hydrology docs](https://virtual-ecosystem.readthedocs.io/en/666-repair-model-attribute-typing-and-docstrings/api/models/hydrology/hydrology_model.html#module-virtual_ecosystem.models.hydrology.hydrology_model) as compared to [old hydrology docs](https://virtual-ecosystem.readthedocs.io/en/latest/api/models/hydrology/hydrology_model.html#module-virtual_ecosystem.models.hydrology.hydrology_model)).

It would be great to get general feedback, as well as everyone looking over the changes to their modules in order to make sure the variables are all typed correctly etc @vgro @davidorme @jacobcook1995 @TaranRallings. 

Fixes #666 

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
The changes look good, and indeed this is a more canonical approach than using fake class variables.

Probably the docs should be updated to explain what goes in `__init__` and what in `_setup` when creating a new model.

davidorme (changes requested):
This is looking good. Can you update the doc API files for the models to include inherited members? I think that's what we want as we currently can't see anything being inherited from the base class.

For example:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/7838fa8fa6dfa884748cd330464d27ae58e781f8/docs/source/api/models/plants/plants_model.md?plain=1#L29-L36

should go to (I think):

````md
```{eval-rst}
.. automodule:: virtual_ecosystem.models.plants.plants_model
    :autosummary:
    :members:
    :exclude-members: model_name
    :inherited_members:
```
````

sallymatson:
> This is looking good. Can you update the doc API files for the models to include inherited members? I think that's what we want as we currently can't see anything being inherited from the base class.
> 
> For example:
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/7838fa8fa6dfa884748cd330464d27ae58e781f8/docs/source/api/models/plants/plants_model.md?plain=1#L29-L36
> 
> should go to (I think):
> 
> ```
> ```{eval-rst}
> .. automodule:: virtual_ecosystem.models.plants.plants_model
>     :autosummary:
>     :members:
>     :exclude-members: model_name
>     :inherited_members:
> ```
> ```

I added this. I'm a bit confused about how the docs work TBH because in the version that I built and linked in the description above, it does seem to be showing? Is that in a different place? Anyways, added `:inherited-members:` to all of them !

davidorme (approved):
This looks good - we really need to update the model creation docs, but we also need to do that to reflect the static model changes as well, so lets do that separately under #676

PR #693, opened 2025-01-22
Make bypass setup and run update invalid in static models

### LINKED ISSUE #687: Prevent not running `_setup` but running `_update` from happening in `static` models
@davidorme , @sallymatson  I think that in total we have 4 cases, with one of them ill-defined:

| Static | _setup | _update | Comments                                                         |
| ------ | ------ | ------- | ---------------------------------------------------------------- |
| No     | Yes    | Yes     | Normal run                                                       |
| Yes    | Yes    | Once    | Update once and freeze                                           |
| Yes    | No     | Once    | Ill defined. Setup is skipped but update is expected to run once |
| Yes    | No     | No      | Use fix data                                                     |

The third one is the problem. As the code is now, it is allowed: you just provide all the variables that need to be populated by init but none of those that need to be populated by update. As a result, the `_setup` method is not run and none of the internals that the `_update` method will need are there.

If this is not a use case, then we need to fix the code to prevent that from happening.

_Originally posted by @dalonsoa in https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/682#issuecomment-2602291783_

dalonsoa:
# Description

This PR removes the possibility of skipping running the `_setup` method but still running `_update` once, which is an invalid configuration for static models.

As an extra, a bug is fixed by which the `_update` method was run in static models if no variables were present and None were required. If None are required, then the updated method should not run. Granted that is unlikely to happen in a real scenario, yet it should not happen ever. 

Fixes #687 

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
Great - I'll add this specification in the docs as well.

vgro (approved):
Looks fine to me

tests/conftest.py
```diff
@@ -592,3 +592,10 @@ def patch_bypass_setup(model: str):
     klass = "".join([w.capitalize() for w in model.split("_")] + ["Model"])
     object_to_patch = f"virtual_ecosystem.models.{model}.{model}_model.{klass}"
     return patch(f"{object_to_patch}._bypass_setup_due_to_static_configuration")
+
+
+def patch_static_config(model: str):
+    """Patch the run_update_check udring the init of the model."""
+    klass = "".join([w.capitalize() for w in model.split("_")] + ["Model"])
+    object_to_patch = f"virtual_ecosystem.models.{model}.{model}_model.{klass}"
+    return patch(f"{object_to_patch}._check_static_config")
```
davidorme: I think it would be easier to work with the model object here and in the other patches above:
```suggestion
def patch_static_config(model: type):
    """Patch the run_update_check udring the init of the model."""
    return patch(f"{model.__module__}.{model.__name__}._check_static_config")
```
```

File: tests/models/abiotic/test_abiotic_model.py [hunk 2]
```diff
@@ -368,12 +373,8 @@ def test_update_abiotic_model(dummy_climate_data, fixture_core_components):
...
 
     # initialise model
-    with (
-        patch_run_update("abiotic") as mock_update,
-        patch_bypass_setup("abiotic") as mock_bypass_setup,
-    ):
-        mock_update.return_value = False
-        mock_bypass_setup.return_value = False
+    with patch_static_config("abiotic") as mock_static_config:
+        mock_static_config.return_value = False, False
```
davidorme: And with the change to the patchers, this becomes:

```suggestion
    with patch_static_config(AbioticModel) as mock_static_config:
        mock_static_config.return_value = False, False
```

tests/models/abiotic/test_abiotic_model.py
```diff
@@ -9,7 +9,12 @@
 import xarray as xr
 from xarray import DataArray
 
-from tests.conftest import log_check, patch_bypass_setup, patch_run_update
+from tests.conftest import (
+    log_check,
+    patch_bypass_setup,
+    patch_run_update,
+    patch_static_config,
```
davidorme, line 16: Why is there all of the patching and checking of the static setup in the AbioticModel testing specifically? Is this just a convenient test case setup for running these checks or do these need to be rolled out to all model tests?

If it is the first - and this is really fussy - it makes the tests harder to follow for their primary purpose. It seems like it would be better to have a separate test file.

davidorme (changes requested):
The logic looks solid but a couple of things I'm not so clear on.

tests/conftest.py
dalonsoa: Much cleaner, indeed. To be honest, it did not occurred to me. I'll make the changes.

tests/models/abiotic/test_abiotic_model.py
dalonsoa: Neither. The patches are also used in hydrology and litter, in addition to both abiotic models. 

The problem is that the data sources used in those models for testing purposes do not honour the requirements for bypassing the setup and running the update. Typically, these models are not static models and include some of the variables that they are meant to initialise or update in either the setup or update methods. So the checks run in `__init__` fail.

Maybe I should have flagged this up and request from everyone to amend their data used in tests, but since that was a lot of work and no one complained when I implemented this in the previous PR, I assumed it was OK.

davidorme: Ooooooh. Bugger. I think this is something we want to revisit because I don't think it helps the clarity of the testing, but perhaps not here.

tests/models/abiotic/test_abiotic_model.py
dalonsoa: Sorted in [e7689cd](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/693/commits/e7689cdd8ad2b30f231a6d7a4182193a98e4bd64)

tests/conftest.py
dalonsoa: Sorted in [e7689cd](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/693/commits/e7689cdd8ad2b30f231a6d7a4182193a98e4bd64)

dalonsoa:
@davidorme , I think I implemented all the changes requested here. Shall we merge, so it does not become stalled?

davidorme (approved):
LGTM

Issue #699, opened 2025-01-27
Soil-abiotic links review


jacobcook1995:
All the links we want between the abiotic models (microclimate and hydrology) and the decay models (soil and litter) have now been implemented (at least for `v1.0.0`). A final thing we should do before initial release is to check that all of these links are (a) correct and (b) passing the information in the most useful form. 

There's already a few issues that fall under this umbrella (#533, #535, #346), but I think there would be a real value in sitting down and going through the whole model flow systematically to check that everything makes sense.

Issue #364, opened 2024-01-17
Resolving constants milestone issues and constants bundling.


davidorme:
This is a cross between a meta-issue and a discussion to tidy up a backlog of constants questions that we've grouped in [the "Constants system" milestone](https://github.com/ImperialCollegeLondon/virtual_rainforest/milestone/2).

This issue is also a summary of the inaugural monthly "GitHub issues" developer meetings to bundle existing issues and clean them up 🎉 🥂 

### Where we are now

* We have now adopted the constants system presented in discussion #162
* There's been some discussion of passing multiple constants classes to a model: `CoreConstants` and `ModelSpecificConstants`  and whether it might be easier to merge core constants into each model constants to provide a simpler namespace (see #342).

### Outstanding issues

We have five outstanding constants issues:

* #129 is about creating core constants but contains still relevant discussions about units. We have a separate milestone [bundling units issues](https://github.com/ImperialCollegeLondon/virtual_rainforest/milestone/4) so I think we can simply close #129.
* #277 is about how we ensure that configured constants are passed down correctly rather than falling back to defaults. I think this is largely outdated now that we are using the constants system, so I think we can close #277.
* #348 is still active and just needs a  PR to redefine constants values through the models.

The remaining two issues are linked and this PR suggests a resolution for both:
*  #342 is the discussion about simplifying model signatures to only need one bundled constants class. Implementations for this were also discussed in #341.
*  #358 is an issue that @vgro had with sharing constants and functions between parallel model development: if a function using constants is defined in one model then currently importing and re-using that function in another model requires also importing the constants and juggling multiple parallel constants classes.

### Sharing constants and functionality

The proposed solution to #358 that we can define model "meta" constants classes and use class inheritance to allow the clean use of different constants classes with shared functions. So:

```python

from virtual_rainforest.core.constants import ConstantsDataclass

class ModelMeta(ConstantsDataclass):
    """A meta constants class created in e.g. SimpleModel."""
    b = 2
    c = 3

class ActualModelSimple(ModelMeta):
    """The actual simple model constants from SimpleModel.

    This is derived from the meta class and only adds a constant specific to this model
    """
    d = 5

class ActualModelComplex(ModelMeta):
    """A more complex model constants from a parallel model.

    This adds two new constants and overwrites an existing constant with a model specific version
    """
    b = 4
    d = 5
    e = 6


def shared_function(constants: ModelMeta):
[...]

Issue #346, opened 2023-12-01
Default temperature units should be Kelvin rather than Celsius 


jacobcook1995:
**Is your feature request related to a problem? Please describe.**
In the current setup of the Virtual Rainforest all microclimate calculations are performed in Kelvin, and the results are then converted into Celsius units to be stored in the data object. Then when the `soil` and `animal` module use this data they convert it back to Kelvin units. This doesn't seem like a sensible thing to be doing.

**Describe the solution you'd like**
If we switched to using Kelvin as the default for data we could massively minimise the number of required conversions. We would basically only have to convert temperature data when it is input, for some of the functions of the plant module, and for plotting. The plant model uses both Celsius and Kelvin units so it will have to make conversions regardless of the temperature unit chosen

PR #700, opened 2025-01-27
Adding inorganic nitrogen pools to the soil model

### LINKED ISSUE #680: Add inorganic nitrogen pools
Similarly to the case for inorganic phosphorus I need to add inorganic nitrogen pools to the soil model

jacobcook1995:
# Description

This PR adds two additional nitrogen pools to the soil model. These pools are nitrate and ammonium. These two inorganic pools can be both taken up and generated by microbes, so the update function had to be modified. Other than this, a similar approach as for the inorganic phosphorus pools was taken.

One complication is that the empirical function I am using for soil moisture impacts on de/nitrification uses different soil moisture units to the hydrology model. I can do this conversion in the `soil` model, but it requires a constant from `HydroConsts`. For the time being I've just duplicated the constant into `CoreConsts` as I didn't want to refactor @vgro's code. I've added this as a TODO, and this is an issue we could tackle as part of #699 (I guess)

Anyway as ever this is probably pretty heavy on the science side, but feedback on code structure etc would be appreciated.

Fixes #680 

## Type of change

## Key checklist

## Further checks


dalonsoa:
How many pools are there?!?! 😆 OK, I'll try to review this asap, but won't be this week, I'm afraid, so move on if other people is happy with it.

virtual_ecosystem/core/constants.py
```diff
@@ -70,6 +70,20 @@ class CoreConsts(ConstantsDataclass):
...
+    soil_moisture_capacity: float = 0.9
+    """Soil moisture capacity, unitless.
+
+    The soil moisture capacity, also known as field capacity or water holding capacity,
+    refers to the maximum amount of water that a soil can retain against the force of
+    gravity after it has been saturated and excess water has drained away. The value is
+    soil type specific, the format here is volumetric relative water content (unitless,
+    between 0 and 1).
+
+    TODO - This constant also exists in the hydrology model. There really should be a
```
vgro, line 82: This is definitely something to talk about in more detail. We don't want such a key parameter in two places and potentially with different values. This relates to the metaconstants discussion in #364.

My hack solution to this is at the moment to load the constants from other models in setup, for example here: https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/2a33090e34121077767507d6f65718b0f40c51fc/virtual_ecosystem/models/abiotic/abiotic_model.py#L167

In the long-term this would ideally be a model input or could possibly even be calculated in `init`as a function of soil type? Let's discuss when we meet about #699

virtual_ecosystem/models/soil/constants.py
```diff
@@ -352,19 +382,19 @@ class SoilConsts(ConstantsDataclass):
     be better pinned down.
     """
 
-    half_sat_don_uptake: float = 0.07
-    """Half saturation constant for uptake of dissolved organic nitrogen (DON).
+    half_sat_ammonium_uptake: float = 0.02275
+    """Half saturation constant for uptake of ammonium.
 
-    [kg N m^-3]. The reference temperature is given by :attr:`arrhenius_reference_temp`,
+    [kg P m^-3]. The reference temperature is given by :attr:`arrhenius_reference_temp`,
```
vgro: why P?

virtual_ecosystem/models/soil/constants.py
```diff
@@ -374,8 +404,8 @@ class SoilConsts(ConstantsDataclass):
     be better pinned down.
     """
 
-    half_sat_dop_uptake: float = 0.02275
-    """Half saturation constant for uptake of dissolved organic phosphorus (DOP).
+    half_sat_nitrate_uptake: float = 0.02275
+    """Half saturation constant for uptake of nitrate.
 
     [kg P m^-3]. The reference temperature is given by :attr:`arrhenius_reference_temp`,
```
vgro: Why P?

virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -241,6 +241,93 @@ def calculate_clay_impact_on_enzyme_saturation(
...
+        thermal_sensitivity: Sensitivity of the factor to changes in temperature
+            [unitless]
+
+    Returns:
+        A factor capturing the impact of soil temperature on the nitrification rate
+        [unitless].
+    """
+
+    # Convert the temperatures to Kelvin
+    soil_temp_in_kelvin = convert_temperature(
```
vgro, line 269: These is an open issue to convert temperatures to Kelvin across all models #346 . Maybe put a TODO here to make sure we update this function when this change is implemented.

virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -241,6 +241,93 @@ def calculate_clay_impact_on_enzyme_saturation(
...
+
+    Returns:
+        A factor capturing the impact of soil temperature on the denitrification rate
+        [unitless].
+    """
+
+    # Convert the temperatures to Kelvin
+    soil_temp_in_kelvin = convert_temperature(
+        soil_temp, old_scale="Celsius", new_scale="Kelvin"
+    )
```
vgro, line 326: see previous comment

virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -241,6 +241,93 @@ def calculate_clay_impact_on_enzyme_saturation(
...
+
+    return factor_at_infinity * np.exp(
+        -thermal_sensitivity / (soil_temp_in_kelvin - minimum_temp)
+    )
```
vgro, line 333: Is there a case where `soil_temp_in_kelvin - minimum_temp` could be zero?

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -282,6 +288,8 @@ def integrate(self) -> dict[str, DataArray]:
                 delta_pools_ordered,
                 self.model_constants,
                 self.core_constants.max_depth_of_microbial_activity,
+                self.core_constants.soil_moisture_capacity,
```
vgro, line 291: Maybe add a TODO here, too

virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -241,6 +241,93 @@ def calculate_clay_impact_on_enzyme_saturation(
...
+
+    Args:
+        effective_saturation: Effective saturation of the soil with water [unitless]
+
+    Returns:
+        A factor capturing the impact of soil moisture on the nitrification rate
+        [unitless].
+    """
+
+    return effective_saturation * (1 - effective_saturation) / 0.25
```
vgro, line 296: Where do you actually calculate effective saturation? Or are you assuming that this is available in `data`?

tests/models/soil/test_env_factors.py
```diff
@@ -203,6 +202,78 @@ def test_calculate_clay_impact_on_enzyme_saturation(dummy_carbon_data):
...
+    )
+
+    effective_saturation = dummy_carbon_data["soil_moisture"][
+        fixture_core_components.layer_structure.index_topsoil_scalar
+    ] / (
+        fixture_core_components.layer_structure.soil_layer_thickness[0]
+        * 1e3
+        * HydroConsts.soil_moisture_capacity
+    )
+
```
vgro, line 244: This is how I get from the initial soil moisture (volumetric relative water content) to mm:

 water content in mm = relative water content / 100 * depth in mm
 Example: for 20% water (0.2) at 40 cm this would be: 20/100 * 400mm = 80 mm

So you reverse this for your calculation, the only difference is that you use a factor of 1000 and I use 100. What's your reference? Could be the volume vs depth? I will need to update that throughout the hydrology model.

This conversion could be included in the hydrology model.

vgro (approved):
Looks good to me, there are a few places where I would suggest adding a TODO so we don't miss updating these after our discussion on #699.

virtual_ecosystem/core/constants.py
jacobcook1995: I did try loading the constants from the hydrology model in setup initially, but the testing started to break (because the soil model gets used as an example for a lot of the core model. Decide to go with this approach rather than having to rewrite the tests in that section. I think we should resolve this issue when we do the overall review (it is marked with a TODO, so we can find it easily)

virtual_ecosystem/models/soil/constants.py
jacobcook1995: typo!

jacobcook1995: typo!

virtual_ecosystem/models/soil/pools.py
```diff
@@ -294,6 +335,12 @@ def calculate_all_pool_updates(
             top_soil_layer_index
         ].to_numpy()
         soil_moisture = self.data["soil_moisture"][top_soil_layer_index].to_numpy()
+        # Calculate the effective saturation of the soil (soil layer thickness needs to
+        # be converted from m to mm here to be consistent with soil moisture units)
+        # TODO - This needs to be reviewed as part of the soil abiotic links review
+        effective_saturation = self.data["soil_moisture"][
+            top_soil_layer_index
+        ].to_numpy() / (soil_moisture_capacity * top_soil_layer_thickness * 1e3)
```
jacobcook1995, line 343: @vgro this where the effective saturation is calculated

virtual_ecosystem/models/soil/env_factors.py
jacobcook1995: With the default setup this only happens at -46.02 C, so I ignored the issue. I shouldn't have though as the minimum temperature is configurable and this function outputting a NaN would stall the simulation. I will add a step that throws an error if a temperature below the minimum occurs

tests/models/soil/test_env_factors.py
jacobcook1995: I think the source of the difference is that you convert the soil layer thickness to millimetres early on in your model, whereas I do the conversion when perform the calculation. I think this is another thing to be addressed when we meet about #699

PR #702, opened 2025-01-28
Updating documentation of static models on config page.

### LINKED ISSUE #686: Documentation for static model
@vgro and I have discussed some initial documentation of the static model for users, with the example data and pypi package. 

I will work on the description of the model, purpose, and go through how to set it up in the config. 

@vgro will add a short example of how to make some tweaks to see results using static.

sallymatson:
# Description

Documentation update for the static model.

The overview is built [here](https://virtual-ecosystem.readthedocs.io/en/686-documentation-for-static-model/using_the_ve/configuration/config.html#static-models) and the example is built [here.](https://virtual-ecosystem.readthedocs.io/en/686-documentation-for-static-model/using_the_ve/virtual_ecosystem_in_static_mode.html)

A few notes:
- The **config files are built programmatically,** assuming that the combination of models in the example does not change. The code can be viewed if wanted.
- The **path setting is still not universal** for as os but I have added a note about that at the start.
- I collapsed the output of `ve_run` so it does not clutter the documentation, and I also collapsed the code for the plots.

@sallymatson I cannot add you as a reviewer, please feel free to comment :-)

Fixes #686 

## Type of change

## Key checklist


docs/source/using_the_ve/configuration/config.md
```diff
@@ -198,15 +198,35 @@ All models (except `core`) accept a boolean configuration option, `static`, that
...
-  - **All present**: The model initialisation process is bypassed.
-  - **None present**: The model initialisation process runs normally.
-- For the update, if the variables that are populated are:
-  - **All present**: Then the update process is bypassed.
-  - **None present**: The update *is run just once*, keeping the same values for those
-  variables throughout the simulation.
+If `static=true`, the initial model behaviour depends on the data variables available in
+the [`Data` object](../data/data.md).
+
+* If the data object has all varibles in the model's `vars_populated_by_init`, then the
```
vgro:
```suggestion
* If the data object has all variables in the model's `vars_populated_by_init`, then the
```

docs/source/using_the_ve/configuration/config.md
```diff
@@ -198,15 +198,35 @@ All models (except `core`) accept a boolean configuration option, `static`, that
 indicates if such a model should be updated every iteration (`static=false`, the
 default behaviour) or not.
```
vgro, line 206: I think it would be useful to add a sentence like 'To run the Virtual ecosystem, one more models can bet static,for example x and y to understand z' just to explain what this is about (or not) and highlight the potential of exploring individual models or combinations of models/interactions without running the whole complex thing.

docs/source/using_the_ve/configuration/config.md
```diff
@@ -198,15 +198,35 @@ All models (except `core`) accept a boolean configuration option, `static`, that
...
-variables available in the [`Data` object](../data/data.md):
-
-- For the initialisation, if the variables that are populated are:
-  - **All present**: The model initialisation process is bypassed.
-  - **None present**: The model initialisation process runs normally.
-- For the update, if the variables that are populated are:
-  - **All present**: Then the update process is bypassed.
-  - **None present**: The update *is run just once*, keeping the same values for those
-  variables throughout the simulation.
+If `static=true`, the initial model behaviour depends on the data variables available in
```
vgro: Maybe 'provided at the start of the simulation' instead of 'available? Just to make super clear what the workflow is?

vgro:
@sallymatson I put together some ideas for an example, please let me know if you think this makes sense and I can create the new input file and add the plotting bits. I'll fix the numbering item issue.

sallymatson:
> @sallymatson I put together some ideas for an example, please let me know if you think this makes sense and I can create the new input file and add the plotting bits. I'll fix the numbering item issue.

I think it's a great approach!! I think it would be helpful for us to figure out the best way to "name" the two approaches because then we can use them in the documentation (on the config page that I've been working on) and here, so that users can quickly understand how the two documents relate to eachother. Otherwise, I think it's clear and thorough !!

vgro:
@sallymatson I updated the example to make it a bit simpler and focus only on the static setting that starts from scratch for now, not the option to provide all the variables at the start (I think it was a bit confusing and might be better in a separate example, for instance an experiment 3).
My question is do you know how to make changes to the config files in the notebook, or no we need to somehow provide the alternative config files with the example?

sallymatson:
> @sallymatson I updated the example to make it a bit simpler and focus only on the static setting that starts from scratch for now, not the option to provide all the variables at the start (I think it was a bit confusing and might be better in a separate example, for instance an experiment 3). My question is do you know how to make changes to the config files in the notebook, or no we need to somehow provide the alternative config files with the example?

I think that we would need to create a separate example config for the run. If we load the config first, then we could so something like this like I just did for some testing in the plants model:

`fixture_config["plants"]["constants"] = {"PlantsConsts": {"leaf_lignin": 100.0}}`

Or, you could open the file in the jupyter notebook and edit it (and just explain that this code does not need to be copied.) But I generally think it would be easier and better practice to have its own config, especially since we wouldn't then know if the config changed for other reasons, if it would impact this tutorial. 

Does that make sense? If you want to upload the example you have and then I can play around with the config, happy to help more!!

docs/source/conf.py
```diff
@@ -247,3 +247,7 @@ class MyReferenceStyle(AuthorYearReferenceStyle):
 hoverxref_roles = ["term"]
 
 hoverxref_role_types = {"term": "tooltip"}
+
+# Allow for longer runtime
```
vgro, line 256: Consider commenting out and replace with local solution.

``` yaml
---
execution:
  timeout: 60
jupytext:
  formats: md:myst
  text_representation:
    extension: .md
    format_name: myst
    format_version: 0.13
    jupytext_version: 1.17.1
kernelspec:
  display_name: Python 3 (ipykernel)
  language: python
  name: python3
language_info:
  codemirror_mode:
    name: ipython
    version: 3
  file_extension: .py
  mimetype: text/x-python
  name: python
  nbconvert_exporter: python
  pygments_lexer: ipython3
  version: 3.10.14
---
```

davidorme:
@vgro OK - I see what the issue is with the static configuration paths:

* The static configurations have been created by editing the output config from running the example data. 
* The data paths in the example config are relative: `file_path = "../data/example_climate_data.nc"` but...
* ... the compiled config files are resolved to give absolute paths on the running machine, given the location of the `example_data` folder.

That is a bad thing to commit to the repo, because the paths are machine-specific: it's where ever you put the example data to run `ve_run`. So removing `/private` just converts it to something that happens to work on the runners.

I think there are three ways forward here:

1. Not commit those config files to the repo and instead come up with a way of creating them programatically to run the doc workbook. That's a non-starter I think - it would be a behind the scenes hack that breaks the workflow of explaining how to use the static model.

1. Edit the compiled static configurations to make the paths relative to those config files. That will then be system independent for CI and running them locally. 
  This has the disadvantage though that if people are trying to play along with the guide, those are a whole bunch of changes that are needed for **us** to make the files work cross platform, but aren't terribly relevant locally. I think that could be ok though - we can add an info box that says "if you're looking at the static config toml on the repo you will see it has relative paths. That matters for us because multiple systems, but you only need to make the changes described in this page".

1. Lastly, we could avoid using the compiled configuration and add small config files to the config directory that enable static mode. The disadvantage here is that users would then need to provide the list of config files to use from that menu of files in `config` rather than just pointing at the whole directory. I don't think that's inherently a bad thing - it is why we designed the config system like that.

vgro:
> @vgro OK - I see what the issue is with the static configuration paths:
> 
>     * The static configurations have been created by editing the output config from running the example data.
> 
>     * The data paths in the example config are relative: `file_path = "../data/example_climate_data.nc"` but...
> 
>     * ... the compiled config files are resolved to give absolute paths on the running machine, given the location of the `example_data` folder.
> 
> 
> That is a bad thing to commit to the repo, because the paths are machine-specific: it's where ever you put the example data to run `ve_run`. So removing `/private` just converts it to something that happens to work on the runners.
> 
> I think there are three ways forward here:
> 
>     1. Not commit those config files to the repo and instead come up with a way of creating them programatically to run the doc workbook. That's a non-starter I think - it would be a behind the scenes hack that breaks the workflow of explaining how to use the static model.
> 
>     2. Edit the compiled static configurations to make the paths relative to those config files. That will then be system independent for CI and running them locally.
>        This has the disadvantage though that if people are trying to play along with the guide, those are a whole bunch of changes that are needed for **us** to make the files work cross platform, but aren't terribly relevant locally. I think that could be ok though - we can add an info box that says "if you're looking at the static config toml on the repo you will see it has relative paths. That matters for us because multiple systems, but you only need to make the changes described in this page".
> 
>     3. Lastly, we could avoid using the compiled configuration and add small config files to the config directory that enable static mode. The disadvantage here is that users would then need to provide the list of config files to use from that menu of files in `config` rather than just pointing at the whole directory. I don't think that's inherently a bad thing - it is why we designed the config system like that.

I thought about putting the experiment files in the config folder but that doesn't work as there are conflicts when running the non-static normal ve_example. I set up the whole notebook to install the example to `/tmp`, so the path should not be a problem here.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,441 @@
...
+## Why use Static Mode?
+
+Imagine standing in a rainforest clearing at dawn. The sun rises, mist lifts, and the
+forest slowly comes to life.
+
+Now imagine you could **pause parts of this scene** — hold
+the soil moisture constant, freeze the behaviour of animals, or keep the microclimate
+unchanged — while the rest of the ecosystem continues to move through time.
+
+This is the essence of the **static mode** in the Virtual Ecosystem model.
```
vgro:
```suggestion
This is the essence of the **Static Mode** in the Virtual Ecosystem model.
```

virtual_ecosystem/example_data/static_config/README.md
```diff
@@ -0,0 +1,7 @@
+# Notes on static configuration files
```
vgro, line 1: This might no longer be relevant

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+components reset to their original state (e.g. eaten leaves magically reappear) —
+like a controlled reset — so any change comes only from the active, evolving parts.
+
+This setup enables **targeted experiments** within the full model framework. For
+example, you can test how vegetation responds to changing temperatures without feedback
+from soil moisture changes, or explore animal behaviour in a fixed vegetation landscape.
+
+By choosing which components evolve over time and which stay fixed, Static Mode helps
+reveal **drivers of ecological change** - helping us understand not just
+**what happens**, but **why**.
```
davidorme: I like the writing style here but it makes very heavy use of bold, which is not recommended for accessibility in screen text. I think the key phrase here is "targeted experiments" - I would genuinely only bold that.

The same is true below - only use bold sparingly, not for every line in a list.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+configurations that best support the processes you want to isolate or understand.
+```
+
+## Using the static mode
+
+This example demonstrates how to configure and run the Virtual Ecosystem with selected
+components held static (for more technical details,
+[see](./configuration/config.md#static-models)). Specifically, we focus on **hydrology**
+by keeping the
+microclimate, vegetation, animals, and soils fixed. This isolates hydrological
```
davidorme:
```suggestion
by keeping the
microclimate, vegetation, animals, litter and soils fixed. This isolates hydrological
```

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+Finally, we compare their outputs to the default setup (`ve_example`) where all
+components run dynamically.
+
+```{note}
+This guide focuses on how to technically configure and run static mode, not on
+interpreting the experimental results. This enables you to adapt the setup to suit your
+own analysis goals.
+
+The example uses `%%bash` for demonstration on Linux, but it is compatible with all
+operating systems.
```
davidorme:
```suggestion
The code below to run the Virtual Ecosystem and plot results uses a mixture of the command line (e.g. `bash` on Linux or MacOS, `powershell` on Windows) and python code. For the code run at the command line, we currently only show the syntax for using `bash`. On Windows, you will need to change the paths (for example, to `C:\Users\username\Temp\...`). 
```

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+[example instructions](./virtual_ecosystem_in_use.md) to familiarise yourself with the
+setup.
+
+```{code-cell} ipython3
+%%bash
+# Remove any existing VE data directory in the /tmp/ directory
+if [ -d /tmp/ve_example ]; then
+  rm -r /tmp/ve_example
+fi
+```
```
davidorme: I think we should hide this cell - it's only there to avoid cockups on developer machines and this is really the only cell with very bash specific commands.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+
+  ```toml
+  [abiotic_simple]
+  static=true
+  ```
+
+Here is how you can do these steps programmatically:
+
+```{code-cell} ipython3
+:tags: [hide-input]
```
davidorme: I think we should completely hide these. We need to edit them programatically to build the page, but I don't think it is a useful general technique for modifying files and makes the whole thing look more scary. Even having them as expandable just raises more questions than it answers.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+output_dir.mkdir(parents=True, exist_ok=True)
+
+# Check that the directory is empty
+if output_dir.exists() and output_dir.is_dir():
+    for file in output_dir.iterdir():
+        if file.is_file():
+            file.unlink()
+
+# Load the original config
+config = toml.load("/tmp/ve_example/out/ve_full_model_configuration.toml")
```
davidorme: It is frustrating that we need two seperate TOML libraries to load and write TOML (and need to detect which version of a package we need depending on the system), but I don't think we should add `toml` to the requirements and use it. It hasn't been touched in two years and never made it to 1.0.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+
+# Make a copy to modify
+new_config = copy.deepcopy(config)
```
davidorme: Not needed.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+config["abiotic_simple"]["static"] = True
+config["plants"]["static"] = True
+config["animal"]["static"] = True
+config["soil"]["static"] = True
+config["litter"]["static"] = True
+config["core"]["data_output_options"]["out_path"] = "/tmp/ve_example/HydroDefault_out/"
```
davidorme: Not needed - directly in the command line, so confusing to do it twice.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+
+```toml
+  [hydrology]
+  initial_soil_moisture = 0.3
+  initial_groundwater_saturation = 0.9
+  static = false
+```
+
+Here is how you can do these steps programmatically:
+
```
davidorme: As above - I think we conceal this completely.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+```
+
+### 1: Run `ve_example` as a baseline
+
+Before setting up static experiments, make sure you can run the model successfully.
+If you haven’t yet installed and executed the example, follow the
+[example instructions](./virtual_ecosystem_in_use.md) to familiarise yourself with the
+setup.
+
+```{code-cell} ipython3
```
davidorme, line 107:
```suggestion
```{code-cell} bash
```

Here and elsewhere, these cells should not be styled as python.

docs/source/using_the_ve/configuration/config.md
davidorme, line 1: I think all the relevant information is here but it goes in hard with technical details and repeats quite a lot - I think it would be easier to follow if it builds up a bit more from first principles.

davidorme (changes requested):
This is good. I've suggested a few changes:

* Some technical changes to the setup of the worked example (not using the `toml` package, syntax and hiding the programmatic config creation). 

  I ran those changes locally to check I wasn't breaking it, so have a commit I can push if those sound sensible. Let me know if you want me to do that.

* Some styling suggestions for accessibility and tweaks to wording.

* I think the overview of static mode is a little hard hitting. Again - I have a revised version which I'm happy to push if that is ok.

davidorme:
@vgro and @sallymatson I've pushed my suggestions to a new PR subbranch off this one #905  and asked for a review.

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
davidorme: Oh bugger. We can't do this, because our use of `jupytext` dictates a single `{code-cell}` language and it automatically enforces this. See https://github.com/mwouts/jupytext/issues/1267#issuecomment-2987731168

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+# Save to a new file
+with open("/tmp/ve_example/static_config/HydroDefault_config.toml", "w") as f:
+    toml.dump(config, f)
+```
+
+#### Run HydroDefault experiment
+
+Now run the model with the new configuration from your command line and make sure that
+the `--outpath` command in the second line is followed by the directory of the new
+output folder that you just created (here `/tmp/ve_example/HydroDefault_out/`), and the
```
arne-exe:
```suggestion
output folder that you just created (here `/tmp/ve_example/HydroDefault_out/`),
```

docs/source/using_the_ve/virtual_ecosystem_in_static_mode.md
```diff
@@ -0,0 +1,440 @@
...
+# Save to a new file
+with open("/tmp/ve_example/static_config/HydroDry_config.toml", "w") as f:
+    toml.dump(config, f)
+```
+
+#### Run HydroDry experiment
+
+Now run the model with the new configuration from your command line. Again, make sure
+that the `--outpath` command in the second line is followed by the directory of the new
+output folder that you just created (here `/tmp/ve_example/HydroDry_out/`), and the
```
arne-exe:
```suggestion
output folder that you just created (here `/tmp/ve_example/HydroDry_out/`),
```

arne-exe (commented):
Hi @vgro, a very nice applied example of how to run the model in static mode. I only had suggestions in 2 places where you repeated the same words.

davidorme (approved):
LGTM

Issue #697, opened 2025-01-24
The `PModel` constants are currently fixed.


davidorme:
The `PlantsModel` uses `pyrealm.pmodel.PModel`, which takes arguments setting `pmodel_consts` and `core_consts`, which are instances of `pyrealm.constants.PModelConst` and `pyrealm.constants.CoreConst`. Most of these are very unlikely to change, but we do need to be able to change them. At the moment (over in branch #659 ), the implementation just has this fixed, but we will need a mechanism to update it. 

That might most easily be a new entries in the `PlantsConsts` object allows an arbitrary dictionary that will get passed to `PModelConst` and `CoreConst` as `kwargs`.

Ultimately, we need to look at the overlap between the VE constants and these constants and synchronise them.

PR #707, opened 2025-01-30
Using pyrealm within the Plants model

### LINKED ISSUE #655: Switch plant model community handling to use `pyrealm`
The plant model currently uses placeholder functionality to generate plant communities from cohort data. Now that this functionality is available through `pyrealm` as of `pyrealm==2.0.0-rc.1`, this needs to be switched over.

### LINKED ISSUE #659: Update plants model to use `pyrealm` flora definition
The implementation of the plant functional types and the Flora object in the Plants Model are cut down placeholders for the more complex definition that is now implemented in `pyrealm` (as of `2.0.0-rc.1`). We need to:

* Switch the `Flora` implementation over to `pyrealm`
* Think about how to configure the PFT definitions. 

    This is actually a bit tricky: these are currently part of the TOML model configuration, but that is increasingly unpleasant for large numbers of PFTs and the set of traits expanded beyond the current placeholder of name and max height. The cleanest way is to provide a CSV dataframe, which `pyrealm` already supports, and add a PFT file path to the configuration.

davidorme:
# Description

This PR is to replace the placeholder internals of the Plants Model with actual pyrealm functionality. That is basically impossible to do in small steps, so this is an uncomfortably sprawling PR.

* Replaces old plants config with new setup for T model parameters
* Replaces flora structure with pyrealm Flora
* Replaces cohort structure with pyrealm Cohorts
* Replaces community structure with pyrealm Community
* Uses pyrealm demography functionality to go from cohort DBH data to allometry
* Uses PModel to estimate GPP and transpiration.

> [!TIP]
> @vgro If you can look at the internals of the Plants Model setup and update to see if anything seems wrong
> @sallymatson Could you look at the canopy, community and functional types to see if the way I'm using `pyrealm` structures is well enough described (for now, anyway).

Fixes #659 
Fixes #655 

## Type of change

## Key checklist
 
## Further checks


virtual_ecosystem/models/plants/community.py
```diff
@@ -1,50 +1,22 @@
...
-is used to hold list of plant cohorts by grid cell and generate those lists from
-variables provided in the data object.
-
-NOTE - much of this will be outsourced to pyrealm.
-
+"""The :mod:`~virtual_ecosystem.models.plants.community` submodule contains
+functionality to generate a dictionary of
+:class:`~pyrealm.demography.community.Community` instances for each grid cell, populated
+with size-structured :class:`~pyrealm.demography.community.Cohorts` instances from
+initial plant cohort inventories for a simulation.
```
sallymatson: I think this could be clearer on the exact structure, and it is important to be explicit here in case people don't go to the pyrealm docs to figure out the relationships between the models. Fundamentally we need to communicated that the `PlantCommunities` instance holds a dictionary of `cell_id` to `PlantCommunity` objects. Each grid cell has only one `PlantCommunity` instance, and that `PlantCommunity` instances holds multiple cohort instances.

The key info that is not completely clear here (IMO) is the 1-1 relationship between a grid cell and plant community.

virtual_ecosystem/models/plants/community.py
sallymatson, line 1: Should this file/submodule now be renamed `Communities`, given that that is the actual VE model that it is built on? And also because it is the most overarching model in the workflow here? Naming it `Community` is confusing to me because that is a pyrealm model, and kind of obfuscates the `PlantCommunities` model because it is upstream of `Community`.

virtual_ecosystem/models/plants/__init__.py
```diff
@@ -1,13 +1,27 @@
...
+* The :mod:`~virtual_ecosystem.models.plants.constants` submodule provides definitions
+  of the main constants used in the model.
+
+* The :mod:`~virtual_ecosystem.models.plants.functional_types` submodule implements the
+  handling of individual plant functional types and the overall flora definition to be
+  used in a simulation.
+
+* The :mod:`~virtual_ecosystem.models.plants.community` submodule implements a community
+  object, defined as a the set of size-structured plant cohorts occurring in a grid
+  cell.
```
sallymatson: The submodule implements a Communities object right?

virtual_ecosystem/models/plants/community.py
```diff
@@ -1,50 +1,22 @@
...
-        init=False, default_factory=lambda: np.array([])
-    )
-    """The canopy area within canopy layers of each individual."""
-    gpp: float = field(init=False, default=0)
-    """The gross primary productivity for each individual."""
-
-
-class PlantCommunities(dict, Mapping[int, PlantCohort]):
+class PlantCommunities(dict, Mapping[int, Community]):
     """A dictionary of plant cohorts keyed by grid cell id.
```
sallymatson: A dictionary of plant Community objects

virtual_ecosystem/core/utils.py
```diff
@@ -56,3 +60,40 @@ def check_outfile(merge_file_path: Path) -> None:
         raise to_raise
 
     return None
+
+
+def split_arrays_by_grouping_variable(
```
sallymatson, line 102: Does this function need a test?

sallymatson (commented):
Exciting that this PR is in!! I had some comments on clarifying the role of `PlantCommunities` vs `PlantCommunity`, you explain it all really well in the plant model docstring but there are a few places where it is referenced that I think could be clarified. The canopy and pft looks pretty clear, a bit harder for me to comment on canopy since I haven't used it, but also seems like there is a bit less model structure there so it should be more straightforward (?). 

Otherwise, the logic looks good to me, and seems well tested. LMK if you want any clarifications on any of the comments!

virtual_ecosystem/models/plants/community.py
davidorme: I agree - this may also disappear into `pyrealm` at some point because that is likely going to want to be able to handle multiple sites.

virtual_ecosystem/core/utils.py
davidorme: Yup  -  looking at this again, this kinda forms a pair with `core.data.Data.confirm_variables_form_data_frame`. They should either both be in `Data` (and hence operate only on data in that `Data` object) or both should be available as a utilities for any arrays. 

For now, I've moved them both into `core.utils`, made a note about this choice and added tests. We can review their use later. That also meant updating the `PlantCommunities.__init__` code.

sallymatson (approved):
Thanks for making the changes - LGTM!

docs/source/api/models/plants/plant_structures.md
```diff
@@ -32,6 +32,7 @@ This page documents three submodules of the `plants` module:
 2. The plant community structures, describing the cohorts of stems of different PFTs
    with different diameters at breast height within a grid cell.
 3. The canopy structure generated in a grid cell by the plant community.
+4. The constants definitions required to run the model
```
vgro, line 36: The first sentence says 'three submodules', I believe this needs updating.

tests/conftest.py
```diff
@@ -110,13 +110,45 @@ def fixture_config():
...
 
         [plants]
-        a_plant_integer = 12
-        [[plants.ftypes]]
-        pft_name = "shrub"
-        max_height = 1.0
-        [[plants.ftypes]]
-        pft_name = "broadleaf"
-        max_height = 50.0
+        [[plants.pft_definition]]
```
vgro, line 113: I assume these are variables that go like this into pyrealm, I would still encourage following our naming convention and make the names more descriptive/longer

virtual_ecosystem/models/plants/__init__.py
```diff
@@ -1,13 +1,27 @@
...
+  of the main constants used in the model.
+
+* The :mod:`~virtual_ecosystem.models.plants.functional_types` submodule implements the
+  handling of individual plant functional types and the overall flora definition to be
+  used in a simulation.
+
+* The :mod:`~virtual_ecosystem.models.plants.communities` submodule provides the
+  :class:`~virtual_ecosystem.models.plants.communities.PlantCommunities` class which
+  maps each grid cell on to a representation of the plant community within that cell.
+
```
vgro, line 25: are the cohorts missing here or are they a lower level?

virtual_ecosystem/models/plants/canopy.py
```diff
@@ -201,3 +70,46 @@ def initialise_canopy_layers(data: Data, layer_structure: LayerStructure) -> Dat
...
+    This function takes a PlantCommunities object and calculates the canopy
+    representation for each community, using the perfect plasticity approximation to
+    calculate the closure heights of the canopy layers.
+
+    Args:
+        communities: The PlantCommunities object to convert
+        max_canopy_layers: The maximum number of permitted canopy layers
+
+    Returns:
+        A dictionary, keyed by cell id, of the canopy representations of each community.
```
vgro: Since you import pyrealm here, it would be useful to give a bit more info on the actual outputs (I assume these are the same as the previous functions returned?)

virtual_ecosystem/models/plants/communities.py
```diff
@@ -0,0 +1,126 @@
...
+
+    A class instance must be initialised using a
+    :class:`~virtual_ecosystem.core.data.Data` object containing the data required to
+    define those cohort instances. The required variables are:
+
+    * ``plant_cohorts_cell_id``,
+    * ``plant_cohorts_pft``,
+    * ``plant_cohorts_n``, and
+    * ``plant_cohorts_dbh``.
+
```
vgro, line 39: could you write out what these are here?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -96,42 +105,42 @@ class PlantsModel(
...
+
+    These data are used to setup the plant communities within each grid cell, using the
+    :class:`~virtual_ecosystem.models.plants.communities.PlantCommunities` class to
+    maintain a lookup dictionary of communities by grid cell.
+
+    The model setup then initialises the canopy layer data within the
+    :class:`virtual_ecosystem.core.data.Data` instance for the simulation and populates
+    these data layers with the calculated community canopy structure for each grid cell.
+    The community canopy representation is calculated using the perfect plasticticy
+    approximation, implemented in the `pyrealm` package. The canoppy variables populated
```
vgro:
```suggestion
    approximation, implemented in the `pyrealm` package. The canopy variables populated
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -164,9 +173,25 @@ def __init__(
...
+        a shorter reference to self.layer_structure.index_canopy."""
+        self.canopies: dict[int, Canopy]
+        """A dictionary giving the canopy structure of each grid cell."""
+        self.filled_canopy_mask: NDArray[np.bool_]
+        """A boolean array showing which layers contain canopy by cell."""
+        self.stem_gpp: dict[int, NDArray[np.float32]]
+        """A dictionary keyed by cell id giving the stem GPP for each cohort in the cell
+        community"""
+        self.stem_transpiration: dict[int, NDArray[np.float32]]
+        """A dictionary keyed by cell id giving the stem transpiration for each cohort
```
vgro: Stem transpiration is a bit of an odd word, is this the transpiration per individual? Just for clarification, no need to change

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -164,9 +173,25 @@ def __init__(
...
+        in the cell community"""
+        self.pmodel_consts: PModelConst
+        """PModel constants used by pyrealm."""
+        self.pmodel_core_consts: CoreConst
+        """Core constants used by pyrealm."""
```
vgro, line 194: is there much overlap with the ve core constants and if yes, should priority be given to ve? Something to think about at some later point

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -223,27 +248,44 @@ def _setup(
...
+            data=self.data, flora=self.flora, grid=self.grid
+        )
+        # This is widely used internally so store it as an attribute.
+        self._canopy_layer_indices = self.layer_structure.index_canopy
 
-        # Update canopy layers
-        # TODO - this initialisation step may move somewhere else at some point.
+        # Initialise the canopy layer arrays.
+        # TODO - this initialisation step may move somewhere else at some point see #442
         self.data = initialise_canopy_layers(
```
vgro: Is this the right way of adding new variables to the data object? I would expect `self.data['layer_heights']=initialise_canopy_layers(...)`

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -223,27 +248,44 @@ def _setup(
...
+        self.pmodel_consts = PModelConst()
+        self.pmodel_core_consts = CoreConst()
 
-        # Run the canopy initialisation - update the canopy structure from the initial
-        # cohort data and then initialise the irradiance using the first observation for
-        # PPFD.
+        # Create and populate the canopy data layers and set the absorption from the
+        # first time index
         self.update_canopy_layers()
         self.set_canopy_absorption(time_index=0)
```
vgro: The new abiotic model also creates some initial canopy absorption, let's check how the order works out and I might be able to delete this in the abotic model

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -317,28 +403,29 @@ def set_canopy_absorption(self, time_index: int) -> None:
...
+            canopy_top_ppfd - np.nansum(absorbed_irradiance, axis=0)
+        )
 
-        # Store the absorbed irradiance in the data object and add the remaining
-        # irradiance at the surface layer level
-        # NOTE - this is only the _PPFD_ at ground level not the SWDown.
-        self.data["canopy_absorption"][:] = absorbed_irradiance
-        ground = np.where(self.data["layer_roles"].data == "surface")[0]
-        self.data["canopy_absorption"][ground] = ground_irradiance
+        self.data["canopy_absorption"] = absorbed_irradiance
```
vgro: Could you please clarify, is the absorbed radiation not returned for each canopy layer?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -353,98 +440,122 @@ def estimate_gpp(self, time_index: int) -> None:
...
 
-        # Get an array where populated canopy layers are one otherwise nan
-        filled_canopy = xr.where(
-            (self.data["layer_heights"] * self._canopy_layer_indices[:, None]) > 0,
-            1,
-            np.nan,
+        # Estimate the light use efficiency of leaves within each canopy layer within
+        # each grid cell. The LUE is set purely by the environmental conditions, which
+        # are shared across cohorts so we can calculate all layers in all cells.
+        pmodel_env = PModelEnvironment(
```
vgro, line 469: Note that these arrays contain all the layers, including above, soil, and nan canopy layers. Would it make sense to select by filled canopy index?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -317,28 +403,29 @@ def set_canopy_absorption(self, time_index: int) -> None:
...
-            .data
+            .to_numpy()
         )
 
         # Calculate the fate of PPFD through the layers
-        absorbed_irradiance = canopy_top_ppfd * self.data["layer_fapar"].data
-        ground_irradiance = canopy_top_ppfd - np.nansum(absorbed_irradiance, axis=0)
+        absorbed_irradiance = self.data["layer_fapar"] * canopy_top_ppfd
+        # Add the remaining irradiance at the surface layer level
+        absorbed_irradiance[self.layer_structure.index_surface] = (
```
vgro: This is an interesting point to discuss, the abiotic model does the same calculation, I was not aware that this is already part of the `canopy_absorption` variable. From a physical perspective I'm not sure if that makes sense to allocate the remaining energy to the surface layer. We could make the assumption that this is absorbed by the topsoil layer or store it in a separate variable or not at all? What do you think?

vgro: `# Net radiation topsoil, [W m-2]
        net_radiation_soil = energy_balance.calculate_net_radiation(
            incoming_radiation=data["topofcanopy_radiation"].isel(time_index=time_index).to_numpy(),
            absorbed_radiation=np.nansum(data["absorbed_radiation"].to_numpy(), axis=0),
            longwave_emission=longwave_emission_soil,
            albedo=abiotic_constants.surface_albedo,
        )`

vgro (commented):
Looks good to me, a few points that we might want to discuss in person. One change that is unclear at the moment is what happened to the Cohorts?

virtual_ecosystem/models/plants/__init__.py
davidorme: That's where it drops into `pyrealm` but I've clarified this.

virtual_ecosystem/models/plants/canopy.py
davidorme: I've linked through to the `pyrealm` `Canopy` class, which gives the details docs on the returned values.

virtual_ecosystem/models/plants/communities.py
davidorme: Done.

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yes - I've changed this to `per_stem_gpp` and `per_stem_transpiration`

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yes - there likely is quite a bit of overlap. We'll need to map overlap and then use our core settings to set `CoreConst`.

Ha. Now that I look at #697, which is there to allow these values not to be fixed, I find I wrote:

> Ultimately, we need to look at the overlap between the VE constants and these constants and synchronise them.

So we even have an issue for this!

virtual_ecosystem/models/plants/plants_model.py
davidorme: It is, yes. Here:

```python
absorbed_irradiance = self.data["layer_fapar"] * canopy_top_ppfd
```

The values in `self.data["layer_fapar"]` are only defined for grid_cell x layer entries that contain leaves, so this product has all the layers.

virtual_ecosystem/models/plants/plants_model.py
davidorme: I did consider it - I'm not sure whether the subsetting and reinsertion is actually a net benefit over sticking oversized arrays through `pyrealm`, so I kept it simple!

virtual_ecosystem/models/plants/plants_model.py
davidorme: That's a mistake on my part - the topsoil layer is clearly the right place to dump it. I think it is cleaner if my array identify the fate of all the radiation through the canopy, partly because at some point animals may be interested in sunflecks and the like, which will become part of this.

davidorme: I've moved those values down into the topsoil layer.

virtual_ecosystem/models/plants/plants_model.py
davidorme: I agree it looks inconsistent with our previous usage. 

I don't think there's anything wrong with it (take `data`, update it and replace it) but we should try and do things _one_ way, so I've updated this to use `data.update_from_dict`.

davidorme:
@vgro Thanks for those comments - I've resolved the simple typos and doc updates and commented on the others. Could you review my responses and either resolve them if you are happy or flag them for more discussion!

I've updated the docs to give more clarity on cohorts - they are part of the pyrealm community objects.

tests/conftest.py
davidorme: These names are currently used internally by `pyrealm` so I'd have to introduce a mapping to fix them in this PR. I'll have a look at renaming them in `pyrealm` but that's parking this issue for now.

virtual_ecosystem/models/plants/plants_model.py
vgro: Oh, I understood that this to be 'per stem', I meant that it is not them stem that does the transpiration, but the leaf :-) Minor point

virtual_ecosystem/models/plants/plants_model.py
vgro: Is the output variable still called `canopy_absorption`? This could be a bit misleading, would shortwave absorption be more precise?

tests/models/plants/conftest.py
```diff
@@ -91,46 +107,99 @@ def fxt_plants_model(plants_data, flora, fixture_core_components):
...
+def fixture_canopy_layer_data():
     """Shared canopy layer data.
 
-    The fixture supplies tuples of layer name, test values and the indices of the
-    vertical layer dimension to insert test values.
-
-    TODO: This is currently convoluted because of the way in which layer_heights is set
-    within the plants model.
+    The fixture supplies a dictionary of data values expected from the cohort data in
+    the plants_data fixture. Each enntry provides a tuple of the variable name to be
```
vgro:
```suggestion
    the plants_data fixture. Each entry provides a tuple of the variable name to be
```

vgro (approved):
This looks good to me. The only remaining issue I have is the name of the `canopy_absorption` which I think we should change if we include the radiation absorbed by the soil.

virtual_ecosystem/models/plants/plants_model.py
davidorme: Aha 😆  I think it is clearer with `per_stem_transpiration` that we don't mean the stem is transpiring, so I think this is OK!

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yup - I agree. Will update.

PR #710, opened 2025-01-30
Allow `[[core.data.variable]]` to be split across files

### LINKED ISSUE #234: `[[core.data.variable]]` tags cannot be used in more than one config file

### LINKED ISSUE #709: Allow data to be configured in different files
At present, we are locked into having a single file containing all `core.data.variable` definitions.

At the moment, that means you can't simply hot swap different e.g. plant configs by changing a single file, you have to edit the data file too. It's much easier to maintain these files if the data related to a model is in with the other config options.

davidorme:
# Description

This PR updates `config_merge` to handle the special case where configuration sources provide identical key paths but the values are both lists. In this case, the config merges the two lists, rather than adding a conflict.

The main (only?) structure giving this is in TOML is the array of tables, which is what we use for `[[core.data.variable]]`. This PR doesn't guarantee that the merged lists make _any_ sense but downstream configuration checkers should handle that. It also means you could in split other array of tables structures (like `[[plants.pft_type]]`) across files. That strikes me as stupid, but not wrong, and I think there are clear advantages of allowing model-specific data to be bundled with model config settings.

Fixes #709 
Fixes #234 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/using_the_ve/data/data.md
+++ b/docs/source/using_the_ve/data/data.md
@@ -210,10 +210,9 @@ file="'../../data/xy_dim.nc'"
 var_name="temp"
 ```
 
-**NOTE**: At the moment,
-`core.data.variable` tags cannot be used across multiple toml config files without
-causing `ConfigurationError: Duplicated entries in config files: core.data.variable` to
-be raised. This means that all variables need to be combined in one `config` file.
+You can include `core.data.variable` tags in different files. This can be useful to
+group model-specific data with other model configuration options, and allow
+configuration files to be swapped in a more modular fashion.
 
 To load configuration data , you will typically use the `cfg_paths` argument
 to pass one or more TOML formatted configuration files to create a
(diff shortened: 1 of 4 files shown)


jacobcook1995 (commented):
Looks sensible to me! I think it might be worth adding a separate plants config `toml` file as part of this PR so that the `test_ve_run` test will confirm that the change is working for the intended use case

jacobcook1995 (approved):

davidorme:
>  I think it might be worth making a separate plants config as part of this PR so that the `test_ve_run` test will confirm that the change is working for the intended use case

That's basically coming in the plants model rejig in #707 anyway, so I'll merge here.

PR #714, opened 2025-01-31
Adds the spin-up step

### LINKED ISSUE #581: Implement model spin-up
From the proposal

- We need a mechanism to allow a model to run for many years while holding everything else constant. The principal use of this is to allow initial estimates of input data to converge to a biologically stable state before starting the ‘actual’ simulation from those stable conditions.
- There are two possible mechanisms for terminating the spin up process: run for a fixed number of years or run until a set of watched variables generate stable results to a given precision. It’s probably good to have both options and they could work together.
- In most cases, the spin up process can be thought of as a set of sequential runs with initial runs that we might want to converge to a steady state before we start the simulation, and then an actual simulation using outputs from those convergence runs. For example: 
  - Run models with all frozen apart from Hydrology for 10 years, save soil moisture. 
  - Run models with all frozen apart from Soil for 500 years, save carbon variables. 
  - Run models with all active for 100 years, using the saved outputs from the first two runs (the _actual_ simulation). 
- That _could_ be done manually using three runs and configs but it would be useful to have a configuration mechanism that: 
  - Specifies a set of spin up “epochs” with different model settings and spin up termination rules
  - Identifies which variables should pass from spin up into the main simulation.
  - In many cases, we will want to restart the ‘actual’ simulation with different scenarios after the burn in, so it might be cleaner to setup a burn in process that runs models and then updates the input data, ready for the actual simulation.
- The immediate use case here is the Hydrology model, but any model could conceivably want to use this process.

dalonsoa:
# Description

Adds the possibility to run spin-up steps before the actual simulation. 

Details of PR TBC

Fixes #581 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/core/module_schema.json
+++ b/virtual_ecosystem/core/module_schema.json
@@ -257,6 +257,7 @@
                      },
                      "variables": {
                         "type": "array",
+                        "description": "Variables to be carried over to the next step.",
                         "items": {
                            "type": "string"
                         }
--- a/virtual_ecosystem/core/module_schema.json
+++ b/virtual_ecosystem/core/module_schema.json
@@ -250,15 +250,22 @@
                      },
                      "timing": {
                         "type": "object",
-                        "description": "Timing options for this spin_up step",
+                        "description": "Timing options for this spin-up step",
                         "allOf": [
                           { "$ref": "#/properties/core/properties/timing" }
                         ]
+                     },
+                     "variables": {
+                        "type": "array",
+                        "items": {
+                           "type": "string"
+                        }
                      }
                   },
                   "required": [
                      "models",
-                     "timing"
+                     "timing",
+                     "variables"
                   ]
                },
                "default": []
--- a/virtual_ecosystem/core/module_schema.json
+++ b/virtual_ecosystem/core/module_schema.json
@@ -233,6 +233,35 @@
                   "surface_layer_height",
                   "subcanopy_layer_height"
                ]
+            },
+            "spin_up": {
+               "description": "A sequence of spin-up steps to run before the normal run.",
+               "type": "array",
+               "items": {
+                  "type": "object",
+                  "properties": {
+                     "models": {
+                        "description": "List of models to spin-up together in this step.",
+                        "type": "array",
+                        "items":{
+                           "type": "string"
+                        }
+                        
+                     },
+                     "timing": {
+                        "type": "object",
+                        "description": "Timing options for this spin_up step",
+                        "allOf": [
+                          { "$ref": "#/properties/core/properties/timing" }
+                        ]
+                     }
+                  },
+                  "required": [
+                     "models",
+                     "timing"
+                  ]
+               },
+               "default": []
             }
          },
          "default": {},
--- a/virtual_ecosystem/core/variables.py
+++ b/virtual_ecosystem/core/variables.py
@@ -478,6 +478,15 @@ def get_variable(name: str) -> Variable:
         raise KeyError(f"Variable '{name}' is not a known variable.")
 
 
+def clear_run_variables() -> None:
+    """Clear the RUN_VARIABLES_REGISTRY, in preparation for another run.
+
+    Typically, this will be used between steps of a spin up process, so each new run
+    starts fresh.
+    """
+    RUN_VARIABLES_REGISTRY.clear()
+
+
 def get_model_order(stage: str) -> list[str]:
     """Get the order of running the models during init or update.
 
--- a/virtual_ecosystem/main.py
+++ b/virtual_ecosystem/main.py
@@ -314,13 +314,28 @@ def _spin_up_run(config: Config, progress: bool) -> None:
 def extract_spin_up_data(config: Config, step: int) -> None:
     """Extract the relevant data from the previous step and updates the config.
 
-    This function extracts the variables indicated in the configuration for the chosen
-    step, picks the last time entry for each of them from the output continous file
-    and puts them into a new file, saving it in that step subfolder. Then, it updates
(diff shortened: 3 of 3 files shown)


dalonsoa:
Maybe is not an issue in real life, but in its current implementation, if a variable is chosen to be carried over from the previous step, it might trigger errors in either de variable system or the the static configuration check.

For example, the following config file runs fine the spin-up but, for the actual simulation it fails with the error below (which was to be expected). I am not sure if this can never happen in a real scenario but I feel this is going to cause trouble. 

Input file:
```toml
[core]
[[core.spin_up]]
models = ["hydrology"]
timing = {start_date = "2013-01-01", update_interval = "1 month", run_length = "2 years"}
variables = ["air_temperature", "relative_humidity"]
[hydrology]
[abiotic_simple]
[animal]
[litter]
[soil]
```

Error
```
ValueError: Variable air_temperature initialised by abiotic_simple already in registry as initialised by data.
```

jacobcook1995 (commented):
@dalonsoa this looks like a sensible approach to me. I unfortunately don't have much to add beyond that, but I am sure others will have more to say

vgro:
> Maybe is not an issue in real life, but in its current implementation, if a variable is chosen to be carried over from the previous step, it might trigger errors in either de variable system or the the static configuration check.
> 
> For example, the following config file runs fine the spin-up but, for the actual simulation it fails with the error below (which was to be expected). I am not sure if this can never happen in a real scenario but I feel this is going to cause trouble.
> 
> Input file:
> 
> ```toml
> [core]
> [[core.spin_up]]
> models = ["hydrology"]
> timing = {start_date = "2013-01-01", update_interval = "1 month", run_length = "2 years"}
> variables = ["air_temperature", "relative_humidity"]
> [hydrology]
> [abiotic_simple]
> [animal]
> [litter]
> [soil]
> ```
> 
> Error
> 
> ```
> ValueError: Variable air_temperature initialised by abiotic_simple already in registry as initialised by data.
> ```

What exactly are the 'variables' in this setup?

alexdewar (commented):
I fear I've lost track of VE development a bit too much to be able to provide too much in the way of useful comments, but it does seem like a sensible approach in general.

Bigger picture, I am wondering if there might be a way to avoid having `RUN_VARIABLES_REGISTRY` as a global variable in general. If, for example, it's only ever used in the body of a main loop, you could just declare it as a variable in that loop and pass it into the functions that need it, then you'd get a clean slate on every iteration anyway. But I suspect it's probably not that simple :wink:

dalonsoa:
> What exactly are the 'variables' in this setup?

They are the variables from that step that will be used in the following step. If we look at [the issue](https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/581) this is trying to solve, there you said:

> Run models with all frozen apart from Hydrology for 10 years, save soil moisture.

So `soil_moisture` will be one of these `variables`.

vgro:
> > What exactly are the 'variables' in this setup?
> 
> They are the variables from that step that will be used in the following step. If we look at [the issue](https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/581) this is trying to solve, there you said:
> 
> > Run models with all frozen apart from Hydrology for 10 years, save soil moisture.
> 
> So `soil_moisture` will be one of these `variables`.

Ok, that makes sense. Does the same error come up when you use variables that are more exclusive to the hydrology model, such as soil moisture? `air_temperature` for example is a bit of a tricky one because it is initialized by the `hydrology` model but updated by the `abiotic` model. I can't quite figure out in this spinup functionality what is going on in terms of static etc, is it equivalent to static with one update?

dalonsoa:
Same problem. `soil_moisture` is declared to be initialised by the `hydrology` model, so when it is there already, in the data object, as it has been loaded from the previous step, the variables system complains.

The models that are declared static - all the models except those we are spinning up in the given step - follow the standard rules for static models (first update if no variables to be initialised are in the data object or frozen if all variables are there). For the actual simulation, none of the models are static - unless they were declared as such in the config file.

The bottom line is that we cannot provide some variables as input that were supposed to be initialised by a model. The variables system prevents it. But even if we modify the variables system to allow that, the input value will be overwritten when the model is initialised in the real simulation (where presumably, it won't be static) with whatever it calculates.

So the only way forward I see is setting up some mechanism within the model - or the data object - that prevents creating a new variable if it is already there. But that might lead to inconsistent results if other variables depend on the calculated one, so I'm not sure of the best way forward. It might require a significant refactoring of the model system.

davidorme:
> The models that are declared static - all the models except those we are spinning up in the given step - follow the standard rules for static models (first update if no variables to be initialised are in the data object or frozen if all variables are there). For the actual simulation, none of the models are static - unless they were declared as such in the config file.

And we have another problem there: because we use a simple boolean flag and the data payload to determine the static mode, we cannot at the moment spin the model up in fully-frozen mode and then run the model in update-once mode.

davidorme:
@vgro and @jacobcook1995 How much of this can we achieve using manual spinup runs using the existing infrastructure? 

So:
* define a specific spin up config using the static system and run that 
* define a "real" run that points to the spun-up variables output for the required variables

My worry here is that anything we try and develop is going to be brittle until we have a better idea of the actual use cases. If we can develop that understanding more, we'll be able to guide this process. If we _can't_ do this manually now, then I think we think about how to fix those blocks for manual spinup before we go about trying to do it programatically?

dalonsoa:
@davidorme check Teams. I've pointed this same thing there and suggested a meeting.

davidorme:
> check Teams.

Too many channels :smile:

dalonsoa:
I'm closing this, for now, as it seems it won't be implemented yet - and if it is, it will need significant changes. We will resume this work when appropriate.

PR #716, opened 2025-01-31
Update config file path resolution

### LINKED ISSUE #715: Extend path resolution in configuration files
Way back when, we added code to resolve paths in configuration files. This is partly so that relative paths are always resolved relative to the specific config TOML they come from, and partly because VE needs resolved paths.  This was implemented in https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/269  - but that PR only handles resolution of file paths in `[core.data.variable]` blocks and specifically notes that this is something we may need to revisit.

That time is now - the PlantsModel needs to load PFT definitions and forcing that through the `data` object is extremely heavy handed and will create a bunch of single use constant "variables" floating around in the `data` object.

The issue here is that the config file values from TOML have a very limited number of types, so paths are just strings, and we only want to try and resolve the strings that are paths and not other things.

* A clever approach would be to use something that types path settings in a detectable way. We could add "I am a path" properties to our JSON Schema or possibly replace our schema setup with `pydantic`, which would be cleaner and more pythonic. The problem though is that we are resolving paths in arbitrary chunks of config _before_ combining and validating them. So it's not clear how we'd match values to the schema. 

* A more pragmatic approach right now is to use the config key name to identify paths. We update the `_resolve_config_paths` function such that:
    * It walks over the entire config file contents,
    * looking for `key, value` pairs where `key.endswith('_path')` and `isinstance(value, str)`
    * and updates those values with the existing resolution mechanism in that function.

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/4cab25cbb957d94cb76486b2f2aba89e9aa6d33a/virtual_ecosystem/core/config.py#L100-L114
 
This does:

* involve updating the `[core.data.variable]` spec to use `file_path` and not `file`, but I think it is worth having a single clear pattern.
* mean we have to avoid using keys ending in `_path` for other uses, but that seems reasonable.

davidorme:
# Description

This PR:

* Updates `core.config._resolve_config_paths` to recursively search a config file payload for values stored under keys ending in `_path` and then resolves those values relative to the config path. The function previously only did this to `file` entries in `[core.data.variable]` tables and recent changes in model configurations now need options to resolve other file paths within the config tree.

* Extends the test suite to check path manipulation and error handling work as expected

* Updates the core model schema to change `file` to `file_path` and then rolls out the change in the `[core.data.variable]` schema specs across example files and tests to get this to apply. This is a bit of a sweeping change but I think it is long term easier to maintain the pattern "`xyz_path` needs resolving" than to have to maintain a list that "these specific key paths need resolving".

Fixes # 715 

## Type of change

## Key checklist

## Further checks


docs/source/development/design/defining_new_models.md
```diff
@@ -336,6 +336,14 @@ Virtual Ecosystem configuration and so can be deleted. You may also need to edit
 properties are required and which provide defaults that will be used to fill missing
 properties.
 
+One last detail of model configurations is that you may want to allow users to specify a
+file path to configure your model. This should not be used for loading the data used to
+run the model but might be used to load further details. If you do want to include a
+file path in your congfiguration, you need to pick a configuration key name that ends in
```
jacobcook1995: It might be worth giving an example here, as it's pretty technical language which I think becomes much more obvious when you give an example of what a non-`data` data file might contain

jacobcook1995 (approved):
LGTM! Just had a minor documentation suggestion

docs/source/development/design/defining_new_models.md
dalonsoa: Agree. I have to admit that without the explanation in the PR description, I would not know what this is suppose to be doing.

virtual_ecosystem/core/config.py
```diff
@@ -97,36 +97,34 @@ def config_merge(
     return dest, conflicts
 
 
-def _resolve_config_paths(config_dir: Path, params: dict[str, Any]) -> None:
+def _resolve_config_paths(config_dir: Path, config_dict: dict[str, Any]) -> None:
```
dalonsoa, line 100: I didn't checked this back in the day, so I mention it here now. This will work totally find in Mac and Linux (bugs aside) but will fail in Windows if any of the paths is in a different unit than the configuration file. Let's say your config file is in `C:` but you have all you data files in a separate drive, eg. `D:`. In this case, there will be an error as it is not possible to stablish relative paths between locations in different units. It's a very specific case, but I mention it because I learn it the hard way!

dalonsoa (changes requested):
The code looks OK, but I think the explanation on what this is for and how to use it needs some improvement. As it is now, it is very much unclear. 

I've also flagged one potential issue, although that might be better addressed in a separate PR.

virtual_ecosystem/core/config.py
davidorme: @dalonsoa  Just to check I understand this. If we have a config file on Windows, then the file could be `C:/config.toml` but contain references to `D:/data/data1.nc`. Within the single config file, does that data path _have_ to be absolute, because there is no way in Windows to get a relative path onto the `D:/` drive from `C:/`? 

If that is the case, then we just need to make sure those `D:/` references don't get tampered with. I think this is another PR.

docs/source/development/design/defining_new_models.md
davidorme: @dalonsoa Yup - I've updated both this section and the `_resolve_config_paths` docstring. Is that now clearer?

jacobcook1995: I guess the part I find difficult is "specify a file path to configure your model", in that I think you can only understand this if you already have a pretty good idea of how the model configuration system works. I guess something expanded on what this means/why you would want to configure the models with additional non-data files would be useful

davidorme: Oh FFS. I hadn't pushed my update to that section. I've pushed it and added a concrete example.

virtual_ecosystem/core/config.py
dalonsoa: Yep. In Windows, you cannot stablish relative paths between locations in different units. It just does not let you. At least, `pathlib` cannot resolve that situation. This is what I get when I tested it a few minutes ago:

```
>>> from pathlib import Path
>>> a = Path.cwd()
>>> a
WindowsPath('C:/Users/dalonsoa')
>>> b = Path(r"D:/agdp.exe")
>>> b
WindowsPath('D:/agdp.exe')
>>> b.exists()
True
>>> b.relative_to(a)
  File "<stdin>", line 1, in <module>
  File "C:\Users\dalonsoa\.pyenv\pyenv-win\versions\3.12.4\Lib\pathlib.py", line 682, in relative_to
    raise ValueError(f"{str(self)!r} is not in the subpath of {str(other)!r}")
ValueError: 'D:\\agdp.exe' is not in the subpath of 'C:\\Users\\dalonsoa'
>>> a.relative_to(b)
  File "<stdin>", line 1, in <module>
  File "C:\Users\dalonsoa\.pyenv\pyenv-win\versions\3.12.4\Lib\pathlib.py", line 682, in relative_to
    raise ValueError(f"{str(self)!r} is not in the subpath of {str(other)!r}")
ValueError: 'C:\\Users\\dalonsoa' is not in the subpath of 'D:\\agdp.exe'
```

docs/source/development/design/defining_new_models.md
jacobcook1995: Oh yeah the new text is much clearer!

virtual_ecosystem/core/config.py
```diff
@@ -100,9 +100,19 @@ def config_merge(
 def _resolve_config_paths(config_dir: Path, config_dict: dict[str, Any]) -> None:
     """Resolve paths in a configuration file.
 
-    Takes the path of a directory containing a given configuration file and resolves any
-    file paths in the configuration file contents, relative to that file location. File
-    paths as string values that have a key ending in ``_path``.
+    Configuration files may contain to keys providing file paths for data and other
```
dalonsoa:
```suggestion
    Configuration files may contain keys providing file paths for data and other
```

dalonsoa (approved):
Much clearer now!

PR #724, opened 2025-02-04
Add nitrogen fixation to the soil model

### LINKED ISSUE #708: Add functionality for nitrogen fixation
We need functionality to capture the process of nitrogen fixation. Discussed this with @davidorme and we agreed that as a first pass all of the fixation should happen in the soil model to begin with (this is something that we might want to revisit down the line to allow a more complex plant associated nitrogen fixation model)

jacobcook1995:
# Description

This PR adds nitrogen fixation to the soil model. There are two distinct types, fixation by free-living microbes and fixation by plant associated microbes. At some point in the future it might make sense for the plant associated fixation to happen within the `plants` model, but for now keeping it the soil seems simplest.

To get this to work I had to add code to the plants model so @davidorme if you could take a look and check that my changes seem reasonable that would be great.

Again this is a PR with fairly dense scientific content, but any feedback on code style etc would be appreciated.

Fixes #708 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -92,6 +93,7 @@ class PlantsModel(
         "leaf_turnover_c_p_ratio",
         "plant_reproductive_tissue_turnover_c_p_ratio",
         "root_turnover_c_p_ratio",
+        "nitrogen_fixation_carbon_supply",
```
dalonsoa, line 96: If this is initially populated by the plants model, it seems more reasonable to make the plants model own this process, I think. You have already mentioned this, so just supporting the idea. 

Having said that, and looking at the code, I wonder why this needs to be initialised by the plants instead of the soil model. Is it to avoid an unfeasible model execution order or something like that?

virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -333,6 +333,54 @@ def calculate_denitrification_temperature_factor(
...
+    thermal_sensitivity: float,
+    cost_equality_temp: float,
+):
+    """Calculate the cost of symbiotic nitrogen fixation in carbon terms.
+
+    The function used here is adapted from an empirical function provided in
+    :cite:t:`brzostek_modeling_2014`. As the function is not defined below zero degrees
+    celsius if a negative temperature is input an infinite cost is returned.
+
+    TODO - I could not sensibly convert this empirically derived function from Celsius
```
dalonsoa: It's a piecewise function with a cut-off threshold, so the way you have handled it seems totally reasonable specially considering that the input `soil_temp` is already in Celsius. I don't see any problem with it.

dalonsoa (approved):
No idea about the science, but the code looks sensible. I've just made a couple of comments.

virtual_ecosystem/models/soil/env_factors.py
jacobcook1995: Yeah I think this TODO more speaks to an anxiety I have about how a lot of the models I am drawing from make use of temperature, rather than anything wrong with the function implementation. I've made a broader issue for this #728, so I think it makes sense to make this a comment rather than a TODO

virtual_ecosystem/models/plants/plants_model.py
jacobcook1995: I guess it needs to be populated by the plants model because this variable represents the amount of carbon plants are prepared to supply to microbes living in the soil to fix nitrogen. At the moment this is just a constant that gets passed from the plants model to the soil model, but down the line it should depend on plant productivity, which means the plant model will need to run before it can be populated.

I think down the line it might make sense for symbiotic nitrogen fixation to be represented within the `plants` module, because there is a grey area here about whose domain this falls into. But when me and @davidorme discussed it we thought it is best to have it contained in the soil model to begin with

PR #725, opened 2025-02-04
Adding python `3.13` as a supported version

### LINKED ISSUE #722: We don't currently support python `3.13`
We don't currently support `python 3.13`. This is a problem as windows users being prompted to install python get pointed to `3.13` as a default which then leaves them unable to install the `virtual_ecosystem` package. I can't see any good reason for us not to support `3.13` so I think we should add it as a supported version

jacobcook1995:
# Description

This PR adds python `3.13` as a supported version so that if people install the most recent version of python they can actually download the package without having to roll back.

This PR would mean that we now support 4 versions of python, which I guess isn't really a problem. As far as I recall we discussed whether there were any syntax advantages to upgrading to `3.11` as a minimum version and we didn't think that there were. `3.10` also has about a year and a half of official support left, so there's no pressing need to drop it.

Fixes #722

## Type of change

## Key checklist

## Further checks


davidorme (approved):
LGTM

PR #726, opened 2025-02-05
Feature/stoichiometry for Animals

TaranRallings:
# Description

This PR contains the transition from a purely carbon mass system to a carbon, nitrogen, phosphorus (cnp) system. 

Key Changes:
- the cnp mass dictionary is now the core mass used in most methods
- mass_current (total mass) is now a property, as is reproductive_mass, as the total values are sometimes needed
- a 'grow' method has been implemented in AnimalCohort to ensure stoichiometric ratios are adhered to
- a lot of tests were reworked and expanded. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -489,16 +491,22 @@ def calculate_litter_additions_from_herbivory(self) -> dict[str, DataArray]:
...
+            if self.leaf_waste_pools[cell_id].mass_cnp["nitrogen"] > 0
+            else 0.0
             for cell_id in self.data.grid.cell_id
         ]
         leaf_c_p = [
-            self.leaf_waste_pools[cell_id].c_p_ratio
+            self.leaf_waste_pools[cell_id].mass_cnp["carbon"]
+            / self.leaf_waste_pools[cell_id].mass_cnp["phosphorus"]
+            if self.leaf_waste_pools[cell_id].mass_cnp["phosphorus"] > 0
+            else 0.0
```
jacobcook1995: Sorry despite what I said earlier using 0.0 here doesn't really work, but obviously using `inf` for the divide by zero case would break the downstream calculations. Not 100% sure what the best approach is here, I guess this is a value that you should never expect, so maybe we just accept this as a weird edge case?

virtual_ecosystem/models/animal/decay.py
```diff
@@ -157,54 +204,95 @@ class LitterPool:
...
-        # Return the net mass gain of detritivory, after considering
-        # digestive efficiencies
-        net_mass_gain = (
-            actual_consumed_mass * detritivore.functional_group.conversion_efficiency
-        )
+        # Apply mechanical efficiency before limiting available mass
+        efficiency = detritivore.functional_group.mechanical_efficiency
+        actual_consumed_mass = actually_available_mass * efficiency
+
+        # Avoid division by zero in nutrient fractions calculation
```
jacobcook1995: You actually take a step to avoid division errors here, is there a reason not to do this in the plant waste products case above?

jacobcook1995 (commented):
I've looked at the exchanges with the litter/soil model and I am happy that information will still get passed into my models in the format I want it in. I'm not approving as I don't feel well placed to give feedback on the overall structure/approach.

I had one specific comment which directly contradicts what I told you on teams earlier, sorry!

dalonsoa (commented):
I have not gone into the detail of the code, but don't have much time today and wanted to give you some early feedback. This very much calls for a dedicated class in which to put together all the `cnp` related manipulations. Something like the following will also facilitate manipulation and type checks:

```python
@dataclass 
class CNP:
    carbon: float
    nitrogen: float
    phosphorus: float

    @property
    def total(self) -> float:
        return self.carbon + self.nitrogen + self.phosphorus

    def __getitem__(self, key):
        """In case you want dictionary-style access."""
        return getattr(self, key)
```

Or, conversely, you can make it inherit from dict and add properties, as needed:

```python
class CNP(dict):

    @property
    def carbon(self) -> float:
        return self["carbon"]

    @property
    def nitrogen(self) -> float:
        return self["nitrogen"]

    @property
    def phosphorus(self) -> float:
        return self["phosphorus"]

    @property
    def total(self) -> float:
        return sum(self.values())
```

Given that `CNP` is something that seems to have some entity as a group, I think it deserves to be treated as such in the code, also facilitating its documentation.

TaranRallings:
> This very much calls for a dedicated class in which to put together all the `cnp` related manipulations. 

@dalonsoa This is a really good idea and I'm annoyed that I didn't think of it! I'll make the adjustments.

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: Okay actually think you want to add the following between the if and the else
```
elif self.leaf_waste_pools[cell_id].mass_cnp["carbon"] > 0 np.inf```

Comment (jacobcook1995) [reply virtual_ecosystem/models/animal/animal_model.py]: (the above might not be valid, but you get the idea)

This means that if there is a carbon flow with no associated nitrogen/phosphorus an infinite value gets returned (which then raises an error downstream). And this makes sense because biological matter always contains N and P.

But this handles the "there's no phosphorus because there's no organic matter at all" case gracefully

**Review by jacobcook1995 — COMMENTED** [review]
I'm happy with how the links to soil/litter now work!

File: virtual_ecosystem/models/animal/cnp.py [hunk 3]
```diff
@@ -0,0 +1,140 @@
...
+        Args:
+            carbon (float): The mass of carbon to subtract.
+            nitrogen (float): The mass of nitrogen to subtract.
+            phosphorus (float): The mass of phosphorus to subtract.
+        """
+        self.carbon -= carbon
+        self.nitrogen -= nitrogen
+        self.phosphorus -= phosphorus
+
+    def to_dict(self) -> dict[str, float]:
```
dalonsoa: Not needed. `dataclasses` have this built in. See https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict

virtual_ecosystem/models/animal/cnp.py
```diff
@@ -0,0 +1,140 @@
...
+        """Subtract the provided amounts of C, N, and P from the current CNP object.
+
+        Args:
+            carbon (float): The mass of carbon to subtract.
+            nitrogen (float): The mass of nitrogen to subtract.
+            phosphorus (float): The mass of phosphorus to subtract.
+        """
+        self.carbon -= carbon
+        self.nitrogen -= nitrogen
+        self.phosphorus -= phosphorus
```
dalonsoa: First, why not just using negative number for subtracting and merging these two functions in a single `update` one?

Second, you might want to force these to be keyword argument-only functions and provide default values of 0 for all of them. Might not save typing, but will make using them more robust and less prone to error as you need to be explicit about what input is what. 

```suggestion
    def add(self, *, carbon: float = 0, nitrogen: float = 0, phosphorus: float = 0) -> None:
        """Add the provided amounts of C, N, and P to the current CNP object.

        Args:
            carbon (float): The mass of carbon to add.
            nitrogen (float): The mass of nitrogen to add.
            phosphorus (float): The mass of phosphorus to add.
        """
        self.carbon += carbon
        self.nitrogen += nitrogen
        self.phosphorus += phosphorus

    def subtract(self, *, carbon: float = 0, nitrogen: float = 0, phosphorus: float = 0) -> None:
        """Subtract the provided amounts of C, N, and P from the current CNP object.

        Args:
            carbon (float): The mass of carbon to subtract.
            nitrogen (float): The mass of nitrogen to subtract.
            phosphorus (float): The mass of phosphorus to subtract.
        """
        self.carbon -= carbon
        self.nitrogen -= nitrogen
        self.phosphorus -= phosphorus
```

So you can do things like `cnp.add(phosphorus=42)` or `cnp.substract(nitrogen=17)`. But, AS I said, I think it will be cleaner to have `cnp.update(phosphorus=42)` and `cnp.update(nitrogen=-17)`.

Finally, does it make sense to have a any of these negative? If not, there should be an error or a warning whenever the total of any of these quantities become negative.

virtual_ecosystem/models/animal/cnp.py
```diff
@@ -0,0 +1,140 @@
...
+    phosphorus: float
+
+    @property
+    def total(self) -> float:
+        """Calculate the total combined mass of C, N, and P.
+
+        Returns:
+            float: The sum of carbon, nitrogen, and phosphorus mass.
+        """
+        return self.carbon + self.nitrogen + self.phosphorus
```
dalonsoa, line 35: See my comment about negative values. This will make no sense if any of the values is negative.

dalonsoa (changes requested):
Very nice change! It looks cleaner and more robust. I only one comment, although it is a big one.

virtual_ecosystem/models/animal/cnp.py
TaranRallings: That makes sense. I'll make the change.

virtual_ecosystem/models/animal/cnp.py
```diff
@@ -52,42 +52,41 @@ def __getitem__(self, key: str) -> float:
...
-        self.nitrogen += nitrogen
-        self.phosphorus += phosphorus
-
-    def subtract(self, carbon: float, nitrogen: float, phosphorus: float) -> None:
-        """Subtract the provided amounts of C, N, and P from the current CNP object.
+        for name, value in {
+            "carbon": self.carbon,
+            "nitrogen": self.nitrogen,
+            "phosphorus": self.phosphorus,
+        }.items():
```
dalonsoa: A little shorter:
```suggestion
        for name, value in self.as_dict().items():
```

dalonsoa (approved):
Just a minor change, but this looks good to me!

PR #729, opened 2025-02-06
Update sphinx to 8.1

### LINKED ISSUE #727: Update `sphinx` to 8.1
This is mostly because I was trying to see if upgrading `sphinx` fixed an issue with resolving references. It doesn't fix the issue, but since the project builds under the most recent `sphinx`, we probably should move to using it.

davidorme:
# Description

This PR:

* Updates `pyproject.toml` to use `sphinx ^8.0.0`, which is currently `8.1.3`. We may need to be careful about the next release as it drop `python 3.10` support. I think that is just that we'd have to run `sphinx` in `python 3.11+` but something to keep in mind.

* Updates `conf.py` to:
   * Mute Sphinx 9 deprecation warnings
   * Remove some odd jupytext cruft that had crept in
   
* Adds the TODO extension to give nice clear TODO sections in docstrings. I've updated `hydrology_model.py` to use the directive and check it works and is rendering.

* Fixes a test bug - `xarray` is also upgraded in this and they have changed the wording on KeyErrors with `Dataset` objects.

The main reason for doing this was I upgraded to see if it fixed a bug. It didn't but we might as well upgrade it.

Fixes #727

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

PR #730, opened 2025-02-06
703 adding growth to the plantsmodel

### LINKED ISSUE #703: Adding growth to the PlantsModel
In the active branch #659 (PR #707), the `PlantsModel.estimate_gpp` method now uses `pyrealm.pmodel.PModel` to estimate GPP properly. It populates a model attribute `self.stem_gpp` that contains per cell arrays of the total stem gpp associated with each of the cohorts in the community (stored in `self.communities`). 

In this issue, we need to update `PlantsModel.allocate_gpp`:

* Use the `pyrealm.demography` implementation of the T Model to actually allocate this  GPP and calculate the $\Delta DBH$ for the stems and hence grow the trees. This should be relatively easy as the individual community objects within `self.communities` are `pyrealm.demography.Community` instances.
* Expose some of the data in the `data` object. Two turnover parameters in `PlantsModel.calculate_turnover` are estimated by the T Model: `self.data["leaf_turnover"]` and  `self.data["root_turnover"]` and I think those values should move into `allocate_gpp`. We need a separate PR to handle the update of the rest of `PlantsModel.calculate_turnover`

The critical bit in the code in #659 is this

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/8e40f50d8153c16f5b7e940bbe3db9b1780b2afd/virtual_ecosystem/models/plants/plants_model.py#L535-L539

That attribute is a per cell array of the stem GPP to be allocated.

sallymatson:
# Description

Adding functionality to `allocate_gpp`, which takes the gpp and then grows the plants 😄 

Fixes #703

## Type of change

## Key checklist

## Further checks


tests/models/plants/test_plants_model.py
```diff
@@ -161,6 +161,27 @@ def test_PlantsModel_estimate_gpp(fxt_plants_model, fixture_core_components):
...
+    """Test the allocate_gpp method."""
+
+    fxt_plants_model.stem_gpp = {
+        cell_id: np.array([55]) for cell_id in fxt_plants_model.communities.keys()
+    }
+    # Allocate GPP
+    fxt_plants_model.allocate_gpp()
+
+    for community in fxt_plants_model.communities.values():
+        # .. TODO: not sure what to test here? Main thing it does is change the
```
davidorme, line 174: Ha! The `.. TODO::` annotation is specific to function docstrings, so they appear nicely styled in the HTML docs. Here, plain old `TODO` will cover it 😄

tests/models/plants/test_plants_model.py
```diff
@@ -161,6 +161,27 @@ def test_PlantsModel_estimate_gpp(fxt_plants_model, fixture_core_components):
...
+        # dbh value, so maybe check that it has grown?
```
davidorme: Ha! The `.. TODO::` annotation is specific to function docstrings, so they appear nicely styled in the HTML docs. Here, plain old `TODO` will cover it 😄
 
 On the question though - yes, just checking they are larger than their state before the call seems reasonable.

tests/models/plants/test_plants_model.py
```diff
@@ -161,6 +161,27 @@ def test_PlantsModel_estimate_gpp(fxt_plants_model, fixture_core_components):
...
+        assert (community.cohorts.dbh_values > 0).all()
+
+    # Check that leaf_turnover and root_turnover are correctly calculated
+    for cell_id in fxt_plants_model.communities.keys():
+        # .. TODO: same as above, this is a bit logic-less
```
davidorme: I think this is hard to avoid though, we could set up something testing for specific values but that is going to be a pain to do. 

We _could_  generate some specific numeric expectations but that does feel a bit circular.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -557,18 +558,41 @@ def allocate_gpp(self) -> None:
...
+        self.data["root_turnover"] = xr.full_like(self.data["elevation"], 0)
+
+        # Loop over each cell
+        for cell_id in self.communities.keys():
+            community = self.communities[cell_id]
+            cohorts = community.cohorts
+
+            cohort_allometry = StemAllometry(
+                stem_traits=community.stem_traits, at_dbh=cohorts.dbh_values
+            )
```
davidorme: This is actually already populated in `community.stem_allometry` but at the moment there's no mechanism by which that gets updated when the DBH increases.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -557,18 +558,41 @@ def allocate_gpp(self) -> None:
...
+
+            cohort_allocation = StemAllocation(
+                stem_traits=community.stem_traits,
+                stem_allometry=cohort_allometry,
+                at_potential_gpp=self.stem_gpp[cell_id],
+            )
+
+            # Grow the plants by increasing cohort dbh
+            cohorts.dbh_values += cohort_allocation.delta_dbh[0]
```
davidorme: I don't think this should need that `[0]`? It should generate an array that maps 1-to-1 onto the cohorts?

davidorme (commented):
A couple of comments, but looks good so far.

tests/models/plants/test_plants_model.py
sallymatson: Oops!!!!

virtual_ecosystem/models/plants/plants_model.py
sallymatson: This operator wasn't working, but changing it to `cohorts.dbh_values = cohorts.dbh_values + cohort_allocation.delta_dbh` does. The former is a 1d array and the second is 2d, which is where the problem was coming in, but with the normal operators it works!

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Oh gotcha. Should I update `community.stem_allometry` here in that case?

jacobcook1995 (approved):
This all looks good in terms of what I'm expecting as input to the litter model. Very happy that a couple of my "guesses" of turnover rates are being killed off!

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yup - think it logically should be done in this method after the `cohort.dbh_values increment`?

davidorme (changes requested):
LGTM - just that recalculation of the StemAllometry to add, I think?

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Okay, @davidorme can you check that the updates are now in the right order? I actually changed it a little bit so that the GPP allocation uses the current `community.stem_allometry`, and then after the new DBH values are calculated, `communtiy.stem_allometry` gets reassigned using the new DBH. This makes sense to me, as allometry is dependent on DBH so the `stem_allometry` property should be updated after GPP is allocated and the tree grows to make it current.... but would love you to double check.

virtual_ecosystem/models/plants/plants_model.py
davidorme: Can you add a comment here about the clash in array dimensions. There's some inconsistency in `pyrealm` that needs ironing out.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -560,20 +561,43 @@ def estimate_gpp(self, time_index: int) -> None:
...
+            )
+            community.stem_allometry = updated_allometry
+
+            # TODO: move leaf/root turnover calculation to pyrealm & call here
+            leaf_turnover = updated_allometry.foliage_mass / community.stem_traits.tau_f
+            # Calculate total turnover from all cohorts in a grid cell
+            self.data["leaf_turnover"][cell_id] = np.sum(leaf_turnover)
+            self.data["root_turnover"][cell_id] = np.sum(
+                cohort_allocation.turnover - leaf_turnover
+            )
```
davidorme: I think this is in the wrong order - the turnover is the loss of the current allometry.
```suggestion
            # TODO: move leaf/root turnover calculation to pyrealm & call here
            leaf_turnover = community.stem_allometry.foliage_mass / community.stem_traits.tau_f
            # Calculate total turnover from all cohorts in a grid cell
            self.data["leaf_turnover"][cell_id] = np.sum(leaf_turnover)
            self.data["root_turnover"][cell_id] = np.sum(
                cohort_allocation.turnover - leaf_turnover
            )
            # Update community allometry with new dbh values
            community.stem_allometry = StemAllometry(
                stem_traits=community.stem_traits, at_dbh=cohorts.dbh_values
            )
 
```

davidorme (changes requested):
I think the order is out but other than that this looks great.

virtual_ecosystem/models/plants/plants_model.py
davidorme: I think the order is the wrong way around - the turnover is the loss of the current mass and then the new allometry has larger stems and more leaf mass, some of which has turned over.

sallymatson: Okay, makes sense!

davidorme (approved):
LGTM

PR #733, opened 2025-02-10
Added step to catch infinite stochiometries

### LINKED ISSUE #732: Litter model needs to handle infinite stoichiometric ratios
In #726 @TaranRallings has started calculating stoichiometries for matter flows being supplied to the litter model (they were previously just constants being fed to the litter model). This is all well and good but introduces the edge case where the input has an infinite stoichiometric ratio, i.e. the input is pure carbon. At present, this edge case (which we only rarely expect) causes the entire model to stall.

The litter model can amend its calculation to treat infinite stoichiometric ratios as cases that are 100% carbon. This will complicate some of litter model calculations but should have minimal effect elsewhere.

I did consider adding a step to just crash the model when an infinite stoichiometric ratio is returned, but I think this is probably overly restrictive

jacobcook1995:
# Description

I've added a new function to the litter model code that should catch any case where input organic matter stoichiometries or chemical proportions (in the case of lignin) go negative. This new function raises an error in this case as we have decided to treat this case as an error.

Fixes #732 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/litter/test_inputs.py
+++ b/tests/models/litter/test_inputs.py
@@ -1,5 +1,6 @@
 """Test module for models.litter.inputs.py."""
 
+from contextlib import nullcontext as does_not_raise
 from logging import ERROR
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


TaranRallings (approved):
LGTM!

PR #751, opened 2025-02-18
Feature/new microclimate model

vgro:
I finally managed to include (most of) the key processes for the `abiotic` model in one work flow. There are a lot of changes compared to the previous version - it might be easier to look at the raw new files following all the changes. I would appreciate any feedback at this point, even if it is just on a subset.

Major changes:
- renamed files and variables to have consistency between models
- simplified processes and a lot less variables in `data`. Best point to start might be `virtual_ecosystem/models/abiotic/microclimate.py`
- added wind speed as output to `abiotic_simple`, so I could get rid of the unnecessary initialization of abiotic variables in the `hydrology` model

Many things still to do, including (not complete list):
- account properly for time step and area, in particular with energy fluxes
- sensible way of iterating towards convergence
- relative humidity is 100%, we need to include some form of advection/outlet
- the indexing/chopping/writing is very clunky and make s the code a bit hard to read

- there are some functions floating around that I might or might not use in the future
- tests are mostly to check dimensions and order of magnitude is right, needs some more thought

- update documentation
- update conftest dummy data

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/old_functions.py
jacobcook1995, line 1: I don't think there's much value in keeping a file like this, when you can just go back in the git history. I would just delete this (and the associated test file) with a commit with a message like "deleting old abiotic model functions". You could also tag that commit on GitHub if you are particularly concerned about finding it in future

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -122,3 +127,115 @@ def find_last_valid_row(array: NDArray[np.float32]) -> NDArray[np.float32]:
...
+        temperature: Temperature, [C]
+        saturated_pressure_slope_parameters: List of parameters to calculate
+            the slope of the saturated vapour pressure curve
+
+    Returns:
+        Slope of the saturated pressure curve, :math:`\Delta_{v}`
+    """
+
+    return (
+        saturated_pressure_slope_parameters[0]
```
jacobcook1995: Are these parameters basically just coefficients to a polynomial (I've never been clear how to handle/name these)?

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -247,562 +236,335 @@ def calculate_longwave_emission(
...
-        ),
-        conductivity_from_ref_height=(
-            current_conductivities["conductivity_from_ref_height"][
-                layer_structure.index_filled_canopy
-            ]
-        ),
-        delta_v_ref=delta_v_ref,
-    )
+    n_layers = len(soil_temperature)
+    new_soil_temperature = soil_temperature.copy()
```
jacobcook1995: Why do you need to use `.copy()` here?

jacobcook1995 (approved):
Can't comment on the science but the overall flow looks sensible to me.

I had a few minor queries, but the biggest thing I would suggest is deleting the `old_functions` file and test file. That information is already stored in git, and keeping them just means having a huge amount of (possibly) redundant information around (it also makes this PR look like a lot bigger change than it is)

virtual_ecosystem/models/abiotic/old_functions.py
vgro: I agree, I was just keeping them because I was still moving functions around. Will delete in the next round.

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: Yes, they are. A better way of dealing with this is probably with `gettattr()` like this (from abiotic_simple):
```lower, upper = getattr(bounds, "soil_temperature")```

File: virtual_ecosystem/models/abiotic/abiotic_tools.py [hunk 3]
```diff
@@ -122,3 +127,115 @@ def find_last_valid_row(array: NDArray[np.float32]) -> NDArray[np.float32]:
...
+    return np.where(
+        air_temperature >= 0,
+        1 / dewpoint_temperature_positive - 273.15,
+        1 / dewpoint_temperature_negative - 273.15,
+    )
+
+
+def calculate_slope_of_saturated_pressure_curve(
+    temperature: NDArray[np.float32],
+    saturated_pressure_slope_parameters: list[float],
```
davidorme: I'd generally type these as tuples - safer to use an immutable type for constants here.

virtual_ecosystem/models/abiotic/abiotic_tools.py
davidorme: I'd generally do something like this to separate the handling of the params from the equation.

```python
    a, b, c, d = saturated_pressure_slope_parameters
    return (
        a * (b * np.exp(c * temperature / (temperature + d))) / (temperature + d) ** 2
    )

File: virtual_ecosystem/models/abiotic/abiotic_tools.py [hunk 4]
```diff
@@ -122,3 +127,115 @@ def find_last_valid_row(array: NDArray[np.float32]) -> NDArray[np.float32]:
...
+
+    Returns:
+        Vapour pressure at soil surface, [kPa]
+    """
+
+    return (
+        saturated_vapour_pressure
+        * np.exp(
+            soil_matric_potential / (gas_constant_water_vapour * soil_temperature)
+        ).squeeze()
```
davidorme: Why is squeeze needed here? Is it to remove singleton dimensions? If so, how do they occur? I'm asking because I'm cautious when a function changes the dimensionality of an input when it returns.

virtual_ecosystem/models/abiotic/energy_balance.py
davidorme: Yeah - from the code below, it looks like `new_soil_temperature = np.full_like(soil_temperature, np.nan)` would be more explicit?

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -247,562 +236,335 @@ def calculate_longwave_emission(
...
+        specific_humidity: specific humidity, [kg kg-1]
+        layer_thickness: Layer thickness, [m]
+        atmospheric_pressure: Atmospheric pressure, [kPa]
+        water_to_air_mass_ratio: Ratio of water vapour to dry air mass
+        dry_air_factor: Complement of water_to_air_mass_ratio, accounting for dry air
+        cell_area: Grid cell area, [m2]
 
     Returns:
-        Factors a_E and b_E for vapour pressure linearisation
+      list with updated specific_humidity, vapour pressure and vapour pressure deficit
```
davidorme:
```suggestion
      A dictionary containing arrays of updated ``specific_humidity``, ``vapour_pressure`` and ``vapour_pressure_deficit`` values.
```

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -128,685 +116,88 @@ def calculate_roughness_length_momentum(
...
-        canopy_height: Canopy height, [m]
+        reference_wind_speed: Reference wind speed above the canopy, [m s-1].
+        reference_height: Reference height above the canopy, [m].
+        wind_heights: Heights where wind speed is to be calculated, [m].
+        roughness_length: Momentum roughness length, [m]
         zero_plane_displacement: Height above ground within the canopy where the wind
             profile extrapolates to zero, [m]
-        roughness_length_momentum: Momentum roughness length, [m]
-        mixing_length_factor: Factor in calculation of mean mixing length, dimensionless
+        min_wind_speed: Minimum wind speed, [m s-1]
```
davidorme: Can you explain this in the docstring? My assumption is that something breaks horribly if it wind speed hits zero?

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -128,685 +116,88 @@ def calculate_roughness_length_momentum(
...
-        np.log((height - zeroplane_displacement) / roughness_length_momentum)
-        + diabatic_correction_momentum,
+    wind_speed = (
+        reference_wind_speed
+        * np.log((heights - zero_plane_displacement) / roughness_length)
+        / np.log((reference_height - zero_plane_displacement) / roughness_length)
     )
-
-    return np.where(wind_profile == 0.0, np.nan, wind_profile).squeeze()
+    return np.where(wind_speed >= 0, wind_speed, min_wind_speed)
```
davidorme: Shouldn't this be something like:

```suggestion
    return np.where(wind_speed > min_wind_speed, wind_speed, min_wind_speed)
```

Otherwise, you could get values below the minimum wind speed, but that are just greater than zero?

davidorme (commented):
This looks good - I think the big picture structures all make sense and link together. Some minor comments on the changes here but I've left off commenting on things like the hard coded Bounds and Consts, which are placeholders.

tests/models/abiotic/test_microclimate.py
```diff
@@ -0,0 +1,139 @@
...
+        time_index=0,
+        time_interval=3600,
+        cell_area=10000,
+        layer_structure=lyr_str,
+        abiotic_constants=AbioticConsts(),
+        abiotic_simple_constants=AbioticSimpleConsts(),
+        core_constants=CoreConsts(),
+    )
+
+    # exp_longwave_emission = lyr_str.from_template()
```
sallymatson: Are these code chunks left intentionally commented instead of deleted?

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -122,3 +127,115 @@ def find_last_valid_row(array: NDArray[np.float32]) -> NDArray[np.float32]:
...
+
+    Args:
+        air_temperature: Air temperature, [C]
+        effective_vapour_pressure_air: Actual vapour pressure, [kPa]
+
+    Returns:
+        Dewpoint temperature, [C]
+    """
+
+    e0_positive = 611.2 / 1000  # e0 in kPa
```
sallymatson: What is our protocol with these kinds of constants? Just wondering if it's worth putting them in the AbioticConstants so the equations are clear, or is that unnecessary given that maybe they are quite obvious values to anyone who would know?

sallymatson (commented):
I looked through all the code, and it seems sensible to me with my limited knowledge of microclimates etc. Made a few very small remarks. I think it'll be great to get this merged in, so that the tweaks can then be identified in smaller chunks!! Congrats on getting so much done already! :)

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: I think this is because mypy hates tuples

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: quite possibly a relict, will adjust

virtual_ecosystem/models/abiotic/abiotic_tools.py
davidorme: It _shouldn't_ do! What happens when it is  a tuple?

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: This is a good point, it happens in a few places and I usually put TODOs there. I will investigate further

virtual_ecosystem/models/abiotic/wind.py
vgro: yes, should not divide by zero

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: I removed the copy but David's suggestion just returns nan everywhere which is weird, I'll look into it

tests/models/abiotic/test_microclimate.py
vgro: yes, I was using this section to check whether the results were in the right order of magnitude and for debugging. I will remove it

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: we usually put these in the model constants and call them as arguments like in the example earlier discussed. I half expected that I would not keep this function so I didn't make the effort at this stage. I actually deleted it now, so no longer an issue here.

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: I cannot reproduce the error here, I think it was a problem with arrays but that might have been my error. I changed it in the constants

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: I found the issue, I made a mistake when selecting the index using `variable[layer_structure.topsoil_index]` instead of `variable[layer_structure.topsoil_index_scalar]`

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: looks like the += was the issue, I actually do not need a new variable here

sallymatson (approved):

vgro:
I merge this branch with many outstanding TODOs  so I can continue to work on other parts of the model. Also, I do not want this PR to get any bigger. The main outstanding problems mentioned in the intro are now outsourced to #780 , #781 , #782

Issue #748, opened 2025-02-18
Create plant constants for global lignin and CN and CP ratios


davidorme:
We need to move many of the values in `calculate_turnover` into PlantConstants, so that they are not hard-coded. Ultimately these may move into Plant Functional Traits, but we need to align that change with `pyrealm`.

We need to add the following constants for `..._lignin` , `..._c_n_ratio` and `..._c_p_ratio`

- "stem_lignin" (replacing "deadwood_lignin")
- "senesced_leaf_lignin" (replacing "leaf_turnover_lignin")
-  "leaf_lignin" (a new addition)
- "plant_reproductive_tissue_lignin" (replacing "plant_reproductive_tissue_turnover_lignin")
- "root_lignin" (replacing "root_turnover_lignin")

At the end of this PR, we can simple replace the hardcoded current assignment values in `calculate_turnover` with the values from the constants files. And we should test that a model configuration successfully alters the resulting value!

PR #765, opened 2025-03-04
747 initial simple mortality

### LINKED ISSUE #747: Initial simple mortality
* Add `per_stem_annual_mortality_probability` to the PlantConstants as a simple per stem probability of death: see e.g. https://nph.onlinelibrary.wiley.com/doi/full/10.1111/nph.20407. We'll need to scale that annual value to the time step of the model.

* Create a new `apply_mortality` method to the `PlantsModel` that uses this constant to apply stochastic mortality. The cohort `n_individuals` is reduced by the number of individuals dying and the biomasses associated with those stems are added to `deadwood_production`. We'd also need to increment the leaf litter and reproductive tissue pools. @jacobcook1995 What is the best way to bring those in?

* Remove deadwood production from calculate turnover.

> [!WARNING]
> Very simplistic - we will need size structured and possible PFT specific mortality and also "caused" mortality (e.g. carbon drought by overgrazing) but we'll start here.

sallymatson:
# Description

Adding in basic mortality. One question I have is how to deal with the potential of no trees dying if the numbers are too small. 

If your update interval is 1 week and you use 10% annual mortality, the mortality PER update interval is 0.19%, so if you don't have a very sizeable tree population, it's possible that it will just always round down and no trees will ever die. 

Should I tried to do something fancier here, or is this simple version ok for now with acknowledged?

Fixes #747 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -625,6 +634,41 @@ def allocate_gpp(self) -> None:
...
+            )
+
+            # Decrease size of cohorts based on mortality
+            cohorts.n_individuals = cohorts.n_individuals - mortality
+
+            # Update deadwood production
+            total_deadwood_mass = np.sum(mortality * community.stem_allometry.stem_mass)
+            self.data["deadwood_production"][cell_id] = total_deadwood_mass * (
+                1 - self.model_constants.percent_stem_mass_attributed_to_lignin
+            )
```
jacobcook1995: I think I explained this badly sorry! `deadwood_production` should be the total mass of wood and `deadwood_lignin` is the proportion of the wood in `deadwood_production` that is lignin

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -625,6 +634,41 @@ def allocate_gpp(self) -> None:
...
+            self.data["deadwood_lignin"][cell_id] = total_deadwood_mass * (
+                self.model_constants.percent_stem_mass_attributed_to_lignin
+            )
+
```
jacobcook1995, line 668: As the lignin content is a proportion this calculation becomes more complex (i.e. it's the change in a proportion). I can send an equation over, but I guess it's worth asking whether that's worth implementing now or later, as at present we assume that all trees have the same lignin content anyway

jacobcook1995 (changes requested):
Hi Sally, just had a couple of comments, I think they mainly stem from me not actually explaining the pools going into litter properly

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Ahhh makes sense! In that case I think I'll just remove the references to `deadwood_lignin` and update that with the rest of #748

jacobcook1995 (approved):
LGTM!

virtual_ecosystem/models/plants/constants.py
```diff
@@ -11,6 +11,9 @@
 class PlantsConsts(ConstantsDataclass):
     """Constants for the :mod:`~virtual_ecosystem.models.plants` model."""
 
+    per_stem_annual_mortality_rate: float = 0.1
+    """Basic annual mortality rate for plants."""
+
```
davidorme: This is fussy - particularly in a temporary simple implementation - but a mortality rate is usually given as a number out of a given population size (5.7 / 1000 for example), where the `per_stem_annual_mortality_probability` is the probability per stem (so is in [0,1]). Probably better to stick with using `probability` not `rate`.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -625,6 +636,36 @@ def allocate_gpp(self) -> None:
...
+        # Loop over each grid cell
+        for cell_id in self.communities.keys():
+            community = self.communities[cell_id]
+            cohorts = community.cohorts
+
+            # Calculate the number of individuals that have died in each cohort
+            mortality = np.round(
+                cohorts.n_individuals * self.per_update_interval_stem_mortality_rate
+            )
```
davidorme: I see the problem with the no mortality at small sizes. I think we should do this statistically:

```python
mortality = np.random.binomial(cohorts.n_individuals,  self.per_update_interval_stem_mortality_rate)
```
The binomial distribution gives the number of "successes" (deaths) out of a number of trials (number of stems) for the given per stem probability of success. This way we get stochastic mortality and every stem _could_ die even with low probabilities.

There is absolutely the right way to do background mortality, but the down side to this is that it makes the testing more problematic because the outcomes are no longer deterministic. We could do something crude - to check that after mortality the number of individuals is <= the previous state or use large enough populations that the chance of = is vanishingly unlikely. 

We can also seed the random number generator - the `distribute_monthly_rainfall` function in the hydrology model does this.

virtual_ecosystem/core/core_components.py
```diff
@@ -151,6 +153,10 @@ def __post_init__(self, config: Config) -> None:
 
         self.n_updates = int((self.end_time - self.start_time) / self.update_interval)
 
+        # Calculate the number of updates in one year
+        seconds_per_year = np.timedelta64(31536000, "s")
+        self.updates_per_year = seconds_per_year / self.update_interval
+
```
davidorme: This is a bit of a hack, but I don't think we should try anything more complex while our time handling is so simplistic.

```suggestion
        # Calculate the number of updates in one year
        # TODO - this is not calendar aware - variable length months and leap years.
        seconds_per_year = np.timedelta64(31536000, "s")
        self.updates_per_year = seconds_per_year / self.update_interval

```

davidorme (changes requested):
A minor complaint about terms and then I think you are right that there is an issue with small numbers. I think drawing mortality stochastically using a binomial distribution is the way forward.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -305,6 +307,12 @@ def _setup(
             (self.layer_structure.n_layers, self.grid.n_cells), False
         )
 
+        # Calculate the per update interval stem mortality rate
+        self.per_update_interval_stem_mortality_probability = (
+            model_constants.per_stem_annual_mortality_probability
+            / self.model_timing.updates_per_year
+        )
+
```
davidorme: Ah. Now... It turns out this is wrong and we need to treat it like compound interest. Which was obvious _after_ this was gently pointed out to me at breakfast by someone who actually works with mortality data.

```python
import numpy as np

annual_p_death = 0.1
n_steps = 12
per_step_p_death = 1 - (1 - annual_p_death) ** (1 / n_steps)


n_indiv_annual = np.full((10000,), fill_value=1e6, dtype="int")
n_indiv_annual -= np.random.binomial(n_indiv_annual, annual_p_death)

n_indiv_naive = np.full((10000,), fill_value=1e6, dtype="int")
for month in np.arange(n_steps):
    n_indiv_naive -= np.random.binomial(n_indiv_naive, annual_p_death / 12)


n_indiv_compound = np.full((10000,), fill_value=1e6, dtype="int")
for month in np.arange(n_steps):
    n_indiv_compound -= np.random.binomial(n_indiv_compound, per_step_p_death)

plt.hist(n_indiv_compound, label="compound")
plt.hist(n_indiv_naive, label="naive")
plt.hist(n_indiv_annual, label="annual")
plt.legend()
plt.show()
```

davidorme (changes requested):
This looks good to me _apart_ from the slight issue of me not knowing any maths...

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Oh! Well that plot is quite convincing! Will make the change

davidorme (approved):
👍

PR #767, opened 2025-03-05
Replace CUE with a logistic expression

### LINKED ISSUE #746: Review microbial carbon use efficiency (CUE)
Currently the temperature-dependent CUE function we used is based on Abramoff et al. (2018), who cited Xu et al. (2014), who then cited Devevre & Horwath (2000). In Devevre and Horwath, CUE came from a single rice paddy study! 
> by soil microorganisms during rice straw decomposition was determined in a rice paddy soil, under aerobic and anaerobic (flooded) conditions at different temperatures (5, 15, and 25C)

The latest study I could find is [Qiao et al. (2019)](https://www.nature.com/articles/s41598-019-42145-6) at the global scale. Mostly from US and Europe and not a lot of tropical studies, but seems to be the most extensive to date. Their parameterised function is also identical so should be easy to adapt.

jacobcook1995:
# Description

This PR replaces the current expression for CUE with a logistic expression. This is mainly so that the CUE can be parameterised to always be between 0 and 1.

@hrlai I've changed the function and constants so that it fits with the new model you suggested [here](https://github.com/ImperialCollegeLondon/ve_data_science/pull/7). Once the data science side is sorted/merged you could add the values for the constants here.

Let me know how that goes! (also don't worry about the broken tests, they are mostly downstream stuff, I can fix them once you've made your changes)

## Type of change

## Key checklist

## Further checks


hrlai:
@jacobcook1995 would you be able to add [this citation](https://www.nature.com/articles/s41598-019-42145-6) to Zotero and then the bibliography file so I can cite the new data? Thanks.

virtual_ecosystem/models/soil/pools.py
```diff
@@ -1123,26 +1123,36 @@ def calculate_maintenance_biomass_synthesis(
...
     """
 
-    return reference_cue - cue_with_temperature * (soil_temp - cue_reference_temp)
+    return 1 / (
+        1
+        + np.exp(
+            -reference_cue_logit
+            - logit_cue_with_temp * (soil_temp - cue_reference_temp)
+        )
+    )
```
hrlai: This inverse-logit function looks fine to me, but is it worth using an existing function in scipy? https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.expit.html

hrlai (approved):
Looks good to me, should be ready to merge once I update the constants and add a reference.

virtual_ecosystem/models/soil/pools.py
jacobcook1995: That's a good catch (I really should be reimplementing standard functions), thanks!

hrlai:
cool, I think I've added everything from this side and it is ready for you to take a final look.

jacobcook1995:
@hrlai looks good! I'll fix the testing and merge it next week

PR #773, opened 2025-03-06
Create plant constants for lignin and C:N / C:P ratios

### LINKED ISSUE #748: Create plant constants for global lignin and CN and CP ratios

sallymatson:
# Description

Fixes #748 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -640,36 +640,55 @@ def calculate_turnover(self) -> None:
...
+        # Lignin concentrations
+        self.data["stem_lignin"] = xr.full_like(
+            self.data["elevation"], self.model_constants.stem_lignin
+        )
+        self.data["senesced_leaf_lignin"] = xr.full_like(
+            self.data["elevation"], self.model_constants.senesced_leaf_lignin
+        )
+        self.data["leaf_lignin"] = xr.full_like(
+            self.data["elevation"], self.model_constants.leaf_lignin
+        )
```
jacobcook1995, line 657: Is there a reason for tracking leaf lignin and well as senesced leaf lignin? I'm guessing in practice they would be the same?

jacobcook1995 (approved):
LGTM!

Just had a science question, which might be more for @davidorme

virtual_ecosystem/models/plants/plants_model.py
davidorme: We certainly need to account for N and P differences in active and senesced leaf leaves, but I agree that different lignin concentrations are not needed. So here I think we should take the mass of the leaf turnover and multiply that by the `leaf_lignin` to give the leaf lignin input to litter.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -640,36 +640,55 @@ def calculate_turnover(self) -> None:
...
         )
-        self.data["deadwood_lignin"] = xr.full_like(self.data["elevation"], 0.545)
-        self.data["leaf_turnover_lignin"] = xr.full_like(self.data["elevation"], 0.05)
-        self.data["plant_reproductive_tissue_turnover_lignin"] = xr.full_like(
-            self.data["elevation"], 0.01
+
+        # Lignin concentrations
+        self.data["stem_lignin"] = xr.full_like(
+            self.data["elevation"], self.model_constants.stem_lignin
+        )
```
davidorme, line 651: I think we should revert this to "deadwood_lignin" to be clear what it is - ideally what we want to do here is take the mortality from the model (so that PR should go in first) and then multiply the biomass of those dead stems by the `stem_lignin` constant to give the inputs to the litter model.

At the moment the code is writing out the constant values, rather than actual masses.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -640,36 +640,55 @@ def calculate_turnover(self) -> None:
...
+        self.data["leaf_lignin"] = xr.full_like(
+            self.data["elevation"], self.model_constants.leaf_lignin
+        )
+        self.data["plant_reproductive_tissue_lignin"] = xr.full_like(
+            self.data["elevation"],
+            self.model_constants.plant_reproductive_tissue_lignin,
+        )
+        self.data["root_lignin"] = xr.full_like(
+            self.data["elevation"], self.model_constants.root_lignin
+        )
```
davidorme, line 664: Same here - I think these need to be masses? @jacobcook1995 ?

davidorme (changes requested):
I think my comments are aiming at a more complex solution than in the issue! I think if we get the mortality in, we can actually populate real values here and that might help clarify what quantities are used.

virtual_ecosystem/models/plants/plants_model.py
jacobcook1995: The soil model uses proportions, so provided lignin is expressed as a proportion of total carbon in the plant model this works

virtual_ecosystem/models/plants/plants_model.py
sallymatson: In the original code, the hard-coded constants for lignin were all proportions. I used the hard-coded values that Jacob had in the plants model, and just moved those values to the constants. So I believe that the calculation of the biomass would then happen downstream in the litter model.

Maybe this would be better to calculate here, but without changing the litter model then I think here it needs to stay proportions.

davidorme: Yup - that's fair. Let's keep the scope small here.

sallymatson: So I think this was also the motivation for changing `deadwood_lignin` to `stem_lignin` in #748 - given that it is not a mass, it makes more sense to clarify that it is technically the lignin content across all stems?

virtual_ecosystem/models/plants/plants_model.py
davidorme: But presumably we should be passing in mass, right?

virtual_ecosystem/models/plants/plants_model.py
davidorme: OK - let's resolve this in a meeting!

virtual_ecosystem/models/plants/plants_model.py
sallymatson: I don't remember the exact rationale, but #748 specifies:

- "senesced_leaf_lignin" (replacing "leaf_turnover_lignin")
- "leaf_lignin" (a new addition)

So shall I remove `senesced_leaf_lignin`, and `leaf_lignin` would then replace `leaf_turnover_lignin` instead?

davidorme: Ah crap. I think `leaf_lignin` is the exception here - we need separate N and P ratios for `leaf` vs `senseced_leaf` but I think @jacobcook1995  is right that the lignin will be the same. I'm not sure whether we want to hang on to them both for symmetry? Partly because I think there's an argument for creating a TissueProperties dataclass at some point to encapsulate these.

virtual_ecosystem/models/plants/plants_model.py
davidorme: OK - the model inputs also take mass _separately_ but we're not currently adding that information. I'm getting confused about the scope of this PR.

davidorme (approved):
LGTM - as we discussed, this is a sensible incremental step towards the more complex solution in future PRs/

PR #775, opened 2025-03-10
Animals - Aquatic Reproduction and External Migration

### LINKED ISSUE #753: Animals - Aquatic Reproduction and External Migration
The animal model needs a system for managing the aquatic life stages of animals like amphibians. 
It also needs a system for managing cohorts leaving the simulation area to return at a later date. 
These are essentially the same problem. 

We do not want to model the aquatic ecology explicitly. The solution should take the cohort out of the normal model sequence, freeze it, modify it's functional group (where appropriate), apply a mortality, and then return it at a later date. 

This basic freezing system could be shared with external migration.

TaranRallings:
# Description

**READY NOW**

This PR handles both aquatic reproduction and external migration from the system. These are closely related problems, which is why I did them together. 

The commonality is that both involve "pausing" a cohort with respect to the update sequence. When a cohort is born aquatically (frogs etc) or externally migrates, it is taken out of the `active_cohorts` pools and is placed in either the `migrated_cohorts` or `aquatic_cohorts` pools. They remain there for a (yet unparameterized) amount of time, either developing or doing migrationy stuff. On return, they experience mortality and are returned to the `active_cohorts` pool in the same grid square they originally left from. 

External migration happens with any cohort that has the external migration trait. My assumption is that this will mostly be seasonal but I have left the seasonality logic undeveloped for now and will have some conversations about that shortly.

Aquatic migration is part of the birth process. An aquatically reproducing organism births a new cohort into stasis. I have coded this trait as direct development even though this is not scientifically correct. It could be expanded upon later. 

As I was working on this, it became clear that my `birth` method was doing far too many things so I have also broken that up into a series of helper methods. The logic (and testing) should now be easier to follow. Blessedly, no big problems were discovered in the process. 

A lot of files were changed but almost all of the real content is in `animal_model.py`.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa:
Just to confirm, should I review or wait, considering the message at the top?

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -769,131 +774,231 @@ def remove_dead_cohort_community(self) -> None:
...
+
+        Args:
+            parent: The parent cohort.
+
+        Returns:
+            Reproductive mass for carbon, nitrogen, phosphorus (kg).
+        """
+        semelparous_loss = self.calculate_semelparous_mass_loss(parent)
+
+        return {
```
dalonsoa, line 816: I seem to remember a carbon-nitrogen-phosphorus dataclass being defined somewhere else (maybe for the soil model?). Maybe we want to reuse that, for consistency (although that might be done in a separate PR).

dalonsoa (approved):
Scientifically speaking, I can't say, but code-wise this seems pretty neat! It would have been best to separate in different PRs the refactor of the `birth` method and the addition of the new functionality, but changes can be followed well, so no problem.

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -1403,3 +1401,20 @@ def find_intersecting_carcass_pools(
...
+
+        Returns: A bool of whether it is time to migrate.
+
+
+        Notes:
+            This method uses Python's built-in :func:`random.random` function.
+
+        """
+
+        return random.random() <= self.constants.seasonal_migration_probability
```
alexdewar, line 1420: I forget... Do we let users provide a random seed to the simulation so that they can make this stuff reproducible? Or is this just intended as a temporary thing?

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -769,131 +774,231 @@ def remove_dead_cohort_community(self) -> None:
...
+            parent_cohort, reproductive_mass
+        )
 
-            # Cap loss to available mass
-            non_reproductive_mass_loss_c = min(loss_c, parent_cohort.mass_cnp.carbon)
-            non_reproductive_mass_loss_n = min(loss_n, parent_cohort.mass_cnp.nitrogen)
-            non_reproductive_mass_loss_p = min(
-                loss_p, parent_cohort.mass_cnp.phosphorus
-            )
+        if number_offspring <= 0:
```
alexdewar: Are there situations in which `calculate_offspring_count` can return numbers <0? That seems a bit weird. Or is it just defensive programming? If it can't ever be <0, I'd just do this for clarity:

```suggestion
        if number_offspring == 0:
```

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -744,23 +749,23 @@ def remove_dead_cohort(self, cohort: AnimalCohort) -> None:
...
+            del self.active_cohorts[cohort.id]
         else:
             raise KeyError(f"Cohort with ID {cohort.id} does not exist.")
 
     def remove_dead_cohort_community(self) -> None:
         """This handles remove_dead_cohort for all cohorts in a community."""
         # Collect cohorts to remove (to avoid modifying the dictionary during iteration)
         cohorts_to_remove = [
-            cohort for cohort in self.cohorts.values() if cohort.individuals <= 0
+            cohort for cohort in self.active_cohorts.values() if cohort.individuals <= 0
```
alexdewar: I'd rewrite the code a bit so that `cohort.individuals` can never be <0.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1076,9 +1181,96 @@ def metamorphose_community(self) -> None:
...
+        dt_float = float(dt / timedelta64(1, "D"))
+        for cohort in list(self.migrated_cohorts.values()):
+            cohort.remaining_time_away -= dt_float
+            if cohort.remaining_time_away <= 0:
+                self.reintegrate_cohort(cohort, source="migrated")
+
+        for cohort in list(self.aquatic_cohorts.values()):
+            cohort.remaining_time_away -= dt_float
+            if cohort.remaining_time_away <= 0:
+                self.reintegrate_cohort(cohort, source="aquatic")
```
alexdewar, line 1209: So this function does have the consequence that `remaining_time_away` might be left as less than zero afterwards. Not sure if you want to just set it to zero in this case? It might be nice if other code can rely on it being >=0 (not sure if other code is actually using it though)

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1076,9 +1181,96 @@ def metamorphose_community(self) -> None:
...
+        """Cycles through all active cohorts and checks for external migration.
+
+        Only calls `trigger_external_migration` for cohorts that are seasonal migrators.
+        """
+        for cohort in list(self.active_cohorts.values()):
+            if (
+                cohort.functional_group.migration_type == "seasonal"
+                and cohort.is_migration_season()
+            ):
+                self.migrate_external(cohort)
```
alexdewar, line 1264: Are migration seasons a relevant concept if the migration type isn't seasonal? If not, I'd be inclined to make `is_migration_season` just return false in this case and document this behaviour in the docstring.

alexdewar (approved):
Looks good! Agree with @dalonsoa: the PR's maybe a bit on the long side, but the code is nice and readable, so it wasn't too hard to review.

I've added a few comments/suggestions. As I've said in the comments, I'd personally tweak things so that `remaining_time_away` and `individuals` can never be <0 (i.e. if they dip below this, explicitly set them to zero). Otherwise it makes their semantics a bit confusing and it's a potential future footgun.

virtual_ecosystem/models/animal/animal_cohorts.py
davidorme: Funnily enough, I've been thinking about this recently and I think we _do_ want to be able to seed a simulation.

I _had_ been thinking that have a problem in that we cannot guarantee that models would always request the same number values from a single RNG and in the same sequence , so we'd very quickly lose reproducibility. But of course, the only way that this divergence would happen if is stochastic differences lead to e.g. different numbers of cohorts through mortality. The method execution order should not change, so with a single seeded generator we _should_ be able to seed the whole simulation?

I had been thinking of something complex like a core RNG factory that models could be use to request per process RNGs, but if the last paragraph is correct, then we can provide `core.rng`, seed that from a `core.random_seed` config and then every model just inherits that RNG?

I'm not 100% convinced that will work but I think it should do?

davidorme: And at the very least, it's a really simple PR to try out!

virtual_ecosystem/models/animal/animal_model.py
davidorme: We've been looking at the same thing in `plants` - we do though have the issue that we _might_ want to run a model without animals (although animals without plants is a bit less feasible 🤣 ) so sharing structures between models is problematic?

Actually it isn't is it -  we don't have to be using the model to import from it. It is only a problem if somebody plugs in a new model that _replaces_ the existing one and doesn't provide the object. And that's clearly somebody else's problem at that point.

I think there might be a case for some `core` structures like this, but that is definitely a different PR!

virtual_ecosystem/models/animal/animal_model.py
TaranRallings: I just checked and it actually can't slip below zero. I'm realizing that I've got a defensive coding habit that I should probably chill out about!

davidorme: > I just checked and it actually can't slip below zero. I'm realizing that I've got a defensive coding habit that I should probably chill out about!

You can also go ultra-pythonic and use "truthiness" rather than `== 0` 😄 

```python
In [314]: @dataclass
     ...: class Test():
     ...:     a: int
     ...: 

In [315]: x = Test(a=1)

In [316]: z = Test(a=0)

In [317]: [z for z in [x, z] if not z.a]
Out[317]: [Test(a=0)]
```

I don't know if that is as clear though!

davidorme: Although that _doesn't_ catch values below zero.

PR #785, opened 2025-03-13
Removed soil_moisture_capacity from hydrology constants

### LINKED ISSUE #768: `soil_moisture_capacity` should be deleted from `hydrology` constants
The `soil_moisture_capacity` currently exists in both `CoreConsts` and `HydrologyConsts` (this is so it is available to the soil model as well). It should only exist in one place so should be deleted from `HydrologyConsts`

(In future `soil_moisture_capacity` will become a variable, I think it's important to make sure it is only defined in one place, so that one of the constants doesn't get missed when this change happens)

jacobcook1995:
# Description

This PR removes the `soil_moisture_capacity` from `HydroConsts` as it is duplicated in `CoreConsts`, which is a possible source of confusion.

Fixes #768

## Type of change

## Key checklist

## Further checks


vgro (approved):
LGTM

PR #795, opened 2025-03-20
move initialisation of aerodynamic resistance to hydrology model

### LINKED ISSUE #794: move calculation of aerodynamic resistance to hydrology model
The aerodynamic resistance variables are used by both the `hydrology` and `abiotic` model. If the virtual ecosystem is run with the `abiotic_simple` model, some of these are missing, so they need to be moved to hydrology.

vgro:
Aerodynamic resistance of canopy is used in the hydrology model but was not available with the simple abiotic model. This PR initialised aerodynamic resistance canopy in the hydrology model and fixes the calculation in the abiotic model. At some point we might need to actually move the actual calculation (#794), but this required a lot of wind related stuff to be provided; ight be worth moving fully to the process-based abiotic model then. For now the variable issue is sorted.

One error remains with the abiotic model which I cannot figure out, @davidorme we chatted about this briefly, have you come across this problem before? I have printed the logs but cannot see a difference:

```tests/models/abiotic/test_abiotic_model.py:77: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

caplog = <_pytest.logging.LogCaptureFixture object at 0x110e5ef10>
expected_log = ((10, "abiotic model: required var 'air_temperature_ref' checked"), (10, "abiotic model: required var 'relative_humidi...iotic model: required var 'layer_heights' checked"), (10, "abiotic model: required var 'wind_speed_ref' checked"), ...)
subset = None

    def log_check(
        caplog: pytest.LogCaptureFixture,
        expected_log: tuple[tuple],
        subset: slice | None = None,
    ) -> None:
        """Helper function to check that the captured log is as expected.
    
        Arguments:
            caplog: An instance of the caplog fixture
            expected_log: An iterable of 2-tuples containing the log level and message.
            subset: Only check a specified subset of the captured log.
        """
    
        # caplog.records is just a list of LogRecord objects, so can use a slice to drop
        # down to a subset of the records.
        if subset is None:
            captured_records = caplog.records
        else:
            captured_records = caplog.records[subset]
    
        assert len(expected_log) == len(captured_records)
    
        assert all(
            [exp[0] == rec.levelno for exp, rec in zip(expected_log, captured_records)]
        )
>       assert all(
            [exp[1] in rec.message for exp, rec in zip(expected_log, captured_records)]
        )
E       assert False
E        +  where False = all([True, True, True, True, True, True, ...])

tests/conftest.py:43: AssertionError
```

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/abiotic/test_abiotic_model.py
+++ b/tests/models/abiotic/test_abiotic_model.py
@@ -370,7 +370,7 @@ def test_setup_abiotic_model(dummy_climate_data, fixture_core_components):
         model.update(time_index=0)
 
     expected_soil_temp1 = lyr_strct.from_template()
-    expected_soil_temp1[lyr_strct.index_all_soil] = np.array([18.701923, 19.989061])[
+    expected_soil_temp1[lyr_strct.index_all_soil] = np.array([18.727418, 19.989483])[
         :, None
     ]
     xr.testing.assert_allclose(model.data["soil_temperature"], expected_soil_temp1)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -546,8 +546,11 @@ def dummy_climate_data(fixture_core_components):
     data["molar_density_air"] = from_template()
     data["molar_density_air"][lyr_str.index_filled_atmosphere] = 38.0
 
+    data["density_air"] = from_template()
+    data["density_air"][lyr_str.index_filled_atmosphere] = 1.255
+
     data["specific_heat_air"] = from_template()
-    data["specific_heat_air"][lyr_str.index_filled_atmosphere] = 29.0
+    data["specific_heat_air"][lyr_str.index_filled_atmosphere] = 1.006
 
     data["attenuation_coefficient"] = from_template()
     data["attenuation_coefficient"][lyr_str.index_filled_atmosphere] = np.array(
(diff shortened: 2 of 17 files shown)


tests/models/abiotic/test_abiotic_model.py
sallymatson, line 1: I checked this out locally and found the issue. The test has an expected log that differed from the actual log in one line.

At line 35 you need to change the expected log from "Adding data array for 'vapour pressure deficit'" to "Replacing data array for 'vapour pressure deficit'". 

This worked locally and I think should fix the problem, but I don't want to do it directly in case it is not the behaviour you expect from the model and could indicate an upstream problem.

vgro: It works for me locally, too. I am not sure why some of the variables are added and others are replaced, Diego wrote this test when he replaced the init/setup use and I did not question it at the time. @dalonsoa could you please comment on this?

vgro: > It works for me locally, too. I am not sure why some of the variables are added and others are replaced, Diego wrote this test when he replaced the init/setup use and I did not question it at the time. @dalonsoa could you please comment on this?

I think we resolved it. The variables that are 'Added' are not in the dummy climate data set in `conftest.py`, and I added `vapour_pressure_deficit` with this PR

sallymatson (approved):
Although I can't speak to the calculations, the code + logic LGTM !

PR #800, opened 2025-03-24
798 sort out units of specific heat air latent heat vapourisation density of air

### LINKED ISSUE #798: Sort out units of specific heat air, latent heat vapourisation, density of air
Specific heat air, latent heat vapourisation, and density of air are used with different units across models which means there might be some inconsistencies in the use and/or some unnecessary conversions.

I suggest storing both the molar and the kg based variable in the data object.

vgro:
There were inconsistencies with the use of a few variables. I fixed `hydrology` and `abiotic`, @jacobcook1995  and @davidorme please check if your models are affected.

Fixes #798 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
The changes look sensible to me!

I don't think any of this touches on my model though, I use a "gas constant" but that's from thermodynamics so doesn't relate to any real gas. I guess @davidorme might be more effected though

davidorme:
I think I'm good too - I care about atmospheric pressure in Pa and vapour pressure deficit in Pa.

vgro:
> I think I'm good too - I care about atmospheric pressure in Pa and vapour pressure deficit in Pa.

Are you happy with me merging this and changing the unit when you read in the data for now? We can also move to Pa everywhere but I would prefer moving that to a separate PR, this would affect a lot of files...

vgro:
> > I think I'm good too - I care about atmospheric pressure in Pa and vapour pressure deficit in Pa.
> 
> Are you happy with me merging this and changing the unit when you read in the data for now? We can also move to Pa everywhere but I would prefer moving that to a separate PR, this would affect a lot of files...

I've created an issue (#802) to do this separately so I can move on with development of other parts. I will address this soon.

Issue #779, opened 2025-03-12
calculate_pH_suitability calculates upper section incorrectly


jacobcook1995:
**Describe the bug**
The function to calculate the microbial pH response returns incorrect values when the pH is above the upper optimum value but below the maximum value for microbial activity. I have given the correct expression in my docs branch but I need to adjust the code to match this

PR #801, opened 2025-03-25
Improved microbial enzyme secretion model

### LINKED ISSUE #760: Microbial enzyme secretion model needs to be improved
At the moment microbes secrete enzymes as a fixed maintenance cost, and these enzymes are assumed to have exactly the same stoichiometry as the group that secreted them. This assumption works for our current purposes, but really cannot capture the complex cost decisions that mycorrhizal fungi make. So this model needs to be improved.

I don't have a worked out solution for this yet, it probably needs some mulling over. I think a sensible aim is to try to capture key differences between taxa, rather than attempting to tackle the very thorny problems of microbial decision making processes. (at least for the time being)

jacobcook1995:
# Description

This PR changes how the calculation of enzyme secretion by microbes so that microbes have an allocate a certain fraction of new biomass to extracellular enzymes (previously it was a fraction of cellular maintenance). The reason behind this change is that it allows enzymes to have different stochiometries to the microbes producing them. This is something that needs to be added before fungi can be split between mycorrhizal and saprotrophic.

The main structural change arises because the microbial groups now need information about the enzyme classes (so that the average stoichiometry of the synthesised biomass can be found). So now a set of `EnzymeConstants` has to be provided for the creation of a `MicrobialGroupConstants` data class. Feedback on this would be appreciated!

I also fixed a bug I noticed with the calculation of microbial pH suitability.

Fixes #760 #779

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/soil/microbial_groups.py
```diff
@@ -99,14 +106,112 @@ class MicrobialGroupConstants:
...
+        / total_enzyme_allocation
+        for substrate, allocation in enzyme_production.items()
+    )
+
+    return {
+        "nitrogen": (c_n_ratio + enzyme_c_n_weighted * total_enzyme_allocation)
+        / (1.0 + total_enzyme_allocation),
+        "phosphorus": (c_p_ratio + enzyme_c_p_weighted * total_enzyme_allocation)
+        / (1.0 + total_enzyme_allocation),
+    }
```
dalonsoa: Please, let someone check my maths, but I think that you can take the dividing `total_enzyme_allocation` out of the sums, in which case it will cancel with the multiplying one  below:
```suggestion
    enzyme_c_n_weighted = sum(
        enzyme_classes[f"{name}_{substrate}"].c_n_ratio
        * allocation
        for substrate, allocation in enzyme_production.items()
    )

    enzyme_c_p_weighted = sum(
        enzyme_classes[f"{name}_{substrate}"].c_p_ratio
        * allocation
        for substrate, allocation in enzyme_production.items()
    )

    return {
        "nitrogen": (c_n_ratio + enzyme_c_n_weighted)
        / (1.0 + total_enzyme_allocation),
        "phosphorus": (c_p_ratio + enzyme_c_p_weighted)
        / (1.0 + total_enzyme_allocation),
    }
```

dalonsoa (approved):
I've just made a small, maths-related comment, but all looks good otherwise. Quite an interesting business!

virtual_ecosystem/models/soil/microbial_groups.py
jacobcook1995: Yes, you are right! I think I was overly focused on the line-by-line logic so I lost the big picture

PR #803, opened 2025-03-27
Average abiotic quantities over relevant soil layers

### LINKED ISSUE #533: Soil model should average abiotic quantities over relevant soil layers

jacobcook1995:
# Description

This PR changes the soil model so that it averages abiotic quantities over the relevant soil layers. The temperature and matric potential averaging works the same as in the litter model. The soil moisture "averaging" is something I just added, so sanity checking that would be really appreciated!

Fixes #533

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -405,3 +407,30 @@ def calculate_leaching_rate(
...
+        layer_structure: The LayerStructure instance for the simulation.
+
+    Returns:
+        The total soil moisture in the microbially active depth [mm]
+    """
+
+    # Find the fraction of each layer that lies within the microbially active zone
+    layer_weights = (
+        layer_structure.soil_layer_active_thickness
+        / layer_structure.soil_layer_thickness
```
vgro, line 435: I _think_ you might need to use the soil layer depth instead of the soil layer thickness here. If both layers are 1m thick, the depth of the first on is 1m, the depth of the second is 2m. if your microbial depth is then 1.2 m, you want all of the first layer but only 0.2 of the second layer, right? At the moment it looks like it always starts from zero?

virtual_ecosystem/models/soil/pools.py
```diff
@@ -415,20 +421,28 @@ def calculate_all_pool_updates(
...
-            top_soil_layer_index
-        ].to_numpy()
-        soil_temperature = self.data["soil_temperature"][
-            top_soil_layer_index
-        ].to_numpy()
-        soil_moisture = self.data["soil_moisture"][top_soil_layer_index].to_numpy()
+        # Find temperature, soil water potential and soil moisture values for the
+        # microbially active depth
+        soil_water_potential = average_water_potential_over_microbially_active_layers(
+            water_potentials=self.data["matric_potential"],
```
vgro, line 427: Is there a reason why you chose DataArray here and convert to numpy in other places?

virtual_ecosystem/models/soil/env_factors.py
jacobcook1995: Okay, I'll add a note explaining what `layer_structure.soil_layer_active_thickness` is as that isn't really clear from the name alone

virtual_ecosystem/models/soil/pools.py
jacobcook1995: Is basically that I want to make use of the layer structure for the matric potentials but in the other cases I can just use the `numpy` arrays

vgro (approved):
LGTM

PR #807, opened 2025-03-28
Feature/794 aero resistance

### LINKED ISSUE #794: move calculation of aerodynamic resistance to hydrology model

vgro:
This PR moves the initialisation of aerodynamic resistances to the hydrology model because the variable will be required for the canopy evaporation. If the VE runs with `abiotic_simple`, these are otherwise not inilialised. The actual calculation of aero resistance in canopy remains in the abiotic model. The implementation there is still tentative, fixing that is a separate issue.
@sallymatson and @davidorme you have seen this before, it is the branch that we lost in the GitHub void.

Fixes #794 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -120,12 +120,41 @@ def run_microclimate(
...
+    # temperature update function. Could have to do with low wind speeds.
+    # aerodynamic_resistance_canopy = energy_balance.calculate_aerodynamic_resistance(
+    #     wind_heights=data["layer_heights"][
+    #         layer_structure.index_filled_canopy
+    #     ].to_numpy(),
+    #     roughness_length=roughness_length,
+    #     zero_plane_displacement=zero_plane_displacement,
+    #     friction_velocity=friction_velocity,
+    #     von_karman_constant=core_constants.von_karmans_constant,
+    # )
```
sallymatson, line 149: Are these things you want to leave in?

sallymatson (approved):
Made one comment.. otherwise, LGTM !

virtual_ecosystem/models/abiotic/microclimate.py
vgro: yes, I have moved this in and out in the past, it produces totally wrong values and I think it is due to the numbers I put in, not the function itself. I don't want to do it in this PR because it's a more general issue with time intervals.

PR #817, opened 2025-04-04
755 animals   vertical occupancy

### LINKED ISSUE #755: Animals - Vertical Occupancy
Animals currently do not distinguish between vertical layers. This needs to be changed such that their trophic interactions only occur where they have vertical overlap in the same way that horizontal overlap is currently required. 

After building the lists of foods that share vertical overlap, the foraging process should isolate foods that share vertical overlap. This should ensure that moles do not eat canopy insects and so on.

TaranRallings:
# Description

This PR contains functionality for animals occupying particular vertical strata in the system. There are three such strata: soil, ground, and canopy. All cohorts and resource pools will have a vertical_occupancy tag specifying which of these strata they occupy. They can occupy more than one.  A trophic interaction can occur anytime a single strata layer is overlapping. 

I added a small refactor of `get_prey` to pull out the "can this trophic interaction even occur?" logic and put it into `can_prey_on`.

The key methods here are:
- get_prey
- can_prey_on
- get_plant_resources
- can_forage_on
- match_vertical

The trophic system expansion I am doing next will expand this to include vertical_occupancy and trophic interactions for a bunch of new things. The rewiring to actually tap into the plant system will occur then as well.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_traits.py
```diff
@@ -70,3 +70,14 @@ class MigrationType(Enum):
...
+
+class VerticalOccupancy(Enum):
+    """Enumeration for vertical occupancy trait."""
+
+    SOIL = "soil"
+    GROUND = "ground"
+    CANOPY = "canopy"
+    SOIL_GROUND = "soil_ground"
+    GROUND_CANOPY = "ground_canopy"
+    SOIL_GROUND_CANOPY = "soil_ground_canopy"
```
dalonsoa: I'm going to suggest a change here and then use that change in in `match_vertical` which I think will make the code more readable.

The idea would be to use flags instead of normal enumerations, parsing the string in the configuration.

```python
from enum import auto, Flag

class VerticalOccupancy(Flag):
    """Enumeration for vertical occupancy trait."""
    SOIL = auto()
    GROUND = auto()
    CANOPY = auto()
    
    @classmethod  
    def parse(cls, occupancy: str) -> "VerticalOccupancy":
        occupancy_list = occupancy.split("_")
        occupancy_flags = getattr(cls, occupancy_list.pop(0).upper())
        for oc in ocuppancy_list:
            occupancy_flags = occupancy_flags | getattr(cls, oc.upper())

        return occupancy_flags
```

This will produce flags that look like:

```
<VerticalOccupancy.SOIL|GROUND|CANOPY: 7>
<VerticalOccupancy.SOIL|GROUND: 3>
<VerticalOccupancy.GROUND|CANOPY: 6>
...
```
In addition to the pure ones.

Then... see the other comment.

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -1418,3 +1436,64 @@ def is_migration_season(self) -> bool:
...
+                VerticalOccupancy.SOIL_GROUND,
+                VerticalOccupancy.GROUND_CANOPY,
+                VerticalOccupancy.SOIL_GROUND_CANOPY,
+            },
+        }
+
+        return (
+            resource_occupancy
+            in occupancy_pairs[self.functional_group.vertical_occupancy]
+        )
```
dalonsoa: If you implement the `VerticalOccupancy` the way I suggest, these whole thing becomes:

```suggestion
        return bool(
            resource_occupancy & self.functional_group.vertical_occupancy]
        )
```

All the logic to define the validity is taken care for by the `Flag` enum. This might also come handy in other cases.

virtual_ecosystem/models/animal/functional_group.py
```diff
@@ -82,6 +84,8 @@ def __init__(
         """The excretion type of the functional group."""
         self.migration_type = MigrationType(migration_type)
         """The migration type of the functional group."""
+        self.vertical_occupancy = VerticalOccupancy(vertical_occupancy)
```
dalonsoa: This will change simply to:
```suggestion
        self.vertical_occupancy = VerticalOccupancy.parse(vertical_occupancy)
```

virtual_ecosystem/models/animal/plant_resources.py
```diff
@@ -39,6 +40,8 @@ def __init__(self, data: Data, cell_id: int, constants: AnimalConsts) -> None:
         """The mass of the plant leaf mass [kg]."""
         self.constants = constants
         """The animal constants, including energy density."""
+        self.vertical_occupancy: VerticalOccupancy = VerticalOccupancy.GROUND_CANOPY
```
dalonsoa: And this will be:
```suggestion
        self.vertical_occupancy: VerticalOccupancy = VerticalOccupancy.GROUND | VerticalOccupancy.CANOPY
```

dalonsoa (changes requested):
I've made a few suggestions - actually only one, but has multiple ramifications. I think it will make the code more readable and robust. Probably you will need to add other bits and pieces to make it work, both in terms of comments and type annotations.

virtual_ecosystem/models/animal/animal_traits.py
davidorme: Ooooh. That's very elegant. ✏️ 📓

TaranRallings: Much more elegant solution, thanks!

dalonsoa (approved):
LGTM!

PR #818, opened 2025-04-04
Adding subcanopy vegetation

### LINKED ISSUE #816: Ground vegetation
We need vegetative biomass at ground level for non-arboreal herbivores. The full T Model isn't well suited to shrubs - the single stem approach doesn't really fit and the limited maximum height will results in huge stem diameters.

For now, in lieu of a specific shrub model, we need a simple calculation of biomass following the available light at ground level. That can be driven by parts of the T Model (yield etc) but without any reproduction. It's just green stuff that appear, like algae, and has no height.

davidorme:
# Description

This PR introduces a simple model of subcanopy vegetation. There is a standing mass of vegetation that is treated as pure leaf biomass - which we can use to estimate LAI via a specific leaf area - and a mass of seedbank biomass, that sprouts new biomass with a give half life.

The PR:

* introduces 2 new plant carbon biomass pools: `subcanopy_vegetation_biomass` and `subcanopy_seedbank_biomass`. These need to be set to initialise the model and simply represent the amount of carbon mass.
* introduces 3 new `PlantsModel` methods: `set_subcanopy_light_capture`, `calculate_light_use_efficiency`  and `calculate_subcanopy_dynamics`.
* Extends the model constants with some subcanopy vegetation parameters

The `set_subcanopy_light_capture` method estimates the leaf area index given the biomass in kg m2 and from this estimates the light transmission through subcanopy using Beer-Lambert and hence the fAPAR of the subcanopy layer. It updates `leaf_area_index` and `layer_fapar` adding the subcanopy values to the surface layer (we may want to revisit this). With this in place, the existing `set_shortwave_absorption` method now includes absorption by the subcanopy.

The `calculate_light_use_efficiency` method simply breaks the P Model calculation out of `calculate_gpp` and stores the results in a new `pmodel` attribute so that the LUE for the surface layer can also be used for subcanopy productivity.

The `calculate_subcanopy_dynamics` method estimates the subcanopy GPP given the LUE and subcanopy PPFD and then partitions it into respiration, growth, allocation to seedbank biomass and some yield losses. The seedbank biomass resprouts at an exponential rate given the elapsed time in the update and produces new vegetation biomass, along with a sprouting loss factor.

If either pool is empty, it refills after update, but if both are empty no subcanopy biomass of either kind is created.

Fixes #816 

## Type of change

## Key checklist

## Further checks


davidorme:
@vgro This is breaking because it is introducing negative surface flows in the hydrology. Any suggestions? My first suspicion is that the plants are sucking the ground dry, but not sure how that sneaks through into surface flow.

davidorme:
> @vgro This is breaking because it is introducing negative surface flows in the hydrology. Any suggestions? My first suspicion is that the plants are sucking the ground dry, but not sure how that sneaks through into surface flow.

OK - that has gone away for some reason 🤷‍♂️

davidorme:
@sallymatson @vgro  and @TaranRallings Could you have a look at this?

Taran, it includes both ground vegetation and a herb/shrub/grass seedbank, both of which you should be able to target as a resource.

tests/models/plants/conftest.py
```diff
@@ -161,10 +187,24 @@ def fixture_canopy_layer_data():
...
+                ]
+            ),
+            [1, 2, 3, 4, 11],
+            # index_filled_canopy, index_surface
+        ),
+        "layer_fapar_canopy_only": (
+            "layer_fapar",
+            np.array(
+                [
+                    [5.86036011e-01, 5.86033790e-01, 5.86046818e-01, 5.93428098e-01],
```
vgro, line 200: What is the first layer here, are we not assuming 3 filled canopy layers? With vegetation in the subcanopy, the indexing will get awkward, something to revisit in the future.

virtual_ecosystem/core/core_components.py
```diff
@@ -85,6 +85,8 @@ class ModelTiming:
     """The configured run length as a pint Quantity."""
     update_interval: np.timedelta64 = field(init=False)
     """The configured update interval."""
+    update_interval_seconds: float = field(init=False)
+    """The configured update interval in seconds."""
```
vgro, line 89: I like this very much!

virtual_ecosystem/models/plants/constants.py
```diff
@@ -60,6 +61,30 @@ class PlantsConsts(ConstantsDataclass):
     root_turnover_c_p_ratio: float = 656.7
     """Carbon to Phosphorous ratio of root turnover."""
 
+    subcanopy_extinction_coef: float = 0.5
+    """The extinction coefficient of subcanopy vegation (unitless)."""
```
vgro:
```suggestion
    """The extinction coefficient of subcanopy vegetation (unitless)."""
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -464,35 +474,63 @@ def set_shortwave_absorption(self, time_index: int) -> None:
...
         """Estimate the gross primary productivity within plant cohorts.
 
-        This method uses the P Model to estimate the light use efficiency of leaves in
-        gC mol-1, given the environment (temperate, atmospheric pressure, vapour
-        pressure deficit and atmospheric CO2 concentration) within each canopy layer.
-        This is multiplied by the absorbed irradiance within each canopy layer to
-        predict the gross primary productivity (GPP, µg C m-2 s-1) for each canopy
-        layer.
+        This method uses estimated light use efficiency from the P Model to estimate the
+        light use efficiency of leaves in gC mol-1, given the environment (temperate,
```
vgro:
```suggestion
        light use efficiency of leaves in gC mol-1, given the environment (temperature,
```

vgro (approved):
Looks sensible to me, the only part I am not sure about is the number of filled layers in the radiation related variables; I don't think there should be any absorption above the canopy?
(I will definitely have to make some big changes in the abiotic model with the introduction of this vegetation layer, but this is not your problem ;-) )

tests/models/plants/conftest.py
davidorme: The subcanopy LAI is using the surface layer - I've deliberately kept the subcanopy below the canopy layers. 

That's wildly artificial of course and is going to become even more wildly artificial when we start getting tree recruitment. I've no idea right now what the initial size of those juvenile trees will be, but pretty tiny. I'd be astonished if they aren't in the surface layer.

virtual_ecosystem/core/core_components.py
davidorme: Seemed like a useful thing that all models might need! Might advertise it once this merges!

tests/models/plants/conftest.py
vgro: Is the below_canopy index now part of index filled canopy?
I think it was the layer_heights_canopy variable that confused me in the first place, why do you have 4 canopy layer heights ? Also the shortwave absorption seems to have too many entries, 3 canopy layers + subcanopy + topsoil =5 not 6? Or is there some vegetation above the first canopy layer?

davidorme:
>  I don't think there should be any absorption above the canopy?

@vgro There shouldn't be. The subcanopy layer is in the surface layer, so adds LAI below the canopy. That might not make sense or be workable. The options are: 

* Goes in surface (which is really ground-effects, right) - what we have now.
* A new specific subcanopy vegetation layer (which we got rid of earlier, right?)
* Integrate it into the canopy layers. The lowest layer in the canopy would then get an additional chunk of LAI and continues to represent all the vegetation between the last closed layer (or the canopy top if the canopy doesn't close) and the surface layer. I don't much like smushing the canopy and subcanopy LAI together though, although with seedlings, they are in practice smushed together.
* We could use up a canopy layer below the canopy bottom, so the subcanopy vegetation appears in the canopy stack, but is kept separate.

Thoughts? In particular, do any of those options reduce the impact on the abiotic model? I think the current solution is probably the most damaging?

tests/models/plants/conftest.py
davidorme: Yup - these test values are tricky because I'm using them to test recalculation at different stages so layers drop in and out, but there _are_ four canopy layers in the first cell. So we have:

[idx=0] above_canopy: always empty
[1-4] canopy_1 to canopy_4: contain LAI + absorbance data in at least one cell
[6-10] canopy_6 to canopy_10: contain no data
[11] surface: contains LAI and absorbance
[12] ground: contains absorbance

TaranRallings (approved):
This LGTM! I think this is a very reasonable solution to our current problems. After I get my trophic expansion done, I'll do a new PR for the rewiring and we can take a few days to make sure things are connected properly.

davidorme:
> This LGTM! I think this is a very reasonable solution to our current problems. After I get my trophic expansion done, I'll do a new PR for the rewiring and we can take a few days to make sure things are connected properly.

@TaranRallings There are several things coming in that will need wiring. In this PR you've got:

* subcanopy leaf biomass 
* subcanopy seedbank/seed/wafting_around_in_the_aether.

In addition to that there is now a data layer containing tree propagules (which will be called `plant_pft_propagules`) and then there will be a bunch of tree seedlings. Those are going to be an issue, because technically they'll be part of the canopy. Just a very low and edible part of the canopy, so we're going to want to figure out how to expose small seedlings to ground herbivory, otherwise they'll only get exposed to canopy herbivory.

TaranRallings:
Sounds good to me! Once we're at the rewiring we can take a look at the whole nest of trophic connections and make sure we've got strong solutions for everything.

vgro:
> > I don't think there should be any absorption above the canopy?
> 
> @vgro There shouldn't be. The subcanopy layer is in the surface layer, so adds LAI below the canopy. That might not make sense or be workable. The options are:
> 
>     * Goes in surface (which is really ground-effects, right) - what we have now.
> 
>     * A new specific subcanopy vegetation layer (which we got rid of earlier, right?)
> 
>     * Integrate it into the canopy layers. The lowest layer in the canopy would then get an additional chunk of LAI and continues to represent all the vegetation between the last closed layer (or the canopy top if the canopy doesn't close) and the surface layer. I don't much like smushing the canopy and subcanopy LAI together though, although with seedlings, they are in practice smushed together.
> 
>     * We could use up a canopy layer below the canopy bottom, so the subcanopy vegetation appears in the canopy stack, but is kept separate.
> 
> 
> Thoughts? In particular, do any of those options reduce the impact on the abiotic model? I think the current solution is probably the most damaging?

I think it is ok for now in the surface layer as we are using the simple model and that will not be affected. Once I get back to the energy balance we can revisit. We will also have to talk about where trees/grass get their water from which might introduce more changes to the layer structure.

davidorme:
@arne-exe Can you see anything that worries you in this? We'd need some rough approximation of subcanopy biomass per m2 and - more problematically I suspect - some kind of mass for subcanopy seedbank.

arne-exe:
> @arne-exe Can you see anything that worries you in this? We'd need some rough approximation of subcanopy biomass per m2 and - more problematically I suspect - some kind of mass for subcanopy seedbank.

Hi @davidorme, could you briefly clarify what subcanopy vegetation refers to? If I understood correctly this would be all standing (living) biomass that is not part of the tree cohorts and is not part of tree recruitment (seedlings and saplings)? If this is the case, that would mean that this seed bank pool is separately from the tree cohort seed bank, and that we'd need validation for the relationship between subcanopy vegetation biomass and its seed bank biomass. 
At first sight, this seems very similar to Sally's approach for calculating reproductive tissue mass (derived from foliage mass through some relationship to be validated). What also remains to be validated is then what proportion of that reproductive tissue mass is given to seeds (i.e., reproductive tissue mass minus allocation to fruits and flowers) 
My guess is that you could use that same approach to get values for the subcanopy seedbank mass, this might be rough but it should be ok until we find data to refine those relationships.

I assumed that leaf area index was a PFT specific trait, which allows to correct foliage area once it's derived from foliage mass (using specific leaf area). So, it's a bit unclear why you calculate LAI here, but perhaps I'm missing a step here. I'll have a look again at the details of how the P model works to see if that clarifies it.

davidorme:
>Hi @davidorme, could you briefly clarify what subcanopy vegetation refers to?

It's basically grass and shrubs, simulated as a biomass without any kind of woody tissue. That's obviously unrealistic, but the main aim for the moment is to provide at least something for ground-dwelling herbivores to eat.

> If I understood correctly this would be all standing (living) biomass that is not part of the tree cohorts and is not part of tree recruitment (seedlings and saplings)? If this is the case, that would mean that this seed bank pool is separately from the tree cohort seed bank, and that we'd need validation for the relationship between subcanopy vegetation biomass and its seed bank biomass.

Yup. With tree propagules, we are explicitly modelling a number of propagules to give recruitments of cohorts with a number of individuals. Here though, we just have a generic biomass, so there is a lossy conversion of seedbank biomass into initial vegetation mass.

> At first sight, this seems very similar to Sally's approach for calculating reproductive tissue mass (derived from foliage mass through some relationship to be validated). What also remains to be validated is then what proportion of that reproductive tissue mass is given to seeds (i.e., reproductive tissue mass minus allocation to fruits and flowers) My guess is that you could use that same approach to get values for the subcanopy seedbank mass, this might be rough but it should be ok until we find data to refine those relationships.

Yup - with tree reproduction we are splitting into propagules vs non-propagule mass, but here we're just allocating to seedbank mass.

>I assumed that leaf area index was a PFT specific trait, which allows to correct foliage area once it's derived from foliage mass (using specific leaf area). So, it's a bit unclear why you calculate LAI here, but perhaps I'm missing a step here. I'll have a look again at the details of how the P model works to see if that clarifies it.

We need to calculate the light absorption through the subcanopy, which is a function of LAI ($f_a = e^{-kL}$), so we take the biomass of vegetation (kg m-2) and multiply through by SLA (m2 kg) to give the LAI in (m2 m-2) as a representation of leaf "depth" for absorption in the layer.

arne-exe:
> 
> Yup. With tree propagules, we are explicitly modelling a number of propagules to give recruitments of cohorts with a number of individuals. Here though, we just have a generic biomass, so there is a lossy conversion of seedbank biomass into initial vegetation mass.
> 

> 
> Yup - with tree reproduction we are splitting into propagules vs non-propagule mass, but here we're just allocating to seedbank mass.
> 

I see, so essentially for both subcanopy biomass and tree cohorts we want values that predict seed mass from foliage mass, and for tree cohorts we also need some average mass of an individual seed to calculate the amount of seeds. For tree cohorts I think we have the datasets to do that, and for subcanopy biomass I need to do some digging. I'll check this tomorrow!

> 
> We need to calculate the light absorption through the subcanopy, which is a function of LAI ( f a = e − k L ), so we take the biomass of vegetation (kg m-2) and multiply through by SLA (m2 kg) to give the LAI in (m2 m-2) as a representation of leaf "depth" for absorption in the layer.

Thanks, that clarifies it!

Issue #750, opened 2025-02-18
Allocation to fruit production


davidorme:
The `pyrealm` package doesn't currently include reproductive tissues in its allocation model. It should do, but isn't there yet. For the moment, we need to handle allocating carbon to reproductive tissue production within VE. We need to define:

- The amount of reproductive tissue on a given stem. We'll start with having a simple fixed fraction of the leaf mass - that then scales with crown area. 
- The turnover of that tissue - a proportion of the tissue that is actively lost to the ground. 
- Propagule mass - we can then actually calculate a number of fruit. 

I don't know if @TaranRallings cares about individual fruits - and he'd also like more than just one gigantic pot of "reproductive tissues", but that at least starts to populate a seedbank.

At the start of the simulation, the existing stems are assumed to have fruit but then at GPP allocation, we need to allocate new fruit costs as the canopy grows and turnover costs. 

That's awkward:

- We can topslice the GPP to allocate to turnover in the known standing mass, but...
- until we calculate DeltaDBH, we don't know the change in leaf mass and hence the GPP required for new fruit.
- I think the easiest solution for now is to impose a delay - new canopy mass is fruitless and then topslice the _next_ GPP with a fruit growth cost from the last iteration. That's hacky but has some biological plausibility. I think the alternative would be that we'd have to iterate the allocation process somehow to solve "deltaDBH that gives leaves enough GPP spare to grow the fruit added to the canopy", which seems worse given we'll have to do this properly at some point!

> [!NOTE]
> - These will be global constant standing fruit mass. No differences in propagule strategy or seasonality etc. Gotta start somewhere

TaranRallings:
Sounds like a reasonable start to me. I don't think I need individual fruits, we'll just need to navigate some relationship between fruit mass:seedmass.

PR #822, opened 2025-04-07
Dividing propagules by PFT and other issues

davidorme:
# Description

These are some changes I think we'll need in the fruit production.

* I've tried to tackle the issue of getting fallen propagules per PFT  and per cell. That involves a loop over cohorts per cell.
* I've applied the same solution to split up the propagules and non propagule mass in the canopy by PFT. Ultimately, I think we need to rethink this bit - at the moment, reproductive mass is kinda plastic until it hits the floor and I think we might need to allocate reproductive tissue to the two classes at a cohort level.

* I spotted a problem with the recalculation of the turnover rates - we were calculating the actual allocation using the annual rate and then only passing a timestep fraction of the productivity into the turnover pools. That meant that - for example with monthly timesteps - each month allocated a years worth of turnover but 1/12 was added to the leaf turnover and the remaining 11/12 of the leaf turnover disappeared. I've hacked the rates when the Flora is added to the model, which isn't elegant but I think will do for now.

* There was also an issue that per stem allocations weren't being scaled up to the number of individuals in a cohort.

Contributes to fixing #750

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/plants/plants_model.py
+++ b/virtual_ecosystem/models/plants/plants_model.py
@@ -732,8 +732,8 @@ class PlantsModel(
                 self.data["canopy_n_propagules"].loc[cell_id, cohort_pft] += (
                     stem_n_propagules * cohort_n_stems
                 )
-                self.data["canopy_n_propagules"].loc[cell_id, cohort_pft] += (
-                    stem_n_propagules * cohort_n_stems
+                self.data["canopy_non_propagule_c_mass"].loc[cell_id, cohort_pft] += (
+                    stem_non_propagules * cohort_n_stems
                 )
 
             # Allocate the topsliced GPP to root exudates with remainder as active
--- a/virtual_ecosystem/models/plants/plants_model.py
+++ b/virtual_ecosystem/models/plants/plants_model.py
@@ -680,34 +680,18 @@ class PlantsModel(
                 stem_allocation.fine_root_turnover * cohorts.n_individuals
             )
 
-            # Reproductive tissue turnover: partitioning into propagules and
-            # non-propagules
-            # TODO: dimension issue in pyrealm, returns 2D array.
+            # Partition reproductive tissue into propagule and non-propagule masses and
+            # convert the propagule mass to number of propagules
+            # 1. Turnover reproductive tissue mass leaving the canopy to the ground
             stem_fallen_n_propagules, stem_fallen_non_propagule_c_mass = (
                 self.partition_reproductive_tissue(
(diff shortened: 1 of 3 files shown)


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -627,53 +661,92 @@ def allocate_gpp(self) -> None:
...
+            cohorts.dbh_values = cohorts.dbh_values + stem_allocation.delta_dbh
 
             # Sum of turnover from all cohorts in a grid cell
-            # TODO: Pyrealm provides annual turnover values. Divide by the number of
-            # updates_per_year to get monthly turnover values is naive and will
-            # overestimate turnover. This should be updated eventually to a more
-            # sophisticated approach.
             self.data["leaf_turnover"][cell_id] = np.sum(
-                cohort_allocation.foliage_turnover / self.model_timing.updates_per_year
+                stem_allocation.foliage_turnover * cohorts.n_individuals
```
sallymatson, line 677: Does this mean that turnover values are per stem? I never realized that.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -627,53 +661,92 @@ def allocate_gpp(self) -> None:
...
+                stem_canopy_n_propagules.squeeze(),
+                stem_canopy_non_propagule_c_mass.squeeze(),
+                cohorts.n_individuals,
+            ):
+                self.data["canopy_n_propagules"].loc[cell_id, cohort_pft] += (
+                    stem_n_propagules * cohort_n_stems
+                )
+                self.data["canopy_non_propagule_c_mass"].loc[cell_id, cohort_pft] += (
+                    stem_non_propagules * cohort_n_stems
+                )
```
sallymatson: Can this for loop be combined with the one above, so we only do the loop once?

sallymatson (commented):
One optimisation question. Otherwise LGTM

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yup - the allocation model the GPP for a representative stem in the cohort and gives the resulting stem allocation.

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yup - not sure how much overhead it saves, but it it's definitely logically cleaner.

sallymatson (approved):

PR #825, opened 2025-04-09
StemStochiometry basic structure.

### LINKED ISSUE #749: Initial stochiometry
> [!CAUTION]
> Take no action yet

With #748 complete, we have a definition of the standing N and P contents of each tree. For example, for Nitrogen

Stem N = stem mass / stem CN ratio
Root N = root mass / root CN ratio
Leaf N = leaf mass / root CN ratio
plant reproductive tissue (PRT) N =  PRT mass / PRT  CN ratio

We can then calculate losses through turnover and the amount on N and P required to support new growth. Unfortunately, we have the same problem of timing as in #750:

- We need to run the allocation model to calculate turnover. We can then use that to calculate how much N and P is lost through turnover (and also discount that by how much we can save through leaf senescence).
- But we also need to allocate carbon costs for the acquisition of that N and P. And that needs to be factored into the amount of carbon available to drive deltaDBH.

There are more complications - there are several mechanisms for CN acquisition. 

- N and P acquired simply through water uptake and transpiration. This isn't necessarily cost free - there is a carbon input into the soil as exudates that supports this.
- Direct mechanisms - ecto- and endo-mycorrhizae and plant associated nitrifying bacteria.

We can simply add a proportional allocation of GPP to exudates and top slice GPP, and transpiration then gives some "passive" N and P. But then we have an "active" N and P acquisition cost - we can lump the different mechanisms within plants for the moment and the soil model allocate that "active" carbon to the different mechanisms. But we have the problem of needing to know the cost of how much extra N and P to seek in order to grow and factoring that cost into calculating deltaDBH.

> [!CAUTION]
> - We don't have a firm plan here yet
> - We're also ignoring N and P limitation (see the comments on #738 about negative N and P concentrations!).
> - We don't have variable CN and CP ratios in response to availability.

sallymatson:
# Description

This PR adds a new module, `Stochiometry`, to track plant NP values. The class stricture includes a baseclass `Foliage` which has subclasses for each tissue type. The `Foliage` models store the element quantity, ideal ratios, and perform simple calculations using these values. The set of tissues for a cohort is stored in the `StemStochiometry` class. This class has further fuctionality which can balance element surplus/deficit, and perform calculations across the whole plant foliage groups. 

The next steps for stochiometry include expanding the initialization so that each PFT can specify the initial ratios (#899), dealing with mortality (#800), and creating more robust tests to ensure all cases are covered (#801). 

Fixes #749 

## Type of change

## Key checklist

## Further checks


davidorme (commented):
Yup - that looks good!

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -733,24 +756,140 @@ def allocate_gpp(self) -> None:
...
+            )
+            stochiometry.n_foliage = (
+                stochiometry.n_foliage
+                + stochiometry.n_for_foliage_growth(stem_allocation.delta_foliage_mass)
+            )
+            stochiometry.n_wood = stochiometry.n_wood + n_for_stem_growth
+            stochiometry.n_reproductive_tissue = (
+                stochiometry.n_reproductive_tissue + n_for_rt_growth
+            )
+            stochiometry.n_roots = stochiometry.n_roots + n_for_roots_growth
```
sallymatson, line 852: Should this be happening here?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -733,24 +756,140 @@ def allocate_gpp(self) -> None:
...
+
+            # Balance the N surplus/deficit with the symbiote carbon supply
+            n_weighted_avg = np.dot(
+                self.data["ecto_supply_limit_n"][cell_id]
+                + self.data["arbuscular_supply_limit_n"][cell_id],
+                cohorts.n_individuals,
```
sallymatson: Should this be `n_individuals * total tissue mass` or `crown_area`?

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,347 @@
...
+
+        Args:
+            cohort: The cohort to reconcile surplus.
+        """
+        if self.n_surplus[cohort] > self.n_tissue_deficit[cohort]:
+            # If there is sufficient surplus N to cover the existing deficit, the
+            # amount of the deficit is subtracted from the surplus which persists until
+            # the next update. All tissue types are updated to the ideal N ratios.
+            self.n_surplus[cohort] = (
+                self.n_surplus[cohort] - self.n_tissue_deficit[cohort]
```
sallymatson: Handle with `.max()` and the subtract to get remaining surplus

tests/models/plants/test_stochiometry.py
davidorme, line 1: This looks good - it's nicely structured and seems really coherent with the rest of the existing test suite.

The only thing I'd add are more comments in the test code to walk through what is happening. These really don't have to be extensive but even `# now create the XYZ object to test ABC` style comments make it much easier to pick tests up if they need updating.

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
+"""The :mod:`~virtual_ecosystem.models.plants.stochiometry` module contains the class
+for managing plant cohort stochiometry ratios. The carbon mass is stored in plant
+alloemetry or allocation, so this class uses those as the anchor weights and stores
```
davidorme:
```suggestion
allometry or allocation, so this class uses those as the anchor weights and stores
```

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+CN and CP ratios.
+
+The class holds current CN and CP ratios for foliage, reproductive tissue, wood, and
+roots on the cohort level. Each tissue also has an idea CN and CP ratio, which is used
```
davidorme:
```suggestion
roots on the cohort level. Each tissue also has an ideal CN and CP ratio, which is used
```

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+as a comparison in the case of any nutrient deficit. Senesced leaves also have fixed CN
+and CP ratios, which is used for leaf turnover.
```
davidorme:
```suggestion
and CP ratios, which are used for leaf turnover.
```

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+    specific plant tissue type (generally foliage, wood, roots or reproductive tissue).
+    The class also holds the ideal ratio of the element for that tissue type. They hold
+    an entry for each cohort in the data class.
+    """
+
+    tissue_name: str
+    """The name of the tissue type."""
+    community: Community
+    """The community object that the tissue is associated with."""
+    # Should this be stored only in stochiometry and not in tissue?
```
davidorme: It feels like we're storing a lot of local references to things, but I think that is something we revisit later. The `pyrealm` demography structures may change, so it seems premature to try and muck around with too much here.

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+
+import numpy as np
+from numpy.typing import NDArray
+from pyrealm.demography.community import Community
+from pyrealm.demography.core import CohortMethods, PandasExporter
+from pyrealm.demography.tmodel import StemAllocation
+
+
+@dataclass
+class Tissue(PandasExporter, CohortMethods):
```
davidorme: A couple of things here:

* This is intended to be used as an abstract base class - I think we can explicitly make it an ABC and that has the advantage of having some built in mechanisms to enforce subclass method definitions. The definition of `core.base_model.BaseModel` uses a lot of this machinery. For example, there we:
    * explicitly decorate `@abstractmethod`s and that automatically implements warnings about needing to be defined in subclasses. Should be more compact and more explicit.
    * implement a `__subclass_init__` method on the ABC that allows the subclass class attributes to be defined as arguments to the subclass (I think we can then drop the subclass `__init__` methods as these only call the `super` and that would be done automatically.). 
     
    @dalonsoa Could you have a quick look at this specific bit (Tissue and then the explicit subclasses below) and see if you can spot a problem with making this switch.
    
* Perhaps quickly test that these mixins work (now that I know that name!).

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+        return (allocation.fine_root_turnover * (1 / self.Cx_ratio)).squeeze()
+
+
+@dataclass
+class StemStochiometry(CohortMethods, PandasExporter):
+    """A class holding the ratios of Carbon to Nitrogen and Phosphorous for stems.
+
+    This class holds the current ratios across tissue type for a community object, which
+    in essence is a series of cohorts. It acts in parallel with StemAllometry, a class
+    attribute of Community.
```
davidorme: Maybe remove explicit references to N and P and say it is a class holding element ratios to carbon mass. There's a little bit of this throughout this file, I think, so a quick sweep might be useful to emphasize that although we are currently doing "only" N and P, this is extendable

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -312,6 +323,116 @@ def _setup(
         self.communities = PlantCommunities(
             data=self.data, flora=self.flora, grid=self.grid
         )
+
+        # Initialize the stochiometries of each cohort
+        self.stochiometries = {
+            cell_id: {
```
davidorme: Could you add a few comments to sketch out what values are being passed into these constructors. The variable naming is all clear enough but just a quick overview of the biology alongside the code would be useful.

I think this is perhaps a bit of a big chunk for a dictionary comprehension? I don't think there are performance implications, more readability.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -790,11 +908,70 @@ def allocate_gpp(self) -> None:
...
+            # Balance the N & P surplus/deficit with the symbiote carbon supply
+            n_weighted_avg = np.dot(
+                self.data["ecto_supply_limit_n"][cell_id]
+                + self.data["arbuscular_supply_limit_n"][cell_id],
+                cohorts.n_individuals,
+            )
+            n_avaiable_per_cohort = n_weighted_avg / sum(n_weighted_avg)
+            n_available_per_stem = np.divide(
+                n_avaiable_per_cohort, cohorts.n_individuals
+            )
```
davidorme:
```suggestion
            n_available_per_cohort = n_weighted_avg / sum(n_weighted_avg)
            n_available_per_stem = np.divide(
                n_available_per_cohort, cohorts.n_individuals
            )
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -790,11 +908,70 @@ def allocate_gpp(self) -> None:
...
+                cohorts.n_individuals,
+            )
+            p_available_per_cohort = p_weighted_avg / sum(p_weighted_avg)
+            p_available_per_stem = np.divide(
+                p_available_per_cohort, cohorts.n_individuals
+            )
+
+            stochiometries["P"].element_surplus = (
+                stochiometries["P"].element_surplus + p_available_per_stem
+            )
```
davidorme: Think we can loop this process?
```suggestion
            # Balance the N & P surplus/deficit with the symbiote carbon supply
            for element in ["N", "P"]:
                element_weighted_avg = np.dot(
                    self.data["ecto_supply_limit_" + element.lower()][cell_id]
                    + self.data["arbuscular_supply_limit_" + element.lower()][cell_id],
                    cohorts.n_individuals,
                )
                element_available_per_cohort = (
                    element_weighted_avg / sum(element_weighted_avg)
                )
                element_available_per_stem = np.divide(
                    element_available_per_cohort, cohorts.n_individuals
                )
                stochiometries[element].element_surplus += element_available_per_stem
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -906,6 +1098,29 @@ def calculate_nutrient_uptake(self) -> None:
...
+        # Calculate N/P uptake (g N/P per stem) due to transpiration. Multiply:
+        # - Per stem transpiration (µmol H2O per stem)
+        # - Conversion factor from µmol H2O to m^3 (1.08015*10^-11)
+        # - Concentration of N/P uptake (kg m^-3)
+        # - Kg to g (1000)
+
+        for cell_id in self.communities.keys():
+            self.stochiometries[cell_id]["N"].element_surplus += (
+                self.per_stem_transpiration[cell_id]
+                * (1.8015 * pow(10.0, -11))
```
davidorme: Here and below:

```suggestion
                * 1.8015e-11
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -906,6 +1098,29 @@ def calculate_nutrient_uptake(self) -> None:
         self.data["plant_nitrate_uptake"] = self.data["dissolved_nitrate"] * 0.01
         self.data["plant_phosphorus_uptake"] = self.data["dissolved_phosphorus"] * 0.01
 
+        # Calculate N/P uptake (g N/P per stem) due to transpiration. Multiply:
+        # - Per stem transpiration (µmol H2O per stem)
+        # - Conversion factor from µmol H2O to m^3 (1.08015*10^-11)
```
davidorme, line 1180: Technically speaking this varies with temperature and atmospheric pressure. Pyrealm supports the scaling, but I'm not sure we fix this in this PR. We do use this fine tuning elsewhere in the model but it is a pretty small scale effect.

Maybe raise a new issue?

davidorme (changes requested):
@sallymatson Looks really good - very clean coding. Changes are mostly either "more comments" or minor structural changes.

The bigger one is the setup of the Tissues abstract class, which I think we can finesse a bit. I've called in @dalonsoa to just check this isn't an "everything looks like a nail" issue.

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+            The carbon to element ratio for the specified tissue.
+        """
+        return self.carbon_mass / self.actual_element_mass
+
+
+class FoliageTissue(Tissue):
+    """A class to hold foliage stochiometry data for a set of plant cohorts."""
+
+    # reclaim_ratio: NDArray[np.float64]
+    """The ratio of the element that can be r eclaimed from the sensced tissue."""
```
arne-exe:
```suggestion
    """The ratio of the element that can be reclaimed from the senesced tissue."""
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -312,6 +323,116 @@ def _setup(
...
+                        ),
+                        RootTissue(
+                            community=self.communities[cell_id],
+                            ideal_ratio=np.full(
+                                self.communities[cell_id].n_cohorts,
+                                model_constants.root_turnover_c_n_ratio,
+                            ),
+                            actual_element_mass=model_constants.root_turnover_c_n_ratio
+                            * self.communities[cell_id].stem_traits.zeta
+                            * self.communities[cell_id].stem_allometry.foliage_mass,
```
arne-exe: Since zeta is the ratio of fine root mass to foliage **area**, is it possible that these lines still need to include SLA to convert foliage mass to area (i.e., root tissue mass = zeta * foliage mass * SLA)?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -312,6 +323,116 @@ def _setup(
...
+                        ),
+                        RootTissue(
+                            community=self.communities[cell_id],
+                            ideal_ratio=np.full(
+                                self.communities[cell_id].n_cohorts,
+                                model_constants.root_turnover_c_p_ratio,
+                            ),
+                            actual_element_mass=model_constants.root_turnover_c_p_ratio
+                            * self.communities[cell_id].stem_traits.zeta
+                            * self.communities[cell_id].stem_allometry.foliage_mass,
```
arne-exe: Same question here.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -790,11 +908,70 @@ def allocate_gpp(self) -> None:
                 )
             )
 
+            # FOURTH, ALLOCATE TO GROWTH:
+            # Grow the plants by increasing the stem dbh
+            # TODO: dimension mismatch (1d vs 2d array) - check in pyrealm
+            cohorts.dbh_values = cohorts.dbh_values + stem_allocation.delta_dbh
+
+            # Subtract the N/P required from growth from the element store, and
+            # redistribute it to the individual tisuses.
```
arne-exe:
```suggestion
            # redistribute it to the individual tissues.
```

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+    @property
+    def carbon_mass(self) -> NDArray[np.float64]:
+        """Get the carbon mass for root tissue.
+
+        Returns:
+            The carbon mass for root tissue.
+        """
+        return (
+            self.community.stem_allometry.foliage_mass * self.community.stem_traits.zeta
+        ).squeeze()
```
arne-exe: Likely needs to be multiplied by SLA too (see other comment on zeta).

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -0,0 +1,421 @@
...
+        """Calculate the nitrogen needed for growth for root tissue.
+
+        Returns:
+            The nitrogen needed for growth for root tissue.
+        """
+        return (
+            allocation.delta_foliage_mass
+            * (1 / self.ideal_ratio)
+            * self.community.stem_traits.zeta
+        ).squeeze()
```
arne-exe: Think you want the change in foliage area here, so also multiply by SLA?

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Yes, that seems logical! Good catch thank you!

virtual_ecosystem/models/plants/stochiometry.py
sallymatson: You're right, this is much nicer. Thanks David. 

I also realized that as long as I decorate the subclasses with `@dataclass` as well then I don't need an `__init__` at all... I had only added it before when I was getting issues that `FoliageTissue` had an extra attribute. But that was because I didn't decorate it with `@dataclass`. So overall, a nice improvement to the code quality :)

virtual_ecosystem/models/plants/stochiometry.py
sallymatson: Thanks for catching these Arne!!

davidorme (approved):
LGTM - basic structure of the class is now very clean looking and all the stretch goals are clearly noted for later PRs.

davidorme:
We should also add the stochiometric data into the CommunityDataExporter...

PR #827, opened 2025-04-11
Recruitment of new tree cohorts

### LINKED ISSUE #824: Recruitment of new tree cohorts
With the addition of #812, we now have per-PFT propagules in the `PlantsModel` , so we need to germinate those and add new tree cohorts.

* Add initial values for the PFT propagule pool to the required data in the model.
* Add a germination rate to convert propagules into new tree cohorts. This should be per PFT, but it can probably start life as a PlantsModel constant like the mortality rate.
* Add those cohorts to the plant communities.
* Worry about the separation of the subcanopy vegetation and the seedlings. The light environment of these two different groups matters a lot to seedling success. If we go with an infinitely thin algal mat model for the subcanopy vegetation, then all seedlings are instantly above the subcanopy. That's easy and wildly inaccurate.
* Whilst we're in this cycle, check to see if empty cohorts are removed following mortality. I don't think they are at present.

davidorme:
# Description

Since #812, we have had allocation of carbon to reproductive propagules in the plant model and numbers of propagules in the model.

This PR is to provide an initial draft of recruitment of new cohorts from those propagules:

- [x] Adds a germination rate to convert propagules into new tree cohorts. This is currently global but should be per PFT at some point, like the mortality rate.
- [x] Implements an `apply_recruitment` method that uses the germination rate to convert propagules into seedlings and adds those seedlings as new cohorts to the Community object in the cell.
- [x] Add initial values for the PFT propagule pool to the required data in the model.
- [x] Updates testing to pass with recruitment and mortality.

Things to worry about later - probably convert each of these to its own issue

- [x] The initial stem diameter is entirely arbitrary - we need to find some kind of initial DBH and stochiometry from the carbon and nutrients available in the propagule. We could set that as a trait, but it would be better to work it out from first principles using the T Model.
    
    [Converted to issue #1024]
- [x] The separation of the subcanopy vegetation and the seedlings. The light environment of these two different groups matters a lot to seedling success. If we go with an infinitely thin algal mat model for the subcanopy vegetation, then all seedlings are instantly above the subcanopy. That's easy to code but wildly inaccurate.

    [This is what we are doing for now - @arne-exe and @davidorme have started discussing a shrub layer which would make this complex, but that's way down the line]

- [ ] Check to see if empty cohorts are removed following mortality of all individuals. I don't think they are at present.
- [ ] I'm still seeing bizarre errors in growth patterns. Something is up with the GPP units, I think.  

Fixes #824 

## Type of change

## Key checklist

## Further checks


davidorme:
@sallymatson This is a bit of a monster now (although there have been worse!). I've reviewed all of the stochiometry stuff though, and tests pass. Could you review?

davidorme:
The initial stem diameter is entirely arbitrary - we need to find some kind of initial DBH and stochiometry from the carbon and nutrients available in the propagule. We could set that as a trait, but it would be better to work it out from first principles using the T Model.

sallymatson (approved):
LGTM. I had reviewed before I started the stoichiometry stuff and it all made sense to me then, and when I added as well.

PR #843, opened 2025-05-07
823 dont allow rainfall to be negative

### LINKED ISSUE #823: Don't allow rainfall to be negative
At the moment, it is possible to get negative rainfall at the surface, which we do not want. This has likely to do with the implementation of the interception and require some adjustment to account for the daily loop.

vgro:
This PR fixes negative rainfall and adds error logging if it should happen again. I had calculated the evaporation and leaf drainage from each canopy layer with the condition that this should not exceeds the interception pool. In the next step, I summed over all layers and subtracted that from the rainfall, I believe this was the problem. Now I sum first to make sure the sum is smaller than the available water.

Fixes #823

## Type of change

## Key checklist

## Further checks


tests/models/hydrology/test_hydrology_model.py
```diff
@@ -323,12 +323,19 @@ def test_setup(
                     atol=1e-4,
                 )
 
+            exp_canopy_evap = model.layer_structure.from_template()
+            exp_canopy_evap = 0
```
vgro, line 327: there should be something in place to get the nan's back in place

jacobcook1995 (approved):
LGTM! The function to automatically catch this type of error if it occurs again feels like a really good idea for helping with calibration

davidorme:
I think we should look again at implementing the bounds checking for this too - running the simulation in bounds checking mode would be really useful for calibration.

vgro:
> I think we should look again at implementing the bounds checking for this too - running the simulation in bounds checking mode would be really useful for calibration.

I agree, but future problem :-)

virtual_ecosystem/models/hydrology/hydrology_tools.py
```diff
@@ -260,3 +261,17 @@ def calculate_psychrometric_constant(
...
+    """Check that precipitation at the surface is not negative.
+
+    Args:
+        precipitation_surface: Precipitation at the surface
+
+    Returns:
+        error if precipitation is negative in any grid cell
+    """
+    if (precipitation_surface < 0.0).any():
+        LOGGER.error("Surface precipitation should not be negative!")
```
sallymatson: Question about this error message. Since (I believe) the precipitation is a calculated/derived value, how would a modeller reconcile this issue? Is it possible to point them to how to solve it (e.g. "Surface precipitation is negative. Consider checking XX input variable to reconcile.") 

Also, should this be a fatal error? It seems here that this error will write to the log but not stop the simulation. But what will that mean going forward for the next runs? 

Not requesting changes, just confirming that this is the best way to alert - I leave final judgement to you! And I guess if the simulation now prevents it, then maybe it's not necessary to change.

sallymatson (approved):

virtual_ecosystem/models/hydrology/hydrology_tools.py
davidorme: We've got a long-standing plan to implement something like this as broader bounds checking. This discussion is old (https://github.com/ImperialCollegeLondon/virtual_ecosystem/discussions/135) but I think we'd now add hard or soft bounds data to `variables.toml`. Those are going to be useful values for users anyway, and then if we run in `bounds` mode, all changes to a variable in `data` can be bounds checked. Hard bounds raise, soft bounds warn.

But - as @vgro says: a problem for another day.

sallymatson: Cool. Seems very valuable for calibration..... maybe something for me to work on once stochiometry is figured out.

vgro: Good point, I agree that the simulation should stop when this happens and some more info should be given to solve the problem.

PR #866, opened 2025-05-19
831 calculation of net radiation needs to find a place

### LINKED ISSUE #831: calculation of net radiation needs to find a place
At the moment the calculation of net radiation is done internally in the `abiotic` model and not returned to `data`. The `hydrology` model will need this variable (currently hard-coded hack). In order to make ve run with the `abiotic_simple` model, we need to find a sensible place for the calculation to be placed. This could be the `plant` model or the `hydrology` model?

vgro:
This PR includes the calculation of net radiation in the simple abiotic model to make sure that it is always available for the hydrology model (it is now a required variable).

Fixes #831 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -708,6 +708,9 @@ def dummy_climate_data(fixture_core_components):
     data["latent_heat_flux"] = from_template()
     data["latent_heat_flux"][flux_index] = 0.0
 
+    data["net_radiation"] = from_template()
+    data["net_radiation"][lyr_str.index_flux_layers] = 20.0
+
     data["molar_density_air"] = from_template()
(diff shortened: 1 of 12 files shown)


davidorme (approved):
LGTM - I'm not convinced that the `leaf_albedo` constant belongs in abiotic constants, but that's part of a bigger picture issue with the constants 😱

vgro:
> LGTM - I'm not convinced that the `leaf_albedo` constant belongs in abiotic constants, but that's part of a bigger picture issue with the constants 😱

Agreed. There are a few plant traits that I currently have as constants in the abiotic/hydrology model which I hope will be variables at some point. I also took a note to cross-check with the plant model at some point what is duplicated, we might want to move some constants around, but like you say, this brings us back to the old constants issue...

PR #867, opened 2025-05-19
Convert plant variables into units used by the soil and litter model

### LINKED ISSUE #863: Plant model doesn't convert to units used by the soil and litter model
When the plant model generates `root_carbohydrate_exudation` and `plant_symbiote_carbon_supply` it gets calculated with units of kg of carbon per grid square (default hectare) per timestep (defaults to month). The soil model then interprets this as having units of kg of carbon per m^2 per day, which causes a massive over estimate of carbon inputs.

The plant model also generates `deadwood_production`, `leaf_turnover`, `root_turnover` and `fallen_non_propagule_c_mass` with these units. In this case the litter model expects input in kg per m^2 per time step units (which again causes a big overestimate).

@sallymatson, I think this is code that you originally wrote. Are you happy for me to go and make the appropriate changes to it? And do you reckon it would make sense to write helper functions like `to_soil_units()` and `to_litter_units()` or just to make the changes in line?

jacobcook1995:
# Description

This changes the calculation of the variables that pass from the plant model to the soil and litter models so that they use the expected units (I probably should have noticed that they didn't when I reviewed this code before 🫠)

@sallymatson, this changes code that you originally wrote so let me know if you are happy with the changes. I also considered making helper functions for `to_litter_units` and `to_soil_units`. If you think those would be useful let me know and I can write them.

Fixes #863

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/plants/test_plants_model.py
+++ b/tests/models/plants/test_plants_model.py
@@ -478,3 +478,16 @@ def test_partition_reproductive_tissue(fxt_plants_model):
         n_propagules * fxt_plants_model.model_constants.carbon_mass_per_propagule
         + mass_non_propagules
     )
+
+
+def test_convert_to_litter_units(fxt_plants_model):
+    """Tests the helper function that converts to litter unit."""
+
+    input_mass = np.array([1e5, 3.4e2, 123.7, 0.007])
+    expected_input_density = [12.345679, 0.0419753, 0.0152716, 8.64198e-7]
+
+    actual_input_density = fxt_plants_model.convert_to_litter_units(
+        input_mass=input_mass
+    )
+
+    assert np.allclose(expected_input_density, actual_input_density)
--- a/tests/models/plants/test_plants_model.py
+++ b/tests/models/plants/test_plants_model.py
@@ -410,7 +410,10 @@ def test_PlantsModel_apply_mortality(fxt_plants_model):
             original_population[cell_id]
             - fxt_plants_model.communities[cell_id].cohorts.n_individuals
         )
-        deadwood_mass = np.sum(mortality * community.stem_allometry.stem_mass)
+        deadwood_mass = (
+            np.sum(mortality * community.stem_allometry.stem_mass)
+            / fxt_plants_model.grid.cell_area
+        )
 
         assert np.all(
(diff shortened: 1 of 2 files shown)


sallymatson:
Hi @jacobcook1995 I do feel like within the plants model keeping the units as total mass is helpuful. Especially with stochiometry, these are values that I will be using in other places in our side of the model. So if possible, I do think writing the helper functions would be better. Could we then do the conversions closer to the source or as part of the intake processing in the soil model? 

LMK if this doesn't make sense!!

davidorme:
I had a quick chat with @jacobcook1995 about this - I think the values going into `data` should have clear units (not just kg), so having helper functions that are only used at the point of input to `data` makes sense. All our model internal values can stay as mass, but we scale to area and time units when we pass the data out for use in other models? That sound ok?

sallymatson:
> I had a quick chat with @jacobcook1995 about this - I think the values going into `data` should have clear units (not just kg), so having helper functions that are only used at the point of input to `data` makes sense. All our model internal values can stay as mass, but we scale to area and time units when we pass the data out for use in other models? That sound ok?

Okay gotcha. I definitely defer to you on how best to handle data units. In that case though, can we still do the conversion somehow outside of `allocate_gpp`? I already find `allocate_gpp` is doing a lot, and adding another conversion directly there feels to me like it will add further confusion/bulk to the workflow.

jacobcook1995:
> Okay gotcha. I definitely defer to you on how best to handle data units. In that case though, can we still do the conversion somehow outside of `allocate_gpp`? I already find `allocate_gpp` is doing a lot, and adding another conversion directly there feels to me like it will add further confusion/bulk to the workflow.

@sallymatson  I am a bit cautious about adding temporary values into the data object, which then get modified later, as the variables in the data object are basically treated as a source of truth by the other models.

I guess we could do something like populate plant model attributes within `allocate_gpp`, and then add another function to actually add them to the data object?

sallymatson (approved):
Thank you, I like this as it is just a bit clearer about why we are doing that conversion in general and within `allocate_gpp`. And thanks for catching it in the first place!

PR #875, opened 2025-05-30
Animals - Trophic Expansion

### LINKED ISSUE #754: Animals - Trophic Expansion
Only simple-herbivory and predation are currently implemented. This needs to be expanded to include:
- omnivory
- detritivory
- complex herbivory (min: leaves, fruits, seeds)
- fungivory 

The current structure of diet and foraging can be expanded to include these categories.

TaranRallings:
# Description

This is the trophic expansion for the Animal Model. It's main job is to expand the types of diets that animals can have and provide the machinery for linking those diets to the right resources. 

Core Changes
- added new `DietTypes` in `AnimalTraits`
- made `get_eaten` methods for carcass and waste pools
- made carcass, detritus, and excrement accessible in the foraging loop
- added `delta_mass_` and `calculate_consumed_mass_` foraging methods for carcasses, detritus, and excrement
- expanded `prey_group_selection` scaling function

Supplemental
- refactored `LitterPool` to align it with other resources - one object per grid cell
-  added `create_new_cohort` method to `AnimalModel` as I was repeating the same functionality in three places

Fixes # (issue)

## Type of change

- [] Optimization (back-end change that speeds up the code)
- [] Bug fix (non-breaking change which fixes an issue)

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -1012,28 +1004,284 @@ def delta_mass_herbivory(
...
+        The search/handling formulation is identical to herbivory; only the resource
+          list and pool class differ.
+
+        Args:
+            litter_pools: All litter pools available in the territory.  Used
+                to compute the denominator of handling time.
+            target_pool: The specific pool from which biomass is sought.
+
+        Returns:
+            Wet mass (kg) of litter requested from ``target_pool`` during the
```
jacobcook1995: I don't think this is really a wet mass as the litter model only tracks the carbon, nitrogen and phosphorus contents

jacobcook1995 (commented):
The changes to the animal -> soil/litter links work for me. Just had one comment about litter wet mass

dalonsoa:
This is a pretty long PR, and I'm at a conference the whole of this week, so please bear with me and I will review it as soon as possible next week.

virtual_ecosystem/models/animal/animal_cohorts.py
TaranRallings: Good catch!

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1008,51 +998,77 @@ def birth_community(self) -> None:
...
+
+            # Live plant resources
+            if diet & DietType.HERBIVORE:
+                plant_list = cohort.get_plant_resources(self.plant_resources)
 
-            # Check the diet of the cohort and get appropriate resources
-            if consumer_cohort.functional_group.diet == DietType.HERBIVORE:
-                plant_list = consumer_cohort.get_plant_resources(self.plant_resources)
+            # Live prey
+            if diet & DietType.CARNIVORE:
```
sallymatson: Discussed briefly in our chat yesterday, but wanted to confirm how an omnivore diet is handled here, because I don't see it as an `if` statement.

sallymatson: To be more specific, want to confirm that if the `DietType` is omnivore, does it return true for both `DietType.HERBIVORE` and `DietType.CARNIVORE`?

virtual_ecosystem/models/animal/animal_traits.py
```diff
@@ -13,19 +13,111 @@ class MetabolicType(Enum):
     ECTOTHERMIC = "ectothermic"
 
 
-class DietType(Enum):
-    """Enumeration for diet types."""
+class DietType(Flag):
```
dalonsoa, line 16: Neat :)

virtual_ecosystem/models/animal/decay.py
```diff
@@ -75,6 +83,71 @@ def add_carcass(self, carbon: float, nitrogen: float, phosphorus: float) -> None
...
+            carbon=-taken_wet * frac_C,
+            nitrogen=-taken_wet * frac_N,
+            phosphorus=-taken_wet * frac_P,
+        )
+        # 2. Transfer handling losses to the decomposed compartment
+        self.decomposed_cnp.update(
+            carbon=missed_cnp["carbon"],
+            nitrogen=missed_cnp["nitrogen"],
+            phosphorus=missed_cnp["phosphorus"],
+        )
```
dalonsoa: Just to confirm this is correct, it replaces the existing `decomposed_snp`, not added to it, right?

virtual_ecosystem/models/animal/decay.py
```diff
@@ -154,6 +236,62 @@ def reset(self) -> None:
...
+        self.scavengeable_cnp.update(
+            carbon=-taken_wet * frac_C,
+            nitrogen=-taken_wet * frac_N,
+            phosphorus=-taken_wet * frac_P,
+        )
+        # Handling losses become decomposed material
+        self.decomposed_cnp.update(
+            carbon=missed_wet * frac_C,
+            nitrogen=missed_wet * frac_N,
+            phosphorus=missed_wet * frac_P,
```
dalonsoa: And same here.

virtual_ecosystem/models/animal/decay.py
```diff
@@ -154,6 +236,62 @@ def reset(self) -> None:
         """
         self.decomposed_cnp = CNP(carbon=0.0, nitrogen=0.0, phosphorus=0.0)
 
+    def get_eaten(
```
dalonsoa: This function looks almost identical - haven't check exactly - to the other `get_eaten` above, so I would suggest to remove duplication. One way of doing it would be using [`mixins`](https://www.pythontutorial.net/python-oop/python-mixin/). It would look something like:

```python
class GetEatenMixin:
    def get_eaten(...):
        # The implementation of the method, which looks identical.

@dataclass
class CarcassPool(GetEatenMixin):
    ...

@dataclass
class ExcrementPool(GetEatenMixin):
    ...
```

I'm not entirely sure how mixins play with `dataclasses`, and you might need to get creative with `mypy` and use a `Protocol` for the mixin [as described in `mypy` docs](https://mypy.readthedocs.io/en/latest/more_types.html#mixin-classes), but it should increase the re-usability of the code.

dalonsoa (approved):
The code itself looks really good to me, and I only have a couple of questions and a suggestion to reduce the redundancy of the code.

virtual_ecosystem/models/animal/decay.py
davidorme: If I understand your suggestion here, I think mixins and `dataclasses` work fine.

I haven't called them "mixins" but I believe that the `PandasExporter` and `CohortMethods` classes in `pyrealm` here are mixins:

https://github.com/ImperialCollegeLondon/pyrealm/blob/develop/pyrealm/demography/core.py

And I use them (multiply!) to share functionality among classes e.g. here

https://github.com/ImperialCollegeLondon/pyrealm/blob/a8a133de9bc1bc69415b2d3ebae950a5120bad37/pyrealm/demography/flora.py#L475

dalonsoa: Yep, they look the same thing.

davidorme: Although it seems like I don't need to declare them as an ABC? Will have to look!

virtual_ecosystem/models/animal/animal_model.py
TaranRallings: I was being lazy using any broad diet categories in that method, that isn't really what those broad categories exist for haha. I have updated the logic to use the specific flags, which should circumvent any omnivory problems.

virtual_ecosystem/models/animal/decay.py
TaranRallings: The actual CNP values are stored in a `CNP` object which has that `update` method. The update method does add the value to the totals rather than replacing them.

virtual_ecosystem/models/animal/decay.py
TaranRallings: This works the same way as above.

virtual_ecosystem/models/animal/decay.py
TaranRallings: The intention is that `decomposed_cnp` is a pool of trophically inaccessible mass to which a portion of food that is attempted to eat falls. The pool is updated rather than replaced because this is a flow into the pool and jacob's soil model handles the flow out.

virtual_ecosystem/models/animal/decay.py
TaranRallings: Ah, good idea. I hadn't seen those mixins before. I have replaced the redundant method with a mixin and the required protocol.

Issue #880, opened 2025-06-03
soil moisture capacity needs to come out of core constants


vgro:
soil moisture capacity/saturation is currently in hydrology model and core, under different names. The hydrology model calls both, this needs to be unified.

jacobcook1995:
I need to fix #770 first

jacobcook1995:
#770 is now fixed, so this issue can now be addressed

PR #881, opened 2025-06-04
Add effective saturation calculation to hydrology tools

### LINKED ISSUE #770: function to calculate the effective saturation should be moved to hydrology tools
Effective soil saturation is currently calculated in two places, in the soil model, and in the calculation of vertical flow in the hydrology model. This is done with two separate functions. Instead this should become a hydrology tool, which then gets used in both places

jacobcook1995:
# Description

The effective saturation calculation has now been added as a function in hydrology tools, so that the hydrology model and the soil model use the same calculation.

The soil model now uses `soil_moisture_saturation` and `soil_moisture_saturation` from `HydroConsts` rather than `soil_moisture_capacity` from `CoreConsts` which should unblock #880

I also changed the core testing to use the litter model as a test case rather than the soil model. This has been annoying me for a while, and I reached breaking point so switching the testing to use the simpler model.

Fixes #770

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -497,90 +497,81 @@ def fixture_core_components(fixture_config):
 
 
 @pytest.fixture
-def dummy_carbon_data(fixture_core_components):
-    """Creates a dummy carbon data object for use in tests."""
+def dummy_litter_data(fixture_core_components):
+    """Creates a dummy litter data object for use in tests."""
 
     from virtual_ecosystem.core.data import Data
+
+    lyr_strct = fixture_core_components.layer_structure
 
     # Setup the data object with four cells.
     data = Data(fixture_core_components.grid)
 
-    # The required data is now added. This includes the five carbon pools: mineral
-    # associated organic matter, low molecular weight carbon, microbial biomass and
-    # necromass carbon and particulate organic matter. It also includes various factors
-    # of the physical environment: pH, bulk density, soil moisture, soil temperature,
-    # percentage clay in soil.
-    data_values = {
-        "soil_c_pool_lmwc": [0.05, 0.02, 0.1, 0.005],
-        "soil_c_pool_maom": [2.5, 1.7, 4.5, 0.5],
-        "soil_c_pool_bacteria": [5.8, 2.3, 11.3, 1.0],
-        "soil_c_pool_saprotrophic_fungi": [0.89, 8.55, 2.21, 4.54],
-        "soil_c_pool_arbuscular_mycorrhiza": [0.65, 1.47, 3.92, 9.04],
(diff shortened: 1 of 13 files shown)


tests/conftest.py
```diff
@@ -497,90 +497,81 @@ def fixture_core_components(fixture_config):
...
+    data["matric_potential"] = lyr_strct.from_template()
+    data["matric_potential"][lyr_strct.index_topsoil] = [-10.0, -25.0, -100.0, -100.0]
+    data["matric_potential"][lyr_strct.index_subsoil] = [-11.0, -29.5, -123.0, -154.1]
 
-    data["air_temperature"] = lyr_str.from_template()
-    data["air_temperature"][lyr_str.index_filled_atmosphere] = np.array(
+    data["air_temperature"] = lyr_strct.from_template()
+    data["air_temperature"][lyr_strct.index_filled_atmosphere] = np.array(
         [30.0, 29.844995, 28.87117, 27.206405, 16.145945]
     )[:, None]
```
vgro, line 576: If the 16.14 C affects your litter activities negatively feel free to change that, this value is too low and an artifact from an old implementation

virtual_ecosystem/models/soil/pools.py
```diff
@@ -419,9 +423,9 @@ def calculate_all_pool_updates(
                 pools are stored in the initial condition vector.
             layer_structure: The details of the layer structure used across the Virtual
                 Ecosystem.
-            soil_moisture_capacity: Soil moisture capacity, i.e. the maximum
-                (volumetric) moisture the soil can hold [unitless].
-            top_soil_layer_thickness: Thickness of the topsoil layer [mm].
+            soil_moisture_saturation: The :term:`soil moisture saturation` [unitless].
+            soil_moisture_residual: The :term:`soil moisture residual` [unitless].
+            top_soil_layer_thickness: Thickness of the topsoil layer [m].
```
vgro, line 428: nice use of glossary here :-)

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -234,20 +234,27 @@ def from_config(
             config, enzyme_classes=enzyme_classes
         )
 
+        # Load hydrology constants
+        hydro_constants = load_constants(config, "hydrology", "HydroConsts")
```
vgro, line 238: This looks like a neat solution to the different constants, I should adopt this to my models...

vgro (approved):
LGTM

tests/conftest.py
jacobcook1995: This stuff purely gets used for unit testing, so no problem with that!

PR #893, opened 2025-06-12
890 mycorrhizal fungi should be obligate rather than able to survive via saprotrophy

### LINKED ISSUE #890: Mycorrhizal fungi should be obligate, rather than able to survive via saprotrophy
At the moment, the two mycorrhizal functional groups can both use plant carbon to grow as well as taking up organic matter from the soil. This is a problem for two reasons

1) it doesn't match the biological reality (particularly for arbuscular mycorrhizal fungi, which lack the ability to even synthesise fatty acids for themselves)
2) it makes the model more complex than it needs to be, i.e. it would be much cleaner to have each fungal function be specific to a single functional group rather than having overlaps of function

I need to rewrite the soil model so that mycorrhizal fungi can take up organic matter, but only so that they can use the nutrients from it. This should also present an opportunity to simplify the code, which is currently pretty convoluted in this area

jacobcook1995:
# Description

This PR changes the function that calculate mycorrhizal carbon acquisition and nutrient uptake so that only plant supplied carbon is used for growth. Nitrogen and phosphorus can still be obtained from soil organic matter, but mycorrhiza will now try and obtain inorganic nutrients first.

I also altered the test parameterisation of the arbuscular mycorrhizal fungi, so that they don't produce enzymes. Down the line I need to considering preventing them from producing enzymes at all.

Fixes #890

## Type of change

## Key checklist

## Further checks


vgro (approved):
LGTM

PR #894, opened 2025-06-13
887 animal model   handling starvation during ontogeny

### LINKED ISSUE #887: Animal Model - handling starvation during ontogeny
**Describe the bug**
Currently `scaling_functions.starvation_mortality` provides mortality rate based on a comparison of the adult body mass of the functional group and the current bodymass of the cohort. This works well for adult cohorts but fails for young cohorts that have not yet reached adult mass as a newborn is treated as a starving adult. 

Following Madingley, this should be updated with the following changes:

- a `largest_mass_achieved` attribute for animal cohorts,
- use this new mass attribute instead of adult mass to baseline the starvation method 
- a check and update method for cohorts to update this mass through ontogeny
- a new `animal_model` method to call `increase_age` and the new mass checking method within the update loop

 
 This is not a perfect solution but it will allow us to compare more cleanly against Madingley and should serve as a first pass to capture ontogeny and starvation together.

TaranRallings:
# Description

This short (for once) PR addresses the issue of cohorts being born into starvation as their adult body mass was used as the reference for whether or not they are starving. I have followed Madingley and implemented a `largest_mass_achieved` attribute that is updated over time until maturity. This provides a rough approximation of the development process. Once maturity is achieved, adult mass becomes the reference for starvation as normal.

This includes:
- `largest_mass_achieved` attribute
- `handle_ontogeny` method in `AnimalModel`
- `update_largest_mass` method in `AnimalCohort`
- two helper methods to make `update` easier to read: `update_community_bookkeeping` and `update_cohort_bookkeeping`

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -120,6 +122,14 @@ def reproductive_mass(self) -> float:
         """Dynamically calculate the current reproductive mass from CNP object."""
         return self.reproductive_mass_cnp.total
 
+    def update_largest_mass(self) -> None:
+        """Update the record of the largest body-mass achieved by this cohort."""
```
dalonsoa: It might be worth to add here an explanation of why this is needed, similar to what you include in the PR description.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1139,6 +1164,16 @@ def increase_age_community(self, dt: timedelta64) -> None:
         for cohort in self.active_cohorts.values():
             cohort.increase_age(dt)
 
+    def handle_ontogeny(self) -> None:
```
dalonsoa, line 1166: Just to confirm, this is not used in the code (aside from tests), yet, right?

dalonsoa (approved):
Looks good. Just a minor suggestion. 

Short PR are the best! :)

virtual_ecosystem/models/animal/animal_model.py
TaranRallings: Ha, oversight on my part. I was calling the cohort level methods with my bookkeeping method, instead of the community methods. Fixed it now, thanks!

Issue #781, opened 2025-03-13
Abiotic model - sensible way of iterating towards convergence


vgro:
To find an equilibrium between air and canopy temperature in the energy balance, the abiotic model has to iterate. At the moment, there is an iterator hardcoded to run the functions 10 times, this needs to be changed. It could be an input set to a specific number or alternatively set to a convergence threshold, for example fluxes change less than 1 W m-2 (following SCOPE model).

vgro:
different solution found

PR #897, opened 2025-06-16
848 abiotic model add latent heat flux to derivative of temperature update

### LINKED ISSUE #848: Abiotic model: Add latent heat flux to derivative of temperature update
The latent heat flux term is missing from the calculation of the derivative of the energy imbalance. This needs to be added in the function that updates temperatures and it also likely requires calculating the latent heat flux internally rather than taking the outputs from plants and hydrology model. I think it could be good to set the bounds, needs some experimenting.

vgro:
The PR started out as adding a missing term to the derivative of the energy balance to fix #848 , I also made some small changes to the structure to make it more readable and updated the documentation.

The docs are built [here](https://virtual-ecosystem.readthedocs.io/en/848-abiotic-model-add-latent-heat-flux-to-derivative-of-temperature-update/virtual_ecosystem/implementation/abiotic_implementation.html)

Fixes #848 #781

## Type of change

## Key checklist

## Further checks


docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
```diff
@@ -58,45 +58,158 @@ display_markdown(
...
+Let $T_i^t$ represent the temperature (°C or K) of the $i^\text{th}$ soil layer at time
+$t$. The soil column is discretized into $n$ layers, each of thickness $\Delta z$ (m),
+and time advances in steps of $\Delta t$ (s).
+
+**Top layer update** (surface boundary condition):
+
+The topmost layer ($i = 0$) is updated using the net ground heat flux $G$
+($\mathrm{W\,m^{-2}}$):
+
+$$T_0^{t+\Delta t} = T_0^t + (\frac{\Delta t}{(\rho c \Delta z)}) G$$
```
jacobcook1995:
```suggestion
$$T_0^{t+\Delta t} = T_0^t + \left(\frac{\Delta t}{\rho c \Delta z}\right) G$$
```

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
```diff
@@ -58,45 +58,158 @@ display_markdown(
...
+
+```{math}
+\begin{aligned}
+T_i^{t+\Delta t} =
+& T_i^t + (\frac{\Delta t}{\Delta z^2}) \alpha (T_{i+1}^t - 2T_i^t + T_{i-1}^t)
+\end{aligned}
+```
+
+This term approximates vertical conduction using the second spatial derivative of
+temperature.
```
jacobcook1995, line 241: Might be worth giving either a citation to or writing out the temperature function that is being differentiated here

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
```diff
@@ -113,109 +226,203 @@ Under steady-state, the balance equation for the leaves in each canopy layer is
 follows (after {cite:t}`maclean_microclimc_2021`):
 
 ```{math}
-    & R_{abs} - R_{em} - H - \lambda E \\
-    & = R_{abs} - \epsilon_{s} \sigma T_{L}^{4} - c_{P}g_{Ha}(T_{L} - T_{A})
+    & R_{abs} - R_{em} - H - Q_{LE} \\
```
jacobcook1995:
```suggestion
    & R_{\text{abs}} - R_{\text{em}} - H - Q_{\text{LE}} \\
```

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
```diff
@@ -113,109 +226,203 @@ Under steady-state, the balance equation for the leaves in each canopy layer is
...
-emissivity of the leaf, $\sigma$ the Stefan-Boltzmann constant, $T_{L}$ the
-absolute temperature of the leaf, $T_{A}$ the absolute temperature of the air
-surrounding the leaf, $\lambda$ the latent heat of vapourisation of water,
-$e_{L}$ the effective vapour pressure of the leaf, $e_{A}$ the vapour
-pressure of air and $p_{A}$ atmospheric pressure. $g_{Ha}$ is the heat
-conductance between leaf and atmosphere, $g_{v}$ represents the conductance
-for vapour loss from the leaves as a function of the stomatal conductance $g_{c}$.
+where:
+
+$R_{abs}$:
```
jacobcook1995: Same \text change applies here and a number of other places, it's basically a way to stop multiple letter subscripts from rendering in a funky way in LaTeX

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
```diff
@@ -113,109 +226,203 @@ Under steady-state, the balance equation for the leaves in each canopy layer is
...
+energy balance at the leaf surface, see previous section.
 
-The air temperature surrounding the leaf $T_{A}$ is assumed to be influenced
-by leaf temperature $T_{L}$, soil temperature $T_{0}$, and reference air
-temperature $T_{R}$ as follows:
+#### Newton Linearization
 
-$$g_{tR} c_{p} (T_{R} - T_{A}) + g_{t0} c_{p} (T_{0} - T_{A}) + g_{L} c_{p} (T_{L} - T_{A})
-= 0$$
+To iteratively solve for the leaf temperature that satisfies the energy balance $EB$, we
```
jacobcook1995: Would generally try and avoid multiple letter variable names as they quickly become hard to interpret once you use them within more complex mathematical notation, think it would be better to call this $E_B$ or $B$ or something similar

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
```diff
@@ -113,109 +226,203 @@ Under steady-state, the balance equation for the leaves in each canopy layer is
...
-gives
+The temperature derivative of the energy balance is calculated analytically as:
 
-$$e_{a} = a_{E} + b_{E} \Delta T_{L}$$
+```{math}
+\frac{\partial EB}{\partial T_L^{\text{old}}} =
+\frac{\rho_a c_p}{r_a} +
+\frac{\rho_a \Delta_v}{r_a + r_s} \lambda +
+4 \epsilon \sigma (T_L^{\text{old}} + 273.15)^3
+```
```
jacobcook1995, line 374: Is the energy balance defined anywhere?

virtual_ecosystem/models/abiotic/constants.py
```diff
@@ -176,3 +176,10 @@ class AbioticConsts(ConstantsDataclass):
 
     initial_flux_value: float = 0.001
     """Initial non-zero fill value for energy fluxes, [W m-2]."""
+
+    numerical_stability_factor: float = 0.1
+    """Factor for numerical stability in temperature update.
```
jacobcook1995: I think's probably worth giving this a more descriptive name, i.e. one focused on what it changes in the model, so like "energy_balance_update_step" or similar

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -433,96 +440,187 @@ def update_soil_temperature(
...
+    new_canopy_temperature = canopy_temperature + numerical_stability_factor * (
+        energy_balance_canopy / derivative_energy_balance_canopy
     )
 
-    # Newton-Raphson step
-    new_canopy_temperature = canopy_temperature - energy_balance_canopy / derivative
-
-    new_air_temperature = air_temperature + relaxation_factor * (
+    alpha = time_interval / (density_air * specific_heat_air * aerodynamic_resistance)
+    new_air_temperature = air_temperature + alpha * (
```
jacobcook1995: I don't understand this step (let's talk about it when we meet later)

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -325,23 +325,47 @@ def run_microclimate(
...
+        current_energy_balance = energy_balance.calculate_energy_balance_canopy(
+            absorbed_radiation_canopy=data["shortwave_absorption"][
+                layer_structure.index_filled_canopy
+            ].to_numpy(),
+            longwave_emission_canopy=longwave_emission_canopy,
+            sensible_heat_flux_canopy=sensible_heat_flux_canopy,
+            latent_heat_flux_canopy=latent_heat_flux_canopy,
+        )
+
+        derivative_energy_balance_canopy = (
```
jacobcook1995: I think there needs to be a loop here (again lets discuss this in detail later)

jacobcook1995 (commented):
I had a bunch of comments which I'm adding so we can discuss them when we meet

virtual_ecosystem/models/abiotic/constants.py
vgro: no longer relevant with scipy.optimise.newton

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -192,195 +196,203 @@ def run_microclimate(
...
-            air_temperature=mean_air_temperature,
-            atmospheric_pressure=atmospheric_pressure[0],  # all layers identical
-            specific_gas_constant_dry_air=core_constants.specific_gas_constant_dry_air,
-            celsius_to_kelvin=core_constants.zero_Celsius,
-        )
-        specific_heat_air = calc_specific_heat(
-            tc=mean_air_temperature,
-        )
+    # -------------------------------------------------------------------------
+    #  Calculate atmospheric background variables using mean air temperature
```
vgro:
```suggestion
    #  Calculate atmospheric background variables
```

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
vgro: Will need to look into this again, I noted this down in a separate issue #919

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
vgro: I was following here the paper that I cite, I understand your point

docs/source/virtual_ecosystem/implementation/abiotic_implementation.md
vgro: At the top, I added a reference to it

vgro:
Thanks @jacobcook1995 for your feedback! I made a few more changes:
- updated documentation to have the same variable names, including the suggested change from EB to dQ/dt. I hope I caugt everything, will have a separate go at this when the insides of the model are ok.
- simplified calculation of net radiation
- incorporated scipy.optimise.newton to solve canopy temperature. It is not converging so I currently write out the best estimate after maxiter. In the course of this deleted my own derivative function.
- changed how air temperature is updated

Variable values are not in the range of physically possible but there needs to be some calibration done. There are also still open TODOs regarding unit conversions and time intervals, will address this separately to not make this more complicated here...

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -433,100 +416,320 @@ def update_soil_temperature(
...
+    nrows, ncols = canopy_temperature_initial.shape
+    solved_temperature = np.empty_like(canopy_temperature_initial, dtype=np.float32)
+
+    if isinstance(density_water, float):
+        density_water_array = np.full(solved_temperature.shape, density_water)
+    else:
+        density_water_array = density_water
+
+    for i in range(nrows):
+        for j in range(ncols):
```
jacobcook1995, line 616: Are these loops so you can solve the problem for each canopy layer for each cell individually?

jacobcook1995 (approved):
LGTM! Just had one small query

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: yes, the scipy function only takes scalar values

jacobcook1995: oh didn't realise that (does make sense). Was going to suggest flagging this as a potential performance bottleneck that you could improve by using a vector rather than scalar solver but if the solver only takes scalars then that is moot really

vgro: I thinks it's really fast but I can add a flag, good point

davidorme: Late to this party, but I suspect that this might be a bottleneck - many definitions of a cell specific function. I think a gridded version of a secant solver (which is what is being used here) might be faster, but this is definitely something to circle back to rather than worry about now.

davidorme: Actually, having said that, I've got a similar gridded root finding in `pyrealm.splash` which has been intermittently bothering me, so I've looked back into this and it is possible to parallelise this. I think this is going to be more efficient because it can operate over all cells and layers simultaneously - and numpy is great at handling those mass calculations rather than iterating over cells and layers. I suspect I'd want @dalonsoa or @alexdewar  to weigh in on the implementation, but I think this is doable.

It makes more sense to use canned versions like `scipy.optimise.newton` - unless they impose restrictions like only scalar inputs.

```python

from typing import Callable

import numpy as np


def secant_method(
    f: Callable,
    x0: NDArray[np.floating],
    x1: NDArray[np.floating],
    tol: float,
    **kwargs,
) -> NDArray[np.floating]:
    """Secant root for array inputs"""

    # Starting difference from root
    unsolved = np.full_like(x0, True, dtype=np.bool_)
    n_iter = 0

    while np.any(unsolved):
        # Calculate fx0 and fx1
        fx0 = f(x0, **kwargs)
        fx1 = f(x1, **kwargs)

        # Update using secant method
        x2 = x1 - fx1 * (x1 - x0) / (fx1 - fx0)

        # Update unsolved values
        unsolved = np.abs(fx0) > tol
        x0, x1 = np.where(unsolved, x1, x0), np.where(unsolved, x2, x1)

        n_iter += 1
        print(f"{unsolved.sum()} unsolved at iteration {n_iter}")

    return x2

def func(x: NDArray[np.floating], b: NDArray[np.floating]):
    return x**2 - 10 + b

# Cell specific variation
b = np.array([[1, 2, 3, 4]])
# Starting points
x0 = np.full_like(b, 0)
x1 = np.full_like(b, 40)

# Solve
root = secant_method(func, x0=x0, x1=x1, tol=1e-8, b=b)

# Plot
x = np.broadcast_to(np.arange(0, 5, 0.01)[:, None], (500, 4))
plt.plot(x, func(x, b=b))
ax = plt.gca()
style = dict(linewidth=0.2, color="black")
ax.vlines(root, ymin=0, ymax=1, transform=ax.get_xaxis_transform(), **style)
ax.hlines(0, xmin=0, xmax=1, transform=ax.get_yaxis_transform(), **style)
plt.show()

```

Figure_1

PR #903, opened 2025-06-18
Updating ve_run verbosity

### LINKED ISSUE #902: Better control of ve_run screen output
The output to the console from `ve_run` is not well structured. The progress bar _always_ prints, the actual short progress summary has to be turned on. This is finicky and right now is a problem for getting clean outputs on #702

davidorme:
# Description

This PR updates the reporting from `ve_run` command to give more control and better defaults. The `ve_run` function now:

* Only writes progress reporting to screen if the log is being redirected to file (otherwise, users are get a ton of logging feedback that we don't want to disrupt).
* Now defaults to writing out a short but information progress report:

```text
Starting Virtual Ecosystem simulation.
Logging to: /private/var/folders/62/cz0ctby96lq7ggnytfbtd__40000gn/T/pytest-of-dorme/pytest-61/test_ve_run_progress_reporting0/log.log
* Loading configuration
* Built core model components
* Initial data loaded
* Models initialised: testing
* Starting simulation
100%|██████████| 24/24 [00:00<00:00, 19354.60it/s]
* Simulation completed
Virtual Ecosystem run complete.
```

This can be simplified using the `ve_run` progress argument to drop (in order): the progress bar, the starred stage input and lastly all of the report. This is exposed via the `ve_run_cli` endpoint using the `-q` flag, which can be repeated to drop those components in order. So @vgro, you can use `-q` in #702 to drop that repeated progress bar (or `-qq` if you just want:

```text
Starting Virtual Ecosystem simulation.
Logging to: /blah/blah/log.log
Virtual Ecosystem run complete.
```

The other changes are:

* There is a new model (`TestingModel` in `models/testing/testing_model.py`). This does nothing and needs no variables and is mostly a way to setup testing of the model running infrastructure (like this PR) with an incredibly lightweight config (one model, no data or variables). @dalonsoa We probably could hack something together to get this only in the test suite, but we do have quite strict requirements for loading models and where it looks at the moment and it just seems like a feature we will use. If and when we add structure to allow people to specify their own model directories (which we need to do) then maybe we can move this back into `tests`.

* There are new tests to check the new arguments work via `ve_run` and `ve_run_cli`.

Fixes #902 (issue)

## Type of change

## Key checklist

## Further checks


docs/source/using_the_ve/virtual_ecosystem_in_use.md
```diff
@@ -89,15 +89,16 @@ for file in example_files:
...
+can be made shorter or completely muted by using the `-q` argument: repeat the argument
+to remove more details (e.g. `-qq` or `-qqq`).
 
 ```{code-cell} ipython3
 %%bash
 ve_run /tmp/ve_example/config \
     --out /tmp/ve_example/out \
-    --logfile /tmp/ve_example/out/logfile.log \
-    --progress \
+    --logfile /tmp/ve_example/out/logfile.log
```
dalonsoa, line 101: This reminds me [something we did in another project](https://github.com/EnergySystemsModellingLab/MUSE_OS/blob/97da23fe3eecce7d03b0f93e4d0e052897d55ef0/src/muse/__init__.py#L42) - also done for jobs in the HPC - where we separate the logs in two files: in one we show the normal log information showing progress (INFO or below), while the other shows important information that a user must pay attention (WARNING and above). This way, warnings are not lost in the mess of other messages and are more easily spotted by users. Just a suggestion for the future, not something to change in this PR.

virtual_ecosystem/models/testing/testing_model.py
dalonsoa, line 1: I think this is totally fine here, and I don't know why it did not occurred to us before. When developing software to control lab equipment, in particular with GUIs, you do need to include in your code dummy devices, sort of digital twins of their physical counterparts, and not just for formal testing, but just to be able to try the tool and develop functionality. So having, formally, simple TestModels that let you run your tool on simpler scenarios without having to resort to full models I think makes total sense.

dalonsoa (approved):
I've a couple of comments, and answer yours about the TestModel, but this looks good to me.

docs/source/using_the_ve/virtual_ecosystem_in_use.md
davidorme: Agreed 💯 . In the unlikely use case that a run is logging to the console, you would see the whole thing, but this mini report does not currently show anything if it breaks. You'd have to go the log.

virtual_ecosystem/models/testing/testing_model.py
davidorme: I thought about trying to test the new stuff using the existing setup (i.e run a 24 month simulation with 6 models) and thought this was now needed! Glad it seems fine to keep it here.

PR #905, opened 2025-06-19
Suggestions for 686

davidorme:
# Description

This PR is a subbranch off #686 to propose my suggested changes there.


vgro (approved):
LGTM, @sallymatson wrote the section in the `config.md` so I think it would be good to get her ok, too.

sallymatson (approved):
LGTM, thanks for going through it !

PR #906, opened 2025-06-19
888 animal model   foraging time budget

### LINKED ISSUE #888: Animal Model - foraging time budget
**Describe the bug**
I think the current foraging structure over multiple diet types may be reusing the activity window time budget for each different type of foraging such that a trophically flexible cohort has multiple full days worth of foraging in a single bout. 

This could be handled with a static proportion of foraging effort stored as an attribute in the cohort or functional group OR could be solved with a dynamic time budget tracked across resources.

TaranRallings:
# Description

This PR fixes a bug where the time allocated to foraging was being reused such that the more types of resources a cohort foraged on, the more time it had to forage. Each diet-type was given the full "work-day" to forage on. 

The fix takes the available time and divides it equally among the different resource types of resources can forage on. So if you consume just animals, animal foraging gets the whole time. If you consume animals and plants, each get half of the time, and so on.

Changes:
- `DietType.count_dietary_categories()`
- `AnimalCohort.diet_category_count`
- Revised time logic in `AnimalCohort.forage_cohort()`
- Revised tests

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Code-wise, this looks totally fine. I'm curious about the science of it, though. I bet there're diet that animals consume only as a last resort if there's nothing else available, right? I understand that might be hard to model, though.

TaranRallings:
> Code-wise, this looks totally fine. I'm curious about the science of it, though. I bet there're diet that animals consume only as a last resort if there's nothing else available, right? I understand that might be hard to model, though.

You're absolutely right. In real life, animal diets are diversely weighted at any particular moment and, broadly, quite adaptive. I spent a big chunk of my doctoral work on this problem in herbivores. 

The plan for animal diets its to stage out the complexity over time, particularly as we see how run-time is looking:
1) flat herbivore/carnivore distinction
2) highly resolved, equally weighted, non-adaptive
3) highly resolved, diversely weighted, non-adaptive
4) highly resolved, diversely weighted, adaptive

We have just moved from stage 1 -> stage 2. It is currently unclear whether our final destination is stage 3 (a standard, defensible approach) or stage 4 (more state-of-the-art). 

Moving to an weighting system with functional groups requires some additional scientific work so I'll tackle it after the basic version is up and running.

PR #912, opened 2025-06-21
Add cohort data exporter from PlantsModel

### LINKED ISSUE #911: Add cohort data exporter from PlantsModel
We want to be able to output a time series of community data across cells and cohorts. 

I think we could do this via the `data` object, but it is a lot of data that is only used internally within the PlantsModel and it's also set of ragged arrays. I think we:

* provide a new plant model config option "cohort_data_output_path" which needs to be passed to the model `__init__`.
* If that is not None then at each time step the cohort details are appended to that file. At the moment, maybe just DBH and crown area. Everything about the tree can be recreated from PFT and DBH anyway, but might be convenient to allow allometry data to be saved out. 
* That's more details, so maybe package this in a CohortOutput class to allow a single argument to the model `__init__` to give more detail (and potentially encapsulate methods).

Although this seems like a "nice to have" feature, it's going to be hugely useful for debugging issues with the plant model, so I think it is worth pushing now.

davidorme:
# Description

This PR introduces the `models.plants.exporter` module that provides a the `CommunityDataExporter` tool for plant community data. This data could be worked into the `data` object and exported by the standard mechanism, but the data is not used by any other model and consists of sets of highly ragged arrays of community data per cell that is much more suited to a data frame style export than the DataArray format.

There are three kinds of data:

* cohort data - one line per cohort per cell
* community canopy data - one line per canopy layer per cell
* stem canopy data - one line per canopy layer per cohort per cell

Each of those different classes has a bunch of attributes that are stored within the `communities`, `canopies` and `stem_allocation` attributes within the Plants Model.

So, this PR:

* Adds new section of configuration for the plants model schema to specify output paths for each kind of a data and an attribute list to subset the data written out.

```toml
[plants.community_data_export]
required_data = ['canopy', ... ]
cohort_attributes = []
community_canopy_attributes = []
stem_canopy_attributes = []
```

* Adds the new exporter class `CommunityDataExplorer` with:
    * an `__init__` method that the settings above directly
    * a `from_config` factory method that creates an instance from the settings loaded in a Config object.
    * a `dump` method that can be called to compile and output the three data types to CSV file. It switches from write to append mode after first use to avoid duplicating column headers.
    * Some private validation methods for paths and attribute subsets.

* Updates `PlantsModel` to require a `CommunityDataExplorer` object at `__init__` and then to call the `dump()` method at the end of model  setup and in each update.

* Adds a bunch of tests in `tests.models.plants.test_explorer.py`

Fixes #911  (issue)

## Type of change

## Key checklist

## Further checks


dalonsoa (commented):
Conceptually, I think this approach makes sense, leaving up to the model to export some of the data relevant for it. My only concern is if the data it exports might be ever be updated by a model run after plats within the same timestep. I wonder if it would be better to have an `export` step that is run for every model after all the updates are done.

davidorme:
> My only concern is if the data it exports might be ever be updated by a model run after plats within the same timestep. I wonder if it would be better to have an `export` step that is run for every model after all the updates are done.

After all the updates are done within a time step. That's an interesting idea that I think might be useful. In this specific case, we're intending to exporting the internal cohort structure and tree allometry, which is only under the control of the plants model, and indeed isn't exposed to the other models through `data`. 

But... I think we could revisit this same structure and make it part of the BaseModel to create model exporter methods for dumping internal model state. I'm just not keen to do that now 😄

virtual_ecosystem/models/plants/exporter.py
```diff
@@ -0,0 +1,374 @@
...
+            # If subset is provided, check the values are all valid
+            if subset:
+                not_found = subset.difference(available)
+                if not_found:
+                    msg = (
+                        f"The {subset_name} exporter configuration contains "
+                        f"unknown attributes: {', '.join(not_found)}"
+                    )
+                    LOGGER.error(msg)
+                    raise ConfigurationError(msg)
```
dalonsoa: Totally a personal thing, but I feel it is more readable to reduce nesting as much as possible. In this case, I would go for:
```suggestion
            if not subset:
                continue
            
            not_found = subset.difference(available)
            if not_found:
                msg = (
                    f"The {subset_name} exporter configuration contains "
                    f"unknown attributes: {', '.join(not_found)}"
                )
                LOGGER.error(msg)
                raise ConfigurationError(msg)
```
You could even replace the second conditional with:
```python
if not not_found:
    continue
```
But that is probably a step too far and mask a bit that the exceptional circumstance is the error, so I would leave it as it is.

virtual_ecosystem/models/plants/exporter.py
```diff
@@ -0,0 +1,374 @@
...
+            community_canopy_attributes = []
+            stem_canopy_attributes = []
+
+        If the "attributes" sections are empty arrays, then all attributes are written
+        to file, but specific fields may be named here to reduce the amount of data
+        exported.
+        """
+
+        # Try and build the arguments as a dictionary from the config, substituting
+        # explicit None values for empty strings
```
dalonsoa, line 242: I might be getting this wrong, but I think this is replacing explicit empty strings by None values, so the opposite of what you say.

virtual_ecosystem/models/plants/exporter.py
```diff
@@ -0,0 +1,374 @@
...
+        The configuration requires that the following details are present in the plants
+        model section of the configuration
+
+        .. code-block:: toml
+
+            [plants.community_data_export]
+            active = true
+            cohort_data_path = "path/to/cohort_data.csv"
+            community_canopy_data_path = "path/to/community_canopy_data.csv"
+            stem_canopy_data_path = "path/to/stem_canopy_data.csv"
```
dalonsoa: The simulation is already producing outputs somewhere. Do you want to allow producing the outputs for the plants internals anywhere else? Wouldn't it be clearer to use "convention over configuration" and output the plants stuff within a folder in the outputs directory with well defined filenames? It will also save you checking if the location is writtable and, possibly, other checks as well.

virtual_ecosystem/models/plants/exporter.py
```diff
@@ -0,0 +1,374 @@
...
+                stem_canopy_attributes=set(cfg["stem_canopy_attributes"]),
+            )
+        except KeyError as excep:
+            LOGGER.error(excep)
+            raise
+
+        # Return the instance
+        return cls(**args)
+
+    def dump(
```
dalonsoa, line 266: This method should be broken down into smaller chunks, which will also make for better readability and easier, more targeted testing testing. 

```python
if self.cohort_data_path:
    self._dump_cohort_data(...)
if self.community_canopy_data_path:
    self._dump_community_canopy_data(...)
if self.stem_canopy_data_path:
    self._dump_stem_canopy_data(...)
```
You can even leave the checking to the individual methods, so each of them is "safe":

```python
def dump(...):
    self._dump_cohort_data(...)
    self._dump_community_canopy_data(...)
    self._dump_stem_canopy_data(...)

def _dump_cohort_data(...):
    if not self.cohort_data_path:
        return
    ...
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -405,6 +433,15 @@ def _update(self, time_index: int, **kwargs: Any) -> None:
         # Calculate the subcanopy vegetation
         self.calculate_subcanopy_dynamics()
 
+        # Run the community data exporter
+        self.exporter.dump(
+            communities=self.communities,
+            canopies=self.canopies,
+            stem_allocations=self.stem_allocations,
+            time=self.model_timing.start_time
+            + time_index * self.model_timing.update_interval,
```
dalonsoa, line 442: It might come handy to add a simple method in the base model that returns the current time:
```python
def current_time(self, time_index: int) -> int: # was this in epoch?
    return self.model_timing.start_time + time_index * self.model_timing.update_interval
```

dalonsoa (changes requested):
I have a couple of comments, but it is broadly looking good.

virtual_ecosystem/models/plants/exporter.py
davidorme: It's certainly in a confusing order.

virtual_ecosystem/models/plants/exporter.py
davidorme: Yup - I think you're right. 

I'm using the empty string to signal not to save that file, which is kinda clunky, so we'd need a config option to turn them on or off. But I think fixing the names is much cleaner and simpler (at the cost of a totally insignificant bit of configuration flexibility).

virtual_ecosystem/models/plants/plants_model.py
davidorme: We need to revisit the timing in more detail but this is a good idea.

virtual_ecosystem/models/plants/exporter.py
davidorme: OK - now uses the main output directory from the config and set file paths.

davidorme:
@sallymatson I've finished updating from @dalonsoa's review now. Diego - your suggestions have simplified the code and testing quite a bit. Nothing hugely changed but just more polished, I think.

dalonsoa (approved):
Much better :)

arne-exe:
Hi @davidorme, I just had a look at this, the exporter looks very neat and will be super useful for looking at the outputs!

PR #917, opened 2025-06-23
Allow animal consumption of soil pools

### LINKED ISSUE #904: Allow animal consumption of soil pools
Need to add functionality to allow animals to consume soil pools. This could be done by the animals just subtracting directly from the relevant pools, but this would result in a sudden step change between simulation steps. What I really want is a continuous consumption rate which can be used within the integration.

I think this can probably be achieved via a similar structure to what is used in the animal-litter consumption case

jacobcook1995:
# Description

This PR adds `SoilPool` objects which animals can consume from. These follow the basic structure of the `LitterPools` which are already used for animal consumption of litter. They mainly serve to convert from the quantities used in the soil model into those used by the animal. They also serve to group the pools available for animal consumption into ones that animals could feasibly specialise in eating (i.e. particulate organic matter, bacteria, fungi), rather than allowing them to feed from every single pool as a discrete resource or to feed from nonsensical pools (e.g. soil enzymes)

As part of the PR I've had to modify the animal model code, so feedback on whether the changes seem sensible/in keeping with the overall code style would be useful.

Fixes #904

## Type of change

## Key checklist

## Further checks


TaranRallings (commented):
Am I following the logic correctly?

We keep the decay classes, the new one mirrors the litter pool.
I forage from the soil pool objects directly.
The new soil code auto-calculates the loss to animals and feeds that back into your new data variables?

jacobcook1995:
> Am I following the logic correctly?
> 
> We keep the decay classes, the new one mirrors the litter pool. I forage from the soil pool objects directly. The new soil code auto-calculates the loss to animals and feeds that back into your new data variables?

Yep exactly! With the idea being that the foraging code does not touch the soil model variables in the data object at all (the possibility for bugs going that route feels limitless)

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -166,8 +184,12 @@ def __init__(
         """The carcass pools in the model with associated grid cell ids."""
         self.leaf_waste_pools: dict[int, HerbivoryWaste]
         """A pool for leaves removed by herbivory but not actually consumed."""
-        self.litter_pools: dict[int, dict[str, Resource]] = self.populate_litter_pools()
+        self.microbial_c_n_p_ratios: dict[str, dict[str, float]]
+        """The CNP ratios of the different microbial functional groups."""
```
dalonsoa, line 189: I think I ask this every time I see CNP somewhere. Were not a dataclass to handle this specific data structure consistently across the codebase?

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -513,6 +585,142 @@ def calculate_total_litter_consumption(
...
+
+        pom_final_stock = array(
+            [
+                soil_pools[cid]["pom"].mass_current
+                / (area * self.core_constants.max_depth_of_microbial_activity)
+                for cid in cell_ids
+            ]
+        )
+
+        pom_consumption_carbon = pom_initial_stock - pom_final_stock
```
dalonsoa, line 622: Is there any risk of this ever becoming negative? If so, what should happen?

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -513,6 +585,142 @@ def calculate_total_litter_consumption(
...
+
+        bacteria_final_stock = array(
+            [
+                soil_pools[cid]["bacteria"].mass_current
+                / (area * self.core_constants.max_depth_of_microbial_activity)
+                for cid in cell_ids
+            ]
+        )
+
+        bacteria_consumption = bacteria_initial_stock - bacteria_final_stock
```
dalonsoa, line 644: Same here.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -513,6 +585,142 @@ def calculate_total_litter_consumption(
...
+        )
+
+        fungi_final_stock = array(
+            [
+                soil_pools[cid]["fungi"].mass_current
+                / (area * self.core_constants.max_depth_of_microbial_activity)
+                for cid in cell_ids
+            ]
+        )
+        fungi_consumption = fungi_initial_stock - fungi_final_stock
```
dalonsoa, line 675: And here...

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -513,6 +585,142 @@ def calculate_total_litter_consumption(
...
+                ectomycorrhiza_consumption
+                / self.model_timing.update_interval_quantity.to("days").magnitude,
+                dims="cell_id",
+            ),
+            "animal_arbuscular_mycorrhiza_consumption": DataArray(
+                arbuscular_mycorrhiza_consumption
+                / self.model_timing.update_interval_quantity.to("days").magnitude,
+                dims="cell_id",
+            ),
+        }
```
dalonsoa: Just for readability, I would use a small helper function:
```suggestion
        days = self.model_timing.update_interval_quantity.to("days").magnitude
        def to_da(array):
            return DataArray(array / days, dims="cell_id")
            
        return {
            "animal_pom_consumption_carbon": to_da(pom_consumption_carbon),
            "animal_pom_consumption_nitrogen": to_da(pom_consumption_nitrogen),
            "animal_pom_consumption_phosphorus": to_da(pom_consumption_phosphorus),
            "animal_bacteria_consumption": to_da(bacteria_consumption),
            "animal_saprotrophic_fungi_consumption": to_da(saprotrophic_fungi_consumption),
            "animal_ectomycorrhiza_consumption": to_da(ectomycorrhiza_consumption),
            "animal_arbuscular_mycorrhiza_consumption": to_da(arbuscular_mycorrhiza_consumption),
        }
```

virtual_ecosystem/models/animal/decay.py
```diff
@@ -344,6 +344,242 @@ def get_eaten(
...
+
+    def _extract_fungi_cnp_mass(
+        self,
+        data: Data,
+        biotic_activity_depth: float,
+        c_n_p_ratios: dict[str, dict[str, float]],
+    ):
+        """Extract the CNP masses of the soil fungi pools.
+
+        Animals are assumed to just generically eat soi fungi rather than being able to
```
dalonsoa:
```suggestion
        Animals are assumed to just generically eat soil fungi rather than being able to
```

virtual_ecosystem/models/soil/pools.py
dalonsoa, line 1: Here we have a collection of subtractions. I guess you are on top of them and it is not possible, but just checking that they cannot lead to negative numbers, ever.

dalonsoa (approved):
I've a couple of suggestions here and there, and a few comments, but nothing major. All looks good!

virtual_ecosystem/models/soil/pools.py
jacobcook1995: The changes in pools that are being calculated here can be negative. In terms of preventing the pools from going negative, I haven't been able to think of a reasonable way to stop this from happening. As the animal and soil model run in sequence the total consumption is calculated based on old soil pool sizes and then subtracted from the soil pools at the next simulation step, so if the abundance of the pool suddenly drops the consumption rate won't allowing the pool to potentially go negative.

This happening would be an indication that something is wrong in how animal consumption has been parameterised, but I'm not really sure what makes sense to do in response. Pools that are negative could be replaced with zero values, but I feel like that just masks the problem, as well as breaking stoichiometric conservation (because the animals receive more CNP from the soil than they remove)

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: There shouldn't be any processes in the animal model that can increase the size of the `SoilPool` pools. These are designed to just track the consumption within one animal model evaluation step, so there shouldn't be any changes from any other models to the pool within the relevant time frame. So, `pom_final_stock` can only ever be equal to or smaller than `pom_initial_stock`

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: see above

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: see above

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: Yeah that could be useful. It would be somewhat tricky as the representation varies between models.

For the microbial groups, the soil model does not track the stoichiometry explicitly it just assumes that each group has a fixed stoichiometry (which is an input parameter). This is basically to keep the amount of pools that the integration has to track lower (there's a huge number anyway).

I guess it probably does make sense to think about a unified approach to this at some point, but I think it probably has to wait until plant stoichiometry is implemented so that we know what all models require the general system to do

PR #921, opened 2025-06-26
880 soil moisture capacity needs to come out of core constants

### LINKED ISSUE #880: soil moisture capacity needs to come out of core constants

vgro:
This PR removes `CoreConsts.soil_moisture_capacity` and uses `HydroConsts.soil_moisture_saturation` instead.

Fixes #880


jacobcook1995 (approved):
LGTM!

PR #922, opened 2025-06-27
Update pyrealm version

### LINKED ISSUE #918: Update pyrealm version
I've got the sequence of updates in a mix - I've made changes to `pyrealm` that I've brought into #824 but those are mixed up with the actual new code for recruitment. So I need to bring some of those updates in to `develop` now to allow #911 to proceed.

So this issue is to bring VE up to date with `pyrealm` - for the moment I'm going to switch over to using  the `develop` branch of `pyrealm` rather than actual releases to give a bit more flexibility while developing both. We can swap back when `pyrealm 2.0.0` drops.

davidorme:
# Description

This PR:

* Swaps `pyrealm` to tracking the `develop` branch rather than specific releases. This will stabilise again, but during this bug-squashing phase it is a pain to keep having to make `pyrealm` release candidates just to fix the next bug.
* Updates to the new light capture model in `pyrealm`reworkin
    * A couple of new PlantsModel properties to capture light making it through the canopy and to the ground
    * Restructure of the light capture part of the model
* I've reworked the tests extensively:
    * The fixture canopy layer data for testing is now generated programatically rather than using fixed values. This is more flexible for future changes to the canopy being used. The aim of these tests is basically to check that shoving everything through `PlantsModel` gives the same results as a simple direct calculation - we're not aiming at a particular gold standard set of values beyond that.
    * The code in that fixture is basically the recipe I would have needed to write to generate the new set of fixed data, and it just seemed more sane and flexible to provide the recipe rather than just the result.
    * I've simplified the structure of the fixture data passing to make the `data_validator` helper simpler and rationalised the test case names to make them more consistent.

> [!CAUTION]
> This PR also adds a deliberate hack to prevent stem DBH being driven to zero or lower. This lead to wild values and `pyrealm` now explicitly blocks it. Now, if the delta DBH for  a stem _would_ force the DBH to become zero or less, the delta DBH is not applied and the stem hangs around as a kind of zombie sapling. This is easier _for now_ than adding explicit mortality for those stems and removing the cohorts. We will revisit this once we've figured out why so much shrinking is happening and when we look at carbon starvation.

Fixes #918

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
lgtm!

PR #928, opened 2025-07-03
Updating NDArray typing

### LINKED ISSUE #910: Revise NDArray typing.
* Make sure all NDArray typed variables specify a dtype.
* Convert all of the existing floating types (mostly `np.float32`, but some `np.float64` and `np.floating`) to the generic `np.floating` type

See the following Teams convo:

@davidorme 
> Numpy array typing
> Diego and Alexander - something that has been in the back of my mind for a while and got raised over in the pyrealm project. At the moment, we are typing (mostly) our numpy array contents using NDArray[...]. We got into the habit at the start of using NDArray[np.float32] for floats and I think that's going to bite us:
>  
> As I understand it, numpy 2.0 has rejigged typing aliases and the typing hierarchy and I'm seeing more instances of mypy complaining about stuff being NDArray[np.float32] and a numpy function returning NDArray[np.floating] .  
>  
> I think the rationale was that nothing we did needed full 64 bit precision so we could keep the memory profile lower using 32 bit, but since machines have 64 bit architecture, unless we enforce the bit depth incredibly carefully, the code is going to be constantly converting data types.
>  
> My suspicion is that we should just use NDArray[np.floating] throughout to keep mypy warning us if we try and shove int or bool etc. data into those variables, but we should otherwise not fuss about bit depth.
>  
> Does that sound right?
> 

 
@alexdewar 
> That sounds right to me. I wouldn't worry too much about the overhead of using 64-bit numbers here as I suspect it'll be pretty negligible with the hardware and models you'll be using. If there are performance problems down the road, I suspect it'll be because of the time spent in the Python interpreter instead, which will require a different approach to fix.
 
@davidorme 
> We might want to only write data out to float32. In my experience that can be a valuable disk space saving approach, but we're likely not dealing with huge datasets.
 
@dalonsoa 
> Make the conversion just at the output layer, then. For actual calculations, the higher precision might come handy, anyway.
 
@alexdewar 
> I'd be inclined to just live with the bigger data files, just so that they match the internal representation of VE as much as possible. You can always trim them down if they get silly big.
 
@davidorme 
> Yeah - I'm not going to rush to that now, but I've used it a lot saving global  climate data files on the RDS, where a 50% file saving is many GB.

davidorme:
# Description

This:

* updates all `NDArray` typing of floats to use `np.floating`
* adds some typing of bare `NDArray`. A couple of utility functions are intended to handle heterogenous types, so I've left those alone.
* updates `np.ndarray(..., dtype='np.float32')` to `np.ndarray(..., dtype='np.float64')` - can't request a generic floating type here, but np.float64 is more consistent.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!

PR #929, opened 2025-07-03
Missed push from #911

### LINKED ISSUE #911: Add cohort data exporter from PlantsModel

davidorme:
# Description

This is a micro PR of a single commit that was not pushed to #912:

* docstring update and explanation of non-obvious feature in test TOML strings.

Close #911 

## Type of change

## Key checklist

## Further checks


dalonsoa (approved):
Short PRs are the best...

Issue #517, opened 2024-07-12
Replace daily lists in HydrologyModel with array implementation


davidorme:
**Is your feature request related to a problem? Please describe.**

At the moment, the hydrology model tracks daily values for variables within a time step by appending daily arrays to a list and then using `np.stack` to combine them at the end to get statistics across the time step.

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/68129075e9bd03f77d612eb5b6b2447d49a59aef/virtual_ecosystem/models/hydrology/hydrology_model.py#L390-L408

Since the number of days is known (and will always be known?), this can be done by declaring a days x grid cell numpy array and inserting daily rows. If the code below is a fair comparison of the two approaches then this is faster and cleaner. I have to admit I was expecting it to be _more_ faster, but it is faster and the advantage greater as the number of cells increases.

```python
In [1]: import timeit
   ...: import numpy as np
   ...: 
   ...: 
   ...: def stacking_arrays(n_cells=10, n_days=30):
   ...: 
   ...:     val = []
   ...:     for day in np.arange(n_days):
   ...:         val.append(np.arange(n_cells))
   ...: 
   ...:     return np.sum(np.stack(val, axis=1), axis=1)
   ...: 

In [2]: def matrix_insertion(n_cells=10, n_days=30):
   ...: 
   ...:     val = np.empty((n_days, n_cells))
   ...:     for day in np.arange(n_days):
   ...:         val[day, :] = np.arange(n_cells)
   ...: 
   ...:     return np.sum(val, axis=0)
   ...: 

In [3]: 
   ...: assert np.allclose(stacking_arrays(),  matrix_insertion())
   ...: 
   ...: number = 1000

In [4]: np.min(timeit.repeat("stacking_arrays()", globals=globals(), number=number))
Out[4]: np.float64(0.02308262512087822)

In [5]: np.min(timeit.repeat("matrix_insertion()", globals=globals(), number=number))
Out[5]: np.float64(0.0157989589497447)

In [6]: np.min(timeit.repeat("stacking_arrays(n_cells=1000)", globals=globals(), number=number))
Out[6]: np.float64(0.05126216681674123)

In [7]: np.min(timeit.repeat("matrix_insertion(n_cells=1000)", globals=globals(), number=number))
Out[7]: np.float64(0.03652237495407462)

In [8]: np.min(timeit.repeat("stacking_arrays(n_cells=5000)", globals=globals(), number=number))
[...]

PR #933, opened 2025-07-07
780 abiotic model   account properly for time step and area in particular with energy fluxes

### LINKED ISSUE #780: Abiotic model - account properly for time step and area, in particular with energy fluxes
At the moment it is not clear what the timescale of the abiotic model is; I only assume that it is more than one hour so the soil, canopy and atmosphere are in equilibrium. The time step needs to be accounted for properly in the microclimate calculations and the time step limits need to be set in the model configuration.

vgro:
This PR aims to sort out the time intervals for the energy balance. The equations really only make sense in a short time interval, an hour or so, so I did a very clumsy for loop implementation as a first step and would appreciate feedback on how to better handle it. At the moment we cannot have hourly input data, so it does not make sense to run it for more than 1 average hour, this is a separate problem.

I have also deleted a lot of bits that are handled by the plant model now (related to absorbed radiation).

Fixes #780 

## Type of change

## Key checklist

## Further checks


davidorme:
@vgro Could you walk through the logic of the time interval value to `run_microclimate`? At the moment `AbioticModel._update` is pinning that value at 3600? So what are the different time interval cases that the logic is intended to handle?

vgro:
I want the timestep that goes in the functions (delta t) to be one hour, the update interval of the model will most likely always be larger than that, so I divide the model update interval by 3600 to get the number of hours per model time interval. If the model time interval is a day or shorter, run for every hour (this will be the option with hourly input data), if not than do only one step

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -197,121 +192,192 @@ def run_microclimate(
...
     ].to_numpy()
 
-    # -------------------------------------------------------------------------
-    #  Calculate atmospheric background variables
-    # -------------------------------------------------------------------------
+    # Get combined evapotranspiration from plant and hydrology model, per hour
+    evapotranspiration = data["canopy_evaporation"] + data["transpiration"]
+
+    # If hourly input data is provided, iterate over day, else equilibrium assumption
+    hourly_time_interval = int(time_interval / core_constants.seconds_to_hour)
```
dalonsoa: Not sure if it is a risk, but if `time_interval < core_constants.seconds_to_hour` this result in `hourly_time_interval = 0`. As that probably makes no sense, given that you use this below below in a denominator, I would do:
 ```suggestion
    hourly_time_interval = max(int(time_interval / core_constants.seconds_to_hour), 1)
```

dalonsoa (commented):
I've made a comment about a potential error, but other than that, I do not see any other, more elegant way of doing this if the equations being solved require smaller timesteps. To be honest, inner loops to handle shorter timesteps is something that I was expecting to happen given the different timescales of the processes you are handling.

davidorme (approved):
Looks good - thanks for clarifying the timing stuff. I do find the setup within `run_microclimate` a bit confusing - it seems like the running mode (iterated vs equilibrium) might be something that is an argument to the function. Having said that - there's a lot of stuff still to come when the hourly data becomes available - so I think that's polishing for later.

vgro:
> Looks good - thanks for clarifying the timing stuff. I do find the setup within `run_microclimate` a bit confusing - it seems like the running mode (iterated vs equilibrium) might be something that is an argument to the function. Having said that - there's a lot of stuff still to come when the hourly data becomes available - so I think that's polishing for later.

Thanks for the feedback. I will definitely revisit the setup; at the moment the model just returns the final state and I would like to treat this more like the hydrology model where monthly means/stats will be returned. I just didn't want to complicate this PR any further (initially only intended to make sure everything has the same unit). Will consider #517 in this context.

PR #948, opened 2025-07-14
Added production of fungal fruiting bodies

### LINKED ISSUE #907: `calculate_new_biomass_average_nutrient_ratios` calculates the weighted averages of the C:N:P ratios incorrectly
**Describe the bug**
In the soil model function `calculate_new_biomass_average_nutrient_ratios` I calculate the weighted average of the stoichiometric ratios relative to the carbon allocation. I should instead calculate the total N and P flows and then find the correct C:N:P ratios based on this

### LINKED ISSUE #932: Fungi should produce fruiting bodies
In the real world, fungal fruiting bodies form a significant part of the diets of some animal taxa. But at present, these bodies are not represented in the Virtual Ecosystem

A certain percentage of fungal biomass synthesis should be allocated to production of fungal bodies which animals can consume.

The basic assumption that I want to use is that fungi allocate to reproduction when they are growing well. This will allow for environmental factors to be captured indirectly, which minimises the parameterisation complexity, i.e. so we don't have to track down parameters for the temperature response of fungal fruiting.

jacobcook1995:
# Description

This PR adds production of fruiting bodies by soil fungi. On the soil side this change is fairly straightforward, fungi allocate a certain proportion of their new biomass to fruiting body production in the same way that they allocate to enzyme production. The production rate across all fungal groups is then summed and passed to the animal model.

The animal model stores the mass of fruiting bodies as a scavengable pool. As well as scavenging the animal model also updates this pool based on the input of new fruiting body biomass from the soil model, and the decay of fruiting bodies. This decay gets passed back to the soil model as low molecular weight carbon (and the equivalent forms of nitrogen and phosphorus). At each time step the animal model also updates the fungal fruiting body abundance in the data object.

@TaranRallings, I've added a reasonable amount of code to the animal model here, so let me know if you are happy with the changes.

Fixes #932. Fixes #907 

## Type of change

## Key checklist

## Further checks

- [] Relevant documentation reviewed and updated


virtual_ecosystem/models/soil/microbial_groups.py
```diff
@@ -178,14 +206,18 @@ def calculate_new_biomass_average_nutrient_ratios(
...
 
-    Microbes have to synthesise cellular biomass as well as extracellular enzymes. This
-    method calculates average nutrient ratio of this total biomass synthesis by
-    calculating the average weighted by the relative production allocation to each
-    enzyme class and cellular growth.
+    Microbes have to synthesise cellular biomass as well as extracellular enzymes, and
+    fungi also allocate to reproductive fruiting bodies. This method calculates average
+    nutrient ratios of this total biomass synthesis using the relative production
+    allocation to each enzyme class, cellular growth and (for fungi) reproductive
+    allocation. Carbon nutrients ratios have to be inversed before averaging as they are
```
dalonsoa: `as they are...` something, but we don't know what :) . There's something missing at the end of this sentence.

dalonsoa (approved):
It's a pretty long PR, but I think all makes sense and is clear. This is becoming more complicated by the minute...

virtual_ecosystem/models/soil/microbial_groups.py
jacobcook1995: Ahh this is something I was having trouble explaining so left intending to come back and then just forgot 🫠. I'll fix that now

jacobcook1995:
> This is becoming more complicated by the minute...

Yes this is the last addition that I intended to make to the soil model before moving onto bug-fixing and calibration, so this should be as most complex as the soil model gets 🤞

TaranRallings (approved):
The animal content looks good to me!

PR #954, opened 2025-07-16
Changed calculation of removal rate of nutrients due to water

### LINKED ISSUE #804: Horizontal flows of water should carry nutrients away from the soil
At the moment only downwards flow of water removes nutrients from the soil. This doesn't really make sense and horizontal flows of water should also carry nutrients away. This should only really happen for flows that exit the model, flows of nutrients between soil cells isn't something we want to tackle at present

jacobcook1995:
# Description

This PR changes the calculation of removal of nutrient due to water so that it averages over all flows of water out of the microbial active soil region. This actually turns out to be just vertical flows, but this isn't an issue as what we want is for the soil nutrient flows to track the hydrology flows.

I've tried to future proof a bit against possible future changes to the hydrology model by referring to "soil nutrient" removal rather than "leaching" (which is vertical flow specific, I think). 

Fixes #804

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/soil/env_factors.py
+++ b/virtual_ecosystem/models/soil/env_factors.py
@@ -496,7 +496,7 @@ def find_water_outflow_rates(
     For the vertical movement, water only leaves this region if it passes into a soil
     layer that is either only partially microbially active or completely inactive.
     Therefore, we only need to consider the bottom two microbially active layers as
-    these are the layers only which can have flows into partially active or inactive
+    these are the only layers which can have flows into partially active or inactive
     layers. How much of a given flow is considered to have exited the microbially active
     region depends on how much of each layer is assumed to be microbially active. If
     water is flowing from a layer that is only partially active, the exit rate will be
--- a/virtual_ecosystem/models/soil/env_factors.py
+++ b/virtual_ecosystem/models/soil/env_factors.py
@@ -493,17 +493,19 @@ def find_water_outflow_rates(
     that the upper layers also have horizontal water movements this function will need
     to change to ensure that nutrient flows properly track the water flows.
 
-    For the vertical movement, water only leaves this region if it passes into a soil
-    layer that is either only partially microbially active or completely inactive.
-    Therefore, we only need to consider the bottom two microbially active layers as
-    these are the only layers which can have flows into partially active or inactive
-    layers. How much of a given flow is considered to have exited the microbially active
-    region depends on how much of each layer is assumed to be microbially active. If
-    water is flowing from a layer that is only partially active, the exit rate will be
-    proportional to the fraction of this layer that is considered to be microbially
-    active. Conversely, if water is flowing to a layer that is partially microbially
-    active the exit rate will be proportional to the inverse of the fraction of the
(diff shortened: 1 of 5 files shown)


virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -467,3 +472,73 @@ def find_total_soil_moisture_for_microbially_active_depth(
...
+    equivalent to the rate at which water escapes from this region. For the upper soil
+    layers, all water flows are vertical rather than horizontal, so this function only
+    considers vertical flows. If the implementation of the hydrology model changes so
+    that the upper layers also have horizontal water movements this function will need
+    to change to ensure that nutrient flows properly track the water flows.
+
+    For the vertical movement, water only leaves this region if it passes into a soil
+    layer that is either only partially microbially active or completely inactive.
+    Therefore, we only need to consider the bottom two microbially active layers as
+    these are the layers only which can have flows into partially active or inactive
```
vgro:
```suggestion
    these are the only layers which can have flows into partially active or inactive
```

virtual_ecosystem/models/soil/env_factors.py
```diff
@@ -467,3 +472,73 @@ def find_total_soil_moisture_for_microbially_active_depth(
...
+    assumption underlying the
+    :func:`~virtual_ecosystem.models.soil.env_factors.calculate_solute_removal_by_soil_water`
+    function). The rate of "new" water refreshing the microbially active column will be
+    equivalent to the rate at which water escapes from this region. For the upper soil
+    layers, all water flows are vertical rather than horizontal, so this function only
+    considers vertical flows. If the implementation of the hydrology model changes so
+    that the upper layers also have horizontal water movements this function will need
+    to change to ensure that nutrient flows properly track the water flows.
+
+    For the vertical movement, water only leaves this region if it passes into a soil
```
vgro: I really like the first paragraph, very clear. From here it gets a bit confusing (from my hydrology perspective) as the water is technically still leaving the soil layers, but it is not accounted for in the nutrient carrying flow? Maybe clarify 'this region'.

vgro (approved):
LGTM, a few minor comments on the docs.

virtual_ecosystem/models/soil/env_factors.py
jacobcook1995: Ahh yeah, I had a hard time working out how to explain this. Basically it's that the water column/"region" that the soil nutrient model is interested in (generally) doesn't cover a whole number of layers. So, water exits this column within a specific layer rather than between them.

To calculate this I tried to split "water leaving the microbially active region" into two components:

1) Water that is considered to be entering the microbially inactive portion of a partially microbially active layer (from the layer above), given by:

flow into layer * (1 - fraction of the layer that is microbially active)

2) Water that is flowing into a microbially inactive layer, which needs to account for the fact that some of the flow is already considered to have left the microbially active region, this is given by:

flow into layer*fraction of old layer that is microbially active

Does that explanation make better sense? If so I'll try and rewrite the relevant part of the docstring in the same way

PR #957, opened 2025-07-21
782 abiotic model   relative humidity is 100 we need to include some form of advectionoutlet

### LINKED ISSUE #782: Abiotic model - relative humidity is 100%, we need to include some form of advection/outlet
At the moment, the way that relative humidity is updated causes relative humidity to be 100% very quickly. Wwe need to include some form of advection/outlet and at the same time check that the current implementation makes correct assumptions and conversions with respect to the air volume that is updated.

vgro:
This PR adds vertical mixing of water below the canopy and ventilation above the canopy so water can leave the system. The unit tests create nice vertical humidity profiles, however, the integrated function still returns 100% humidity. I think this has to do with the time interval and the input values and will likely become more realistic when we do the calibration. Any insight are welcome.

I also did a bit of a tidy in the test files, can be skipped.
Fixes #782

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM (with the obvious caveat that I don't understand the underlying science that well)

Issue #959, opened 2025-07-21
Sort out time interval in energy balance


vgro:
There is a persisting problem with the time interval and the temperature and vapor update. Needs some more thinking about the steady state assumption of the time step >1h

PR #960, opened 2025-07-21
958 vertical mixing of temperature

### LINKED ISSUE #958: Vertical mixing of temperature
The abiotic model currently calculates air temperatures for all layers but does not consider vertical mixing. This can be implemented similar to the humidity mixing.

vgro:
This PR adds vertical mixing to the temperature update. I also moved the functions for mixing and ventilation to `wind.py` and run them outside the time loop together with the wind profile because they do currently not change over time.

The problem with the time_interval persists, will be addressed at a later stage #959

Fixes #958 and #964

## Type of change

## Key checklist

## Further checks


vgro:
Sorry one of the test doesn't pass on windows, I'll fix that first

jacobcook1995 (approved):
LGTM (aside the very strange test failure)

davidorme:
I don't believe that the Windows test failure is to do with rounding errors. I've no idea what _is_ causing it and it will be a pain to debug without a Windows machine to hand, so I think going to lower precision is pragmatic for now. However, I think we should raise this fix as an issue because _something_ is happening here.

vgro:
> I don't believe that the Windows test failure is to do with rounding errors. I've no idea what _is_ causing it and it will be a pain to debug without a Windows machine to hand, so I think going to lower precision is pragmatic for now. However, I think we should raise this fix as an issue because _something_ is happening here.

I create #966

PR #962, opened 2025-07-21
Fixing incorrect plant growth

### LINKED ISSUE #961: Incorrect calculations in estimating plant growth.
**Describe the bug**

The plants model growth is clearly bugged - stems rapidly decrease to zero DBH although are now clipped to stop them shrinking into negative DBH. This issue is to document the bug hunting.

davidorme:
# Description

This PR contains fixes for incorrect plant growth calculations. The culprits are:

1. The `PlantsModel._setup` updated the turnover traits for the time step of the model, but did so in the wrong direction. The trait values are expressed in the models as reciprocals, and so $\tau_r = \tau_r / 12$ increases the speed of turnover in the model by a factor of $12^2 = 144$  compared to the correct $\tau_r = \tau_r \times 12$.

1. The `PlantsMode._setup` was not updating respiration rates for the time step of the model at all - so the full annual respiration costs were being removed at each time step. These rates are straightforward factors depending on masses, so **are** simply divided by the number of time steps.

1. The subcanopy vegetation biomass (should be kg C) was being updated with µg C values, so implausibly then contained 10 million tons of carbon per m2. That pushed the equations in the `abiotic_simple`  so far outside bounds that canopy temperature plummeted to -20°C and the plants then all slowly shrank to death.

There are likely to be more issues, but hopefully this gets the model into territory where we can calibrate it.

I had to update the test values in one of the stochiometry tests - I'm assuming these are code-derived values from the previous aberrant behaviour.

Fixes #961

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
LGTM

PR #970, opened 2025-07-24
967 reactivate aerodynamic resistance calculation

### LINKED ISSUE #967: reactivate aerodynamic resistance calculation
The aerodynamic resistance was et to a constant because other parts of the model were not sorted. Now that the temperature update seems stable, this can be addressed.

vgro:
This PR reactivates the functions for `aerodynamic_resistance` and `friction_velocity`. These used to be constants because there was a bug that was hard to identify with other pieces still largely missing. Now resolved. I also moved the function to `wind.py` from `energy_balance.py` and updated the docs.

Fixes #967 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -363,3 +363,50 @@ def advect_water_from_toplayer(
...
+
+
+def calculate_aerodynamic_resistance(
+    wind_heights: NDArray[np.floating],
+    roughness_length: NDArray[np.floating],
+    zero_plane_displacement: NDArray[np.floating],
+    wind_speed: NDArray[np.floating],
+    von_karman_constant: float,
+) -> NDArray[np.floating]:
+    r"""Calculate aerodynamic resistance in canopy, [s m-1].
```
jacobcook1995, line 378: As a minor stylist point, I don't know if you really need to include the units in the first line of the docstring, when they are also provided in the `Returns:` section.

Obviously, up to you though

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -363,3 +363,50 @@ def advect_water_from_toplayer(
...
+    where :math:`z` is the height where the aerodynamic resistance needs to be
+    calculated, :math:`d` is the zero plane displacement height, :math:`z_{m}` is the
+    roughness length of momentum, :math:`\kappa` is the von Karman constant, and
+    :math:`u(z)` is the wind speed at height :math:`z`.
+
+    Args:
+        wind_heights: Heights where wind speed is to be calculated [m].
+        roughness_length: Momentum roughness length, [m]
+        zero_plane_displacement: Height above ground within the canopy where the wind
+            profile extrapolates to zero, [m]
```
jacobcook1995: what does extrapolates to zero mean?

jacobcook1995 (approved):
LGTM! (just a couple of minor queries)

virtual_ecosystem/models/abiotic/wind.py
vgro: The wind speed follows a log profile in the canopy and is expected to reach zero at a height above the ground. If there is no vegetation, d =0. It's like an offset due to the roughness. I made this a bit clearer.

virtual_ecosystem/models/abiotic/wind.py
vgro: I started doing that at some point and kept it for consistency, I am not attached :-) I have a task in the near future to go through the doc strings and docs again to make sure all variable names are consistent, references complete, etc. I can address this then.

virtual_ecosystem/models/abiotic/wind.py
jacobcook1995: Makes sense, yes I think clarifying this would be useful!

PR #971, opened 2025-07-25
Stoichiometry recruitment 

sallymatson:
# Description

This PR creates an `add_cohorts` method for Stochiometry, which allows us to add new cohorts to match David's recruitement PR #827. 

When we merged in Stochiometry, this caused the tests to fail for that PR because while community cohorts were being added nothing was happening for stochiometry, leading to dimension mismatches. 

This PR should fix that issue. 

I am not convinced this is the cleanest code in the world but I really wanted to get in a draft before going on leave. @davidorme, completely up to you how to proceed - if you have major comments and want to wait for me to get back, I'll do it asap when I return. Otherwise, feel free to modulate on this code and then merge it in so you can get recruitement fully off your plate!! Unfortunately, recruitement cannot be merged without this kind of (relatively major) fix. So I hope this is helpful. 

If you decide to merge with outstanding issues, feel free to add as todos for me to finish when I get back. 

Next step will be making a similar method for mortality :) 

## Type of change

## Key checklist

- [ x] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [ x] All tests pass: `$ poetry run pytest`

## Further checks

### DIFF
--- a/virtual_ecosystem/models/plants/plants_model.py
+++ b/virtual_ecosystem/models/plants/plants_model.py
@@ -737,7 +737,7 @@ class PlantsModel(
 
             # GPP for each later is estimated as (value, dimensions, units):
             #    LUE                (n_active_layers, 1)          [gC mol-1]
-            #    * cohort fAPAR     (n_active_layers, n_cohorts)  [-]
+            #    * cohort fAPAR     (n_active_layers, per_layer_transpiration_mm)  [-]
             #    * canopy top PPFD  scalar                        [µmol m-2 s-1]
             #    * stem leaf area   (n_active_layers, n_cohorts)  [m2]
             #    * time elapsed     scalar                        [s]
@@ -1079,6 +1079,17 @@ class PlantsModel(
 
                 # Add recruited cohorts
                 community.add_cohorts(new_data=cohorts)
+
+                self.stochiometries[cell_id]["N"].add_cohorts(
+                    new_cohort_data=cohorts,
+                    flora=self.flora,
+                    element="N",
+                )
+                self.stochiometries[cell_id]["P"].add_cohorts(
+                    new_cohort_data=cohorts,
+                    flora=self.flora,
+                    element="P",
+                )
 
     def update_cn_ratios(self) -> None:
         """Update the C:N and C:P ratios of plant tissues.
--- a/virtual_ecosystem/models/plants/plants_model.py
+++ b/virtual_ecosystem/models/plants/plants_model.py
@@ -737,7 +737,7 @@ class PlantsModel(
 
             # GPP for each later is estimated as (value, dimensions, units):
             #    LUE                (n_active_layers, 1)          [gC mol-1]
-            #    * cohort fAPAR     (n_active_layers, per_layer_transpiration_mm)  [-]
+            #    * cohort fAPAR     (n_active_layers, n_cohorts)  [-]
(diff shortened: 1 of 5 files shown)


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -737,7 +737,7 @@ def estimate_gpp(self, time_index: int) -> None:
 
             # GPP for each later is estimated as (value, dimensions, units):
             #    LUE                (n_active_layers, 1)          [gC mol-1]
-            #    * cohort fAPAR     (n_active_layers, n_cohorts)  [-]
```
davidorme: Not sure why this is changed? It is just giving the expected shape of that input which is layers x cohorts.

davidorme (changes requested):
This looks good. 

I'd kinda like to be able to reuse the `pyrealm` `CohortMethods` mixin here, but I think that requires serious structural changes to work cleanly - essentially we'd need to separate the elements within tissues. That might be worth doing in the long run  - we could have `dict[str, Element]` within a tissue and then `Element.ideal_ratio` etc. But - that's a thing to note and park.

In the short term, I think I'd prefer the tissue classes to have their own `add_cohort` methods as a new abstract method of the `Tissue` ABC. It's a thing a new tissue _has_ to have and if we define it there, then the section in the `StemStochiometry.add_cohorts()` will automatically handle new tissues.

virtual_ecosystem/models/plants/stochiometry.py
```diff
@@ -413,7 +415,68 @@ def default_init(
...
+                    mass = (
+                        new_stem_allometry.foliage_mass[0][i]
+                        * new_stem_traits.zeta[i]
+                        * new_stem_traits.sla[i]
+                    )
+                    ratio = self.extra_pft_traits.traits[new_cohort_data.pft_names[i]][
+                        f"root_turnover_c_{element.lower()}_ratio"
+                    ]
+                tissue.actual_element_mass = np.append(tissue.actual_element_mass, mass)
+                tissue.ideal_ratio = np.append(tissue.ideal_ratio, ratio)
```
davidorme: So - move the per tissue code into `XYZTissue.add_cohorts` and then do this.

```suggestion
        for i in range(new_cohort_data.n_cohorts):
            for tissue in self.tissues:
                tissue.add_cohorts(stem_allometry, extra_pft_traits)
```

davidorme (changes requested):
I meant to include a suggestion showing what I meant, but clicked submit review too early. 

Another thing that has just occurred to me: I think `StemStochiometry.add_cohorts` should loop over the elements internally? We'd _never_ want to just add data on one element, so having to call it multiple times to make sure each element is handled doesn't make sense.

If this is enforced by an outer loop in the calling code, then we might want to revisit the looping over elements structure. That might fit in with the comment above about having an explicit `Element` class, so might be a rework for later, but its definitely something that bothers me!

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Not sure either, will remove this ahah

virtual_ecosystem/models/plants/stochiometry.py
sallymatson: Of course. This is so so much better. Thanks David!

davidorme (approved):
LGTM.

Issue #975, opened 2025-07-30
Test abiotic model with no canopy


vgro:
Update `dummy_climate_data_varying_canopy` to have a grid cell without vegetation.

PR #976, opened 2025-07-30
973 test abiotic model with different number of canopy layers

### LINKED ISSUE #973: Test abiotic model with different number of canopy layers
Abiotic tests currently use the simple dummy climate data in conftest which has homogeneously filled canopy layers.

The tests need to run with the different canopy layers and return NaN correctly. At the same time, the values need to be adjusted to more realistic monthly fluxes. This will hopefully help identify whit we still have problems with the time interval in the energy balance.

vgro:
This PR updates the tests in the abiotic folder to run with a varying number of canopy layers. This required introducing a few extra functions that deal with NaN and fixing how the layer thickness is calculated. I also fixed a sign error in the temperature updated.

You will see some slightly convoluted masking in some of the tests, that is to check that the true values are not NaN.

I haven't introduced a non-vegetated grid cell, this will be addressed in #975 .

Fixes #973

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/abiotic/test_abiotic_model.py
+++ b/tests/models/abiotic/test_abiotic_model.py
@@ -45,7 +45,7 @@ SETUP_MANIPULATIONS = (
 
 
 def test_abiotic_model_initialization(
-    caplog, dummy_climate_data, fixture_core_components
+    caplog, dummy_climate_data_varying_canopy, fixture_core_components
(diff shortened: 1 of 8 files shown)


jacobcook1995 (approved):
LGTM!

PR #981, opened 2025-07-31
Unpin the update interval in hydrology

### LINKED ISSUE #980: Unpin the update interval in `hydrology`
The `hydrology` model is currently pinned on a 30 day time step. We think we can remove this pin and calculate the actual number of days instead.

davidorme:
# Description

This PR:

* Relaxes the `hydrology` model to allow intervals other than 30 days and adds code to calculate the actual number of days in the update interval. It warns when the update interval is not a whole number of days.
* Updates the docs to describe the internal calculation.
* Updates the tests - there was a test that using `1 week` would fail - I've retained the `1 week` test case but updated it to show that it works. The resulting test values seem sensible: the model gets the same precipitation in fewer days, so the vertical flow and runoff are higher, but there is less time for soil evaporation, which is lower.

There are also a couple of requires `mypy` fixes - the `numpy` plugin for `mypy` has been deprecated and the new system picks up a couple of things.

@jacobcook1995 There are a couple of weird typing things here from `litter`: two functions that claimed to return `dict[str, NDArray]` but didn't and hadn't been caught for some reason. 

Fixes #980 

## Type of change

## Key checklist

- [X] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`

## Further checks

- [X] Code is commented, particularly in hard-to-understand areas
- [X] Tests added that prove fix is effective or that feature works
- [X] Relevant documentation reviewed and updated


.pre-commit-config.yaml
```diff
@@ -5,16 +5,16 @@ repos:
       - id: check-merge-conflict
       - id: debug-statements
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.5
+    rev: v0.12.3
```
jacobcook1995: Is there a reason for downgrading the pre-commit hooks?

virtual_ecosystem/models/litter/chemistry.py
```diff
@@ -358,7 +358,7 @@ def calculate_P_mineralisation(
         self,
         decay_rates: dict[str, NDArray[np.floating]],
         active_microbe_depth: float,
-    ) -> dict[str, NDArray[np.floating]]:
+    ) -> NDArray[np.floating]:
```
jacobcook1995, line 361: Huh, I'm really surprised that what was there before got through `pre-commit`!

jacobcook1995 (approved):
LGTM! (Honestly really surprised that that litter model typing ever passed CI)

.pre-commit-config.yaml
davidorme: No idea why that happened. Will check.

virtual_ecosystem/models/litter/chemistry.py
davidorme: I thought the same thing. I get the impression that the `numpy` mypy plugin was a little bit string and sealing wax and it is now more built in at the ground level.

virtual_ecosystem/models/hydrology/hydrology_model.py
```diff
@@ -419,13 +418,15 @@ def _update(self, time_index: int, **kwargs: Any) -> None:
...
-
-        days: int = 30
+        # Determine number of days
+        days_float: float = self.model_timing.update_interval_seconds / 86400
+        days: int = int(days_float // 1)
+
+        # Check if the number of days is exact and warn if not
+        if not np.allclose(days_float % 1, 0):
+            LOGGER.warning(
+                "Update interval is not a whole number of days ({n_days}) using floor."
```
vgro: What is floor?

vgro (approved):
LGTM!

Could you please delete the TODO in line 18 in hydrology_model.py

virtual_ecosystem/models/hydrology/hydrology_model.py
davidorme: Basically round down to nearest integer. The code used to use the `floor` function. I'll clarify.

PR #985, opened 2025-08-04
Release 0.1.1a12

jacobcook1995:
# Description

Bumping version and switching back to using a pypi version of pyrealm to allow a new release to be made

This also migrates the project to use:

*  Python 3.11+ to match `pyrealm` - that also follows SPEC0 recommendations for what to support.
* Poetry 2.1 in CI - we'd pinned on a really outdated version (1.5.1)
* Migrates `pyproject.toml` away from the old `poetry` specific structure and to the PEP621 compliant version (https://peps.python.org/pep-0621/).

However, we've run into the buffers on Windows - something wrong with the pathing on CI Runners:
https://github.com/abatilo/actions-poetry/issues/92


davidorme:
@alexdewar This sounds a little like the issue you found with the poetry install action - any suggestions?

davidorme (approved):
👍

PR #986, opened 2025-08-04
Animal-Plant piping class draft

### LINKED ISSUE #908: Animal - Animal-Plant Piping Update
The trophic dynamics of the animal model need to be directly coupled to the plant resources that are now available from the plant model.

TaranRallings:
# Description

There are two basic options that I'm exploring for making this work.

1) Aggregate by plant functional type into classes containing different resource types. (plant_resources_2.py)
2) Aggregate by resource type into classes containing different plant functional types. (plant_resources_3.py)

I suspect that the second approach will be more computationally effective. Most trophic relationships are done by resource type, not functional type. This makes it easier to sort through, say, a dictionary of fallen seeds in a grid cell to find the appropriate seeds than sorting through a list of plant functional types to see if each has seeds and then if you can even eat those specific seeds. 

Feedback I am looking for:
1) Which approach makes more sense to you?
2) Help sorting through the data format for plants. I am having some trouble figuring out what exactly the plant data looks like when I get access to it such that I can allocate it properly. The draft code I've provided is sort of half made up to give a sense of what I'm trying to do. 
`
   

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


davidorme:
> I suspect that the second approach will be more computationally effective. Most trophic relationships are done by resource type, not functional type. This makes it easier to sort through, say, a dictionary of fallen seeds in a grid cell to find the appropriate seeds than sorting through a list of plant functional types to see if each has seeds and then if you can even eat those specific seeds.

At the moment, the PlantsModel isn't exposing very much at the plant functional type level through `Data`.

* The two propagule arrays _do_ have a PFT dimension - they potentially have different sizes.
* `layer_leaf_mass` is aggregated across PFTs, although we could disaggregate by PFT
* The subcanopy vegetation does not have PFTs at the moment - so`subcanopy_vegetation_biomass` and `subcanopy_seedbank_biomass` are not structured that way.

The main worry I have right now is the stoichiometry. Each cohort will have stochiometric ratios. We can take the mass weighted average across resources to give "average" leaves/fruit - either across all PFTs like in the `layer_leaf_mass` or by PFTs like in the propagules. That's certainly the easiest solution! Ultimately though, animals might want to selectively feed on the most nutritious trees/fruit. That would force us to move to a per cohort resource rather than per PFT. The ugly thing there is that once you're at a cohort level, the data arrays are ragged which is all sorts of not very `xarray` but that may be something we just have to solve.

I don't have strong feelings about the two alternative implementations but I think your intuition about the integrated one is right. We're going to have to have that logic _somewhere_ and I think it is going to be cleaner to set up resources where there is an established mapping between the existing animal foraging requirements and the available resources.

Minor things:
* I don't think you have to handle trapping missing variable names. If they aren't defined, the PlantsModel is screwed up.
* IIRC, the values in the data layers are per cell not per m2, so you don't have to scale by cell area.

TaranRallings:
Ah interesting, so for now I can ignore PFTs for everything but propagules?

I am happy to do the easiest and simplest version right now and worry about more guided foraging patterns in the future. 

I will go with the resource-type integrated classes for now then.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1444,6 +1462,7 @@ def forage_community(self, dt: timedelta64) -> None:
             #  Build resource collections based on diet flags
             plant_list: list[Resource] = []
             prey_list: list[AnimalCohort] = []
+            # mushroom_list: list[Resource] = []
```
jacobcook1995, line 1465: why has this be added commented out?

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1491,6 +1509,7 @@ def forage_community(self, dt: timedelta64) -> None:
             cohort.forage_cohort(
                 plant_list=plant_list,
                 animal_list=prey_list,
+                mushroom_list=[],
```
jacobcook1995, line 1512: I think you would want to populate this based on `self.fungal_fruiting_bodies`

jacobcook1995 (commented):
Approach to fungal fruiting bodies looks sensible to me, I just had a couple of queries

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: (or something similar)

PR #988, opened 2025-08-06
959 sort out time interval in energy balance

### LINKED ISSUE #959: Sort out time interval in energy balance

vgro:
This PR replaced the placeholder '1's with the actual time interval in the microclimate update.

To make this work with the varying number of canopy layers, I also fixed how the above ground layer thickness is calculated, a few directions of fluxes, and adjusted the vertical mixing - basically all the steps where a value from below is called and NaN has to be skipped.

I also tidied the tests a bit to check if the values fall in a sensible range instead of specific values and introduced a step to mask the nan values in the canopy. The next PR will be testing with a grid cell that has no vegetation at all.

Fixes #959 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -381,7 +382,7 @@ def run_microclimate(
             ),
             dry_air_factor=abiotic_constants.dry_air_factor,
             cell_area=cell_area,
-            time_interval=1.0,  # TODO core_constants.seconds_to_hour,
+            time_interval=core_constants.seconds_to_hour,
```
jacobcook1995, line 405: What does using the time_interval as `seconds_to_hour` mean, does it mean you are always running it for 3600s?

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -302,27 +303,76 @@ def mix_and_ventilate(
...
+
+            if not np.isfinite(center_val):
+                continue
+
+            # ---- Upward flux ----
+            upper_i = i - 1
+            while upper_i >= 0 and not np.isfinite(input_variable[upper_i, j]):
+                upper_i -= 1
+
+            if upper_i >= 0 and np.isfinite(mixing_coefficient[upper_i, j]):
```
jacobcook1995: Should this be indented within the while block above? Because the while block above doesn't seem to do anything other than decrease `upper_i`

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -302,27 +303,76 @@ def mix_and_ventilate(
...
+                    flux_up = 0.0
+            else:
+                flux_up = 0.0
+
+            # ---- Downward flux ----
+            lower_i = i + 1
+            while lower_i < n_layers and not np.isfinite(input_variable[lower_i, j]):
+                lower_i += 1
+
+            if lower_i < n_layers and np.isfinite(mixing_coefficient[lower_i, j]):
```
jacobcook1995: Same question about the while block here

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -302,27 +303,76 @@ def mix_and_ventilate(
...
+
+        # ---- Ventilation for top layer ----
+        if not np.isfinite(input_variable[0, j]):
+            continue
+
+        lower_i = 1
+        while lower_i < n_layers and not np.isfinite(input_variable[lower_i, j]):
+            lower_i += 1
+
+        if (
```
jacobcook1995: Same while block question

jacobcook1995 (approved):
Had a few queries, but the general structure looks sensible to me

virtual_ecosystem/models/abiotic/microclimate.py
vgro: yes, the update interval should always be 3600 seconds independent of the VE time interval. There is a conversion before the loop starts

virtual_ecosystem/models/abiotic/wind.py
vgro: Good question, but no, the block should not be indented. The purpose of the `while` loop is only to find the nearest valid layer above/below the current one. Once it finds a finite value (or exits when it can't), the code that follows uses that upper_i value to calculate the flux.

From what I understand, if I indented the `if` block inside the `while`, it would try to calculate the flux on every invalid layer on the way up which is what we want to avoid.not what we want. You only want to calculate it once you've found a valid upper_i, or skip it if none is found.

I added comments inline and also added an explanation in the function docstring, I hope that makes sense

jacobcook1995: Ahh right, I think it might be better to find these indices using `numpy` if possible. Generally loops in python aren't very efficient, so doing stuff in vectorised form is better (though I don't think it's a massive deal in this case).

I *think* this loop can be replaced with

```upper_i  = np.argmax(np.isfinite(input_variable[:, j]))```

which finds the first finite index in the array `input_variable[:, j]`. The lower two loops seem a bit more tricky

Comment (davidorme) [reply virtual_ecosystem/models/abiotic/wind.py]: I think the looping over cells can be replaced with array-based calculation - and I think that should be much faster, which will be critical here because it is having to loop at such a short time scale. I'm seeing if I can figure it out but I've added another couple of questions.

File: virtual_ecosystem/models/abiotic/wind.py [hunk 5]
```diff
@@ -302,27 +313,81 @@ def mix_and_ventilate(
...
+
+            # ---- Upward flux ----
+            # Look for nearest valid input value above the current layer
+            upper_i = i - 1
+            while upper_i >= 0 and not np.isfinite(input_variable[upper_i, j]):
+                upper_i -= 1
+
+            # If a valid upper value and a mixing coefficient are found, compute flux
+            if upper_i >= 0 and np.isfinite(mixing_coefficient[upper_i, j]):
+                delta_up = input_variable[upper_i, j] - center_val
```
davidorme: Should one of `delta_up` or `delta_down` be reversed? Assuming a positive or negative gradient through the canopy, one of the `delta` values is going to yield negative values and then gets clipped by `flux_up < 0`

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -302,27 +313,81 @@ def mix_and_ventilate(
...
+            delta_vent = input_variable[0, j] - input_variable[lower_i, j]
+            change = ventilation_rate[j] * delta_vent * time_interval
+
+            # Limit ventilation to avoid negative top  TODO check consistency
+            max_change = 0.1 * abs(input_variable[0, j])
+            change = np.clip(change, -max_change, max_change)
+
+            # Prevent ventilation that would reduce the top value TODO check consistency
+            if change < 0:
+                change = 0.0
```
davidorme: This is effectively 
```suggestion
            max_change = 0.1 * abs(input_variable[0, j])
            change = np.clip(change, 0, max_change)
```

davidorme (changes requested):

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -302,27 +313,81 @@ def mix_and_ventilate(
...
+        lower_i = 1
+        while lower_i < n_layers and not np.isfinite(input_variable[lower_i, j]):
+            lower_i += 1
+
+        if (
+            lower_i < n_layers
+            and np.isfinite(input_variable[lower_i, j])
+            and np.isfinite(ventilation_rate[j])
+        ):
+            delta_vent = input_variable[0, j] - input_variable[lower_i, j]
```
davidorme: Is this intentionally calculating the difference from the original rather than the mixed value?

virtual_ecosystem/models/abiotic/wind.py
```diff
@@ -302,27 +313,81 @@ def mix_and_ventilate(
...
+                    mixing_coefficient[lower_i, j] * delta_down / layer_thickness[i, j]
+                )
+                # Prevent flux that would reduce center_val TODO check consistency
+                if flux_down < 0:
+                    flux_down = 0.0
+            else:
+                flux_down = 0.0
+
+            # ---- Update value ----
+            change = (flux_up + flux_down) * time_interval / layer_thickness[i, j]
```
davidorme: Is this a double correction for layer thickness? The fluxes have already been divided through by thickness.

virtual_ecosystem/models/abiotic/wind.py
davidorme: @vgro I've pushed a commit with an array-based version of the `mix_and_ventilate` function. It passes the test.

The array calculations do make some assumptions about the canopy structure - namely that within a column (cell) there is always a value at the top (above canopy, index=0) and at the bottom (surface, index=-1) and that if there are canopy layers then they will always be stacked down from the second layer (index=1) and any missing values will be contiguous np.nan values between the bottom layer of the canopy and the surface layer. If those assumptions _are_ broken then the canopy model is wildly broken, so it seems reasonable.

vgro: Thanks @davidorme , I have made a few adjustments based on that and your other comments

virtual_ecosystem/models/abiotic/wind.py
vgro: I removed the clipping step to allow for flows in both directions

virtual_ecosystem/models/abiotic/wind.py
vgro: no, that was a mistake. Fixed now :-)

vgro:
I have made the suggested changes and added an extra step to ensure that the values stay within realistic limits (especially important for atmospheric humidity). I tried to vectorize where possible but couldn't find a way around the last for loop.

davidorme (changes requested):
This looks good. I did have some difficulty getting my head around the new bit to redistribute values outside limits. I think that's partly because that process is embedded within `mix_and_ventilate` so it's hard to visualise/troubleshoot.

I've put in a PR (#1011) that breaks this part out into its own standalone clamping function and tests it separately - I can't get rid of the loop, but I think it can be calculated more simply. This feels like a better unit structure for the functionality. If that PR looks good to you, then I think this is ready to go?

vgro:
> This looks good. I did have some difficulty getting my head around the new bit to redistribute values outside limits. I think that's partly because that process is embedded within `mix_and_ventilate` so it's hard to visualise/troubleshoot.
> 
> I've put in a PR (#1011) that breaks this part out into its own standalone clamping function and tests it separately - I can't get rid of the loop, but I think it can be calculated more simply. This feels like a better unit structure for the functionality. If that PR looks good to you, then I think this is ready to go?

Looks good to me, I just merged it. Could you please approve this PR when you're happy with it?

davidorme (approved):
👍

PR #990, opened 2025-08-07
989 foraging methods refactor to reduce resuse

### LINKED ISSUE #989: Foraging methods refactor to reduce resuse
There is significant redundancy in the foraging system whereby the foraging methods for non-animal resources are reused while being almost identical. This was tolerable with few such resources but as I am expanding to soils, mushrooms, seeds, etc it is blowing up. This is particularly the case with the `delta_mass` and `calculate consumed_mass` methods. The goal is to streamline into a reusable core structure for these foraging methods. 

This issue is so that I behave responsibly and make this a SMALL PR ahead of the piping expansion instead of making you poor souls review a massive PR with 18 different changes.

TaranRallings:
# Description

This is a small refactor to prepare the foraging system for all of the new resource pools I am introducing. It collapses the many identical `delta_mass` and `calculate_consumed_mass` methods into something more DRY. 

Predation is unchanged.

Non-predation consumption now keys off two core methods:
- `default_consumed_resource_mass`
- `forage_resource_list`

Small, resource-specific, methods that serve to call `forage_resource_list` are implemented to aid in bug hunting and facilitate later expansions. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -931,213 +932,106 @@ def delta_mass_predation(
...
-            The mass to be consumed from the target plant by the herbivore (in kg).
+            Mass (kg) to consume from target.
         """
-        F = self.F_i_k(plant_list, target_plant)  # Adjusting this call as necessary
-
-        consumed_mass = target_plant.mass_current * (1 - exp(-(F * adjusted_dt)))
-        return consumed_mass
+        F = self.F_i_k(resource_list, target)
+        consumed = target.mass_current * (1.0 - exp(-F * adjusted_dt))
+        return max(consumed, 0.0)
```
dalonsoa: If `F` is positive (I assume `adjusted_dt` is positive), then `consumed` can not be negative, so the `max(consumed, 0.0)` bit is not really necessary:

```suggestion
        return target.mass_current * (1.0 - exp(-F * adjusted_dt))
```

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -931,213 +932,106 @@ def delta_mass_predation(
...
-                )
+            conv_eff = self.functional_group.conversion_efficiency
+            for elem in total_gain:
+                total_gain[elem] += gain_cnp[elem] * conv_eff
 
-            # Add the litter to the appropriate herbivory waste pool
-            herbivory_waste_pools[plant.cell_id].add_waste(plant_litter_cnp)
+            if herbivory_waste_pools and litter_cnp:
+                if resource.cell_id not in herbivory_waste_pools:
+                    raise KeyError(f"Missing waste pool for cell {resource.cell_id}")
```
dalonsoa: Why or when could this happen?

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -931,213 +932,106 @@ def delta_mass_predation(
 
         return total_consumed_mass
 
-    def calculate_consumed_mass_herbivory(
+    def default_consumed_resource_mass(
```
dalonsoa: I'm not sure if there's going to be ever a consumed resource mass other than this one, but if there is not and the only point here is to re-use this same code by other methods of the class, the usual approach is to just make it "private" and remove the `default`. Otherwise, it will be understood that there is an option for not being default:

```suggestion
    def _consumed_resource_mass(
```

dalonsoa (approved):
Just a couple of small comments, but otherwise it looks good, so approving.

virtual_ecosystem/models/animal/animal_cohorts.py
TaranRallings: I am probably overdoing it here but in general this isn't me anticipating a specific problem as much as it is laying a groundwork to help the inevitable long road of debugging.

virtual_ecosystem/models/animal/animal_cohorts.py
TaranRallings: Ah, good to know, thanks!

PR #995, opened 2025-08-13
Implements disturbance model prototype

dalonsoa:
# Description

Implements an initial proof of concept version of the disturbance framework. To keep it as general as unintrusive as possible, as well as benefitting from the whole infrastructure already in place to discover an initialise models, it has been implemented as a `DisturbanceBase` model, inheriting from `BaseModel`, and just imposing some restrictions/limitations. 

As a sort of summary, disturbance models (those inheriting from `DisturbanceBase`):

- Cannot be `static`: By definition, disturbances are dynamic, so it does not make much sense to make them static.
- Cannot initialise variables during init or first update: That's what normal models do - disturbances only disrupt what's already there. This is also important to define the execution order since other models should not depend on variables initialised by the disturbances.
- New `disturbed_models` class attribute: Take the list of models as input in the `from_config` method, so they can fetch those models that, potentially, they will be directly affecting, if any.

A `DisturbanceTesting` model has been added and included in the tests, using the existing `Testing` model as the model to disturb (even though it does nothing).

**NOTE**: While adding this functionality I've discovered and fixed a few bugs (37317734 and  28228f1e). If this PR is not merged, those commits will need to be incorporated added as a separate PR.

Fixes # (issue)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/core/test_config.py
+++ b/tests/core/test_config.py
@@ -896,8 +896,10 @@ def test__resolve_config_paths_file_locations(
             },
             does_not_raise(),
             {
-                "file1_path": str(Path("path/to/config/file.txt").absolute()),
-                "other_path": str(Path("path/to/config/file2.txt").absolute()),
+                "file1_path": str(Path("path/to/config/file.txt").resolve().absolute()),
+                "other_path": str(
+                    Path("path/to/config/file2.txt").resolve().absolute()
+                ),
                 "foo": "bar",
                 "baz": 6,
             },
@@ -913,8 +915,8 @@ def test__resolve_config_paths_file_locations(
             },
             does_not_raise(),
             {
-                "file1_path": str(Path("path/to/config/file.txt").absolute()),
-                "other_path": str(Path("path/file2.txt").absolute()),
+                "file1_path": str(Path("path/to/config/file.txt").resolve().absolute()),
+                "other_path": str(Path("path/file2.txt").resolve().absolute()),
                 "foo": "bar",
                 "baz": 6,
             },
@@ -929,9 +931,9 @@ def test__resolve_config_paths_file_locations(
             },
             does_not_raise(),
             {
-                "file1_path": str(Path("path/to/config/file.txt").absolute()),
+                "file1_path": str(Path("path/to/config/file.txt").resolve().absolute()),
                 "nested": {
-                    "other_path": str(Path("path/file2.txt").absolute()),
+                    "other_path": str(Path("path/file2.txt").resolve().absolute()),
                     "foo": "bar",
                 },
                 "baz": 6,
--- a/virtual_ecosystem/core/base_model.py
+++ b/virtual_ecosystem/core/base_model.py
@@ -760,3 +760,77 @@ class BaseModel(ABC):
             )
             LOGGER.error(error)
             raise error
+
+
+class BaseDisturbance(
+    BaseModel,
+    ABC,
+    model_name="_base_disturbance",
+    model_update_bounds=("1 day", "1 year"),
+    vars_required_for_init=tuple(),
+    vars_populated_by_init=tuple(),
+    vars_required_for_update=tuple(),
+    vars_updated=tuple(),
+    vars_populated_by_first_update=tuple(),
+):
+    """Base class for disturbances in the virtual ecosystem.
+
+    This class extends BaseModel to include disturbance-specific functionality and
+    configuration.
+    """
+
+    def __init_subclass__(
+        cls,
+        model_name,
+        model_update_bounds,
+        vars_required_for_init,
+        vars_updated,
+        vars_required_for_update,
+        vars_populated_by_init,
+        vars_populated_by_first_update,
+    ):
+        """Initialise subclasses deriving from BaseDisturbance."""
+        if vars_populated_by_first_update or vars_populated_by_init:
+            raise ValueError(
+                "Disturbance models cannot define "
+                "vars_populated_by_first_update or vars_populated_by_init."
+            )
+        super().__init_subclass__(
+            model_name,
+            model_update_bounds,
+            vars_required_for_init,
+            vars_updated,
+            vars_required_for_update,
+            tuple(),
+            tuple(),
+        )
+
+    def __init__(
+        self,
+        data: Data,
+        core_components: CoreComponents,
+        static: bool = False,
+        models: dict[str, BaseModel] | None = None,
+        **kwargs: Any,
+    ):
+        """Initialise the disturbance with shared data and core components.
+
+        Args:
+            data: A :class:`~virtual_ecosystem.core.data.Data` instance containing
+                variables to be used in the disturbance.
+            core_components: A
+                :class:`~virtual_ecosystem.core.core_components.CoreComponents`
+                instance containing shared core elements used throughout models.
+            static: A boolean flag indicating if the disturbance is static.
+            models: A dictionary of model instances that the disturbance will affect
+                directly, if any.
+            **kwargs: Further arguments to the disturbance.
+        """
+        if static:
+            raise ValueError(
+                "Disturbance models cannot be static. Please, use a non-static "
+                "disturbance model."
+            )
+        super().__init__(data, core_components, static, **kwargs)
+        self.models: dict[str, BaseModel] = models or {}
+        """A dictionary of model instances that the disturbance will affect directly."""
--- a/virtual_ecosystem/core/base_model.py
+++ b/virtual_ecosystem/core/base_model.py
@@ -779,6 +779,13 @@ class BaseDisturbance(
     configuration.
     """
 
+    disturbed_models: tuple[str, ...] = tuple()
+    """A tuple of model names that this disturbance will affect directly.
+
+    This is used to ensure that the disturbance is applied to the correct models and
+    to validate the disturbance configuration.
+    """
+
     def __init_subclass__(
         cls,
         model_name,
@@ -788,6 +795,7 @@ class BaseDisturbance(
         vars_required_for_update,
         vars_populated_by_init,
         vars_populated_by_first_update,
+        disturbed_models: tuple[str, ...] = tuple(),
     ):
         """Initialise subclasses deriving from BaseDisturbance."""
         if vars_populated_by_first_update or vars_populated_by_init:
@@ -804,6 +812,7 @@ class BaseDisturbance(
             tuple(),
             tuple(),
         )
+        cls.disturbed_models = disturbed_models
 
     def __init__(
         self,
@@ -814,6 +823,8 @@ class BaseDisturbance(
         **kwargs: Any,
     ):
         """Initialise the disturbance with shared data and core components.
+
+        This method **must** be called in the ``__init__`` method of all subclasses.
 
         Args:
             data: A :class:`~virtual_ecosystem.core.data.Data` instance containing
@@ -828,9 +839,30 @@ class BaseDisturbance(
         """
         if static:
             raise ValueError(
-                "Disturbance models cannot be static. Please, use a non-static "
-                "disturbance model."
+                "Disturbance models cannot be static. Please, set the static flag "
+                "to False."
             )
         super().__init__(data, core_components, static, **kwargs)
-        self.models: dict[str, BaseModel] = models or {}
+
+        models = models or {}
+        if not set(models.keys()).issubset(self.disturbed_models):
+            raise ValueError(
+                f"Disturbance model {self.model_name} is configured to affect "
+                f"{self.disturbed_models}, but the provided models are "
+                f"{list(models.keys())}."
+            )
+        self.models: dict[str, BaseModel] = {
+            name: models[name] for name in self.disturbed_models
+        }
         """A dictionary of model instances that the disturbance will affect directly."""
+
+    @classmethod
+    @abstractmethod
+    def from_config(
+        cls,
+        data: Data,
+        core_components: CoreComponents,
+        config: Config,
+        models: dict[str, BaseModel] | None = None,
+    ) -> BaseModel:
+        """Factory function to unpack config and initialise a model instance."""
--- a/virtual_ecosystem/core/base_model.py
+++ b/virtual_ecosystem/core/base_model.py
@@ -809,8 +809,8 @@ class BaseDisturbance(
             vars_required_for_init,
             vars_updated,
             vars_required_for_update,
-            tuple(),
-            tuple(),
+            vars_populated_by_init=(),
+            vars_populated_by_first_update=(),
         )
         cls.disturbed_models = disturbed_models
 
--- a/virtual_ecosystem/main.py
+++ b/virtual_ecosystem/main.py
@@ -47,16 +47,37 @@ def initialise_models(
 
     Raises:
         InitialisationError: If one or more models cannot be properly configured
+
+    Returns:
+        A dictionary of configured models, keyed by model name.
     """
+    from .core.base_model import BaseDisturbance
 
     LOGGER.info("Initialising models: {}".format(",".join(models.keys())))
 
     # Use factory methods to configure the desired models
     failed_models = []
     models_cfd = {}
+    disturbance_models: list[tuple[str, type[BaseDisturbance]]] = []
+
+    # Loop through the models and try to configure them
+    # If a model is a disturbance model, it will be configured later
     for model_name, model_class in models.items():
+        if issubclass(model_class, BaseDisturbance):
+            disturbance_models.append((model_name, model_class))
+            continue
         try:
             this_model = model_class.from_config(data, core_components, config)
+            models_cfd[model_name] = this_model
+        except (InitialisationError, ConfigurationError):
+            failed_models.append(model_name)
+
+    # If there are disturbance models, configure them now
+    for model_name, model_class in disturbance_models:
+        try:
+            this_model = model_class.from_config(
+                data, core_components, config, models_cfd
+            )
             models_cfd[model_name] = this_model
         except (InitialisationError, ConfigurationError):
             failed_models.append(model_name)
--- a/virtual_ecosystem/main.py
+++ b/virtual_ecosystem/main.py
@@ -172,6 +172,16 @@ def ve_run(
         for model_name in variables.get_model_order("init")
     }
 
+    # Add models not handled by the variable system, like those that do not init
+    # any variables or require any variables to be set up, eg. the testing model.
+    init_sequence.update(
+        {
+            model_name: model_class
+            for model_name, model_class in config.model_classes.items()
+            if model_name not in init_sequence
+        }
+    )
+
     models_init = initialise_models(
         config=config,
         data=data,
--- a/virtual_ecosystem/main.py
+++ b/virtual_ecosystem/main.py
@@ -116,6 +116,7 @@ def ve_run(
         progress: A Progress enum instance setting the level of output to be printed to
             the console when ve_run is running.
     """
+    from .core.base_model import BaseDisturbance
 
     # Mute the progress information when the log is written to stdout.
     if logfile is None:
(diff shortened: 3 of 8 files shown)


virtual_ecosystem/models/disturbance_testing/disturbance_testing_model.py
```diff
@@ -0,0 +1,91 @@
...
+        data: Data,
+        core_components: CoreComponents,
+        static: bool = False,
+        **kwargs: Any,
+    ):
+        """Plants init function.
+
+        The init function is used only to define class attributes. Any logic should be
+        handled in :fun:`~virtual_ecosystem.plants.plants_model._setup`.
+        """
```
jacobcook1995, line 45: Guess this bit should change so it doesn't refer to the Plants model (though I guess this won't be in the api documentation)

virtual_ecosystem/core/base_model.py
```diff
@@ -760,3 +760,109 @@ class attribute are present in the :attr:`~virtual_ecosystem.core.data.Data`
...
+    This class extends BaseModel to include disturbance-specific functionality and
+    configuration.
+    """
+
+    disturbed_models: tuple[str, ...] = tuple()
+    """A tuple of model names that this disturbance will affect directly.
+
+    This is used to ensure that the disturbance is applied to the correct models and
+    to validate the disturbance configuration.
+    """
```
jacobcook1995, line 787: I guess disturbances could affect properties of models directly (i.e. by changing the properties of the plant communities recorded in the plants model), but also by changing the values in the data object (i.e. by changing amount of  labile inorganic phosphorus in the soil). In the latter case what information would we be looking to record here? It would be a disturbance of the soil model (which updates the soil nutrient availabilities), but would have a pretty immediate impact on the plants. (and then there would be downstream inputs on herbivory etc)

jacobcook1995 (commented):
The general approach seems sensible to me. Just had a query about how we would define "disturbed models"

virtual_ecosystem/models/disturbance_testing/disturbance_testing_model.py
dalonsoa: This is copypasting the testing model, so wrong there, wrong here... I'll fix both.

virtual_ecosystem/core/base_model.py
dalonsoa: Here you should include only the models that the disturbance model will need direct access to - in your example, soil. It is not like the variables attributes, which are somewhat just informative (a model could modify any variable in the data object). Here, the models in `disturbed_models` are the ones that would need to be modified by the disturbance. If that modification has knock on effects somewhere else, that's for the normal flow of the simulation to handle. If the disturbance directly affects soil and plants (eg. by modifying their internal state in both cases, so the changes are synchronised) then both models will need to be listed here.

jacobcook1995: That makes sense. Would that mean if a model doesn't have an internal state that is worth modifying (i.e. the soil and litter models write pretty much all relevant variables to the data object), then it would never have to be listed in `disturbed_models`?

jacobcook1995: I guess that does make sense, because in one case you need the disturbance model to change the behaviour of the disturbed model, but in the other the disturbance model is just another model which effects the same variable

dalonsoa: > That makes sense. Would that mean if a model doesn't have an internal state that is worth modifying (i.e. the soil and litter models write pretty much all relevant variables to the data object), then it would never have to be listed in `disturbed_models`?

Very likely - there's nothing to disturb in the model, after all. For simple disturbances, disturbance models are pretty much the same as any other model: they just update some variables in the data object in some way.

vgro (approved):
I agree with Jacob that the general structure makes sense if this is just another model that changes something in the data object. It is not 100% clear to me how any processes would be changed in models, or if that is an option/needed at all?

Would each type of disturbance have a separate model, like `FireModel`, `HuntingModel`, etc, or would this include all sorts of different disturbances and the specific one is selected as an input?

dalonsoa:
> I agree with Jacob that the general structure makes sense if this is just another model that changes something in the data object. It is not 100% clear to me how any processes would be changed in models, or if that is an option/needed at all?

In theory, modifying the models themselves was supposed to be an option, even multiple models at once, as far I understood the discussion. If you're just creating a disturbance by modifying the data object, you just need another, normal model doing that. There's no need to have anything else added. 

> Would each type of disturbance have a separate model, like `FireModel`, `HuntingModel`, etc, or would this include all sorts of different disturbances and the specific one is selected as an input?

Yes, each disturbance should have its own model. If disturbances are meant to be interrelated or had to run in a specific order, then things might get a bit messy as that option is not implemented in this PR, but if they are independent, then this is fine as it is.

dalonsoa:
A minimal example on how to change a process from within a disturbance (very minimal, and not using this PR code) would be:

```python
class A:
    def __init__(self, x, y):
        self.x = x
        self.y = y

    def operation(self):
        return self.x + self.y


class Disturbance:
    def __init__(self, models, operation):
        self.models = models
        self.operation = operation

    def update(self):
        self.models["A"].operation = lambda: self.operation(
            self.models["A"].x, self.models["A"].y
        )


if __name__ == "__main__":
    # Example usage
    a_instance = A(5, 10)
    models = {"A": a_instance}

    disturbance = Disturbance(models, operation=lambda x, y: x * y)
    print(a_instance.operation())  # Output will be 15 (5 + 10)
    disturbance.update()
    print(a_instance.operation())  # Output will be 50 (5 * 10)
    a_instance.x = 2
    print(a_instance.operation())  # Output will be 20 (2 * 10)
```

So, after applying the disturbance, the `A` model does something different when calling its `operation` method.

tests/core/test_config.py
```diff
@@ -896,8 +896,10 @@ def test__resolve_config_paths_file_locations(
             },
             does_not_raise(),
             {
-                "file1_path": str(Path("path/to/config/file.txt").absolute()),
-                "other_path": str(Path("path/to/config/file2.txt").absolute()),
+                "file1_path": str(Path("path/to/config/file.txt").resolve().absolute()),
+                "other_path": str(
+                    Path("path/to/config/file2.txt").resolve().absolute()
```
alexdewar, line 901: I had to look this up [in the docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve), but it seems like `resolve()` already makes paths absolute, so the extra `absolute()` isn't needed.

It seems like this pattern is used in other places, so might be worth doing a global find-and-replace.

virtual_ecosystem/core/base_model.py
```diff
@@ -760,3 +760,109 @@ class attribute are present in the :attr:`~virtual_ecosystem.core.data.Data`
...
+        self,
+        data: Data,
+        core_components: CoreComponents,
+        static: bool = False,
+        models: dict[str, BaseModel] | None = None,
+        **kwargs: Any,
+    ):
+        """Initialise the disturbance with shared data and core components.
+
+        This method **must** be called in the ``__init__`` method of all subclasses.
```
alexdewar, line 827: Are double backticks needed? Would this work?

```suggestion
        This method **must** be called in the `__init__` method of all subclasses.
```

virtual_ecosystem/core/base_model.py
```diff
@@ -760,3 +760,109 @@ class attribute are present in the :attr:`~virtual_ecosystem.core.data.Data`
...
+                "vars_populated_by_first_update or vars_populated_by_init."
+            )
+        super().__init_subclass__(
+            model_name,
+            model_update_bounds,
+            vars_required_for_init,
+            vars_updated,
+            vars_required_for_update,
+            tuple(),
+            tuple(),
```
alexdewar: For readability:

```suggestion
            vars_populated_by_init=(),
            vars_populated_by_first_update=(),
```

virtual_ecosystem/main.py
```diff
@@ -194,6 +226,29 @@ def ve_run(
...
+    # Ensure that all disturbance models are at the end of the update sequence
+    models_update = {
+        model_name: model_instance
+        for model_name, model_instance in models_update.items()
+        if not isinstance(model_instance, BaseDisturbance)
+    } | {
+        model_name: model_instance
+        for model_name, model_instance in models_update.items()
+        if isinstance(model_instance, BaseDisturbance)
+    }
```
alexdewar, line 250: This works. I guess another way to do it would be to sort `models_update` upfront so all the `BaseDisturbance` models are at the end.

virtual_ecosystem/models/disturbance_testing/disturbance_testing_model.py
```diff
@@ -0,0 +1,91 @@
...
+
+class DisturbanceTesting(
+    BaseDisturbance,
+    model_name="disturbance_testing",
+    model_update_bounds=("1 day", "1 year"),
+    vars_required_for_init=tuple(),
+    vars_populated_by_init=tuple(),
+    vars_required_for_update=tuple(),
+    vars_updated=tuple(),
+    vars_populated_by_first_update=tuple(),
```
alexdewar, line 22: I'd probs just use `()` for an empty tuple, but it's not important

alexdewar (approved):
LGTM!

It might be nice to have a minimal example beyond the test model that actually disturbs something, but that's more of a documentation thing and probs out of scope for this PR.

virtual_ecosystem/core/base_model.py
dalonsoa: This is copy paste from the `testing` model, so I assume it works as it is. If it is not, we need to change it in multiple places.

virtual_ecosystem/main.py
dalonsoa: I would put a big chunk of this code in a separate function, to be honest. I've just made something quickly that works, but there's room for improvement.

virtual_ecosystem/models/disturbance_testing/disturbance_testing_model.py
dalonsoa: I've used the same convention than in the testing model - copypaste and rename, so would leave it as it is.

tests/core/test_config.py
dalonsoa: Let's leave it for now and open an issue to change it wherever relevant, as it is independent on this PR.

TaranRallings (approved):
This LGTM and the macro questions I had were already asked.

virtual_ecosystem/core/base_model.py
davidorme: It is RST vs Markdown. At some point I'd like to move the docstrings over to markdown but I don't think there's a good replacement for `sphinx` `autodoc` yet, which can only handle RST docstrings.

virtual_ecosystem/models/disturbance_testing/disturbance_testing_model.py
```diff
@@ -0,0 +1,91 @@
...
+        return inst
+
+    def _setup(self, **kwargs: Any) -> None:
+        """Placeholder function to setup the testing model."""
+
+    def spinup(self) -> None:
+        """Placeholder function to spin up the testing model."""
+
+    def _update(self, time_index: int, **kwargs: Any) -> None:
+        """Placeholder function to update up the tesying model."""
```
davidorme, line 88:
```suggestion
        """Placeholder function to update up the testing model."""
```

Typo - presumably again from the testing model 😄

davidorme (changes requested):
This looks good, but honestly I'm not convinced that subclassing `BaseModel` and mixing disturbances in with the other models is the way to go. 

1. There are a lot of abstract methods in the `BaseModel` that are 'real' model specific. 
    * That's a lot of clutter on the `DisturbanceModel` and we then we need to check that specific 'real' model only attributes - like variable lists and `static` aren't misdefined.  
    * We then need to put in code in the `ve_run` function to ensure that the models get separated out and then put back together in a disturbance last order.
    * I also strongly suspect that we are going to want to add disturbance specific model attributes (we already have one, but there might be attributes on the timing of disturbance execution).
2. I think there are also configuration reasons not to want to lump disturbances in with 'real' models. We could have a config set up to run a model and then want to rerun it with disturbances. If we have a separate disturbance configuration then we can just drop in a new config file to the same initial setup; as it stands this would have to be configured by providing different model definitions.

I think we can reuse a huge amount of the `BaseModel` code, but I just feel that making the disturbances part of the real model dispatch is creating a tension that will come back to bite us and we'd be better off making a clear separation. There's a real cost of complexity there - we need a new `DisturbanceModel` class and registry etc -  but it does simplify the way the disturbance models integrate into execution and it gives us more flexibility.

I think I'm arguing for something that launches off a TOML config like:

```toml
[disturbances.logging]
run_at = [50, 100]
[disturbances.hunting]
run_every = [1]
```

Thoughts?

dalonsoa:
I'm afraid I don't follow your reasoning, @davidorme . Reusing the `BaseModel` as the base for the `DisturbanceModel` gives us all the ecosystem for model discovery, validation and definition. Disturbance models can be as simple or complex as needed, and certainly can have their own variables, just like any other model (I might need to make a small change in the current implementation). Granted that there are things in the `BaseModel` that we do not need, but I don't see why that is a problem.

Maybe it will help me to see your point want want you really need if you give me a couple of realistic use cases where the implementation I suggest is not fit for purpose and an alternative version would be better.

davidorme:
I think the basic operation of a `Disturbance` is that it should be able to dive into the model at a point in time (or possibly repeated points in time) and alter the model data. So it needs access to the data and to the models (because not all model state is in the data - `plants` and `animals` being the culprits here). In configuring a set of disturbances, users should say:

* which disturbance is happening,
* when each disturbance happens - probably using some kind of common API for defining specific model time step or some pattern of repeated disturbance,
* what order they occur in. There isn't a natural order for these, so with multiple disturbances happening in a single time step we probably need both fixed order (sequence as in the config) or randomised. This isn't something that `BaseModels` can do.

It's true that there is some overlap with what `BaseModel` subclasses do, but I think we underestimate how different a model is from what a disturbance should do. I've had a scan through the `BaseModel` attributes and methods:

**Applicable to `Disturbance`**

I think the following are required to run a disturbance.

`__init__`
`model_name` 
`_check_model_name`
`from_config`
`vars_updated` 
`vars_required_for_update` 
`update`
`_check_variables_attribute`

I might argue to rename `update` to `disturb` etc, but that isn't a good enough reason to need a new class :smile:

**Not applicable**

Disturbances do not have update bounds or a speed - they act at a point in time - so the following attributes and methods don't apply:

`model_update_bounds`
`_check_model_update_bounds`
`_check_update_speed'`

Disturbances do not initialise or populate variables, only ~disturb~ update.

`vars_required_for_init`
`vars_populated_by_init`
`vars_populated_by_first_update`
`check_init_data`

Disturbances do not have a static mode:

`_check_static_config`
`_bypass_setup_due_to_static_configuration`
`_run_update_due_to_static_configuration`

Several method steps are not required (or at least are only required because of the redundant need for a static mode):

`_setup` : Disturbances are never static so we don't need to isolate `__init__` functionality within `__setup__`
`_update` : Again, disturbances are not static so we can just use define update directly
`spinup`: Disturbances will never spin up.
`cleanup` : To be fair, nothing currently uses `cleanup` but disturbances would not.

I'd also argue that `__init_subclass__` is dubious - about half of the `BaseModel` checking is of things that aren't relevant to disturbances. 

I also think that configuration differences could make life very confusing. For example, no `BaseModel` would accept a `run_at_step=5` option but any disturbance could. Equally a `Disturbance` should never accept a `update_speed` configuration. Of course, the models all have different configuration options, but I think merging the _common_ configuration options of `BaseModels` with the (currently poorly defined) common options of `Disturbances` is likely confusing.

That all seems like a lot of baggage to carry around and I don't think it makes for an easy developer experience in making new disturbances (and these are likely to be much more bespoke and varied than models). I think a slimmed down API (compared to BaseModel) and a separation of the process make it easier to maintain and write.

TaranRallings:
I am at that difficult crossroads where both positions look fairly compelling. I think I am suffering from a lack of concrete examples. 

Another point to consider may be what relationship we want this disturbance class to have with management. Do we want management to be a different class, facilitating feedbacks, that implements these disturbance tools in a responsive way or something else? Or can management be integrated within either of these proposed disturbance frameworks?

vgro:
I agree with Taran here, both options seem reasonable. I guess the choice comes down to clarity vs. consistency / short-term simplicity vs. long-term extensibility. I’d lean slightly towards Diego's option for now, assuming that we can set the timing and so on in a similar easy way to the constants in the config for example, but I’m happy to go with the consensus.

davidorme:
I've been thinking about the timing side of things and I think there's a problem with the `model_update_bounds`. If we piggyback `BaseModel` then we have to set something in `model_update_bounds` on a `BaseDisturbance` instance. But because disturbances act at a point in time - which we'll need to control through a separate `DisturbanceTiming` system - they can act potentially act at very wide intervals - like logging could be monthly or every 100 years. 

So we have to put very wide bounds on `model_update_bounds` or we can potentially get a disturbance model refusing to run because of a selected model tick length, even though that tick length isn't relevant to the whether the disturbance should act because that is scheduled differently.

I'm just really concerned that in our guide to creating disturbances, we are going to have to explicitly say - "define a model as in the model guide  but all of these methods/attributes have to be present but must be empty, null, set to this arbitrary wide value".

dalonsoa:
We can - and probably should - nullify in `DisturbanceBase` all methods of `BaseModel` that are not relevant, so the end user does not need to worry about them. And those do not behave exactly as we need, we can override. Benefiting from the `BaseModel` ecosystem in place does not mean passing the user the need to handle a lot of irrelevant stuff. That's what `DisturbanceBase` is for.

davidorme:
I'm sorry but I don't get why this isn't a code smell - I see that there is a lot of the initialisation code that is similar but we are creating an inheritance structure where we are actively nulling and overriding two thirds of the instance attributes and methods of the BaseModel.  And 50% of the class attributes - which we cannot hide or nullify - have to be populated with meaningless placeholders. 

https://luzkan.github.io/smells/refused-bequest

dalonsoa:
It is a code smell.

And I think it points to where we should go, but you’re not going to like it. There are a lot of commonalities - in particular the discovery and initialisation part - but also a lot of things that should not be there. If we don’t want to re-write a lot of code, I think we are going to need an important refactoring of the code base. Maybe creating a BaseCommon class that contains the common functionality, and from where BaseModel and BaseDisturbance inherit. Maybe something else.

I also think that, based on the discussion, the api that you need for the disturbance, specially in terms of times, is far from clear. I think you need come up with a concrete set of use cases from where the right implementation can be defined.

sallymatson:
I'm worried that using this approach will lead to quite a confusing user experience down the line and prevent other developers/modellers from using the disturbance class. From our team discussions I feel like we've agreed that this is an important feature, and also is one that is quite critical to making the model useful for varied applications and end-users. 

Correct me if I'm wrong but it seems like the benefit of inheritance here is more on the backend, but the drawbacks are on the frontend, and I guess I'd lean toward making life easier for users than for ourselves..... Although, (as of now at least ahah) I'm not writing this code so easier said than done, of course !

sallymatson:
> It is a code smell. And I think it points to where we should go, but you’re not going to like it. There are a lot of commonalities - in particular the discovery and initialisation part - but also a lot of things that should not be there. If we don’t want to re-write a lot of code, I think we are going to need an important refactoring of the code base. Maybe creating a BaseCommon class that contains the common functionality, and from where BaseModel and BaseDisturbance inherit. Maybe something else. I also think that, based on the discussion, the api that you need for the disturbance, specially in terms of times, is far from clear. I think you need come up with a concrete set of use cases from where the right implementation can be defined. Message ID: ***@***.***>

This actually makes a lot of sense to me. And supports a convo David and I had offline, mainly that disturbances aren't specced out enough.

davidorme:
> And I think it points to where we should go, but you’re not going to like it.

I'd _much much_ rather hear this now than push something that then comes back to bite users heavily in the bum 😄 

I actually suspect that the discovery and initialisation requirements of the Disturbances is much slimmer than the BaseModel - it might actually work better as an entirely parallel system - but I think we need to spec that out "a bit more".

dalonsoa:
So... shall we leave this on hold for the time being?

dalonsoa:
Start new approach, so closing this one.

Issue #536, opened 2024-08-06
Add check for nan values before running the soil model


vgro:
**Is your feature request related to a problem? Please describe.**
The soil model integration function runs into an endless loop when it is fed with NaN values without giving any indication of what is going wrong. This has happened a few times when changes were made in other models and it took quite some time to find the bug.

**Describe the solution you'd like**
We would like the functions that are affected to check if there are NaN values in the input before it starts and throw an error message and exit the model run.

PR #1001, opened 2025-08-18
Changed litter model to use exact solution for the litter pool size

### LINKED ISSUE #969: Environmental variables missing from `LitterModel.vars_required_for_update`
The `LitterModel` does not currently record the three environmental variables it needs to update, specifically `air_temperature`, `soil_temperature` and `matric_potential`.

I don't think this is causing any problems at the moment, so there's no urgency to fix this. But it should be fixed as it could cause bugs in the future

### LINKED ISSUE #979: Integration of the litter model needs to be improved
I currently integrate the litter model using the simplest possible method, i.e. I find the total decay of each pool by

$$D = k P\Delta t ,$$

where $k$ is the decay rate, $P$ is the pool size and $\Delta t$ is the time step. This approximation only works when $\Delta t$ is small relative to the rate at which the pool sizes are changing, which given we use a month time step as default is only true when the pools are close to equilibrium (or when the forest floor is basically frozen 🫠)

I'm unsure what I was thinking with the above as the dynamics are simple enough that they can be integrated to obtain an analytical solution (which I even provided to @hrlai [here](https://github.com/ImperialCollegeLondon/ve_data_science/issues/60)). I think I should try and extend this analytical solution to also include litter input, which should stop the weird behaviour we are seeing

Alternatively, I could convert the whole litter model into rates, and integrate the model using `scipy`. The big problem with doing this is that this require animal consumption to be expressed as a rate rather than a total amount, which would introduce a new way for litter pools to reach negative mass. So, I don't think I should go down this route unless I can't obtain an analytic solution

### LINKED ISSUE #991: Lignin units are incorrect in the litter model
As @hrlai pointed out, the relevant unit to use for lignin proportion is [kg of lignin C per kg of C] as the litter model masses are all expressed as carbon rather than total masses. However, the dostrings currently give the unit for lignin proportion as [kg lignin per kg C], so the docstrings need to be changed to reflect the actual units that the model uses

jacobcook1995:
# Description

This PR changes the litter model to use an exact analytic solution for the pool size with time. The new model approach is [in this part](https://virtual-ecosystem.readthedocs.io/en/979-integration-of-the-litter-model-needs-to-be-improved/virtual_ecosystem/theory/soil/litter_theory.html) of the docs.

While making this change I realised that the way I was averaging nutrient ratios was incorrect. I decided it made most sense to fix that at the same time, but that means that this PR has ballooned in size

@vgro, this fixes the issue of the negative litter pools crashing `ve_run` when the `abiotic` model is included. Unfortunately, it still doesn't run, but it's now a soil model error (linked to #536) so that's progress

@hrlai, I've add you as a reviewer but don't worry about looking at the code, if you could sense check the documentation + check that it fits with the models that you are fitting for the parameterisation, that would be great!

Fixes #979, Fixes #991, Fixes #969

## Type of change

## Key checklist

## Further checks


tests/models/litter/test_chemistry.py
```diff
@@ -58,21 +64,27 @@ def test_calculate_new_pool_chemistries(
...
 ):
     """Test that the function to calculate the lignin updates works as expected."""
 
     expected_lignin = {
-        "above_structural": [-0.0027373, 0.001130172, -0.022171178, -0.029274812],
-        "woody": [-0.00419457, -0.0021217, 0.00224272, 0.00012606],
-        "below_structural": [-0.00025662, 0.01270806, -0.00153633, -0.03044408],
+        "above_structural": [0.4972631215, 0.1011306546, 0.6799674901, 0.6813676608],
+        "woody": [0.4958054, 0.7978783, 0.3522427, 0.350126],
+        "below_structural": [0.49974337, 0.26270880, 0.74846363, 0.71955458],
```
vgro, line 79: :-)

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -152,38 +165,32 @@ def combine_input_sources(data: Data) -> dict[str, NDArray[np.floating]]:
...
     root_phosphorus = data["root_turnover_c_p_ratio"]
     deadwood_phosphorus = data["deadwood_c_p_ratio"]
     reprod_phosphorus = data["plant_reproductive_tissue_turnover_c_p_ratio"]
 
     return {
-        "leaf_mass": leaf_total,
-        "root_mass": root_total.to_numpy(),
-        "deadwood_mass": deadwood_total.to_numpy(),
-        "reprod_mass": reprod_total.to_numpy(),
+        "leaf_mass": leaf_total / update_interval,
```
vgro, line 190: I did not follow the units all the way through, just checking if this is correct if you return pool size/proportions here?

jacobcook1995: Yep this one should be a rate, as the masses in `LitterInputs` are all mass per time. I could change the variable if you think that would make this clearer?

vgro: It is described in the docstring, that should be fine, I just didn't see that in the changed version

vgro (approved):
Looks sensible to me, I cannot comment much on the science

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -120,6 +120,63 @@ the responsiveness of the split to changes in the product of lignin proportion a
...
 
+## Litter decay dynamics
+
+The decay of all litter pools are assumed to follow linear kinetics, with the rate of
+change of litter pool $j$ being given by
+
+$$\frac{dP_j}{dt} = I_j - K_j(T,\psi,L)P_j,$$
+
+where $I_j$ is the rate of input to pool $j$ (in carbon terms), $T$ is the temperature,
+$\psi$ and $K_j(T,\psi,L)$ is the decay rate of the pool. The decay rate of the pool
```
hrlai: Missing definition for psi?

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -120,6 +120,63 @@ the responsiveness of the split to changes in the product of lignin proportion a
...
+will always depend on temperature, but only for the below-ground pools will it be
+affected by soil water potential. Further, lignin content only affects the decay rates
+of pools which contain lignin (i.e. structural and lignin). As an example, the decay
```
hrlai: "lignin" is not a pool?

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -120,6 +120,63 @@ the responsiveness of the split to changes in the product of lignin proportion a
 carbon:nitrogen ratio, and $s_P$ parametrises the responsiveness of the split to changes
 in the product of lignin proportion and carbon:phosphorus ratio.
 
+## Litter decay dynamics
+
+The decay of all litter pools are assumed to follow linear kinetics, with the rate of
+change of litter pool $j$ being given by
```
hrlai: Can we be more explicit like "...the rate of change in the size, $P$, of litter pool..."

hrlai (approved):
Looks good, I've read the theory document only and it makes sense.

jacobcook1995:
Great, thanks both of you! I've made the suggested docs changes and will go ahead and merge now

PR #1004, opened 2025-08-20
536 add check for nan values before running the soil model

### LINKED ISSUE #536: Add check for nan values before running the soil model

jacobcook1995:
# Description

This PR adds a step to check for unexpected NaN values (which can stall the soil model integration) before the soil model integration starts. If NaNs are detected the an error is raised, informing the user which variables contain unexpected NaN values.

For variables that don't have a layer structure, unexpected NaN values are any NaN values. For variables that pertain to the soil layers, they are NaN values in any soil layer For the `soil_temperature` (the only above ground variable the soil model cares about), unexpected NaN values are any NaN values in the surface layer.

Fixes #536

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/soil/test_soil_model.py
+++ b/tests/models/soil/test_soil_model.py
@@ -40,7 +40,6 @@ REQUIRED_INIT_VAR_LOG = (
     (DEBUG, "soil model: required var 'soil_p_pool_secondary' checked"),
     (DEBUG, "soil model: required var 'soil_p_pool_labile' checked"),
     (DEBUG, "soil model: required var 'pH' checked"),
-    (DEBUG, "soil model: required var 'bulk_density' checked"),
     (DEBUG, "soil model: required var 'clay_fraction' checked"),
 )
 POST_SETUP_LOG = (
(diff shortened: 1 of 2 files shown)


jacobcook1995:
If you run `ve_run` with `abiotic` rather than `abiotic_simple`, the soil model raises the following  error several steps in

```
ValueError: Soil model integration cannot proceed because the following variables have unexpected NaN values: {'air_temperature', 'litter_C_mineralisation_rate', 'root_carbohydrate_exudation', 'soil_temperature', 'plant_symbiote_carbon_supply', 'litter_N_mineralisation_rate', 'litter_P_mineralisation_rate'}
```

This is a meaningful improvement as previously the soil model integration would just stall, and this error message gives information as to what the bad input is

vgro (approved):
LGTM

PR #1005, opened 2025-08-20
Removed bulk density from the soil model as it isn't actually used

### LINKED ISSUE #1002: Bulk density is defined as part of the soil model setup but not used
Title.

In the full `abiotic` model bulk density is defined as an input constant. Given that I think it makes sense for bulk density to be deleted as variable for now, so that if @vgro wants to convert bulk density to a variable she can do so without have to build on top the stuff I built and abandoned for the soil model

jacobcook1995:
# Description

This PR removes all uses of bulk density from the soil model associated files as it isn't actually used there (@vgro uses it in a different way in the abiotic/hydrology models)

Fixes #1002

## Type of change

## Key checklist

## Further checks


vgro (approved):
LGTM

PR #1006, opened 2025-08-21
Provide real CN and CP ratios to litter model

### LINKED ISSUE #900: Account for NP during mortaltiy

sallymatson:
# Description

The two main functional changes in this PR are adding stoichiometry to mortality, and removing the placeholder CN/CP turnover ratios that get passed to the litter model with calculated values. This required various other changes in the code including:

- In the litter model, preventing the metabolic rate from going below zero. This was done with a naive fix, a full solution will come in #1010.
- Move the plant turnover data object variable initialisation into a separate function `reset_update_vars` which gets run at the beginning of each update. @davidorme this is different than we discussed, but all of these variables were getting reset each update, just dispersed across `apply_mortality` and `allocate_gpp`, wherever they were encountered first. To avoid any sequencing issues/confusion, I think that a single update function that runs at the beginning is best (allows us to then always add turnover with += in case one or the other function runs first.) 
- Add turnover element mass variables to the data object, for use in calculations across the plants model. Before we only had the CN and CP ratios, but when N/P are being added from both turnover and mortality this gets really messy and confusing. Best to keep things in masses so that they can be used and then convert only at the very end. 
- Litter pre-processing: for a similar reason to above, I removed the calls to `convert_to_litter_units` from within the update sequence and just converted at the very end. I think that we should move this conversion to litter, and I also think there's a case for passing turnover element mass instead of ratio (#1073). The unit choices were made before the plants model was very robust so it made sense to just use whatever was needed in litter as the final units, but at this point storing CN/CP ratios and converting within plants to different units (not mass, which is the unit we use) doesn't make much sense to me. @davidorme @jacobcook1995 if you both agree, I could get rid of the ratios in plants and move all the conversion to litter.. feel free to give thoughts. 

Fixes #900 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -1229,11 +1289,60 @@ def calculate_turnover(self) -> None:
...
+                    out=np.full_like(
+                        self.data["plant_reproductive_tissue_turnover"],
+                        np.inf,
+                        dtype=float,
+                    ),
+                    where=self.data[f"plant_rt_turnover_{element}_mass"] != 0,
+                )
+            )
+
+        # Convert turnover pools to litter units
```
jacobcook1995, line 1357: Happy for this section to move within the litter model, if that's cleaner

virtual_ecosystem/data_variables.toml
```diff
@@ -1160,6 +1160,69 @@ name = "root_turnover_c_p_ratio"
 unit = "-"
 variable_type = "float"
 
+[[variable]]
+axis = ["spatial"]
+description = "Mass of Nitrogen in foliage turnover"
+name = "leaf_turnover_n_mass"
+unit = "kg N"
+variable_type = "float"
```
jacobcook1995, line 1168: Are these needed in addition to `"leaf_turnover_c_n_ratio"` etc?

jacobcook1995 (approved):
Happy with all the changes you've made in the litter model code. I guess we probably want to pin down exactly what information gets passed between the models (I don't know if that makes sense on this PR or on a later PR)

virtual_ecosystem/data_variables.toml
sallymatson: This is what should (eventually, maybe in a future issue) replace the ratios. I'll create a new issue for this, as I think it's not necessary to do in this PR.

jacobcook1995: Makes sense!

davidorme: We should take the opportunity of that PR to synchronise our naming styles for these kinds of variable - at the moment these vars differ in odd ways. Something like `pool_name`+`_C_mass|_N_mass|_P_mass`?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -517,6 +531,53 @@ def _setup(
     def spinup(self) -> None:
         """Placeholder function to spin up the plants model."""
 
+    def reset_update_vars(self) -> None:
+        """Define variables used by the plants model during update."""
+
+        # Initialize variables that hold one value per cell
+        cell_template = xr.full_like(self.data["elevation"], 0)
```
davidorme, line 556: It might be worth adding this as another template to the core components. I think people are using this pattern elsewhere. But that isn't one for this PR.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -517,6 +531,53 @@ def _setup(
...
+
+        self.data["leaf_turnover"] = cell_template.copy()
+        self.data["leaf_turnover_c_n_ratio"] = cell_template.copy()
+        self.data["leaf_turnover_c_p_ratio"] = cell_template.copy()
+        self.data["leaf_turnover_n_mass"] = cell_template.copy()
+        self.data["leaf_turnover_p_mass"] = cell_template.copy()
```
davidorme: I think this function would be cleaner using:

```suggestion
reset_vars = [
    "leaf_turnover", 
    "leaf_turnover_c_n_ratio", 
    "leaf_turnover_c_p_ratio", 
    "leaf_turnover_n_mass",
    "leaf_turnover_p_mass", 
    ...
]
for var in reset_vars:
                self.data[var] = cell_template.copy()
```

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -1029,19 +1064,44 @@ def allocate_gpp(self) -> None:
...
                 cohort_fractions = cohorts.n_individuals / sum(cohorts.n_individuals)
-                element_per_stem = (
-                    total_supply * cohort_fractions
-                ) / cohorts.n_individuals
+                element_per_stem = np.divide(
+                    total_supply * cohort_fractions,
+                    cohorts.n_individuals,
+                    out=np.zeros_like(cohort_fractions),
+                    where=cohorts.n_individuals != 0,
+                )
```
davidorme, line 1105: Could you comment this - elements per stem handling empty cohorts.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -1029,19 +1064,44 @@ def allocate_gpp(self) -> None:
...
                 stoichiometries[element].element_surplus += element_per_stem
 
+                # Add the N and P turnover masses to the data object
+                self.data[f"leaf_turnover_{element.lower()}_mass"][cell_id] += np.sum(
+                    cohorts.n_individuals
```
davidorme, line 1110: Can we standardise on using only upper case for clarity? So vars would be `blah_blah_N_mass` throughout? I actually think that's better generally, as well as cleaner here.

davidorme (approved):
LGTM - some thoughts, but again nothing blocking and I think this re-arrangement is cleaner.

virtual_ecosystem/models/plants/plants_model.py
sallymatson: Going to add all these todos to a general cleanup / clarify issue

sallymatson: https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/1073

PR #1012, opened 2025-08-28
Feature/soil trophic piping

TaranRallings:
# Description

I have made a separate PR for the whole soil foraging system. The new plant piping will be done separately. This portion was expedited as Jacob is going on holiday. 

There shouldn't be anything actually new here, just new versions of existing things to handle the integration of mushrooms, soil fungi, pom, and bacteria. I have gone with mushrooms as the diet name for fruiting bodies because underscores don't work with my naming system.

New Traits
- specific diets for for fungal fruiting (mushrooms), soil fungi, pom, and bacteria

New methods:
- delta_mass methods for fungal fruiting, soil fungi, pom, and bacteria
- get_pools methods for those same resources

Modified methods
- forage cohort
- forage community

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/scaling_functions.py
```diff
@@ -177,6 +177,16 @@ def prey_group_selection(
         result["excrement"] = (0.0, 0.0)
     if diet_type & DietType.DETRITUS:
         result["litter"] = (0.0, 0.0)
+    if diet_type & DietType.MUSHROOMS:
+        # mushroom pool
+        result["fungal_fruiting_bodies"] = (0.0, 0.0)
+    if diet_type & DietType.FUNGI:
```
jacobcook1995, line 183: Is it worth calling this diet type something more specific like SOIL_FUNGI, to really differentiate it from MUSHROOMS?

virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -1070,10 +1077,102 @@ def delta_mass_excrement_scavenging(
...
+        """
+
+        return self.forage_resource_list(
+            resources=soil_fungi_list,
+            adjusted_dt=adjusted_dt,
+            calculate_consumed_mass=self._consumed_resource_mass,
+            herbivory_waste_pools=None,
+        )
+
+    def delta_mass_pomivory(
```
jacobcook1995, line 1126: love the name!

jacobcook1995 (approved):
This seems to be completely consistent with how the previous litter model piping is done. (so there might still be issues, but if there are they will be broader than this PR).

Happy for this to be merged once tests are added

virtual_ecosystem/models/animal/scaling_functions.py
TaranRallings: I think this is a solid idea but I'll need to do a passthrough later on to clean up the diet language and can do it then, if that works for you.

jacobcook1995: Yep that works for me (I guess this is mainly something that will cause you confusion, as I doubt I will be looking at this code)

PR #1015, opened 2025-08-29
Draft for plant resources rework

TaranRallings:
# Description

This is a draft of a new plant resources interface for sitting inside the animal model. I would like it to do a few things:

- create one plant resource object per resource type per grid cell (following the logic of the SoilPool class)
- pool mass between pfts where appropriate
- interface with foraging
- communicate mass information with Data

I've tried fiddling with the data import stuff on my own but I think it will be much more efficient if you can help me out on that front. I have no problem doing all of the foraging interface but the plant data end is pretty arcane for me.

Can you provide feedback on:
- overall form (does the use of one class for these multiple resources make sense?)
- how to construct the `_extract_mass_from_data` method (or a better solution for the same problem)

Thanks!

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/plant_resources_v2.py
```diff
@@ -0,0 +1,158 @@
...
+            else dict(cnp_proportions)
+        )
+
+        # Derived masses
+        self.mass_current: float = 0.0
+        self.mass_stoich: dict[str, float] = {
+            "carbon": 0.0,
+            "nitrogen": 0.0,
+            "phosphorus": 0.0,
+        }
```
davidorme, line 70: Is it necessary to track this? The `get_eaten()` method tracks off `mass_current` and the `cnp_proportions` - it doesn't seem like these masses need to be tracked separately (unless it is used for targeting resources, but I would think that animals target resources based on their expectations of stochiometry, rather than rather god-like knowledge of the total pool).

davidorme (commented):
I've run through the sketch in v2 and tried to unpack it into a working model of a draft API  in `plant_resources_v3.py`. If I understand correctly, we need to:

* Identify a set of variables in data that provide a resource mass and some metric of elemental composition (masses or ratios - we could probably use either).
* Identify a variable used to track consumed mass from that resource.
* The resource is always structured by cell_id but may also be structured vertically and by PFT. We probably need to keep PFTs as separate resources.
* We then need a mechanism by which an array can be parcelled out as individual cells using something that supports the Resource protocol.
* We then need to be able to pull consumed mass back across cells and write it back to data.

The file I've pushed hinges on two key mechanisms:

1. The ArrayResources class can be used to produce cell specific resources using indexing: so `plant_resource_0 = array_resource_instance[0]` gives a PlantResource for cell 0.
2. The total mass consumed in `plant_resource_0` is a sneaky view back into an array across cells in the originating ArrayResources instance - so the total mass consumed is auto compiled.

The file contains some code at the end showing a rough workflow for the new classes. I think this should drop into your existing framework relatively easily?

@jacobcook1995 How does this compare to what is being used for soil array resources - can we unify to a common standard here?

@dalonsoa Could you have a quick look at the logic here - I think using array views to track the data generated by the child `PlantResource` back into the parent `ArrayResources` seems sensible but are there better ways to achieve this?

virtual_ecosystem/models/animal/plant_resources_v2.py
TaranRallings: A dictionary of this kind is just the standard format of passing mass around in the animal model. The way I did CNP in the plant get_eaten is a non-standard implementation. I wasn't sure how you were handling CNP exactly so I built in the bits required for either route. Ultimately, as long as get_eaten returns a dict of that kind, it should work for the animal model.

TaranRallings:
This looks like a very promising way to break the flow down! I'll leave fiddling with foraging interface for after we hear from Jacob and Diego.

virtual_ecosystem/models/animal/plant_resources_v2.py
davidorme: > I wasn't sure how you were handling CNP exactly so I built in the bits required for either route.

Neither am I, yet 😄

However it comes in (mass or ratio) I think handling it internally as proportions and only tracking total mass is easier. It's not like the animals can alter the ratios? 🤞

virtual_ecosystem/models/animal/plant_resources_v3.py
```diff
@@ -0,0 +1,344 @@
...
+        vertical_occupancy: VerticalOccupancy,
+        data: Data,
+        vertical_layers: DataArray | None = None,
+        pft: str | None = None,
+        # cell_id: int,
+    ):
+        self.data = data
+        self.mass_var = mass_var
+        self.n_ratio_var = n_ratio_var
+        self.p_ratio_var = p_ratio_var
```
sallymatson, line 76: Can we name these CN & CP ratios just for continuity?

davidorme:
@TaranRallings Do you prefer to deal with masses of C, N, P or - as is the case in the code at the moment - with carbon mass and then CN and CP ratios? @sallymatson and I want to standardise how we write C+N+P to the `Data` object on either masses (which we mildly prefer) or ratios.

jacobcook1995:
So the above could definitely work for the litter pools, and seems like a much cleaner approach. I think it also could work for the soil pom pool.

With some basic modification I think it could work for the soil bacteria pool (the issue there is that the stoichiometry comes is a constant rather than being recorded in the data object).

Getting it to work for the soil fungi pool would be much more tricky, because in that case you have to combine three fungal pools into a single animal available pool (+ weighting the stoichiometric ratios appropriately)

Fungal fruiting bodies would need a pretty big refactor to work in this framework (because they are primarily updated in the animal model, which could change but that would require thought). Carcasses and excrement would have a similar problem I think

Given all that I'm not sure what the best approach is. I think it definitely makes sense to treat litter pools as plant resources, that follow the same framework. But I'm unsure whether it makes more sense to try and fit the other pools within the same overarching framework, or just to treat them as something meaningfully different that is handled differently?

virtual_ecosystem/models/animal/plant_resources_v3.py
```diff
@@ -0,0 +1,344 @@
...
+        self.mass_consumed_var = mass_consumed_var
+        self.vertical_occupancy = vertical_occupancy
+        self.vertical_layers = vertical_layers
+        self.pft = pft
+
+        for var in (mass_var, n_ratio_var, p_ratio_var, mass_consumed_var):
+            if var not in data:
+                raise ValueError(
+                    f"Cannot initialise ArrayResource: {var} not found in data object"
+                )
```
dalonsoa, line 86: This should not be needed. This resource would be initialised by a model - presumably the `AnimalModel` - which will have among its required variables (possibly required by init ones?) the ones indicated here. So, by this point, all this validation should have been done already.

virtual_ecosystem/models/animal/plant_resources_v3.py
```diff
@@ -0,0 +1,344 @@
...
+for array_res in herbivory_resources:
+    array_res.write_herbivory()
+
+data["leaf_mass_consumed"]
+data["subcanopy_mass_consumed"]
+
+# And then, before starting the next loop:
+
+for array_res in herbivory_resources:
+    array_res.set_mass_and_elemental_ratios()
```
dalonsoa, line 344: Where would all of these be happening? Within the `AnimalModel` `update` or somewhere else?

Also, that triply nested `for` loop looks super inefficient. although it is not obvious how to improve it as none of the functionality is vectorised...

dalonsoa (commented):
Sorry for the slow reply - I hadn't spotted this. 

The approach looks quite sensible and I do not see any obvious improvement. I have a couple of comments, though, because I'm not entirely sure where this would be used.

virtual_ecosystem/models/animal/plant_resources_v3.py
davidorme: True - there's a point at which putting belt and braces of error checking on all your code makes it slow down too much. We could lose this - any crash is then a result of the code of the `AnimalModel` being wrong - it isn't a user facing calibration.

virtual_ecosystem/models/animal/plant_resources_v3.py
davidorme: From this line down, it would be within `AnimalModel._update` (although probably as  a call to an encapsulating method?):

>    `# Get a tuple of all herbivory array resources`

Everything above that is the PlantsModel pushing Plant data into the `Data` object,  in `PlantsModel.__init__` and `PlantsModel._update`

And I agree the triply nested loop looks bad but we'd need to vectorise the animal model to avoid it.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -422,12 +435,30 @@ def _setup(
...
+            ),
+            ArrayResources(
+                data=self.data,
+                mass_var="subcanopy_seedbank_biomass",
+                n_ratio_var="subcanopy_seedbank_n_ratio",
+                p_ratio_var="subcanopy_seedbank_p_ratio",
+                mass_consumed_var="subcanopy_seedbank_biomass_consumed",
+                vertical_occupancy=VerticalOccupancy.GROUND,
+            ),
+        )
```
davidorme, line 466: I'd probably do this - should be easier to maintain?

```suggestion
        # TODO: plant resource
        plant_resources = [
             (
                 "subcanopy_vegetation_biomass",
                 "subcanopy_vegetation_n_ratio",
                 "subcanopy_vegetation_p_ratio",
                 "subcanopy_vegetation_biomass_consumed",
                  VerticalOccupancy.GROUND,
            ),
            (
                "subcanopy_seedbank_biomass",
                "subcanopy_seedbank_n_ratio",
                "subcanopy_seedbank_p_ratio",
                "subcanopy_seedbank_biomass_consumed", 
                VerticalOccupancy.GROUND,
            ),
        ]
        
        self._self._plant_array_resources = (
            ArrayResources(
                data=self.data,
                mass_var=mvar,
                n_ratio_var=nvar,
                p_ratio_var=pvar,
                mass_consumed_var=mcvar,
                vertical_occupancy=vocc
            ) for mvar, nvar, pvar, mcvar, vocc in plant_resources
        )
        
```

TaranRallings: Oh yeah, that's much better!

davidorme:
I'm closing this as it is superseded by #1318

PR #1023, opened 2025-09-04
Adopt pytest-xdist

### LINKED ISSUE #941: Use `pytest-xdist` to parallelise running tests
I noticed that the unit test suite now takes a few minutes to run. (This is a good sign -- the project has great test coverage!) You could speed it up by parallelising the test running though, which will help with development and shorten CI run times. This does assume that your tests *can* be run in parallel (if not, that's a sign they're messing with global state which they probably shouldn't do anyway).

You could consider using the [`pytest-xdist`] package for this. For an example, see here: https://github.com/ImperialCollegeLondon/FROG/blob/main/pyproject.toml

davidorme:
# Description

This PR follows @alexdewar's suggestion to use `pytest-xdist` to speed up testing. We're still going to run into `test_ve_run` being slow, but it should speed up testing generally. It also adds `--jobs auto` to the `sphinx` building options to distribute the running of MyST notebooks. 

Fixes #941

## Type of change

## Key checklist

## Further checks


alexdewar (approved):
LGTM! I think this makes sense.

.github/workflows/ci.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       - name: Build docs using sphinx
         run: |
           cd docs
-          poetry run sphinx-build -W --keep-going source build
+          poetry run sphinx-build --jobs auto -W --keep-going source build
```
alexdewar, line 84: Would it also work if you just ran `make` here? Or will it not run without the `poetry run` prefix?

davidorme: I think `poetry run make` would work - but I think we probably do want to run the local `sphinx-build` in a slightly different way to the CI? So fail fast and not keep going locally but find everything wrong on CI? Doesn't make a huge difference but I'll leave as is for now.

PR #1032, opened 2025-09-09
Subcanopy vegetation turnover and stoichiometry

### LINKED ISSUE #965: Subcanopy vegetation should turnover
At present, subcanopy vegetation just grows and grows and the mass will only decrease via herbivory. There should be some kind of turnover to add to the litter module.

davidorme:
# Description

This PR updates the subcanopy representation to handle stoichiometry and turnover:

1. Adds turnover and stoichiometric ratios constants for the subcanopy vegetation and seedbank
2. Moves the subcanopy representation from a pure array based representation in `Data` into its own module and class.
3. Adds a simple `SubcanopyStochiometry` data class to represent subcanopy biomass pools with some simple helper methods to move biomass between pools.
4. Implements the `Subcanopy` class that has methods to:
    * initialise the subcanopy
    * calculate growth, reproduction and turnover dynamics in the biomass pools
    * calculate the leaf area index and light absorption.

Fixes #965

## Type of change

## Key checklist

## Further checks


davidorme:
@sallymatson Could you cast a quick eye over the proposed `subcanopy` module and see if the logic and structure look sane. If that all looks good then I'll integrate it into the model and update the testing.

virtual_ecosystem/models/plants/subcanopy.py
```diff
@@ -0,0 +1,379 @@
...
+    cells and provides properties to report the CN and CP ratios. It also provides
+    methods to add and remove masses from the class and to remove excess nutrients above
+    ideal ratios.
+    """
+
+    def __init__(
+        self,
+        carbon_mass: NDArray[np.floating],
+        nitrogen_mass: NDArray[np.floating],
+        phosphorous_mass: NDArray[np.floating],
```
sallymatson: In the existing Stoichiometry implementation we decided to have the structure element-agnostic, with the reasoning that it would better allow us to scale to other nutrients down the line. While this does make things easier/more readable in one sense, I don't think it's ideal to have such a different fundamental design.... thoughts?

sallymatson:
Other than the comment above I think that everything else looks very clean and simple!

virtual_ecosystem/models/plants/subcanopy.py
davidorme: I did think about that. I think my rationale here is that this class is very simple and extending it would be easy. 

We could nest the nutrients:

```python
@dataclass
class SubcanopyNutrient:
    name: str
    ideal_ratio: float
    values: NDNDArray[np.floating]
```

And then have `SubcanopyStochiometry` as:

```
carbon_mass: NDArray[np.floating]
nutrients: dict[str, SubcanopyNutrient]
```

That feels a little heavy handed, but it is more in line?

davidorme: I think I'd feel obliged to do more defensive programming so you can't e.g. combine instances with different nutrients? So it would get more complex, but perhaps not too much.

davidorme: But it would also potentially smooth off the rough edge of the excess nutrient transfer having zero carbon and hence possibly NA ratios...

sallymatson: I trust your judgement on which is better. I think I'd lean toward this implementation, if just for continuity's sake. But I also get the argument for simplicity and how it is is super useable. In fact a bit jealous as stoichiometry is quite complicated, so keep this part simple where possible is worth a lot

davidorme: @sallymatson How about the new version - there is one spot that is hard-coded but that is currently bigger than I want to think. I've captured most of the generalisable stuff and I think it is more complex but cleaner in some ways.

davidorme:
@sallymatson I'd forgotten there were some doc fixes to push. I think this is a good point to merge this if it all looks good to you .

sallymatson (approved):
LGTM, splitting into its own section is also helpful for clarity.

davidorme:
@sallymatson @jacobcook1995 

I've updated this PR to add in those variables we talked about: lignin concentrations for litter and then the nutrient uptakes. Does this look ok? 

The specific changes from the version Sally approved are:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/compare/473d0a4..459f7c5

jacobcook1995 (approved):
Happy with this on the soil side!

sallymatson (approved):

PR #1043, opened 2025-09-15
1033 bug in accumulated runoff

### LINKED ISSUE #1033: Bug in accumulated runoff
**Problem:**

There is an assumption that all water from upstream runs through the grid cell in one time step, however, the way it is updated in the loop is adding too much water: 

`New_accumulated runoff(t1) =  surface_runoff(t1, me) + previous_accumulated_runoff(t0, upstream cells) `

then set 
`Previous_accumulated_runoff(t1) = new_accumulated_runoff(t1) `

```
New_accumulated runoff(t2)
=  surface_runoff(t2, me) + previous_accumulated_runoff(t1, upstream cells) 
= surface_runoff(t2, me)  + surface_runoff(t1, me)  + previous_accumulated_runoff(t0, upstream cells) +  previous_accumulated_runoff(t1, upstream cells) 
```
So we keep adding the same water over and over again 

**Possible solutions:** 
1. move water only a shorter distance, like one grid cell per time step, make sure inflow and outflow is considered.

`Flow_through_cell (t1)=next highest neighbour(t0)-me(t1) `

2. Make water run through the whole grid, adjust how previous_accumulated is handled, I believe it should be the surface runoff from the previous timestep only. Maybe better to call it `upstream_runoff`, definitely explain variables better in documentation 

`New_accumulated runoff(t1) =  surface_runoff(t1, me) + upstream_runoff(t0) `

then set 
`net upstream_runoff(t1) = surface_runoff(t1, upstream cells) `

`New_accumulated runoff(t2) =  surface_runoff(t2, me) + upstream_runoff(t1) `

This means that the water will magically disappear at the lowest points with the new time step.

@lelavathy , @sallymatson and I started a discussion that we might want to move to a proper river routing implementation, which would make this problem go away (see #1034 ) but for now decided to fix the bug first.

vgro:
This PR fixes the water accumulation bug described in #1033 and reorganizes how the streamflow is calculated:

- there is no longer a link to the previous time step, only surface and subsurface flows from the current time step contribute to the stream flow.
- I  renamed the accumulated flows to `channel_inflow` and the function `accumulate_horizontal_flow` `route_horizontal_flow`.
- I tried to explain all this better in the docs
- I added a different DEM to the implementation documentation to show the actual connections, but there is currently an issue with the mapping, see note below.
- I added a test to do a monthly water balance between precipitation and the stream flow in the lowest grid cells (if no lowest grid cell, the next lowest group of cells)

NOTE: there seems to be a problem with the mapping of the elevation on the grid, likely because I did not provide a projection, but need to keep an eye on, see [docs built here](https://virtual-ecosystem.readthedocs.io/en/1033-bug-in-accumulated-runoff/virtual_ecosystem/implementation/hydrology_implementation.html).

Fixes #1033

## Type of change

## Key checklist

## Further checks


docs/source/virtual_ecosystem/implementation/hydrology_implementation.md
```diff
@@ -404,6 +406,14 @@ grid = Grid(
 )
 data = Data(grid=grid)
 data["elevation"] = elevation
+ele_plot = DataArray(
```
vgro, line 406: I added this plot to show that the data changes orientation when it is mapped on the grid. Any suggestions what's going on?

davidorme: Oh man, we so shouldn't have used an NxN grid in the example data. I'll try and look into what is going on in the data loader.

docs/source/virtual_ecosystem/implementation/hydrology_implementation.md
```diff
@@ -429,39 +439,60 @@ drainage_map = calculate_drainage_map(
...
+cells. The subsurface flow contribution for each cell is calculated similarly to surface
+flow:
+
+* Local subsurface runoff generated in the cell during the current timestep.
+* Contributions from subsurface runoff from upstream cells during the same timestep.
+  This allows the model to account for below-ground movement of water through soil and
+  groundwater pathways.
+
+The total river discharge at each cell is the sum of surface and subsurface horizontal
+flows (Q):
```
sallymatson: Just to clarify. Shouldn't river discharge NOT be a sum of flow? In my (new) understanding, flow represents how much water flows through each cell in the timestem. Whereas the river discharge should only account for water that *terminates* in that cell. Is that correct? 

So the sum of the flow across the whole grid will be greater than the sum of river discharge across the whole grid.

virtual_ecosystem/models/hydrology/above_ground.py
```diff
@@ -380,38 +380,50 @@ def find_upstream_cells(lowest_neighbour: list[int]) -> list[list[int]]:
...
-        drainage_map: Dict of all upstream IDs for each grid cell
-        current_flow: (Sub-)surface flow of the current time step, [mm]
-        previous_accumulated_flow: Accumulated flow from previous time step, [mm]
+        drainage_map: Dict mapping each cell ID -> list of upstream cell IDs
+        surface_runoff: Surface runoff for this timestep, [mm]
+        subsurface_runoff: Subsurface runoff for this timestep, [mm]
 
     Returns:
-        accumulated (sub-)surface flow, [mm]
+        Total streamflow contribution at each grid cell, [mm]
```
sallymatson: Can the names be a bit more standardized? The return variable is called `channel_inflow` but here you say the return is total streamflow; I actually think that `total_streamflow` is a bit more descriptive at least in how I understand what is going on

sallymatson: But in general would just prefer if this terminology somewhat mirrors the code/variables below, not as picky on which it is

virtual_ecosystem/models/hydrology/hydrology_model.py
```diff
@@ -395,7 +379,7 @@ def _update(self, time_index: int, **kwargs: Any) -> None:
         above ground runoff.
 
         Total river discharge is calculated as the sum of above- and below ground
-        horizontal flow and converted to river discharge rate in m3/s.
+        horizontal flow contributios and converted to river discharge rate in [m3 s-1].
```
sallymatson: Sp

sallymatson (commented):
I think it looks good but I just have a few clarifications about some terminology / explanations. 

Another question (not to fix in this PR necessarily) but with this idea of flow and the upstream cells - are upstream cells just ANY upstream cell in the grid? Because that would not accurately represent flow; e.g. if there is a lake basin then cells north of the river would not flow through cells south of the river, even if the elevation is higher. So is there some sense of directionality built into the flow maps? 

I can/will also check that in the code but just wanted to raise it before it slips my mind.

vgro:
> I think it looks good but I just have a few clarifications about some terminology / explanations.
> 
> Another question (not to fix in this PR necessarily) but with this idea of flow and the upstream cells - are upstream cells just ANY upstream cell in the grid? Because that would not accurately represent flow; e.g. if there is a lake basin then cells north of the river would not flow through cells south of the river, even if the elevation is higher. So is there some sense of directionality built into the flow maps?
> 
> I can/will also check that in the code but just wanted to raise it before it slips my mind.

Thank you so much for your comments! I made some changes to the docs and variables names, I hope it it now clearer what is what. We have the runoff and local generation in each grid cell, and we have the inflow which is everything coming from above, and we have the river discharge which is the sum of all. I also added the definitions to the glossary and linked to it in the documentation.

Regarding your question above, The way the drainage map is calculated should include direction, we first find the lowest neighbor for each grid cell, so where the water flows, and then identify upstream cells backwards.

docs/source/virtual_ecosystem/implementation/hydrology_implementation.md
vgro: the river discharge does still include all the upstream cells, but not from the previous timestep, that was the bug. So the sum of river discharge will be larger than the rainfall-evaporation, but the local generation sum should be the same (over a longer period)

virtual_ecosystem/models/hydrology/above_ground.py
vgro: I made an attempt to do this, please tell me if it is still not clear, this should be really easy and I'm happy to adjust more if needed

vgro:
@sallymatson  I just read this definition in a paper which sounds very clear, I could adopt this (basically rename what is now 'total_river_discharge' to 'total_runoff': 

'Based on that, hereinafter we refer to river discharge, Q, to indicate the amount of water passing a particular point of a river (in m3 s−1), whereas runoff, R, is regarded as the depth of water produced from a drainage area during a particular time interval (in mm). The difference between the two quantities is related to the routing processes that allow runoff to transform into river discharge.'

docs/source/virtual_ecosystem/implementation/hydrology_implementation.md
```diff
@@ -429,39 +444,68 @@ drainage_map = calculate_drainage_map(
...
 
-The accumulated surface runoff is the calculated in each grid cell as the sum of current
-runoff and the runoff from upstream cells at the previous time step.
+### Horizontal flow and river discharge
+
+We track horizontal water fluxes in two pathways — surface and subsurface — and then
+combine them into total river discharge.
+
+#### Surface flow ($Q_{surface}$)
+
```
lelavathy, line 448: The terminology used to describe flow components should be standardized for consistency. For example, some sections refer to surface flow and subsurface flow, while others use surface runoff and subsurface runoff.

lelavathy:
> @sallymatson I just read this definition in a paper which sounds very clear, I could adopt this (basically rename what is now 'total_river_discharge' to 'total_runoff':
> 
> 'Based on that, hereinafter we refer to river discharge, Q, to indicate the amount of water passing a particular point of a river (in m3 s−1), whereas runoff, R, is regarded as the depth of water produced from a drainage area during a particular time interval (in mm). The difference between the two quantities is related to the routing processes that allow runoff to transform into river discharge.'

lelavathy:
> > @sallymatson I just read this definition in a paper which sounds very clear, I could adopt this (basically rename what is now 'total_river_discharge' to 'total_runoff':
> > 'Based on that, hereinafter we refer to river discharge, Q, to indicate the amount of water passing a particular point of a river (in m3 s−1), whereas runoff, R, is regarded as the depth of water produced from a drainage area during a particular time interval (in mm). The difference between the two quantities is related to the routing processes that allow runoff to transform into river discharge.'

yes agree, total runoff (mm) is the right term water balance comparisons

docs/source/virtual_ecosystem/implementation/hydrology_implementation.md
```diff
@@ -429,39 +444,68 @@ drainage_map = calculate_drainage_map(
...
+* Upstream subsurface inflow: subsurface runoff generated in upstream cells during the
+  same timestep.
+
+#### Total river discharge ($Q_{total}$)
+
+The total flow passing through a cell’s channel ({term}`total river discharge`) is the
+sum of these two pathways:
+
+$$Q_{total} = Q_{surface} + Q_{subsurface}$$
+
```
lelavathy, line 470: If we follow Figure 9, subsurface runoff (Q2) and interflow (Q3) are currently not implemented. This means that:    Qtotal= Q surface + Q sub baseflow (Q4) + Q baseflow (Q5) right?

vgro: this is now outdated, I changed it to:
Qtotal = Rsurface + Rsubsurface

vgro: Sorry I just made the changes as you are still reading it, I will wait for you to complete the review :-)

lelavathy: I’ve completed the review. I’m a bit confused with the some of the flow terminology. It's okay, we can discuss this in our meeting.

vgro:
> > I think it looks good but I just have a few clarifications about some terminology / explanations.
> > Another question (not to fix in this PR necessarily) but with this idea of flow and the upstream cells - are upstream cells just ANY upstream cell in the grid? Because that would not accurately represent flow; e.g. if there is a lake basin then cells north of the river would not flow through cells south of the river, even if the elevation is higher. So is there some sense of directionality built into the flow maps?
> > I can/will also check that in the code but just wanted to raise it before it slips my mind.
> 
> Thank you so much for your comments! I made some changes to the docs and variables names, I hope it it now clearer what is what. We have the runoff and local generation in each grid cell, and we have the inflow which is everything coming from above, and we have the river discharge which is the sum of all. I also added the definitions to the glossary and linked to it in the documentation.
> 
> Regarding your question above, The way the drainage map is calculated should include direction, we first find the lowest neighbor for each grid cell, so where the water flows, and then identify upstream cells backwards.

I just printed the drainage map out in the docs, and it looks like it has the direction, but I am not sure that it captures all upstream cells correctly.

docs/source/virtual_ecosystem/implementation/hydrology_implementation.md
vgro: created #1054

sallymatson (approved):
LGTM

PR #1045, opened 2025-09-18
v0.1.1 alpha 13 release

jacobcook1995:
# Description

This PR bumps the version so that we can release a new version. My motivation for doing a release now is so that I have a version of the model after the litter model integration bug has been fixed (mainly for demonstration purposes)


vgro (approved):
Thanks for keeping up with this!

PR #1060, opened 2025-09-29
Retire the plant schema [pft_definition] option

### LINKED ISSUE #1038: Retire the `pft_definition` option in the PlantsModel schema
We now have the much clearer option to configure plant functional types in a single CSV file, rather than a clunky and long-winded array of arrays in TOML. We should only retain _and then document_ one input route and the CSV is far superior so:

* Remove that from the model config
* Update tests to use some shared CSV trait files

Note that the set of variables in the schema are currently the only "documentation" of the required fields, so #1037 needs to happen first!

davidorme:
# Description

This PR simplifies the config options for the PlantsModel by removing the TOML array `[[plants.pft_definition]]` approach completely and only using the `pft_definitions_path` option.

Fixes #1038

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
LGTM!

PR #1065, opened 2025-10-01
Add documentation for nutrient split between litter pools

### LINKED ISSUE #1046: Document the split of nutrients between litter pools
The litter model currently makes a strong simplifying assumption about how nutrients are split between metabolic and structural litter pools (i.e. that there is a fixed ratio for the division). This is neither mentioned nor justified within the docs, which needs to change

jacobcook1995:
# Description

@hrlai, this adds documentation for the split between of nutrient between pools. Please let me know whether it makes sense (if it doesn't make sense to you it won't make sense to anyone really).

One thing I was wondering is whether it's worth including the equations for how the split is calculated in here? Let me know whether that seems like a good idea or just clutter.

The relevant docs can be viewed [here](https://virtual-ecosystem.readthedocs.io/en/1046-document-the-split-of-nutrients-between-litter-pools/virtual_ecosystem/theory/soil/litter_theory.html#split-of-nutrient-inputs-between-pools)

Fixes #1046


hrlai:
Hey @jacobcook1995 yes I think it would benefit (at least someone like me) to have the equations. For now I struggle a bit in parsing the text descriptions.

jacobcook1995:
@hrlai, ah yes that was a bit of a wall of text. I've broken up the text a bit and added in equations, which should hopefully make it clearer. Let me know what you think!

hrlai (approved):
Looks good, added very minor edits.

I think it'd be good to add a one-liner interpretation of $r_m$, and I have come up with a one-liner based on the following flow chart. I hope this is the right interpretation.

image

jacobcook1995:
Yeah the part in the square brackets is a bit hard to explain.

It basically boils down to the fact that the flow to each pool has already been determined, so the ratio of input chemistries ($\rho$) can only be maintained (consistent with the total input) by changing the chemistry of the input flows ($r_m$ and $r_s$). So the first term in the brackets represents the nutrient input to the metabolic pool and the second term the nutrient input that has to be added to the structural pool to maintain the fixed ratio ($\rho$). The solution to the equation is then the only solution that doesn't involve nutrients being created or destroyed.

jacobcook1995:
@hrlai , I've tried to explain the above in the docs, let me know if it makes sense or needs further changes

hrlai:
Ah I see. Thanks, looks good to me now.

hrlai (approved):

Issue #1035, opened 2025-09-11
[META] Data input, output and validation documentation


davidorme:
I've just had a long chat with @arne-exe about data preparation and input formats and validation. That threw up a whole bunch of related issues that are largely to do with documentation of the model inputs and their format and validation. 

There's a ton of stuff here so I've made this into a meta issue to gather sub-issues and provide an overview that could form the draft of an intro to model setup in the documentation.

There are basically four ways to get inputs into models:

1. **Model configuration values**
    
     These are values that are set in the model configuration TOML files. An example would be the maximum number of canopy layers in the simulation. 

    *Format*: entries in a TOML configuration file.

    *Validation*: Currently these values have some simple validation because loading TOML does this automatically - it is really just that numbers are numbers and strings are strings, but the VE model code should then have more sophisticated value checking (for example that a file path exists or that a number isn't negative).

    *Documentation*: Essentially no systematic documentation. Users _can_ read the model schemas to found out names and types but that is not acceptable for most users and so we need actual human readable descriptions of the model schema with example inputs. We do provide examples in the `example_data` but it needs linking in, probably at the top of the model implementation pages.

    #1039

    *Output*: These are fixed in the configuration - they don't output.

2. **Model constants**

    These are per model sets of values, generally things that take a single value that is held constant across the whole simulation, but potentially something slightly more complex (sets of coefficients). An example would be the `wind_reference_height` in the `abiotic` model. 

    In some ways (potentially very many ways) these are like a concealed set of model configuration values and we possibly (probably?) (certainly?) should long-term move integrate these.

    *Format*: This is not well documented but model constants are configured through the model TOML configuration file. The problem here is that the structure of the TOML required for constants is _not_ included in the `model_schema.json`: the names and types of constants are simply not there. 

    *Validation*: Constant values are handled by the constant loading process - which currently only checks for missing or mistyped constant names - and then any further value validation (e.g. sign, bounds etc) is down to model specific checking.

    *Documentation*: The constants objects _are_ documented but only in the API documentation. For example: https://virtual-ecosystem.readthedocs.io/en/latest/api/models/abiotic/abiotic_constants.html

    #1040

    *Output*: These are fixed in the configuration- they don't output.

3. **Model data variables**

    This is using the formal sense of variables that go into the central model `Data` object. These are data arrays that are structured along one or more of the Virtual Ecosystem dimensions and are things like the temperature or litter biomass. At the moment have we four critical dimensions within the VE:

[...]

PR #1079, opened 2025-10-10
Replace configuration and constants with pydantic.

### LINKED ISSUE #1086: Create new configuration classes
Create a proof of concept that creates the new configuration classes and then demonstrates the inclusion of the those classes in the model registry and the resulting documentation.

davidorme:
# Description

> [!WARNING]
> I know this is a big change but:
>  * Our configuration system is a mess and confronting the data team with this has shown up the weaknesses (see #1035). 
> * I think the problems are as much structural as documentary - we could doc what we have, but it would be a dumpster fire.
> * Honestly, it I'm not sure we even _can_ configure all our models from where we are now. I'm not sure how easy it will be to write any papers using the system as is.
>
> So - if you agree with the assessment below - I think we do need to start doing this now but I think we can do it incrementally _to some degree_.

So, this is somewhere between a PR and a discussion. I've pushed the code side of things much further than a simple discussion because I wanted to prove to myself that what I was suggesting would work. 

### The problem

For each model, we have:

**The `module_schema.json` file** which defines the "formal" part of the configuration.

* It is written in JSON_schema, which is frankly no fun even for the dev team. What it _does_ do is provide validation rules on what comes in to the TOML.
* We do not document these data required by these schema cleanly and systematically, and because there is no autodoc on these files, there are all sorts of synchronisation problems if we did.

**The `constants.py` file** defines data classes that contain model specific constant settings.

* These aren't actually too bad - they _do_ autodoc so we can create API outputs quite nicely.
* However they are disconnected from the configuration in a really ugly way. Each module scheme defines a `constants` slot which then accepts **completely arbitrary input** that is not checked by config validation.
* Constant classes are created in model workflow only in `Model.from_config` and it is only at this point that the constant class gets anything passed in via the constants slot in the config. If the configured inputs aren't ok, it fails moderately helpfully,  but there is no mechanism for validation rules, so there's no sophisticated guidance to help the user debug their config.
* Because the constants classes are only created during model init, they are **completely isolated from other models**. As a result, we have ended up with some models having to create their own instances of  constants from other models or have defined core constants locally.

### Suggested solution

* We merge our constants into our model schema. This is already implicit in that `constants` slot in the schema. That way we have a single thing that actual _is_ the config, not a cut and shut of two pieces.
* We ditch JSON Schema and move it all into one of the purpose built validation/serialisation packages. I've had a look at a some popular options (`mashumaro`, `marshmallow`, `pydantic`) and `pydantic` is _incredibly_ close to what we have for `constants.py` already so porting it to that was actually pretty quick.
* The alternative - moving the constants out of data classes into JSON Schema - doesn't bear thinking about: doc problems, horrible to write and maintain and some of our constants classes have Python functionality in addition to their data storage.
* Doing this means that, when a model config is loaded, the whole model config class is loaded **including the constants**  and stored centrally in a single Config object. We can pass that configuration to all models and they **can all access all others constants and configuration settings**.
* The documentation becomes _massively_ easier as we can docstring everything, and can also use the `pydantic` `dump_model` methods to dynamically insert example TOML into the docs. We have a single integrated system. For example, we can include this in a notebook cell:

```python
In [13]: from virtual_ecosystem.models.abiotic.config import AbioticConfig
    ...: import tomli_w
    ...: 
    ...: print(tomli_w.dumps({"abiotic": AbioticConfig().model_dump()}))
[abiotic]
static = false

[abiotic.constants]
leaf_emissivity = 0.98
leaf_albedo = 0.15
bulk_density_soil = 1175.0
wind_reference_height = 10.0
latent_heat_vap_equ_factors = [
    1918460.0,
    33.91,
]
zero_plane_scaling_parameter = 7.5
substrate_surface_drag_coefficient = 0.003
roughness_element_drag_coefficient = 0.3
roughness_sublayer_depth_parameter = 0.193
max_ratio_wind_to_friction_velocity = 0.3
drag_coefficient = 0.2
min_windspeed_below_canopy = 0.1
min_roughness_length = 0.01
light_extinction_coefficient = 0.01
soil_thermal_conductivity = 1.206
specific_heat_capacity_soil = 881
surface_albedo = 0.125
soil_emissivity = 0.95
saturated_pressure_slope_parameters = [
    4098.0,
    0.6108,
    17.27,
    237.3,
]
dry_air_factor = 0.378
initial_flux_value = 0.001
```

See [this example page](https://virtual-ecosystem.readthedocs.io/en/config_via_pydantic/using_the_ve/configuration/new_config.html) using code object to document config directly.

* Loading data via `pydantic` _automatically_ validates types but we can (incrementally) then refine our config classes to add more validation. This is much much easier in Pydantic than it would be in JSON Schema. The pydantic system automatically gathers errors into a neat report object when it loads, so we can make configuration robust in a _much_ better way.

```python
from pydantic import BaseModel 

class Config(BaseModel):
    cant_be_negative: float = Field(gt=0, default=2)
``` 

* One big change is that config settings are accessed by dot notation, not dict, so `myconf.cant_be_negative` rather than `myconf['cant_be_negative']`. That is then consistent with constants - which are already a class.

### Demonstration

* I've pushed a new `model_config.py` module for all models (I've ducked details in `soil` pending possibly some details out to CSV, but everything else is there). 
* I've also pushed a new `core.configuration.py` submodule that defines some common core classes and utility functions.
* I've added a demo page rendering the PlantsModel configuration and added in links to the APIs.
* There is a test that dynamically creates a Config model from a modifiable list of models (everything at the moment), writes it to file and then reads it back in.  It can also load the compiled file from the example data.
* I've added config discovery to the module info in `core.registry.py` to check that it's fine to drop into registration - seems to be.
* That means we do now actually have a **full configuration template** 🤯 

<details>
<summary> 👈 👈  Click to unhide an actual full configuration document!</summary>

```toml
[core]
static = false

[core.constants]
placeholder = 123.4
standard_pressure = 101.325
standard_mole = 44.642
molar_heat_capacity_air = 29.19
gravity = 6.6743e-11
stefan_boltzmann_constant = 5.6703744191844314e-08
von_karmans_constant = 0.4
max_depth_of_microbial_activity = 0.25
meters_to_mm = 1000.0
molecular_weight_air = 28.96
gas_constant_water_vapour = 461.51
seconds_to_day = 86400.0
seconds_to_hour = 3600.0
characteristic_dimension_leaf = 0.01
specific_gas_constant_dry_air = 287.05
molecular_weight_ratio_water_to_dry_air = 0.622
conductance_to_resistance_conversion_factor = 40.9
density_water = 1000.0
fungal_fruiting_bodies_c_n_ratio = 10.0
fungal_fruiting_bodies_c_p_ratio = 75.0
fungal_fruiting_bodies_decay_rate = 0.013862943611198907

[core.grid]
grid_type = "square"
cell_area = 8100.0
cell_nx = 9
cell_ny = 9
xoff = -45.0
yoff = -45.0

[core.data_output_options]
save_initial_state = false
save_continuous_data = true
save_final_state = true
save_merged_config = true
out_path = "."
out_initial_file_name = "initial_state.nc"
out_folder_continuous = "."
out_continuous_file_name = "all_continuous_data.nc"
out_final_file_name = "final_state.nc"
out_merge_file_name = "ve_full_model_configuration.toml"

[core.layers]
soil_layers = [
    -0.25,
    -1.0,
]
canopy_layers = 10
above_canopy_height_offset = 2.0
surface_layer_height = 0.1
subcanopy_layer_height = 1.5

[core.data]
variable = [
    { file_path = "placeholder", var_name = "placeholder" },
    { file_path = "placeholder", var_name = "placeholder" },
]

[abiotic]
static = false

[abiotic.constants]
leaf_emissivity = 0.98
leaf_albedo = 0.15
bulk_density_soil = 1175.0
wind_reference_height = 10.0
latent_heat_vap_equ_factors = [
    1918460.0,
    33.91,
]
zero_plane_scaling_parameter = 7.5
substrate_surface_drag_coefficient = 0.003
roughness_element_drag_coefficient = 0.3
roughness_sublayer_depth_parameter = 0.193
max_ratio_wind_to_friction_velocity = 0.3
drag_coefficient = 0.2
min_windspeed_below_canopy = 0.1
min_roughness_length = 0.01
light_extinction_coefficient = 0.01
soil_thermal_conductivity = 1.206
specific_heat_capacity_soil = 881.0
surface_albedo = 0.125
soil_emissivity = 0.95
saturated_pressure_slope_parameters = [
    4098.0,
    0.6108,
    17.27,
    237.3,
]
dry_air_factor = 0.378
initial_flux_value = 0.001

[animal]
static = false
functional_group_definitions_path = ""
density_scaling_method = "madingley"

[animal.constants]
density_scaling_method = "damuth"
madingley_biomass_scaling_terms = [
    0.6,
    300000.0,
]
birth_mass_threshold = 1.5
flow_to_reproductive_mass_threshold = 1.0
dispersal_mass_threshold = 0.8
energy_percentile_threshold = 0.5
tau_f = 0.5
sigma_f_t = 1.0
alpha_0_herb = 1e-11
alpha_0_pred = 1e-06
phi_herb_t = 0.1
b_herb = 0.7
b_pred = 0.05
M_herb_ref = 1.0
M_herb_0 = 0.7
h_herb_0 = 0.7
M_pred_ref = 1.0
sigma_opt_pred_prey = 0.7
theta_opt_min_f = 0.01
theta_opt_f = 0.1
sigma_opt_f = 0.02
N_sigma_opt_pred_prey = 3.0
h_pred_0 = 0.5
m_tol = 1.6
c_tol = 6.61
m_tsm = 1.53
c_tsm = 1.51
M_disp_ref = 1.0
V_disp = 0.0278
o_disp = 0.48
beta_responsive_bodymass = 0.8
semelparity_mass_loss = 0.5
u_bg = 0.001
lambda_se = 0.003
lambda_max = 1.0
J_st = 0.6
zeta_st = 0.05
metamorph_mortality = 0.1
carbon_excreta_proportion = 0.9
nitrogen_excreta_proportion = 0.1
decay_rate_excrement = 0.25
scavenging_rate_excrement = 0.25
decay_rate_carcasses = 0.0625
scavenging_rate_carcasses = 0.25
migration_mortality = 0.1
aquatic_mortality = 0.1
aquatic_residence_time = 60.0
migration_residence_time = 60.0
seasonal_migration_probability = 0.083

[animal.constants.damuths_law_terms.mammal]
HERBIVORE = [
    -0.75,
    4.23,
]
CARNIVORE = [
    -0.75,
    1.0,
]
OMNIVORE = [
    -0.75,
    3.0,
]

[animal.constants.damuths_law_terms.bird]
HERBIVORE = [
    -0.75,
    5.0,
]
CARNIVORE = [
    -0.75,
    2.0,
]
OMNIVORE = [
    -0.75,
    3.0,
]

[animal.constants.damuths_law_terms.invertebrate]
HERBIVORE = [
    -0.75,
    5.0,
]
CARNIVORE = [
    -0.75,
    2.0,
]
OMNIVORE = [
    -0.75,
    3.0,
]

[animal.constants.damuths_law_terms.amphibian]
HERBIVORE = [
    -0.75,
    5.0,
]
CARNIVORE = [
    -0.75,
    2.0,
]
OMNIVORE = [
    -0.75,
    3.0,
]

[animal.constants.metabolic_rate_terms.endothermic]
basal = [
    41900000000.0,
    0.69,
]
field = [
    908000000000.0,
    0.7,
]

[animal.constants.metabolic_rate_terms.ectothermic]
basal = [
    41900000000.0,
    0.69,
]
field = [
    149000000000.0,
    0.88,
]

[animal.constants.energy_density]
meat = 7000.0
plant = 18200000.0

[animal.constants.conversion_efficiency]
HERBIVORE = 0.1
CARNIVORE = 0.25
OMNIVORE = 0.175

[animal.constants.mechanical_efficiency]
HERBIVORE = 0.9
CARNIVORE = 0.8
OMNIVORE = 0.85

[animal.constants.prey_mass_scaling_terms.endothermic]
mammal = [
    1.0,
    1.0,
]
bird = [
    1.0,
    1.0,
]

[animal.constants.prey_mass_scaling_terms.ectothermic]
invertebrate = [
    1.0,
    1.0,
]
amphibian = [
    1.0,
    1.0,
]

[animal.constants.cnp_proportion_terms.mammal]
carbon = 0.5
nitrogen = 0.3
phosphorus = 0.2

[animal.constants.cnp_proportion_terms.bird]
carbon = 0.4
nitrogen = 0.3
phosphorus = 0.3

[animal.constants.cnp_proportion_terms.invertebrate]
carbon = 0.4
nitrogen = 0.2
phosphorus = 0.4

[animal.constants.cnp_proportion_terms.amphibian]
carbon = 0.4
nitrogen = 0.2
phosphorus = 0.4

[hydrology]
static = false
initial_soil_moisture = 0.5
initial_groundwater_saturation = 0.9

[hydrology.constants]
soil_moisture_residual = 0.175
soil_moisture_saturation = 0.51
saturated_hydraulic_conductivity = 3.5e-05
hydraulic_gradient = 1.31
van_genuchten_nonlinearily_parameter = 1.598
stream_flow_capacity = 5000.0
intercept_parameters = [
    0.935,
    0.498,
    0.00575,
]
veg_density_param = 0.046
groundwater_capacity = 500.0
bypass_flow_coefficient = 1.0
air_entry_water_potential = -3.648
extinction_coefficient_global_radiation = 0.74
max_percolation_rate_uzlz = 2.7
groundwater_loss = 1.0
reservoir_const_upper_groundwater = 20.0
reservoir_const_lower_groundwater = 20.0
initial_aerodynamic_resistance_surface = 12.5
initial_aerodynamic_resistance_canopy = 12.1
drag_coefficient_evaporation = 0.2
intercept_residence_time = 86400.0
initial_stomatal_conductance = 1000.0
pore_connectivity_parameter = 0.5
air_entry_potential_inverse = 0.042
m_to_kpa = 9.804

[litter]
static = false

[litter.constants]
litter_decomp_reference_temp = 40.0
litter_decomp_offset_temp = 31.79
litter_decomp_temp_response = 3.36
litter_decay_water_potential_optimum = -10.0
litter_decay_water_potential_halt = -28800.0
moisture_response_curvature = 1.0
litter_decay_constant_metabolic_above = 0.08
litter_decay_constant_structural_above = 0.021714285714285714
litter_decay_constant_woody = 0.006666666666666667
litter_decay_constant_metabolic_below = 0.1
litter_decay_constant_structural_below = 0.02702702702702703
cue_metabolic = 0.45
cue_structural_above_ground = 0.55
cue_woody = 0.55
cue_structural_below_ground = 0.45
lignin_inhibition_factor = -5.0
max_metabolic_fraction_of_input = 0.85
metabolic_split_nitrogen_sensitivity = 0.018
metabolic_split_phosphorus_sensitivity = 0.0011613
structural_to_metabolic_n_ratio = 5.0
structural_to_metabolic_p_ratio = 5.0

[soil]
static = false
enzyme_class_definition_path = ""
microbial_group_definition_path = ""

[soil.constants]
reference_cue_logit = 0.094
cue_reference_temp = 20.0
logit_cue_with_temperature = -0.039
soil_microbe_water_potential_optimum = -3.0
soil_microbe_water_potential_halt = -15800.0
microbial_water_response_curvature = 1.47
min_pH_microbes = 2.5
lowest_optimal_pH_microbes = 4.5
highest_optimal_pH_microbes = 7.5
max_pH_microbes = 11.0
base_soil_protection = 0.694
soil_protection_with_clay = 1.36
solubility_coefficient_lmwc = 0.05
solubility_coefficient_ammonium = 0.05
solubility_coefficient_nitrate = 1.0
solubility_coefficient_labile_p = 0.005
necromass_decay_rate = 0.23104906018664842
maom_desorption_rate = 1e-05
lmwc_sorption_rate = 0.001
necromass_sorption_rate = 0.6931471805599453
litter_leaching_fraction_carbon = 0.0015
litter_leaching_fraction_nitrogen = 0.0015
litter_leaching_fraction_phosphorus = 0.0001
organic_proportion_litter_nitrogen_leaching = 1.0
organic_proportion_litter_phosphorus_leaching = 1.0
ammonium_mineralisation_proportion = 0.9
tectonic_uplift_rate_phosphorus = 0.0
ammonia_volatilisation_rate_constant = 8.64e-05
nitrification_rate_constant = 0.08639999999999999
denitrification_rate_constant = 0.21600000000000003
nitrification_optimum_temperature = 311.15
nitrification_maximum_temperature = 343.15
nitrification_thermal_sensitivity = 12
denitrification_infinite_temperature_factor = 93.34598
denitrification_minimum_temperature = 227.12999999999997
denitrification_thermal_sensitivity = 308.56
nitrogen_fixation_cost_zero_celcius = 59.19651970522086
nitrogen_fixation_cost_infinite_temp_offset = -0.8034802947791453
nitrogen_fixation_cost_thermal_sensitivity = 0.27
nitrogen_fixation_cost_equality_temperature = 50.28
free_living_N_fixation_reference_rate = 4.106776180698152e-06
free_living_N_fixation_reference_temp = 293.15
free_living_N_fixation_q10_coefficent = 3.0
primary_phosphorus_breakdown_rate = 2.2831050228310502e-07
secondary_phosphorus_breakdown_rate = 7.407407407407407e-05
labile_phosphorus_sorption_rate = 0.0016666666666666668
ammonium_deposition_rate = 4.1067761806981517e-07
phosphorus_deposition_rate = 1.3689253935660508e-08
nitrogen_fixer_supply_fraction = 0.3
ectomycorrhiza_supply_fraction = 0.25

[plants]
static = false
pft_definitions_path = ""
cohort_data_path = ""

[plants.community_data_export]
required_data = []
cohort_attributes = []
community_canopy_attributes = []
stem_canopy_attributes = []
float_format = "%0.5f"

[plants.constants]
per_stem_annual_mortality_probability = 0.1
per_propagule_annual_recruitment_probability = 0.2
dsr_to_ppfd = 2.04
stem_lignin = 0.545
senesced_leaf_lignin = 0.05
leaf_lignin = 0.1
plant_reproductive_tissue_lignin = 0.01
root_lignin = 0.2
subcanopy_extinction_coef = 0.5
subcanopy_specific_leaf_area = 14.0
subcanopy_respiration_fraction = 0.1
subcanopy_yield = 0.6
subcanopy_reproductive_allocation = 0.1
subcanopy_sprout_rate = 0.1
subcanopy_sprout_yield = 0.5
subcanopy_vegetation_turnover = 1.0
subcanopy_seedbank_turnover = 0.25
subcanopy_seedbank_c_n_ratio = 20.0
subcanopy_seedbank_c_p_ratio = 50.0
subcanopy_vegetation_c_n_ratio = 20.0
subcanopy_vegetation_c_p_ratio = 50.0
subcanopy_vegetation_lignin = 0.2
subcanopy_seedbank_lignin = 0.2
root_exudates = 0.5
propagule_mass_portion = 0.5
carbon_mass_per_propagule = 1.0

[abiotic_simple]
static = false

[abiotic_simple.constants]
saturation_vapour_pressure_factors = [
    0.61078,
    7.5,
    237.3,
]
initial_net_radiation = 10.0

[abiotic_simple.bounds]
air_temperature = [
    -20.0,
    80.0,
    -1.27,
]
relative_humidity = [
    0.0,
    100.0,
    5.4,
]
vapour_pressure_deficit = [
    0.0,
    10.0,
    -252.24,
]
wind_speed = [
    0.001,
    100.0,
    -0.1,
]
soil_temperature = [
    -10.0,
    50.0,
]

```
</details>

### Next steps

1. This PR is isolated from everything else at the moment, so could be merged.

1. We can add the new config build step into `ve_run` alongside the existing one - models don't use that, but we can iron out issues from tests on loading to that new style. It should still take dict or TOML inputs, so should be fine.
1. We can add Config into BaseModelABC definition - it's already there in CoreComponents, but I think it makes sense to make it a top level argument. Again - that can be added in parallel.
1. We _should_ then be able to update models one at a time to move from the current system to using the new one.
1. We should end up with a system where the old configuration is redundant and we can rip it out.

## Type of change

## Key checklist

## Further checks


dalonsoa:
I think this is the right move. Pydantic is way superior to JSON schemas for serialisation, more readable and, as pointed out, "documentable". To be honest, I do not know why we did not choose something like this back in the day, because we have been using Pydantic for other projects for quite a while.

davidorme:
> To be honest, I do not know why we did not choose something like this back in the day

@dalonsoa I think we basically pilfered the `Constants` classes from `pyrealm` and that then left a gap for the rest of the config. That was a mis-step because `pyrealm` is intended to be used within Python scripts and notebooks, where defining a Constants class instance in code makes sense, but VE needs to have a user friendly command line interface to a file format that runs validation.

 I don't remember why we used hand tooled JSON Schema rather than wrapping it up in `pydantic`. We'd have made the jump a lot earlier if we'd started there!

davidorme:
@dalonsoa We're going to go ahead with putting this through. Could have review this? 

I wouldn't spend any time looking at the `model_config.py` files themselves - they are basically just turning the JSON schema in pydantic and moving the model constants classes inside the config object. There are definitely things wrong with the structures in there, but they are as the models currently require, so fixing those up comes later.

I think the main things I'd like checked are the technical details in `registry.py` and `configuration.py`.

docs/source/using_the_ve/configuration/new_config.md
```diff
@@ -0,0 +1,310 @@
...
+
+config = CoreConfig()
+
+display(
+    Markdown(
+        "```toml\n"
+        + tomli_w.dumps({"core": json.loads(config.model_dump_json())})
+        + "```"
+    )
+)
```
davidorme: Worth noting here that `model_dump_json()` handles serialisation of Python objects, so to get a serialised TOML document, we need to dump a serialised JSON string, reload it to get a dict and then pass it into `tomli_w`. Which is a fiddle.

davidorme: Found a better way of doing it.

virtual_ecosystem/core/registry.py
```diff
@@ -184,6 +192,46 @@ def register_module(module_name: str) -> None:
                 class_name,
             )
 
+    # Find and register the model configuration
+    try:
+        config_submodule = import_module(f"{module_name}.model_config")
+    except ModuleNotFoundError:
+        raise RuntimeError("Model does not provide a model_config submodule.")
```
dalonsoa: Maybe indicate the offending model in the message:

```suggestion
        raise RuntimeError(f"Model `{module_name}` does not provide a model_config submodule.")
```

virtual_ecosystem/core/registry.py
```diff
@@ -184,6 +192,46 @@ def register_module(module_name: str) -> None:
...
+    # ModelConfigSection subclasses, which need to be filtered out.
+    model_config_subclasses = {
+        class_name: class_obj
+        for class_name, class_obj in getmembers(config_submodule)
+        if isclass(class_obj)
+        and issubclass(class_obj, ConfigRoot)
+        and not issubclass(class_obj, ModelConfigSection)
+        and class_obj is not ConfigRoot
+        and class_obj is not ModelConfigRoot
+    }
```
dalonsoa: Not sure if it has any advantage, but you can also do:

```suggestion
    model_config_subclasses = dict(
        getmembers(
            config_submodule, 
            lambda obj: (
                if isclass(obj)
                and issubclass(obj, ConfigRoot)
                and not issubclass(obj, ModelConfigSection)
                and obj is not ConfigRoot
                and obj is not ModelConfigRoot
            )
        )
    )
```

dalonsoa: Thinking about it, I wonder if it won't be better to just enforce a naming convention related to the model name instead of "searching" for a single config object. So, something like the following (in a separate function, otherwise this becomes enormous):

```python
def get_config_class(module_name):
    name = ''.join(word.capitalize() for word in module_name.split("_")) + "Config"

    try:
        config_submodule = import_module(f"{module_name}.model_config")
        model_config_class = getattr(config_submodule, name)
        assert issubclass(model_config_class, ModelConfigRoot)
    except ModuleNotFoundError:
        raise RuntimeError("Model does not provide a model_config submodule.")
    except AttributeError:
        raise RuntimeError(
            f"Root config class {name} not found for model {module_name} in its `model_config` module."
        )
    except AssertionError:
        raise RuntimeError(
            f"Model {module_name} config class does does inherit from `ConfigRoot`."
        )

    return model_config_class
```

Or something along these lines. I feel this is cleaner.

virtual_ecosystem/core/configuration.py
```diff
@@ -0,0 +1,151 @@
+"""Configuration system elements for pydantic.
+
+The :mod:`~virtual_ecosystem.core.config` module is used to read in the various
+configuration files, validate their contents, and then configure a ready to run instance
+of the virtual ecosystem model. The basic details of how this system is used can be
+found :doc:`here </using_the_ve/configuration/config>`.
+
+The validation of configuration documents is done using JSONSchema documents associated
+with the different model components. See the :mod:`~virtual_ecosystem.core.schema`
+module for details.
```
dalonsoa: No longer true...

virtual_ecosystem/core/configuration.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Args:
+        model_name: The name of the model as it would appear in a configuration file.
+        config_object: A ModelConfig instance
+    """
+
+    def __init__(
+        self,
+        model_name: str,
+        config_object: type[ModelConfigRoot] | type[ModelConfigSection],
```
dalonsoa: If it can be either one of these... then it must be a `ConfigRoot`, right?

```suggestion
        config_object: type[ConfigRoot],
```

virtual_ecosystem/core/configuration.py
```diff
@@ -0,0 +1,151 @@
...
+placeholder value can be written despite not being an existing file.
+
+.. TODO: Fix autodoc
+    This generates a bizarre set of autodoc link failures that try and build links from
+    the text elements from the Annotated pattern. Currently tackled using nitpick 
+    ignore.
+"""
+
+
+class ModelConfigHTMLTable:
```
dalonsoa: I don't think this deserves a class. A simple recursive function will work fine. You just need to be careful of including some html bits in the top call and treat rows as strings all the time, rather than elements in a list:

```python
def model_config_to_html(model_name, config_object, is_table=True):
    rows = ""
    ...
    if isinstance(field_info.annotation, ModelMetaclass):
        rows += model_config_to_html(model_name, config_object, False)

    ...
    return f"<table><tbody>{rows}<tbody><table>" if is_table else rows
```
I prefer to use a functional approach whenever possible.

virtual_ecosystem/core/configuration.py
```diff
@@ -0,0 +1,151 @@
...
+    are used as nested fields within the root configuration but can be only one
+    :class:`ModelConfigRoot` class per model. This base model sets common shared
+    attributes across models: currently just the shared ``static`` option.
+    """
+
+    static: bool = False
+    """The model static mode setting."""
+
+
+class ModelConfigSection(ConfigRoot):
```
dalonsoa: I'm not entirely sure why or when you need to make this distinction. `ModelConfigRoot` makes sense to have it as a separate bit, but the sections... What's exactly the use case?

dalonsoa (commented):
I'm made a few comments on the two files you mentioned, only, not approving or rejecting, yet.

virtual_ecosystem/core/configuration.py
davidorme: Yup - forgot to update these docstrings.

virtual_ecosystem/core/configuration.py
davidorme: It's primarily to be able to unambiguously identify a single root config class within modules. The difference is basically because of `core`, which should not have a `static` setting. If we we move to a naming convention for identifying the root config class in any given module rather than just using class identity (as suggested below) then we can remove this and rename the others to `Configuration(BaseModel)` and `ModelConfigurationRoot(Configuration)`.

I've done this in the edits.

virtual_ecosystem/core/configuration.py
davidorme: Agreed - it was a function and then I just couldn't work out how to avoid jamming extra `<table>` tags in for nested tables. It was probably late | early | before coffee

virtual_ecosystem/core/registry.py
davidorme: I think that's right. 

I'd probably go further and not even have a `module_name` component. Model developers have to provide `xyz_model.model_config.ModelConfiguration` as the root configuration class. That's what I've implemented in the edits.

davidorme:
@dalonsoa Thanks. I've just changed the simple stuff. On the more interlocking things, I have:

* Simplified the pydantic base classes to just two: `Configuration` and `ModelConfigurationRoot`
* Updated root model configuration registration to look explicitly for `XYZ.model_config.ModelConfiguration`
* Update the new `model_config` modules to match that expectation.

If that all sounds good - are you happy to approve? I'm sure the actual model config definitions could be massively improved but that seems like we can do it incrementally.

dalonsoa (approved):
This looks much cleaner now!

PR #1096, opened 2025-10-17
Add configuration to BaseModel.from_config

### LINKED ISSUE #1089: Add the new configuration object into BaseModelABC definition
Add the resulting generated config model into BaseModelABC definition. We already have the current `Config` object in CoreComponents, but I think it makes sense to make it a top level argument. Again - that can be added in parallel.

davidorme:
# Description

This PR:

* Adds a new `configuration` argument to the `BaseModel.from_config()` method that feeds the new style compiled configuration object into the models.
* Updates the various `ActualModel.from_config()` implementations to match.
* Within models,  extracts the model specific configuration settings and uses them to set `static` from the new model configuration as a proof of concept.
* Updates testing to add fixtures and build configurations where needed.

The PR edits lots of files, but they are basically all the same thing repeated across the different model implementations.

Fixes #1089 (issue)

## Type of change

## Key checklist

## Further checks


davidorme:
@dalonsoa One more PR on this and then I'll be handing off the PRs on roll out to specific models to the model owners, so you're off the hook 😄

dalonsoa (approved):
Looks good! As far as I can tell, it is just hooking the new configuration style into the workflow, without removing the old configuration just yet, right?

davidorme:
Exactly. Should have said that in the description. The models all still use the old style but can be hooked in one by one until eventually the old style is unused and we can then rip out the old Config, constants classes and JSON schema.

PR #1103, opened 2025-10-21
Switch core over to new configuration

### LINKED ISSUE #1101: Switch `core` over to new configuration.
Move the core model over to using the new configuration classes.

davidorme:
# Description

This PR updates the whole of the `core` model to using the new configuration system. 

The big changes are:

* The `core.core_component` and `core.grid` modules have been updated to use data provided from `core.model_config` classes, removing all usage of the old `Config` dictionary.
* The `core.model_config` module has been updated to improve the docstrings and to provide better validation and class structure.
* The `soil.model_config` has been updated. I'd been planning to move forward with moving soil config into CSV files and had left a lot of structure out of this file, which is now provided. I still think we should move to CSV, but not within the #1085  meta PR

The other changes are utility fixes, improvements

* Made the new configuration classes frozen and also stopped them silently ignoring unknown inputs.
* Added a TOML output method to `CompiledConfiguration`
* Some of the core tests were config validation within the setup - these tests are all now more appropriate at the level of the `core.model_config` classes since that is one of the key advantages of the new system. Bad configs basically can't be created.

Everything else is test fixes:
* Core stuff and the core fixtures all swap out the old style Config dict for the new classes.
* Other model tests are updated to fix. These mostly maintain a dual system for now until they can be swapped over. Several of the `Model.from_config` tests currently throw in a test case with bad config inputs. As noted above - this is no longer a thing under the new system - the configuration classes auto-validate on loading before they even get near a `from_config` method, so a bad Configuration object is no longer possible. I've propped these tests up for now but those test cases can be retired as models are updated.

Fixes #1101 

## Type of change

## Key checklist

## Further checks


davidorme:
@jacobcook1995 Do you think you'd have time to look at this before you leave? There are quite a lot of files, but it is basically swapping out configuration settings to maintain the same set of tests. 

I've edited tests that incorporated config validation, since that now all happens much earlier.

jacobcook1995:
@davidorme if it's ready to review now I can take a look today

docs/source/conf.py
```diff
@@ -184,6 +184,16 @@ class MyReferenceStyle(AuthorYearReferenceStyle):
     ("py:class", "PosixPath"),
     ("py:class", "virtual_ecosystem.core.configuration.placeholder_validator"),
     ("py:class", "PydanticUndefined"),
+    # Actually just generally intersphinx is janky when it comes to pydantic
```
jacobcook1995, line 187: Outside the scope of this PR but I am wondering if there's any alternative to `intersphinx`, given it seems to be janky for both `pydantic` and `numpy`, neither of which are obscure packages?

tests/core/test_model_config.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+
+@pytest.mark.parametrize(
+    argnames="surf, subc, raises",
+    argvalues=[
+        (0.1, 2, does_not_raise()),
+        (1.2, 1.21, does_not_raise()),
+        (1.2, 0.1, pytest.raises(ValidationError)),
+    ],
```
jacobcook1995, line 62: Might be worth adding a repeated value case here as that's one of the things that your custom soil depths checker checks

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -307,7 +305,6 @@ def _setup(
         """Function to setup up the soil model."""
 
         self.model_constants = model_constants
-        self.core_constants = core_constants
```
jacobcook1995, line 310: Is this change because the core constants are now always an attribute of the BaseModel, so by inheritance they are a property of the SoilModel? (or am I misunderstanding this change?)

virtual_ecosystem/models/soil/model_config.py
```diff
@@ -375,6 +376,45 @@ class SoilConstants(Configuration):
...
+    half_sat_nitrate_uptake: float = Field(default=0.02275)
+    max_uptake_rate_labile_p: float = Field(default=0.0025)
+    half_sat_labile_p_uptake: float = Field(default=0.02275)
+    turnover_rate: float = Field(default=0.005)
+    activation_energy_turnover: float = Field(default=20000)
+    reference_temperature: float = Field(default=12.0)
+    c_n_ratio: float = Field(default=5.2)
+    c_p_ratio: float = Field(default=16)
+    enzyme_production: dict[str, float] = Field(default={"pom": 0.005, "maom": 0.005})
+    reproductive_allocation: float = Field(default=0.0)
```
jacobcook1995, line 415: This is definitely a massive improvement to the JSON schema. I guess further improvements are best left to further down the line

jacobcook1995 (approved):
LGTM! A few minor comments + queries but nothing that should prevent merging

docs/source/conf.py
davidorme: I think its the whole of `autodoc` really. All of the aliases and typing hints are hard to get right, but there are just so many inexplicable failures and the diagnostics on how to fix them are non-existent.

Ideally, I'd like to move to a full Markdown system (no RST docstrings) that supports MyST, and there's no good solution yet. I keep an eye on `mkdocs` but it would be a big jump at the moment - no MyST support for example and would need to change all the links.

virtual_ecosystem/models/soil/soil_model.py
davidorme: `core_constants` were _always_ a property of the base model and present in the soil model by inheritance 😄 

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/8ad15e114a0c7500592f5e30b8b24cbe668e5aaf/virtual_ecosystem/core/base_model.py#L244-L245

virtual_ecosystem/models/soil/model_config.py
davidorme: Yup - it is _wildly_ easier to write, maintain, use and gives us so much more freedom to add validation.

Actually, I am 99% sure that `pydantic` is actually implementing the validation using auto generated JSON Schema under the hood but we _never need to look at it again_ 🥳

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: Ahh whoops 🫠

tests/core/test_model_config.py
davidorme: Done - actually it allowed identical values so I've also updated the validator to disallow it.

Issue #1104, opened 2025-10-22
Document new configuration system


davidorme:
We need to add docs to describe the new configuration system:

* Replacing the existing schema stuff in the guide to creating new models.
* Update user facing demo to something more complete.

PR #1122, opened 2025-10-27
Move configuration use in main over to new configuration

### LINKED ISSUE #1118: Move configuration use in main over to new configuration
The current state of develop replaces the use of the old style config in generating the models, but I haven't updated the rest of the core stuff in main. That needs to drop out before `Config` etc can be retired in #1091

davidorme:
# Description

This PR removes the last remaining uses of the old configuration system and strips out all remaining calls to old `Config` objects in the model code and tests.

The PR **does not**

*  Remove the old implementations yet - the modules are still there but are unused. That will be the next PR (#1091) and should remove `core/constants.py`, `core/constants_loader.py` and `core/config.py` and all associated tests. That PR should also migrate any appropriate tests of the old system onto the new system.
* Systematically update the docs - thats for another PR too (#1104).

A lot of files are affected. The broad changes are:

* Tests of input validity when using `model.from_config` - because the new system pre-validates, these checks are redundant  and can be removed.
* Removal of the now superseded `from_config(..., config, ...)` argument from the BaseModel and then everywhere else. This was needed in parallel during the switch over but #1116, #1114,  #1115 and #1124 have replaced all cases where this input was used.
* Removing any imports from the modules that will be removed.
* Skipping the testing of the old system - this uses autouse fixtures to skip the whole test module. I've retained them rather than deleting them to allow me to cannibalise them in the next PR to strip the code.
* Removing expected schema and constant loading logger entries from tests, since these are no longer emitted.
* A few cases of un-typing variables in animal tests - some checking was moaning about partially typed functions. We don't type the tests, so I've simply retained the types as comments.
* Removal of fixtures providing old-style configs.
* A couple of missed cases where old configs were still being used in tests.
* Crippled the old style `config_merge` in `core/config.py`  - it was looking for schema that no longer exist. The same is true for `load_constants` - it was looking for the now removed constants classes from the registry. Both function have had the problematic code commented out.
* Added an info class attribute onto the `CompiledConfiguration` class that gets populated with a list of the models and BaseModel subclasses  used in it. This was present in the old `Config` and is used in `main` for getting run order and other things.
* Updated the core `load_data_config` to use the new configuration and also simplified the signatures of functions in `core/data.py` that were passing in whole config dicts but only needed a single path out of them.
* Stripped schema and old style constants dataclasses from the Model registry.

That's about it 😬 

Fixes #1118

## Type of change

## Key checklist

## Further checks


davidorme:
@sallymatson Finally got docs fixed. Could you have a look? It's very sprawling so please add as many "What the hell is happening here" questions as you need! Or we can chat IRL 😄

sallymatson (approved):

PR #1138, opened 2025-11-06
Avoiding BaseModel imports in model __init__

### LINKED ISSUE #1136: Simplify model registration structure.
This is an experimental branch to test out an option.

At the moment, all models automatically import their BaseModel in `__init__.py`. That, **I think** is purely because the submodules containing the BaseModel have variable names, and so importing the BaseModel at module `__init__` simplifies searching for it.

However, that approach basically forces the import of the most complex part of the model at `__init__`. That makes it very hard to share components between models without getting circular imports.

This branch is to explore removing that import.

davidorme:
As noted in #1136, we currently use a pattern where we import the `BaseModel` subclass of a model within the model `__init__.py`. If I recall correctly, this is simply to make it easier to find that model class - the import is not used within `__init__.py` (and in fact we have to mark them all to placate `ruff` warning us about unused imports.

That has the serious downside that it makes it impossible to share components between models without importing the BaseModel subclass module, which is often complex. The end result is that importing anything from one model drags in a whole load of imports and makes circular imports very hard to avoid.

The motivating example here is that the `abiotic` and `abiotic_simple` configs are extremely similar and we use `abiotic_simple` methods in `abiotic`: we want to share these definitions, but it is finicky because of these import chains. 

I _think_ it gets a whole lot simpler if we simply adjust the `register_model` function to look for the BaseModel in it's actual source file rather than expecting to import it from the model root namespace. That's really not much more complex.

We could do this by expecting to always look for the same file name (`plants.model`) etc. That would be cleaner. But for the moment, the pattern `plants.plants_model` works just fine.

Fixes #1136

## Type of change

## Key checklist

## Further checks


davidorme:
@dalonsoa Does this make sense to you? It just seems simpler and less intrusive to me and I think it'll make it easier to use model components across models.

virtual_ecosystem/core/registry.py
```diff
@@ -149,6 +110,63 @@ def register_module(module_name: str) -> None:
...
+    # Trap missing and multiple models
+    if len(models_found) == 0:
+        msg = f"Model object not found in {module_name}"
+        LOGGER.critical(msg)
+        raise RuntimeError(msg)
+
+    if len(models_found) > 1:
+        msg = f"More than one model defined in in {module_name}"
+        LOGGER.critical(msg)
+        raise RuntimeError(msg)
```
dalonsoa: Wasn't the case that all models follow the name convention that if the model short name is `my_model_name` then the model calss is `MyModelNameModel`? If that's the case, and as done with the pydantic configuration classes, this whole thing can be reduced a lot.

virtual_ecosystem/core/registry.py
```diff
@@ -149,6 +110,63 @@ def register_module(module_name: str) -> None:
...
+
+    # Trap models that do not follow the requirement that the BaseModel.model_name
+    # attribute matches the virtual_ecosystem.models.model_name
+    # TODO - can we retire the model_name attribute if it just duplicates the module
+    #        name or force it to match programmatically.
```
dalonsoa, line 154: Probably.

dalonsoa (approved):
Have a minor comment, but the change makes total sense. Not sure the original reason for the design, to be honest.

virtual_ecosystem/core/registry.py
davidorme: Agreed - I've updated it to look for `model_name.model_name_model.ModelNameModel`

virtual_ecosystem/core/registry.py
davidorme: I'll park it for now. Again.

PR #1139, opened 2025-11-07
Standalone configuration models for abiotic and abiotic simple models

### LINKED ISSUE #1137: Resolve shared constants between abiotic and abiotic simple to allow configuration.
The `abiotic` and  `abiotic_simple` models reciprocally share constants and bounds and at the moment one or the other is not configurable.

davidorme:
# Description

This PR restructures the configuration of the `abiotic` and `abiotic_simple` models to provide a single standalone configuration for each model. The two models no longer need to supply both `abiotic_constants` and `abiotic_simple_constants`. This is important to make the models fully configurable - users need to be able to configure all settings for , for example, the `abiotic` model from within the `[abiotic]` configuration section.

It does this by:

* Creating a small shared constants configuration dataclass (`AbioticSharedConstants`) that contains only soil and leaf emissivity. These constants are required by both models.
* The `AbioticConstants` and `AbioticSimpleConstants` then build on that shared class to add model specific configuration. For the simple model, that is just `initial_net_radiation` - there was also a  `saturation_vapour_pressure_factors` field but that was completely unused in the model code. For the full model, there are a lot of other constants.
* The Abiotic configuration now imports and includes the AbioticSimpleBounds as the `[abiotic.bounds]` configuration section. So, these bounds are defined in a single place but are now available directly from within a single model config.

The PR then updates the model code to remove all of the parallel usage of `abiotic` and `abiotic_simple` constants and just uses the single model configuration.

Closes #1137

## Type of change

## Key checklist

## Further checks


vgro (approved):
LGTM

Issue #1154, opened 2025-11-20
Energy balance for understory vegetation


vgro:
The understory vegetation does currently absorb a lot of light but has no energy balance. As a result, the soil below freezes. There needs to be a full energy balance of this vegetation layer including heat transfer to the soil.

PR #1155, opened 2025-11-20
1022 get canopy temperatures to converge

### LINKED ISSUE #1022: get canopy temperatures to converge
The current implementation of the Newton method for canopy temperatures produces some wild values when running the full model. This had likely to do with instabilities in the Newtin method implementation.
I want to try to get back to the version that does not use the scipy methods to see if that works and rebuild from there.

vgro:
It turns out that the Newton method converges when the input parameters are sensible, as confirmed by new logging message.

The aerodynamic resistance seems to have quite a strong influence and is currently extremely high due to the canopy structure. My tentative solution is to set a fixed resistance for the canopy (to be solved later in calibration - reopened #967) to find out what else is causing the problems (the model runs now much longer until the soil freezes, to be addressed in #1154).

Also renamed a parameter in `wind.py` which was wrong in the reference...

Fixes #1022 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/abiotic/energy_balance.py
+++ b/virtual_ecosystem/models/abiotic/energy_balance.py
@@ -458,7 +458,7 @@ def solve_canopy_temperature(
                     ),
                     density_air=np.array([[density_air[i, j]]], dtype=np.float64),
                     aerodynamic_resistance=np.array(
-                        [[aerodynamic_resistance[i, j]]], dtype=np.float64
+                        [[aerodynamic_resistance[i]]], dtype=np.float64
                     ),
                     latent_heat_vapourisation=np.array(
                         [[latent_heat_vapourisation[i, j]]], dtype=np.float64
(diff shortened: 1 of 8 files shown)


virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -495,9 +499,27 @@ def tracked_func(x):
...
+            "row": i,
+            "col": j,
+            "converged": converged,
+            "final_value": solved_temperature[i, j],
+            "best_estimate": best_estimate[0],
+            "history": iteration_history,
+        }
+    )
+
+    LOGGER.info(
```
jacobcook1995: Maybe it's worth having a different message here if all cells converge?

jacobcook1995 (approved):
Yep as we discussed agree that it makes sense to set fix the resistance for now so that the errors it causes don't get in the way of trouble shooting elsewhere

PR #1156, opened 2025-11-20
atmospheric pressure and co2 only in true atmosphere

### LINKED ISSUE #484: Abiotic simple - atmospheric pressure and CO2 should only be filled for filled/true above ground layers
At the moment, atmospheric pressure and CO2 are copied from the reference value above the canopy to all above ground layers, including the layers that have no canopy.

The solution is the adjust the `run_microclimate` function so that it only fills for filled/true above ground layers.

vgro:
We used to fill the whole atmosphere with values before we had the layer structure in place. Now updated in simple abiotic model.

Fixes #484 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM!,

Issue #835, opened 2025-04-24
Calculations of limits of mycorrhizal supply for first model step is a very rough estimate


jacobcook1995:
The function I use to calculate the maximum amount of nutrients that mycorrhizal fungi will supply to plants (`calculate_symbiotic_supply_limits`), requires information about the abiotic environment that isn't available at initialisation.

So, for variable initialisation the supply limits are calculated under the assumption that conditions are good (i.e. temperature is 25C and soil moisture is optimal for microbial growth). This could lead to a significant over estimate of the amount of nutrients symbiotes can provide for the first model time step.

I don't think this is an issue that we should try and fix now (the only fix I can see is changing the order at which things are calculated in the abiotic and hydrology models). I'm mainly just write out the issue in case we come across weird behaviour relating to this when we are calibrating the model.

jacobcook1995:
This is at least part of reason why mycorrhizal fungi can end up with negative populations, as the initial estimates for the amount plants can take are double what they would be if the correct temperatures and soil water potential were used

jacobcook1995:
The `calculate_symbiotic_supply_limits` function has been retired entirely, so this issue in no longer relevant

PR #1163, opened 2025-11-27
Change mycorrhizal supply calculation to prevent negative populations

jacobcook1995:
# Description

Negative mycorrhizal populations were occurring after the initial time step because the calculation of the maximum amount of nutrient that they can supply to the plant was using guesses for temperature and matric potential. This resulted in mycorrhizal fungi promising far more nutrient than they could provide (even from their existing biomass). I have partially addressed this overpromising by using mean annual temperature as a guess for the soil temperature (which is a fairly common assumption) and by calculating initial values for matric potential based on the initial values for soil moisture.

This partly addresses #835, but does not fully resolve it, as there is still no initial value for soil temperatures, and the initial value of matric potential could be significantly off. This is arguably an issue that won't be completely fixed until we have decided an approach to model spin up.

@vgro, one thing we should decide is whether the calculations make sense in their current location, or are better moved to the hydrology model init? (happy to do the work to move them if so)

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/soil/test_soil_model.py
+++ b/tests/models/soil/test_soil_model.py
@@ -41,6 +41,8 @@ REQUIRED_INIT_VAR_LOG = (
     (DEBUG, "soil model: required var 'soil_p_pool_labile' checked"),
     (DEBUG, "soil model: required var 'pH' checked"),
     (DEBUG, "soil model: required var 'clay_fraction' checked"),
+    (DEBUG, "soil model: required var 'soil_moisture' checked"),
+    (DEBUG, "soil model: required var 'mean_annual_temperature' checked"),
 )
 POST_SETUP_LOG = (
     *REQUIRED_INIT_VAR_LOG,
@@ -76,6 +78,9 @@ def test_soil_model_initialization(
         enzyme_classes=enzyme_classes,
         soil_moisture_saturation=fixture_hydrology_constants.soil_moisture_saturation,
         soil_moisture_residual=fixture_hydrology_constants.soil_moisture_residual,
+        air_entry_potential_inverse=fixture_hydrology_constants.air_entry_potential_inverse,
+        van_genuchten_nonlinearily_parameter=fixture_hydrology_constants.van_genuchten_nonlinearily_parameter,
+        m_to_kpa=fixture_hydrology_constants.m_to_kpa,
     )
 
     # In cases where it passes then checks that the object has the right properties
@@ -162,6 +167,9 @@ def test_soil_model_initialization_bounds_error(
             enzyme_classes=enzyme_classes,
             soil_moisture_saturation=fixture_hydrology_constants.soil_moisture_saturation,
             soil_moisture_residual=fixture_hydrology_constants.soil_moisture_residual,
+            air_entry_potential_inverse=fixture_hydrology_constants.air_entry_potential_inverse,
+            van_genuchten_nonlinearily_parameter=fixture_hydrology_constants.van_genuchten_nonlinearily_parameter,
+            m_to_kpa=fixture_hydrology_constants.m_to_kpa,
         )
 
     # Final check that expected logging entries are produced
@@ -195,6 +203,9 @@ def test_soil_model_all_pools_positive(
         enzyme_classes=enzyme_classes,
         soil_moisture_saturation=fixture_hydrology_constants.soil_moisture_saturation,
         soil_moisture_residual=fixture_hydrology_constants.soil_moisture_residual,
+        air_entry_potential_inverse=fixture_hydrology_constants.air_entry_potential_inverse,
+        van_genuchten_nonlinearily_parameter=fixture_hydrology_constants.van_genuchten_nonlinearily_parameter,
+        m_to_kpa=fixture_hydrology_constants.m_to_kpa,
     )
 
     assert soil_model._all_pools_positive()
@@ -564,6 +575,7 @@ def test_order_independance(
         "animal_ectomycorrhiza_consumption",
         "animal_arbuscular_mycorrhiza_consumption",
         "decay_of_fungal_fruiting_bodies",
+        "mean_annual_temperature",
     ]
     for not_pool in not_pools:
         new_data[not_pool] = dummy_carbon_data[not_pool]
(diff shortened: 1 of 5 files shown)


virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -315,9 +327,12 @@ def _setup(
         self.microbial_groups = microbial_groups
         self.enzyme_classes = enzyme_classes
 
-        # Store the two required hydrology constants
+        # Store the four required hydrology constants
```
vgro: might be worth dropping the number here so you don't have to change this everytime you change the constants
```suggestion
        # Store required hydrology constants
```

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -661,13 +678,29 @@ def calculate_symbiotic_supply_limits(self, init: bool) -> dict[str, DataArray]:
...
             )
         else:
-            # Want to establish the maximum for optimal conditions, so use the water
-            # potential optimum from the constants, and arbitrarily select a soil
-            # temperature of 25C as "optimal"
-            soil_temperature = np.full_like(self.data["pH"].to_numpy(), 25.0)
-            soil_water_potential = np.full_like(
-                self.data["pH"].to_numpy(),
-                self.model_constants.soil_microbe_water_potential_optimum,
+            # As soil temperature hasn't be calculated yet we assume that it matches the
```
vgro, line 657: soil temperature is initialised in the abiotic simple model but no values are assigned

```
# create soil temperature array
self.data["soil_temperature"] = self.layer_structure.from_template()
```

I think it would make sense to just add the mean annual temperature there. I'm happy for you to do that if that helps here

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -626,13 +641,15 @@ def calculate_symbiotic_supply_limits(self, init: bool) -> dict[str, DataArray]:
...
-        temperature and matric potential are known. At present these are only found once
-        the abiotic and hydrology models (respectively), are updated. Instead a
-        temperature of 25C and optimal water potential (set in the constants) are used.
-        Down the line we need to decide whether this inaccuracy is acceptable, or
-        whether the stage at which basic abiotic information gets calculated needs to
-        change.
+        TODO - This function currently calculates matric potential based on soil
+        moisture when being used within model initialisation. At some point this should
+        be migrated to the hydrology model init/setup as that is a more natural home for
+        this calculation
```
vgro: I agree it would make sense to move this to the hydrology init

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -661,13 +678,29 @@ def calculate_symbiotic_supply_limits(self, init: bool) -> dict[str, DataArray]:
...
-            # Want to establish the maximum for optimal conditions, so use the water
-            # potential optimum from the constants, and arbitrarily select a soil
-            # temperature of 25C as "optimal"
-            soil_temperature = np.full_like(self.data["pH"].to_numpy(), 25.0)
-            soil_water_potential = np.full_like(
-                self.data["pH"].to_numpy(),
-                self.model_constants.soil_microbe_water_potential_optimum,
+            # As soil temperature hasn't be calculated yet we assume that it matches the
+            # mean annual temperature (a common assumption for deeper soil layers)
+            soil_temperature = self.data["mean_annual_temperature"].to_numpy()
```
vgro: note that the mean annual temperature has no `layer` dimension

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: Oh yeah that massively simplifies the work flow on my side!

vgro (changes requested):
Looks sensible to me, I agree that moving the initialization to the abiotic and hydrology model might make it simpler and I'm happy for you to do that. The set of variables that are initialized with values was purely based on my requirements, there is no reason why that shouldn't include more variables.
Let me know when you made the changes and I'll approve the PR

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: Actually, scratch that I would need both the soil temperature and the air temperature (for the surface temperature) to be populated. I guess populating the soil temperature with the mean annual temperature is an acceptable initial guess but that doing the same with air temperature wouldn't be?

jacobcook1995:
I've moved the matric potential calculation into the soil hydrology init, let me know if that all seems reasonable!

I kept the soil temperature stuff unchanged as I didn't think populating the air temperature with the mean annual temperature seemed like a sane thing to do (though I agreed that it seemed reasonable for the soil temperature).

virtual_ecosystem/models/soil/soil_model.py
vgro: yeah that's true. The `abiotic` model initializes everything but the `abiotic_simple` doesn't. Ideally they would have the same set of variables initialized and updated...I think for now your solution is good, it might be worth changing the init of the simple model in case that comes up elsewhere, but that can be my TODO.

tests/models/soil/test_soil_model.py
```diff
@@ -76,6 +78,9 @@ def test_soil_model_initialization(
         enzyme_classes=enzyme_classes,
         soil_moisture_saturation=fixture_hydrology_constants.soil_moisture_saturation,
         soil_moisture_residual=fixture_hydrology_constants.soil_moisture_residual,
```
vgro, line 80: I don't think these hydrology constants are needed any longer if you move everything to the hydrology model?

vgro (approved):
LGTM

tests/models/soil/test_soil_model.py
jacobcook1995: I use these to calculate the effective saturation, which some of the soil model processes need. I guess this makes sense to stay (at least for now) as the alternative would be added "effective_saturation" as something hydrology outputs to the data object

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: Okay, I'll add a TODO to the docstring of abiotic simple model init about this

PR #1170, opened 2025-12-02
removed understorey vegetation from leaf area index sum

### LINKED ISSUE #1169: Remove understorey vegetation from LAI sum in abiotic simple
The abiotic simple model is based on observations of temperatures at 1.5 m below a canopy. At the moment the model sums the LAI over all layers, including the understorey layer which has a very high LAI. That causes the soil to be freezing. Although the understorey vegetation takes some light from hitting the soil, the LAI is currently high and the cooling unrealistic.

I propose removing the understorey vegetation from the summing step. This does not solve the issue of a canopy being lower than 1.5m, but it gives a correct representation of the initial assumptions.

vgro:
The abiotic simple model is based on observations of temperatures at 1.5 m below a canopy. At the moment the model sums the LAI over all layers, including the understorey layer which has a very high LAI. That causes the soil to be freezing. Although the understorey vegetation takes some light from hitting the soil, the LAI is currently high and the cooling unrealistic.

I propose removing the understorey vegetation from the summing step. This does not solve the issue of a canopy being lower than 1.5m, but it gives a correct representation of the initial assumptions.

Fixes #1169 

## Type of change

## Key checklist

### DIFF
--- a/virtual_ecosystem/models/abiotic_simple/microclimate_simple.py
+++ b/virtual_ecosystem/models/abiotic_simple/microclimate_simple.py
@@ -101,6 +101,9 @@ def run_simple_microclimate(
     output = {}
 
     # Sum leaf area index over all canopy layers, [m m-1]
+    # TODO This step excludes the understorey vegetation, assuming that the relationship
+    # between LAI and the variables is purely based on the vegetation above the 1.5m
+    # reference height.
     leaf_area_index_sum = data["leaf_area_index"][
         layer_structure.index_filled_canopy
     ].sum(dim="layers")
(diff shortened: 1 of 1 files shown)


virtual_ecosystem/models/abiotic_simple/microclimate_simple.py
```diff
@@ -101,7 +101,9 @@ def run_simple_microclimate(
     output = {}
 
     # Sum leaf area index over all canopy layers, [m m-1]
-    leaf_area_index_sum = data["leaf_area_index"].sum(dim="layers")
+    leaf_area_index_sum = data["leaf_area_index"][
+        layer_structure.index_filled_canopy
+    ].sum(dim="layers")
```
jacobcook1995, line 109: Only query would be whether it's worth adding a TODO to revisit this at some point when you are further along with the vegetation heat transfer model that you mentioned? (I'm not 100% on that though)

jacobcook1995 (approved):
LGTM, and can confirm that that change gives me more reasonable temperatures to work with

PR #1171, opened 2025-12-02
Change the data variable C, N, P variables to triplets

### LINKED ISSUE #1130: Change CNP triplets to singular DataArrays
- Any place where CNP values are exported for a tissue/biomass pool, this should be changed to a `DataArray` object.

sallymatson:
# Description

A variable refactor to eliminate lots of cluttery stoichiometric data variables. Any previous C, N, and P triplets (i.e. values or ratios for the same tissue type) will be combined into one DataArray that adds a dimension for CNP and can be easily accessed using `.loc[CELL_ID, "C"]` or "N" / "P" as relevant. I also tried to standardize some naming conventions e.g. deadwood -> stem turnover. 

I haven't done much on the plant reproductive tissue, given that we've developed a new concept and I think it'll be much easier to process that in a separate pull request. 

Fixes #1130

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -557,53 +610,51 @@ def spinup(self) -> None:
...
-        for var in by_pft_vars:
+        # Initialize variables that are stored by cell x PFT x CNP
+        pft_cell_template = xr.DataArray(
+            data=np.zeros((self.grid.n_cells, self.flora.n_pfts, 3)),
+            coords={
+                "cell_id": self.data["cell_id"],
+                "pft": self.flora.name,
+                "element": ["C", "N", "P"],
+            },
+        )
```
davidorme: This is duped from `__init__`. Is it worth creating it there and storing it as an instance attribute? We could also use the same template to get `cnp_template` above (cell x CNP) by simply dropping the PFT dimension when we use it with non-PFT structured variables? It's more compact - arguably a little more obscure, so maybe having two separate templates in `__init__` would be cleaner (cell x pft x CNP, cell x CNP).

virtual_ecosystem/models/plants/subcanopy.py
```diff
@@ -437,6 +438,11 @@ def calculate_dynamics(
         # variables are created in the first update, so easier to just write afresh.
         coords = {"cell_id": self.data["cell_id"].data}
 
+        cnp_template = xr.DataArray(
+            data=np.zeros((len(self.data["cell_id"].data), 3)),
+            coords={"cell_id": self.data["cell_id"], "element": ["C", "N", "P"]},
+        )
```
davidorme: Also duped from the above, but since this is a separate function, we'd need to pass it in as an argument. That might not be crazy - the template should be identical throughout the run.

davidorme (approved):
Looks good to me - a couple of nitpicky questions about recreating templates where we could create them at `__init__`, but otherwise no problems.

sallymatson:
Thank you @davidorme I'll do this today! 

@jacobcook1995 I realized it would also be great for you to have a look if you have time, and give any feedback on if this will be ok to incorporate eventually into litter & soil? If you have any q's, lmk!

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -1473,6 +1479,53 @@ def partition_reproductive_tissue(
 
         return n_propagules, non_propagule_mass
 
+    def old_stoichiometry_ratios_to_depricate(self) -> None:
+        """Temporary function to initialise variables for the litter and soil models.
```
jacobcook1995, line 1481: These actually only go to the litter model so it's only that model that will need to be altered for you to depreciate all these variables

virtual_ecosystem/models/plants/subcanopy.py
jacobcook1995: Would it be sane to move this template to `core`? It's different to the `pft` templates in that I think other models (e.g. soil + litter) might want to use it at some point

jacobcook1995 (approved):
LGTM!

Didn't have much feedback, think what is here would completely work in the soil + litter model (though the litter model logic would have to change to work with masses instead of ratios).

In terms of the depreciation of the variables that get passed to litter, I'm happy to help with that process if that's helpful

virtual_ecosystem/models/plants/subcanopy.py
davidorme: We already have _one_ template in `core.grid` somewhere. It might make sense to build this out into a small set of templates. I think there are another couple of use cases floating around. Can chat on Thursday 😄

davidorme: The tricky bit is the templates with a PFT axis. That's not really core configuration, but we do need to put it into core to validate PFT axes on load.

jacobcook1995: yeah maybe this is something that's sensible long term but not for this PR

sallymatson:
@davidorme I added the templates as a dictionary class attribute. Can you have a quick look to ensure it's a decent structure (albeit likely temporary)?

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -477,6 +509,40 @@ def _setup(
...
+        )
+        cnp_template = xr.DataArray(
+            data=np.zeros((self.grid.n_cells, 3)),
+            coords={"cell_id": self.data["cell_id"], "element": ["C", "N", "P"]},
+        )
+        self.data_object_templates = {
+            "cnp_pft": cnp_pft_cell_template,
+            "cnp": cnp_template,
+            "cell": xr.full_like(self.data["elevation"], 0),
+        }
```
davidorme:
```suggestion
        self.data_object_templates = {
            "cnp_pft": xr.DataArray(
                data=np.zeros((self.grid.n_cells, self.flora.n_pfts, 3)),
                coords={
                    "cell_id": self.data["cell_id"],
                    "pft": self.flora.name,
                    "element": ["C", "N", "P"],
                },
            ),
            "cnp": xr.DataArray(
                data=np.zeros((self.grid.n_cells, 3)),
                coords={"cell_id": self.data["cell_id"], "element": ["C", "N", "P"]},
            ),
            "cell": xr.zeros_like(self.data["elevation"]),
        }
```

davidorme (commented):
I wouldn't bother with the intermediate variables, but otherwise looks good!

PR #1190, opened 2025-12-09
Fix/animal model run

TaranRallings:
# Description
This is a few small changes that fix bugs and improve the run of the animal model. 

Most important:
- a bug where the active_cohorts list was being reinitialized (thus wiping it), leading to all cohorts being treated as in hibernation
- a switch from damuth to madingley for population density means much higher densities.
- a couple little run-time error fixes that don't change anything else

The animal model mortality rates outstrip births by quite a lot BUT now there are enough individuals and cohorts to get us through a run of the example data with plenty to look at. More bug hunting can be done from here. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -232,6 +232,18 @@ def grow(self, resource_intake: dict[str, float]) -> dict[str, float]:
...
+        eps = 1e-12
+        for element in ("carbon", "nitrogen", "phosphorus"):
+            value = resource_intake[element]
+            if value < 0.0:
+                if value > -eps:
+                    resource_intake[element] = 0.0
+                else:
+                    raise ValueError(
+                        f"grow produced negative waste for {element}: {value}"
+                    )
```
dalonsoa, line 245: Just a suggestion to reduce indentation and make this visually simpler (in my opinion):

```suggestion
            if value >= 0.0:
                continue

            elif value < -eps:
                raise ValueError(
                    f"grow produced negative waste for {element}: {value}"
                )
            
            resource_intake[element] = 0.0       
```
It might be less explicit of what is being checked here, so no strong feelings either way.

dalonsoa (approved):
LGTM!

nickwctan (approved):
Looks good. Thanks for working on adding my suggestions from yesterday's meeting in such a short time! Appreciate it. I imagine you would add these values generated here to exporter.py?

Baizurah:
Looks good, thanks!

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -184,11 +184,11 @@ def __init__(
...
         """Animal communities with grid cell IDs and lists of AnimalCohorts."""
-        self.active_cohorts: dict[uuid.UUID, AnimalCohort] = {}
+        self.active_cohorts: dict[uuid.UUID, AnimalCohort]
         """A dictionary of all active animal cohorts and their unique ids."""
-        self.migrated_cohorts: dict[uuid.UUID, AnimalCohort] = {}
+        self.migrated_cohorts: dict[uuid.UUID, AnimalCohort]
         """A dictionary of all migrated animal cohorts and their unique ids."""
-        self.aquatic_cohorts: dict[uuid.UUID, AnimalCohort] = {}
+        self.aquatic_cohorts: dict[uuid.UUID, AnimalCohort]
         """A dictionary of all aquatic animal cohorts and their unique ids."""
```
davidorme, line 192: I wouldn't expect this change to make any difference? This is definitely the right thing to do - we want to separate the typing and docstrings in `__init__` and the logic in `_setup`. But `_setup` should only run once as part of the `__init__`, so I'm not sure how this changes the runtime behaviour.

davidorme (approved):
Looks good - but I don't understand why the model changes make any difference!

virtual_ecosystem/models/animal/animal_model.py
davidorme: OK - it's the sequencing of when `super().__init__()` is called.

PR #1193, opened 2025-12-11
Restructure of BaseModel initialisation flow

### LINKED ISSUE #1191: Restructure of BaseModel initialisation flow.
At the moment, we have the following flow when a model initialises:

* `RealModel.__init__()`:
     * Sometimes stuff before `super().__init__()`
     * `super().init()`
         * `RealModel._setup()`
     * Sometimes stuff after  `super().__init__()`

As that suggests, we are inconsistent about where the call to `super().__init__()` goes - the docs and several models have it as the first line, but logically that means that any attributes declared in `__init__` are typed _after_ `RealModel._setup()` has run, which is probably populating most or all of those attributes. I think `super().__init__()` is logically better placed at the end of model `__init__`. 

*But* I think it's much cleaner to split the running of `RealModel._setup()` out from `super().__init__()` completely and have an explicit required workflow like this:

* `RealModel.__init__`:
     * Run `super().__init__()` to do core model attributes.
     * Define, type and doc attributes for specific model.
     * Run `RealModel._setup()` if required.

I think this is much more obvious about what happens where and it separates the concerns of `super().__init__()` - which are to do with checking the run state of the model and setting core attributes - from the concerns of the specific model setup.

There's another strong advantage - at the moment, because the `RealModel._setup()` call is buried in `super().__init__()`, the actual model specific arguments are buried in `kwargs` and don't appear in the model init signature. They are documented in `_setup` but that isn't currently shown in the API pages. If we move to the suggested structure then we can doc the `__init__` args in place.

It is a little longer but basically just because we have to explicitly give the `_setup()` args rather than passing them around as `kwargs` - I don't think this is bad. I think it is logically cleaner and the documentation is much clearer.

### Current structure

```python
class RealModel(BaseModelABC):

    def __init__(
            self,
            data: Data,
            core_components: CoreComponents,
            static: bool = False,
            **kwargs: Any,
        ):
        """
        Init the model.

        Args:
            data: The data object to be used in the model.
            core_components: The core components used across models.
            static: Run the model in static mode
            kwargs: Extra setup args
        """


        maybe: int
        declare: int
        attributes: int
        here: int

        super().__init__(data, core_components, static, **kwargs)

        # Everything below here is typed _after_ setup has run.
        unless: int
        declare: int
        attributes: int
        here: int

    def _setup(
        self,
        flora: Flora,
        cohort_data: pandas.DataFrame,
        extra_pft_traits: ExtraTraitsPFT,
        model_constants: PlantsConstants = PlantsConstants(),
        pyrealm_config: PyrealmConfig = PyrealmConfig(),
        **kwargs: Any,
    ):
        """Setup implementation for the Plants Model.

        Args:
            flora: A flora containing the plant functional types used in the plants
                model.
            cohort_data: A data frame containing the initial cohort data.
            extra_pft_traits: Additional traits for each plant functional type, keyed by
                PFT name.
            model_constants: Set of constants for the plants model.
            pyrealm_config: Configuration options to the pyrealm package.
            **kwargs: Further arguments to the setup method.
        """

        dothings()
```

### Suggested structure

```python
class RealModel(BaseModelABC):

    def __init__(
            self,
            data: Data,
            core_components: CoreComponents,
            static: bool = False,
            flora: Flora,
            cohort_data: pandas.DataFrame,
            extra_pft_traits: ExtraTraitsPFT,
            model_constants: PlantsConstants = PlantsConstants(),
            pyrealm_config: PyrealmConfig = PyrealmConfig(),
        ):
        """
        Init the model.

        Args:
            data: The data object to be used in the model.
            core_components: The core components used across models.
            static: Run the model in static mode
            flora: A flora containing the plant functional types used in the plants
                model.
            cohort_data: A data frame containing the initial cohort data.
            extra_pft_traits: Additional traits for each plant functional type, keyed by
                PFT name.
            model_constants: Set of constants for the plants model.
            pyrealm_config: Configuration options to the pyrealm package.
        """

        # Do the super() stuff
        super().__init__(data, core_components, static)

        # Everything is typed here _before_ setup
        declare: int
        attributes: int
        only: int
        here: int

        if self.run_setup:
            self._setup(
                flora = flora,
                cohort_data = cohort_data,
                extra_pft_traits = extra_pft_traits,
                model_constants = model_constants,
                pyrealm_config = pyrealm_config,
            )

    def _setup(
        self,
        flora: Flora,
        cohort_data: pandas.DataFrame,
        extra_pft_traits: ExtraTraitsPFT,
        model_constants: PlantsConstants,
        pyrealm_config: PyrealmConfig,
    ):
        """Setup implementation for the Plants Model.

        See __init__ docs for argument description
        """

        dothings()
```

davidorme:
# Description

This PR switches over to the more explicit sequencing of the model `__init__` described in #1191:

```python
def __init__(self, data, core_components, static, model_arg) -> None:

    # Set the core stuff including static status flags
    super().__init__(data, core_components, static)
    
    # Declare and type attributes
    self.model_attribute: type
    """Model attribute declarations"""
    
    # Run the setup if the model is not in deep static mode
    if self._run_setup:
        self._setup(model_arg=model_arg)
```

This:
* is clearer about what happens when, rather than hiding away the execution of `_setup`.
* ditches the `kwargs` components of the BaseModel arguments and has an explicit list of model arguments. This identified a couple of testing issues where old model arguments were still in test code, because `kwargs` lets them sneak through.
* as a result, all the model specific arguments are directly specified in the model `__init__` signature.

To do this:

* I've lightly modified the BaseModel, including changes to the static flagging functions to set model attributes rather than passing around boolean return values. That has included inverting `_bypass_setup` to `_run_setup` to sit alongside `_run_update` and to give the more intuitive `if self._run_setup:` above.

* Updates all of the model files to the new structure above. I've re-ordered the sequence of methods to be consistent (`__init__`, `_setup`, `from_config`) across models.

* Updates the class docstrings to actually document all of the constructor args. It would be nice if `ruff` validated that class docstrings contained all the `__init__` args, but that doesn't seem easy to do. May be we should move Args inside `__init__`.

* Provides some minor fixes to the tests:
    * Update to the new `_run_setup` value from `_bypass_setup` (`True` <-> `False` swaps)
    * Removes some rogue arguments to model `__init__`

> [!IMPORTANT]
> This also revealed a bigger issue with the use of patching of the static settings in model `__init__` and `_setup` testing, which was masking some missing init vars. 
>
> I think this is being done because the tests use compiled sets of test data that include variables created by init and update, so the static settings were being patched to avoid raising configuration errors. That's a problem because it breaks the explicit variable checking  - for example, in the abiotic model it was masking incomplete required vars lists - the variable might not be in the required list, but is in the data anyway, so the test runs. 

I've fixed one case here because it was the easiest way to get tests to pass, but we should separately remove all of this patching and simply subset the more inclusive datasets down to the actual alleged required variables.

Fixes #1161

## Type of change

## Key checklist

## Further checks


docs/source/development/design/defining_new_models.md
```diff
@@ -452,34 +475,78 @@ You should end up with something like this:
...
+
+```{code-block} ipython3
+def _setup(self, pond_data_path: Path, constants: FreshwaterConstants) -> None:
+    """Set up the freshwater model."""
+
+    self.pond_data_path = pond_data_path
+    self.constants = constants
+
+    # Load the data using an extra method on the class
+    self.load_pond_data(self.pond_data_path)
```
dalonsoa: I think this is an unfortunate example. You shouldn't want data to be loaded in a setup method...

docs/source/development/design/defining_new_models.md
```diff
@@ -452,34 +475,78 @@ You should end up with something like this:
...
+
+    # Populate a variable using a user defined method
+    self.data["pond_temperature"] = calculate_pond_temperature(
+      data=self.data, constants=self.constants, time_index=0
+    )
+```
+
+### The `_update` method
+
+The `_update` method must then be define to calculate the changes in the model state
```
dalonsoa:
```suggestion
The `_update` method must then be defined to calculate the changes in the model state
```

docs/source/development/design/defining_new_models.md
```diff
@@ -535,32 +602,14 @@ def from_config(
 
 ### Other model steps
 
-There are four functions that must be included as part of the model class. Only the
-`update` function is used at present. The other functions need to be included, but
-there's no need to include any particular content within them (i.e. they can just be
-function definitions with docstrings).
+There are currently two other method that must be included as part of the model class.
```
dalonsoa, line 678: I wonder if we should remove them from everywhere and then add them back if/when needed.

tests/models/abiotic/test_abiotic_model.py
```diff
@@ -274,20 +264,20 @@ def test_setup_abiotic_model(
...
-            core_components=fixture_core_components,
-        )
+    init_data = Data(grid=dummy_climate_data_varying_canopy.grid)
+    for var in AbioticModel.vars_required_for_init:
+        init_data[var] = dummy_climate_data_varying_canopy[var]
+
+    model = AbioticModel(
+        data=init_data,
+        core_components=fixture_core_components,
+    )
```
dalonsoa, line 280: Nice. This hack about mocking stuff is on me, I'm afraid.

virtual_ecosystem/models/abiotic/abiotic_model.py
```diff
@@ -185,12 +162,7 @@ def _setup(
         all time steps of the simulation. All variables are added directly to the
         self.data object.
 
-        Args:
-            model_constants: Set of constants for the abiotic model.
-            pyrealm_core_constants: Additional configuration options to the pyrealm
-                package.
-            bounds: A set of bounds to be applied to abiotic variables.
-            **kwargs: Further arguments to the setup method.
+        See __init__ for argument descriptions.
```
dalonsoa, line 163: Is there no complain from ruff about not describing the arguments of a method in the docstring?

dalonsoa (approved):
Looks nice. Definitely more consistent across the code. Happy to unmock the mocks in the tests, as that is my mess.

docs/source/development/design/defining_new_models.md
davidorme: I hadn't thought of that - will swap it out.

Actually _some_ of the models do have non-`Data` data. The animals and plants models both side-load community structure data into their own model states through separate paths, and that _does_ happen in `_setup`.

docs/source/development/design/defining_new_models.md
davidorme: I think the `_spinup` step at least is still on @vgro and @jacobcook1995's radars, so I'm tempted to leave for now.

virtual_ecosystem/models/abiotic/abiotic_model.py
davidorme: Only if you actually have an `Args:` section. Similarly, there's no `ruff` checking that an `Args:` section in the class docstring matches the `__init__` arguments.

Honestly, I think that by itself is a strong case for putting the class docstring in `__init__`, but I'm not going there now!

tests/models/abiotic/test_abiotic_model.py
davidorme: I've got #1223 up - it isn't quite complete. There are a couple of tests that are now failing - was planning to look into that this morning.

davidorme:
@dalonsoa I realised that the class attributes section of the "defining_new_models.md" process was woefully out of date. I've updated it and gone into some detail about the resolution of execution order and the core `Data` object. That then makes it sensible to _specifically_ talk about additional data, so I've kept the `pond_data_path` example but added a tip on why you might do that. 

Could you give the most recent commit 5f903073 a quick look to see if it all makes sense?

docs/source/development/design/defining_new_models.md
```diff
@@ -526,9 +577,24 @@ def _setup(self, pond_data_path: Path, constants: FreshwaterConstants) -> None:
...
 
+:::{admonition} Additional data inputs to a model
+:class: tip
+
+Most of the data in a Virtual Ecosystem simulation is loaded into the central `Data`
+object and shared between the models. However, you may need to load additional data to
+initialise your model that is only used within the model and not shared through the
+`Data` object.
+
+It is absolutely fine to configure and load data like this within the model `_setup`
```
dalonsoa: I would take the actual data loading outside of `_setup` and do it in `from_config` and pass loaded objects (numpy arrays, xarrays or whatever). As it is now, it is impossible to create an instance of those models without having an actual file with valid data lying around. It is generally considered an [anti-pattern](https://dzone.com/articles/fixing-the-constructor-anti-pattern).

dalonsoa (approved):
Aside from my comment, which if addressed should be done so in a separate PR, this looks good.

docs/source/development/design/defining_new_models.md
davidorme: Ohhhh - got completely the wrong end of the stick with your last comment. I think all the existing models do as you say and load from file in `from_config`. I'll update this example to avoid the anti-pattern. It can show the `from_config` method actually doing something, so a useful restructure. I won't ask for a re-re-review 😄

Issue #1197, opened 2025-12-15
Mycorrhizal uptake model needs to be rethought


jacobcook1995:
At the moment mycorrhizal uptake of nutrients happens with the following process:

1) Mycorrhizal fungi predict the maximum amount of nutrients they can provide the plants (based on soil nutrient availability)
2) Plants decide how much of this they want to take
3) Mycorrhizal fungi provide this amount to the plants

Step 1) happens at the end of one soil model timestep, and step 3) happens during the next time step. This creates the problem that the conditions can change such that the amount promised cannot be provided, *but* the plants have already used them.

The way I currently address this is by mycorrhizal fungi sacrificing biomass to make sure that the provided amount is met (which stops nutrients being created from thin air). However, this can cause mycorrhizal populations to go negative (if the nutrient deficit exceed the amount of nutrient held in their total biomass).

I've previously tried to avoid this issue by making the mycorrhizal fungi more conservative about what they promise. I don't think this can ever fix the problem entirely though, as a month time step is long enough to see 100 fold changes in soil nutrient availabilities (if you start far enough away from a steady state). Forcing mycorrhizal fungi to only supply plants with 1% of the maximum amount they could supply is effectively the same thing as turning them off entirely.

The only possible solution that I can see is changing the order, so that:
1) Plants decide how much carbon to supply to the mycorrhizal fungi (this could be adjusted based on previous returns)
2) Based on this carbon supply and the soil conditions mycorrhizal fungi decide how much nutrients to supply to the plants
3) Plants then assimilate these nutrients that were provided in the previous time step

This order would introduce a lag between plants investing carbon in symbiotic nutrient acquisition and receiving the nutrients (at present it is immediate), but I think that's probably worth it to avoid the possibility of creating nutrients from nowhere or negative mycorrhizal populations.

(There's then the possibility of adding functionality to each side of the exchange to change supply based on previous returns, etc. But that feels like something to sort out once we are into calibration/validation)

This is something that needs discussion in person because it would involve altering both models. I'm mainly just putting this up here now as something to think about in more detail in the new year

davidorme:
I think this makes a lot more sense - and as you say it then opens up memory effects for calibrating investment.

jacobcook1995:
With #1232 initial estimates for the full soil environment are now available (regardless of abiotic model chosen), so when implemented the above I should considering making use of this. Rather than approximating soil temperatures based on the mean annual temperature

PR #1200, opened 2025-12-15
Small fixes to prevent negative microbial populations from propagating further errors

### LINKED ISSUE #1198: Add sensible handling of negative mycorrhizal populations
Mycorrhizal fungal populations can go negative, as described in #1197. That issue won't be quick to fix properly (as it requires changing processes in multiple models). 

In the meantime I need to add sensible handling of negative mycorrhizal populations, so that they don't crash the model. (they will be kept as negative values as a clear indicator that something is going wrong with the model)

jacobcook1995:
# Description

Two changes here.

1) Negative microbial populations now no longer lose a negative amount of biomass to maintenance. Instead the biomass loss rate is zero
2) The animal available microbial consumption pools now treat any negative microbial pool as zero (previously this only happened for the two mycorrhizal pools)

These fixes are needed not just for the mycorrhizal problems (which a better long term solution for is described in #1197), but also because animal consumption will be potentially able to drive microbial pools negative. 

Fixes #1198

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/soil/test_pools.py
+++ b/tests/models/soil/test_pools.py
@@ -544,6 +544,30 @@ def test_calculate_maintenance_biomass_synthesis(
     assert np.allclose(actual_loss, expected_loss)
 
 
+def test_calculate_maintenance_biomass_synthesis_negative(
+    dummy_carbon_data, averaged_soil_temp, functional_groups
+):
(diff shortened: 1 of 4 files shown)


vgro (approved):
LGTM

PR #1223, opened 2025-12-15
Retire patching in `__init__` testing

### LINKED ISSUE #1221: Mocking masking model init variable testing
Over in #1191, I found that there was a issue with the use of patching in tests of the static settings in model __init__ and _setup, which was masking some missing init vars.

This is being done because the tests use compiled sets of test data that include variables created by init and update, so the static settings were being patched to avoid raising configuration errors. That's a problem because it breaks the explicit variable checking - for example, in the abiotic model it was masking incomplete required vars lists - the variable might not be in the required list, but is in the data anyway, so the test runs.

davidorme:
# Description

@vgro and @jacobcook1995 

Your models both used the `patch` routines in testing - I've summarised the changes needed for each set of tests below. If you could have a look and see that it makes sense?

### Abiotic

* Created a new fixture that just provides the init data, now used throughout `test_abiotic_model.py`
* Removed patched calls to `AbioticModel()`
* Merged two very similar tests
* Removed an unneeded initial definition of the `soil_temperature` data array from an empty template in `AbioticModel._setup()`. The variable is populated with real data by a later function.
* Added four variables to `AbioticSimpleModel.vars_required_for_init`. These aren't strictly used by the model but this is currently the only way to stop the static model checking from freezing up: see the [comment below](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1223#issuecomment-3665895109)

### Abiotic simple

* Created a new fixture that just provides the init data, now used throughout `test_abiotic_simple_model.py`
* Removed patched calls to `AbioticSimpleModel()`
* Added seven missing variables to `AbioticSimpleModel.vars_required_for_init`
* Added two missing variables to `AbioticSimpleModel.vars_required_for_update`

### Hydrology

* Add `fixture_hydrology_init_data`, now used in `test_hydrology_model.py` calls
* Removed patched calls to `HydrologyModel()`
* Added two missing variables to `HydrologyModel.vars_required_for_init`

### Litter

* Add `fixture_litter_init_data`, now used in `fixture_litter_model` and `test_litter_model.py` calls 
* Removed patched calls to `LitterModel()`
* Simplified expected log creation and shared across tests.

### Soil model

* Removed unneeded patching in `fixture_soil_model`

Fixes #1221

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -183,7 +183,8 @@ def run_microclimate(
     )
 
     # Aerodynamic resistance soil, [s m-1]
-    aerodynamic_resistance_soil = data["aerodynamic_resistance_surface"].to_numpy()
+    # VIVI - this was ... = data["aerodynamic_resistance_surface"].to_numpy()
+    aerodynamic_resistance_soil = aerodynamic_resistance_understorey
```
vgro: There is a difference between `surface` (which I should rename `soil` now) and `understorey` here, please leave as it is for now if possible, I am actively working on this code

davidorme: OK - let's pause this until the variable names settle. That'll also give a chance for #1191 to clear and reduce this to just the patching changes.

davidorme:
Oh _super_. A heisenbug. 👿 Stochastic failures of `test_ve_run`.

davidorme:
> Oh _super_. A heisenbug. 👿 Stochastic failures of `test_ve_run`.

This turned out to be an issue with variable requirements and the model initialisation order sequencing:

There's an issue with the variable config on the Hydrology and Abiotic models. The hydrology model populates these four variables when it initialises (they are in `HydrologyModel.vars_populated_by_init`)
 
`specific_heat_air, density_air, latent_heat_vapourisation, aerodynamic_resistance_canopy`
 
The abiotic model then has the same variables in `AbioticModel.vars_updated`
 
This is a problem because - at the moment - those variables aren't in `AbioticModel.vars_required_for_init`. So if the model initialisation sequence runs Hydrology first, then the static model checking for the `AbioticModel._update` method finds the variables and objects that some but not all updated variables are present. I think the problem goes away if we add those variables to `AbioticModel.vars_required_for_init` because then those variables are specifically excluded from the variables expected for the update step.

This feels like a hack to me - those variables aren't really required for init. I think the issue is that one model populates the variables and another updates them, and I don't think the static model update variable checking anticipated that possibility.

vgro (approved):
LGTM.
I wonder if the weird behavior could be resolved by moving the initialisation of variables  between the models, but I think for now this does the job.

virtual_ecosystem/main.py
```diff
@@ -71,6 +71,7 @@ def initialise_models(
             models_cfd[model_name] = this_model
         except (InitialisationError, ConfigurationError):
             failed_models.append(model_name)
+            # LOGGER.error(err)
```
jacobcook1995: I guess this should be deleted?

jacobcook1995 (approved):
LGTM (with one minor comment)!

PR #1224, opened 2025-12-16
Add flow of excrement decay into the soil model

### LINKED ISSUE #1177: Excrement and carcass decay doesn't currently go anywhere
At the moment the animal model provides rates of excrement and carcass decay ("decomposed_excrement_carbon", etc), which the soil model just doesn't use at all. This a really weird oversight on my behalf, but should be quite simple to rectify.

I think it makes sense to wait until #1171 is merged to do this, so that I can pass two variables between the models rather than six

jacobcook1995:
# Description

This PR changes the soil model so that it now actually uses the excrement and carcass flows. For this exchange I've now started using the CNP triplets so that I only need to use 2 variables for the exchanges rather than 6.

Fixes #1177 

@sallymatson and @TaranRallings, I've added both of you as reviewers because I want to check that I have used the CNP triplets correctly and that I haven't messed up anything in the animal model. No rush at all with reviewing this, I just wanted to get the PR made before I went on annual leave

## Type of change

## Key checklist

## Further checks


TaranRallings (approved):
Straightforward enough, LGTM!

tests/models/soil/conftest.py
```diff
@@ -109,6 +109,36 @@ def dummy_carbon_data(fixture_core_components):
...
+
+    data["decomposed_carcasses_cnp"].loc[:, "C"] = np.array(
+        [0.0002844, 0.00011089, 2.459e-5, 0.003891]
+    )
+    data["decomposed_carcasses_cnp"].loc[:, "N"] = np.array(
+        [1.544e-6, 7.520e-5, 1.433e-5, 0.0002582]
+    )
+    data["decomposed_carcasses_cnp"].loc[:, "P"] = np.array(
+        [8.935e-7, 1.932e-6, 4.928e-6, 3.769e-6]
+    )
```
sallymatson: I don't think it particularly matters, but you could condense this into something like 

```
data["decomposed_carcasses_cnp"].loc[:, ["C", "N", "P"]] = np.array([
    [0.0002844, 1.544e-6, 8.935e-7],
    [0.00011089, 7.520e-5, 1.932e-6],
    [2.459e-5, 1.433e-5, 4.928e-6],
    [0.003891, 0.0002582, 3.769e-6],
])
```

sallymatson (approved):
LGTM ! Left one minor comment but not blocking by any means

tests/models/soil/conftest.py
davidorme: Or even...

```python
    data["decomposed_carcasses_cnp"] = DataArray(
        data=[
            [0.0002844, 1.544e-6, 8.935e-7],
            [0.00011089, 7.520e-5, 1.932e-6],
            [2.459e-5, 1.433e-5, 4.928e-6],
            [0.003891, 0.0002582, 3.769e-6],
        ],
        coords={"cell_id": data["cell_id"], "element": ["C", "N", "P"]},
    )
```

Edited to fix the rotation of the input array, which @sallymatson had handled. My point is that we don't need to declare the data as zeroes and then over-write it.

PR #1227, opened 2025-12-16
1181 canopy evaporation is not returned properly

### LINKED ISSUE #1162: Surface precipitation should not get negative
The current `ve_run` example data with the `abiotic` model  on `1154-energy-balance-for-understory-vegetation` produces this error:

**ValueError: Surface precipitation should not be negative! Consider checking that the canopy water balance is correct.**

1. checkout `1154-energy-balance-for-understory-vegetation'
2. install ve_example and change config to use [abiotic]
3. `ve_run`

- OS: Sequoia
- Version 15.7.2

This is likely an issue with the canopy interception.

### LINKED ISSUE #1181: canopy evaporation is not returned properly
The canopy evaporation variable is filled with zeros in all layers which seems wrong. Check that sums are correct over daily values and that only true values are taken into account

vgro:
The problem was that part of the canopy water balance were averaged over all canopy layers and therefore returned 1D variables - for the rest to work we require 2D variables. This PR updates interception and canopy evaporation to 3D and includes the understorey layer. I dropped the `leaf_drainage` for simplification - we assume that all water falls down in one time step (1 day) so it was a bit pointless.

I also got rid of the understorey aerodynamic resistance for now assuming that it is the same as the canopy, and I renamed the aerodynamic resistance of the soil for clarification (see [this comment](https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1223/changes/BASE..e914f903c1fec1b4de6efe83be8f80ced4a4288a#r2622358785)) by @davidorme .

I believe this also fixes #1162 .

Fixes #1181

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/data_variables.toml
```diff
@@ -215,6 +215,13 @@ name = "layer_heights"
 unit = "m"
 variable_type = "float"
 
+[[variable]]
+axis = ["spatial"]
+description = "Interception"
```
jacobcook1995: Think this needs a longer description

davidorme (approved):
LGTM

jacobcook1995 (approved):
LGTM, just one small thing

virtual_ecosystem/data_variables.toml
davidorme: Oh hell, they _all_ need a longer description. And units. And proper axes. I think we're gunna need a bigger PR 😄

davidorme:
@vgro When this merges, does that unstick the problem with the naming changes in #1223?

vgro:
> @vgro When this merges, does that unstick the problem with the naming changes in #1223?

yes it should, see intro :-)

PR #1242, opened 2026-01-05
Fixing `ve_run` command line configuration 

### LINKED ISSUE #1241: Bug in passing `--param` to `ve_run_cli`
The current code relies on `_parse_command_line_params` modifying the passed in `override_params` dictionary in place. Unfortunately, the value of that object is replaced here:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/a37f83f369a15accdc89d2a6418ff709e9ae33be/virtual_ecosystem/entry_points.py#L54-L58

The `merge_config_dictionaries` function **deep copies** its inputs, so the returned `override_params` object is not the same as the passed in one. When the function returns, the original dictionary is untouched and so none of the changed parameters are enforced.

davidorme:
# Description

The `ve_run` command is supposed to take configuration overrides as `ve_run ... --param plants.constants.value=0.6` but those values are currently silently swallowed: the dictionary passed into `_parse_command_line_params` was intended to be updated but was redefined within the function and so the updates never propagated back out into the configuration.

This PR:

* Fixes that bug - the CLI configuration parsing function now explicitly returns a dictionary of command line configuration overrides and the handling of the output path is now included as part of the CLI config string parsing, rather than as a pre-existing dictionary entry.
* Updates the naming of functions and arguments aways from `params` to `config` to make it more obvious that this is modifying the model configuration.
* Fixes broken tests - surprisingly few.

It also updates `test_ve_run` to only run 2 months of simulation. We simply don't need a 24 month simulation as part of the unit testing: two updates is enough to verify that the basic model flow is working and drastically reduces the run time of that test (on my laptop: 180 seconds down to 25 seconds). This was the motivating case for the command line override that revealed the bug.

The Actions testing runtime for the OS/Python combos drops by about 2/3rds: worst case is Windows down from 10' to 3'30". The docs build is now the bad citizen in the actions runtime.

@alexdewar This likely breaks [ImperialCollegeLondon/virtual_ecosystem_snakemake_template](https://github.com/ImperialCollegeLondon/virtual_ecosystem_snakemake_template)?

Fixes #1241

## Type of change

## Key checklist

## Further checks


alexdewar (approved):
Makes sense to me, although I've got a couple of questions:

1. Is there a test that checks that this override functionality works? If not, would it be worth adding one?
2. Have you got any idea how long this has been broken for?

The snakemake template has sadly been broken for a while now 😞 and I haven't got round to fixing it... I can though, if that would be useful to you. I had a PR with mysterious CI failures on Windows that I could debug easily (https://github.com/ImperialCollegeLondon/virtual_ecosystem_snakemake_template/pull/51) and then I got distracted with other things. I suppose it might be worth checking whether it works with the version of VR in this current PR though!

davidorme:
> 1. Is there a test that checks that this override functionality works? If not, would it be worth adding one?

There isn't a test -  `tests/core/test_configuration_builder.py::test_ConfigurationLoader_load_configuration_data` passes in the parsed dictionary but there isn't something that checks the data makes it from the CLI to the final configuration. Would be worthwhile, I agree.

> 2. Have you got any idea how long this has been broken for?

I think... always? The dictionary merging function (was `config_merge`, now  `merge_configuration_dicts`) has always started with:

```python
    dest = deepcopy(dest)
    source = deepcopy(source)
```
And I think that is where it goes off the rails and breaks the mutation of the original passed in dictionary. We just never tested it and the snakemake template was at a stage where we wouldn't spot if a config change made any differences?

alexdewar:
> And I think that is where it goes off the rails and breaks the mutation of the original passed in dictionary. We just never tested it and the snakemake template was at a stage where we wouldn't spot if a config change made any differences?

Ah ok. Yes, that's what I was curious about!

davidorme:
@alexdewar  I've added a really simple test that just checks that the CLI input makes it through to the `ve_run` call. The integration of that data is already tested elsewhere, and this is a short and clean check that the input makes it.

alexdewar (approved):
LGTM!

Btw @dalonsoa is still away this week, so you might not want to hold out for a review from him, depending on how soon you want this merged.

PR #1244, opened 2026-01-06
1243 sort out abiotic constants for calibration

### LINKED ISSUE #1243: sort out abiotic constants for calibration
Check that all constants in `abiotic` and `hydrology` are actually used before starting the calibration process.

vgro:
This PR removed unused `abiotic` and `hydrology` constants.

Fixes #1243

### DIFF
--- a/virtual_ecosystem/models/abiotic_simple/model_config.py
+++ b/virtual_ecosystem/models/abiotic_simple/model_config.py
@@ -33,7 +33,7 @@ class AbioticSimpleConstants(AbioticSharedConstants):
     initial_net_radiation: float = 10.0
     """Initial value for net radiation per layer, W m-2.
     
-    TODO This is currently set to an arbitrary value."""
+    This is currently set to an arbitrary value, only used in test."""
 
 
 class AbioticSimpleBounds(Configuration):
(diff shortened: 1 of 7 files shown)


virtual_ecosystem/models/abiotic_simple/model_config.py
```diff
@@ -33,7 +33,7 @@ class AbioticSimpleConstants(AbioticSharedConstants):
     initial_net_radiation: float = 10.0
     """Initial value for net radiation per layer, W m-2.
     
-    TODO This is currently set to an arbitrary value."""
+    This is currently set to an arbitrary value, only used in test."""
```
jacobcook1995, line 36: If this is only used in testing should this be here? Or would it be better moved to `tests/abiotic_simple/conftest.py` (or similar)?

jacobcook1995 (approved):
One minor query but otherwise looks good to go!

virtual_ecosystem/models/abiotic_simple/model_config.py
vgro: It's a bit of a weird one, I was going to delete it but it is the only `AbioticSimpleConstant` and I wanted to keep the general test.

jacobcook1995: oh right didn't realise this was the context, yeah can see not wanting to delete this until you are completely sure that the `abiotic_simple` model isn't going to have any of it's own constants

davidorme: I think arguably would be better to rename it to `placeholder` or something, just to be clear that it isn't a real value?

vgro: I'll include this in the next PR

PR #1245, opened 2026-01-06
1172 adjust flux layers index to account for understorey vegetation

### LINKED ISSUE #1172: Adjust flux_layers index to account for understorey vegetation
The flux layer index currently misses the understorey vegetation which means it needs to be added manually which is a bit clunky.

This change will affect mostly the models `litter`, `soil` and `abiotic`, and a bit the `plants`

vgro:
This PR updates the flux layer index to include the understorey vegetation. I had initially seen this being used in multiple places but could only find very few code snippets now, could you please check that I did not miss anything in plants and soil?

Fixes # 1172

## Type of change

## Key checklist


jacobcook1995 (approved):
I just use the surface and top soil layer indices and don't use the flux ones anywhere in soil or litter. So from my side it's good to merge

PR #1266, opened 2026-01-20
Simplified mycorrhizal uptake model

### LINKED ISSUE #1197: Mycorrhizal uptake model needs to be rethought

jacobcook1995:
# Description

This PR simplifies the mycorrhizal uptake model by reducing the amount of information being passed backwards and forwards between the `soil` and `plants` models.

Instead, the mycorrhizal fungi allocate a fixed (but configurable) portion of their nutrient uptake to supply the plants. And the total amount supplied is passed to the plants at the next model time step.

Getting this to work requires making an initial estimate of the amount of nutrients supplied (in the soil model init). This estimate is based on the initial state of the model with the assumption that the microbes take up nutrients at the maximum possible rate.

I've added documentation justifying the new model. Let me know if it makes sense!

Fixes #1197

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -620,68 +620,90 @@ def calculate_dissolved_nutrient_concentrations(self) -> dict[str, DataArray]:
...
+            f"{full_name}_{nut}_supply": updated_soil_pools[
+                f"new_{abbreviation}_{nut}_supply"
+            ]
+            * self.grid.cell_area
+            * self.core_constants.max_depth_of_microbial_activity
+            for nut in ["n", "p"]
+            for abbreviation, full_name in zip(
+                ["amf", "emf"], ["arbuscular_mycorrhizal", "ectomycorrhizal"]
+            )
+        }
```
davidorme, line 647: I think this is one "for" too many in a dict comprehension 😄 I don't think the code below is faster (and it requires importing `itertools`) but I think it is easier to understand

```suggestion
        var_combinations = itertools.product(["n", "p"], [["amf", "arbuscular_mycorrhizal"],["emf", "ectomycorrhizal"]])
        
        return {
            f"{full}_{nut}_supply": updated_soil_pools[
                f"new_{abbr}_{nut}_supply"
            ]
            * self.grid.cell_area
            * self.core_constants.max_depth_of_microbial_activity
            for nut, (abbr, full) in var_combinations
        }
```

davidorme (approved):
Yup - LGTM. I think this is cleaner code, easy to explain and a reasonable approximation, and it avoids a whole lot of unpleasantness from the previous version. One minor niggle on code.

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: I think this is me finally coming around to python comprehensions and just going over the top with them haha. Thanks for the suggestion, that looks a lot cleaner!

PR #1269, opened 2026-01-21
Converted virtual_ecosystem_in_static_mode.md to a juypter notebook

### LINKED ISSUE #1268: Store virtual_ecosystem_in_static_mode.md as a Jupyter notebook
`virtual_ecosystem_in_static_mode.md` calls `ve_run` multiple times which means it takes more than 5 minutes to execute, significantly increasing docs build times.

The point is to show conceptionally how the static model works so there's very little value in showing the most recent possible output here. 

It makes more sense to store this as an executed notebook so that they contents aren't run every time the docs build. 

(we would have to remember to rerun this page if there are changes in `ve_run`'s API

jacobcook1995:
# Description

This Jupyter notebook is set so that it doesn't get reevaluated every time the CI runs.

@vgro could you check that [the page](https://virtual-ecosystem.readthedocs.io/en/1268-store-virtual_ecosystem_in_static_modemd-as-a-jupyter-notebook/using_the_ve/virtual_ecosystem_in_static_mode.html) still looks as your originally intended in the built docs, cheers!

Fixes #1268

## Type of change

## Key checklist

## Further checks


davidorme (changes requested):
I think the thing this is missing is how the dev team update this page when needed. Ideally that would be a function we could run - maybe a python file in the docs root?

davidorme (approved):
Thanks - looks good.

PR #1270, opened 2026-01-21
Trophic record exporter

### LINKED ISSUE #1247: Animal Model - Trophic interaction output
We need the data output for generating foodwebs.

Animal Cohorts need to be modified such that a record is maintained of all of the foraging events of that cohort in the time step. This record is then output with the rest of the data dump and then cleared for the next time step. I think this will likely work best as a dictionary of cohort IDs/pool-names and CNP mass totals. The data team can then transform the cohort IDs into functional groups when they do the foodweb generation. This, i think, is the most flexible approach.

TaranRallings:
# Description

This PR builds on the existing cohort exporter to export a second csv file detailing every trophic interaction that occurs over the course of the model run. The purpose of this data export is so that the data team can build food web time-series. 

The way it works is that each cohort has a new dict attribute called trophic_record. Every trophic interaction the cohort has is stored in this dict. At the end of each time step, this information is dumped into the csv and then the attribute is reset. 

New additions:
- trophic_record cohort attribute
- write-to-record functionality built into foraging system
- trophic_output system built into exporter, mirroring the existing cohort_output system
- supporting tests

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/exporter.py
```diff
@@ -128,13 +135,22 @@ def from_config(
...
             exporter.cohort_attributes = set()
             exporter.float_format = config.float_format
-            exporter._output_mode = "w"
-            exporter._write_header = True
+
+            # Internal state (match your refactor)
+            exporter._cohort_output_mode = "w"
+            exporter._trophic_output_mode = "w"
+            exporter._write_cohort_header = True
+            exporter._write_trophic_header = True
```
davidorme: Are these separated because it is possible that you might start outputting only to cohorts and not get interactions until later. For example, if you only had adult varanids in the simulation, and they just get hungrier and hungrier until some baby varanids turn up to eat?

davidorme (approved):
LGTM - just one question.

virtual_ecosystem/models/animal/exporter.py
TaranRallings: My thinking at the time was simply that keeping them distinct helps with debugging and updating down the road.  I suppose it could be used for that as well.

PR #1284, opened 2026-01-26
1283 add longwave downward radiation

### LINKED ISSUE #1283: add longwave downward radiation
Longwave downward radiation (LWD) is required to close the energy balance.

TODO:

- add LWD to example data
- add LWD to input data
- add LWD to required vars in abiotic model
- add LWD to canopy eneergy balance
- add LW_canopy to understorey energy balance
- add LW_understorey to lowest canopy layer
- handle energy flux from understorey to soil

vgro:
This PR adds longwave downward radiation to the energy balance.
I haven't updates the documentation because there will be some more changes over the next weeks.

Fixes #1283

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -238,10 +260,82 @@ def run_microclimate(
...
+        stefan_boltzmann_constant=core_constants.stefan_boltzmann_constant,
+        zero_Celsius=core_constants.zero_Celsius,
+        seconds_to_hour=core_constants.seconds_to_hour,
+        return_fluxes=True,
+    )
+
+    if not isinstance(new_energy_balance_canopy, dict):
+        to_raise = ValueError("The energy balance has not returned any fluxes!")
+        LOGGER.error(to_raise)
+        raise to_raise
```
jacobcook1995, line 307: These three lines get flagged as not being unit tested. I guess it's a massive pain to test something like this in a really long function though, so maybe it isn't worth bothering with

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -238,10 +260,82 @@ def run_microclimate(
...
+
+    longwave_emission_canopy = new_energy_balance_canopy["longwave_emission_canopy"]
+    latent_heat_flux_canopy = new_energy_balance_canopy["latent_heat_flux_canopy"]
+    sensible_heat_flux_canopy = new_energy_balance_canopy["sensible_heat_flux_canopy"]
+
+    net_radiation_canopy = (
+        shortwave_absorption_canopy
+        - longwave_emission_canopy
+        + downward_longwave_radiation
+        + longwave_emission_understorey * abiotic_constants.leaf_emissivity
```
jacobcook1995, line 317: Should this be negative if it is an emission (not familiar with the science here so very likely to be wrong)

jacobcook1995 (approved):
Couple of minor queries, but otherwise LGTM

virtual_ecosystem/models/abiotic/microclimate.py
jacobcook1995: Guess you would have to mock `energy_balance.calculate_energy_balance_residual` to return something that isn't a `dict` and then check that the right error is raised.

vgro: Good point, but looking at it now, it's probably not necessary anymore

virtual_ecosystem/models/abiotic/microclimate.py
vgro: It is an emission from a different source that is added to the balance

virtual_ecosystem/models/abiotic/microclimate.py
vgro: It turns out I'm getting typing problems if I remove this... This will change again very soon, so I'll take a note and fix it when the full subdaily features are there, #1285

jacobcook1995: Yep that makes sense (hadn't thought that it might be a way to shut `mypy` up)!

PR #1291, opened 2026-01-27
Clearer documentation of the output files

### LINKED ISSUE #1287: Document the output files more clearly
> Note to self: what is the purpose of the final_state.nc file compared to the all_continuous_data.nc file? I assume the final state is captured within the continuous data?
> There are more variables in final_state (189) compared to continuous (160). It seems like most of these extra ones are initial/startup values, so I wonder if all_continuous_data.nc is all we need? 

 _Originally posted by @arne-exe in [#156](https://github.com/ImperialCollegeLondon/ve_data_science/issues/156#issuecomment-3803567068)_

The intention of the two files is that the `all_continuous_data.nc` file contains time series of variable values at each timestep. It can be configured to save only variables of interest. `The final_state.nc` is intended to record the complete state of the main data object at the end of the simulation.

I think we need to revisit these output options but we should document these current output options more clearly.

jacobcook1995:
# Description

I've added extra description to the `virtual_ecosystem_in_use` notebook explaining what the output files are. This felt like the most appropriate place in the user documentation, but if anyone has alternative suggestions I am open to them.

@arne-exe This is the [section of the docs](https://virtual-ecosystem.readthedocs.io/en/1287-document-the-output-files-more-clearly/using_the_ve/virtual_ecosystem_in_use.html#model-outputs) to look at, let me know if that makes sense or if you think any additional details are needed

Fixes #1287


arne-exe:
Line 390: the use of ```to be calculated``` here is a bit confusing. If the value is constant it shouldn't change. If the value is only initial, then different wording may be more clear, perhaps something like ```that will be updated during each time step``` (although this doesn't really work for constants). Or perhaps just end the sentence after ```initial or constant values across the grid cells```.

Line 391-392: replace ```provide a time series of data at a reference height above the canopy that are used to that drive (or force) the behaviour of the model through time``` by ```provide a time series of data that are used to drive (or force) the behaviour of the model through time```

Line 517-519: This seems to replicate what is said in lines 390-392.

docs/source/using_the_ve/virtual_ecosystem_in_use.ipynb
```diff
@@ -2732,11 +2732,20 @@
    "metadata": {},
    "source": [
     "### Model outputs\n",
+    "Two further netCDF files are output by the simulation:\n",
```
arne-exe:
```suggestion
    "Two further NetCDF files are output by the simulation:\n",
```

docs/source/using_the_ve/virtual_ecosystem_in_use.ipynb
```diff
@@ -2732,11 +2732,20 @@
...
     "\n",
-    "The continuous data and final state datasets contain variables describing the\n",
-    "model state through the simulation process. These can be visualised as\n",
-    "spatial grids, individual time series within grid cells and as the three\n",
-    "dimensional structure of the vertical layers within the simulation.\n",
+    "`final_state.nc` gives the final state of the model, i.e. is effectively the equivalent\n",
+    "of `initial_state.nc` for the final time step. None of the data contained within it has\n",
+    "a time dimension so the variables contained within it can be plotted as shown above.\n",
```
arne-exe: I find the second sentence a bit confusing and would suggest removing it.
If you want to refer to the plotting options above, you could use ```equivalent of initial_state.nc for the final time step (see above for plotting suggestions)```.

Also, it's not clear if the final state is included as the last timestep in the all_continuous_data.nc file.
Below you mention that all_continuous_data.nc does not include the initial state data, so please also clarify this for the final state.

From this documentation section, I would want to know 1) what kind of data/variables does it contain, and 2) why/when do I want to use each of the respective files. At the moment, I'm guessing that final_state.nc is a subset of all_continuous_data.nc, so its purpose is not clear. It might make it easier to understand outputs if we focus on all_continuous_data.nc for most/primary cases, and final_state.nc only when you're interested in comparing the initial and final states.

docs/source/using_the_ve/virtual_ecosystem_in_use.ipynb
```diff
@@ -2732,11 +2732,20 @@
...
+    "\n",
+    "`all_continuous_data.nc` contain variables describing the model state through the\n",
```
arne-exe:
```suggestion
    "`all_continuous_data.nc` contains variables describing the model state through the\n",
```

docs/source/using_the_ve/virtual_ecosystem_in_use.ipynb
```diff
@@ -2732,11 +2732,20 @@
...
+    "simulation process. It does not contain the initial state and starts after the first\n",
+    "time step, this is to avoid duplicating the output given in `initial_state.nc`. To keep\n",
+    "the size of this file down only variables updated by a model are included, which means\n",
+    "things like soil pH and input climatic data are not included. As this data contain in\n",
```
arne-exe:
```suggestion
    "things like soil pH and input climatic data are not included. As the data contained in\n",
```

arne-exe (approved):
Hi @jacobcook1995 and @davidorme, thanks for following up on this.
I've added my suggestions.

Related to output files: I was going through the output files and realised that there doesn't seem to be a straightforward (if any?) way to access the units of each variable.
The NetCDF file does have an attribute ```units``` but this field seems to be empty.
I think it would be a really nice addition (and an essential one perhaps?) if the output data contains info on the units.

docs/source/using_the_ve/virtual_ecosystem_in_use.ipynb
jacobcook1995: Yep good idea to emphasise `final_state.nc` less. I'll try and reframe this section to mainly talk about the continuous data, which is probably what people want to work with most of the time, and just mention the final state file as something people might want to use for a specific purpose

davidorme: I mean honestly, I think we set this up with the mindset that the final state was the main thing and people could track individual variables of interest. In practice, we're writing everything to continuous and so the final state is redundant. I think the way to sell it is those two modes: what does the forest look like after 50 years (final state is your snapshot) vs dynamics of forest over 50 years (use continuous vars).

jacobcook1995:
That's useful feedback thanks, I've made changes to address it.

On the units of the variables, the best place to look is the [new and improved variables table](https://virtual-ecosystem.readthedocs.io/en/feature-docs-ui-update/virtual_ecosystem/implementation/variables.html), which give the units for each variable (though none of them are validated so some of them are likely incorrect). That table isn't merged into develop yet so you need to be on the `feature-docs-ui-update` documentation branch to see it, it also might move around a bit as I restructure the docs (sorry!)

In terms of including the units in the output files I think that's a good idea in principle, but probably not something we can add at the moment. We would basically want a systematic way to add units to the data object, because people adding them all over the place in the code would make errors pretty likely. I guess the best thing would be for me to make an issue for this as something we want to address eventually (does that sound sensible to you @davidorme?)

davidorme:
> In terms of including the units in the output files I think that's a good idea in principle, but probably not something we can add at the moment. We would basically want a systematic way to add units to the data object, because people adding them all over the place in the code would make errors pretty likely. I guess the best thing would be for me to make an issue for this as something we want to address eventually (does that sound sensible to you @davidorme?)

We can just add them directly from the variables registry? Once #1292 is complete they should all be synced up.

jacobcook1995:
Ahh yes that makes sense, so basically just a case of making sure that `data_variables.toml` is regularly reviewed?

jacobcook1995:
Arne's outstanding comment is covered by #1292 and #1293 and #1295, so I will go ahead and merge this now

PR #1294, opened 2026-01-28
Write variable metadata to netcdf files

### LINKED ISSUE #1293: Write units to NetCDF files
We should be able to use the data variables registry to populate the units attributes of NetCDF files. At the moment that data is incomplete and inaccurate (see #1292) but we can work on these in parallel.

davidorme:
# Description

This PR:

* Adds a loop into the two NetCDF output files that adds the `Variable.unit` and `Variable.description` attributes from `ve.core.variables.KNOWN_VARIABLES` to each of the variables in the output data. 
* We only need to add these on write to file (and not as variables are created), so it's much easier just to tag it on at the last moment. 
* Annoyingly there is a circular import issue, so the functions have internal imports of `KNOWN_VARIABLES`. I don't like that but for now it'll stand.

Fixes #1293

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/data_variables.toml
+++ b/virtual_ecosystem/data_variables.toml
@@ -1383,3 +1383,10 @@ description = "Average C, N, and P masses of each seed in a PFT"
 name = "fallen_seeds_cnp"
 unit = "kg"
 variable_type = "float"
+
+[[variable]]
+axis = ["spatial"]
+description = "NEEDS DESCRIPTION"
+name = "root_turnover_c_p_ratio"
+unit = "NEEDS UNITS"
+variable_type = "float"
--- a/tests/core/test_data.py
+++ b/tests/core/test_data.py
@@ -13,6 +13,14 @@ from tests.conftest import log_check
 from virtual_ecosystem.core.exceptions import ConfigurationError
 
 
+@pytest.fixture(autouse=False, scope="module")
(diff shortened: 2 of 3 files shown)


jacobcook1995 (approved):
Ahh that's really straightforward!

virtual_ecosystem/data_variables.toml
```diff
@@ -1383,3 +1383,10 @@ description = "Average C, N, and P masses of each seed in a PFT"
 name = "fallen_seeds_cnp"
 unit = "kg"
 variable_type = "float"
+
+[[variable]]
+axis = ["spatial"]
+description = "NEEDS DESCRIPTION"
+name = "root_turnover_c_p_ratio"
```
jacobcook1995, line 1390: Why does this need to be added? I thought I deleted this variable entirely, did I miss it somewhere?

davidorme:
This is derailed by the use of registries for the variable validation system - the state of those sits outside `pytest` and so we'd need to wrap nearly all tests and fixtures in code to get the `KNOWN_VARIABLES` into the right test for the state. And that could be execution order dependent - which is why  we get those failures above. That would be an incredibly brittle, kludgy crappy fix, would take ages and add a huge maintenance load. So - it's honestly simpler to move the variables stuff into explicit variables within the model - #1296

Issue #1248, opened 2026-01-08
[META-META!] Feature branch for docs update


davidorme:
This issue is to bundle the documentation issues around a feature branch that will be used to hold the documentation updates, using the `pydata-sphinx-theme`.

The intention here is to collect doc content and styling updates off  the `develop` branch until everything stops moving.

PR #1296, opened 2026-01-28
Alternative implementation of `variables.py`

### LINKED ISSUE #1295: Problem with KNOWN_VARIABLES
In #1293, I added variable metadata from `data_variables.toml` to NetCDF outputs. That PR adds the metadata when data is dumped to file, but I think though - well, why not do it when the data is added to the `Data` object. That has the advantage of being in one place, but _also_ allows us to catch any attempt to add unknown variables to the `Data` object. Which seems like a good thing

The problem here though is that trying to do this breaks all the tests. This is because `KNOWN_VARIABLES` is a bit of a vapourware variable: if a test creates a Data object (many of them do) then if a given test hasn't run `register_all_variables`, all the variables are unknown. But if Data or the test runs `register_all_variables` then all the test where it has been run previously fail, because you can't re-register variables. That makes all tests incredibly brittle and contingent on the state of `KNOWN_VARIABLES`

I think checking known variables within `Data` is clearly a good thing to do, but doing that with `KNOWN_VARIABLES` as a registry is horrific. I think the better solution here is that `known_variables` becomes a return value of `register_all_variables`, that could sensibly be generated as an attribute of `Data`. The rest of the variables discovery functions can then take `known_variables=Data.known_variables` as an argument.

That does mean that all tests with `Data` in will load the variables file, but we can probably patch that at the package level. I don't think it'll be a huge runtime bump on tests though.

My sketch code is below - I'll push the broken code into a branch on this PR for reference.

```python
class Data:
    """The Virtual Ecosystem data object.

    This class holds data for a Virtual Ecosystem simulation. It functions like a
    dictionary but the class extends the dictionary methods to provide common methods
    for data validation etc and to hold key attributes, such as the underlying spatial
    grid.

    Args:
        grid: The Grid instance that will be used for simulation.

    Raises:
        TypeError: when grid is not a Grid object
    """

    def __init__(self, grid: Grid) -> None:
        # Set up the instance properties
        if not isinstance(grid, Grid):
            to_raise = TypeError("Data must be initialised with a Grid object")
            LOGGER.critical(to_raise)
            raise to_raise

        # Local import to avoid circular import issue
        from virtual_ecosystem.core.variables import KNOWN_VARIABLES, Variable

        self.known_variables: dict[str, Variable] = KNOWN_VARIABLES
        """A dictionary of known variables."""

        self.grid: Grid = grid
        """The configured Grid to be used in a simulation."""
        self.data = Dataset()
        """The :class:`~xarray.Dataset` used to store data."""
        self.variable_validation: dict[str, dict[str, str | None]] = {}
        """Records validation details for loaded variables.

        The validation details for each variable is stored in this dictionary using the
        variable name as a key. The validation details are a dictionary, keyed using
        core axis names, of the :class:`~virtual_ecosystem.core.axes.AxisValidator`
        subclass applied to that axis. If no validator was applied, the entry for that
        core axis will be ``None``.
        """

    def __repr__(self) -> str:
        """Returns a representation of a Data instance."""

        if self.data:
            return f"Data: {list(self.data.data_vars)}"

        return "Data: no variables loaded"

    def __setitem__(self, key: str, value: DataArray) -> None:
        """Load a data array into a Data instance.

        This method takes an input {class}`~xarray.DataArray` object and then matches
        the dimension and coordinates signature of the array to find a loading routine
        given the grid used in the {class}`virtual_ecosystem.core.data.Data` instance.
        That routine is used to validate the DataArray and then add the DataArray to the
        {class}`~xarray.Dataset` object or replace the existing DataArray under that
        key.

        Note that the DataArray name is expected to match the standard internal variable
        names used in Virtual Ecosystem and this is enforced against the dictionary of
        known variables.

        The method also adds unit and description metadata to from the known variables
        database to attributes as they are written to the data object.

        Args:
            key: The name to store the data under
            value: The DataArray to be stored

        Raises:
            TypeError: when the value is not a DataArray.
        """

        if not isinstance(value, DataArray):
            to_raise = TypeError(
                "Only DataArray objects can be added to Data instances"
            )
            LOGGER.critical(to_raise)
            raise to_raise

        if key not in self.known_variables:
            msg = f"Attempt to add unknown variable to data: '{key}'"
            LOGGER.critical(msg)
            raise ValueError(msg)

        if key not in self.data.data_vars:
            LOGGER.info(f"Adding data array for '{key}'")
        else:
            LOGGER.info(f"Replacing data array for '{key}'")

        # Add variable_metadata from known variables database - these needs to be done
        # for both adding and replacing variables as the science models do not attempt
        # to persist array attributes during calculations.
        variable = self.known_variables[key]
        value.attrs.update({"unit": variable.unit, "description": variable.description})

        # Validate and store the data array
        value, valid_dict = validate_dataarray(value=value, grid=self.grid)
        self.data[key] = value
        self.variable_validation[key] = valid_dict
```

davidorme:
# Description

I was trying to implement a couple of things:

* The data science team wanted units and description in NetCDF outputs. The obvious route there is to add attributes to the `xarray.Dataset` in `Data` from the `KNOWN_VARIABLES` when the `Data.__set_item__` method ran. Then they're dumped automatically. You could also add them right before outputting them...
* Whilst I was in that method, I thought it made sense to check that only things in `KNOWN_VARIABLES` can be added to `Data` - it turns out this is quite important for checking that developers do actually update the `data_variables.toml` (or are alerted to changes in names or deprecated variables)!

There's a problem though - because `KNOWN_VARIABLES` is a registry, it has a state that is dependent on the particular sequence of code execution. That's really hard in tests - the `test_variables.py` explicitly has a bunch of fixtures to reset the registries constantly because of this. If we want to use `KNOWN_VARIABLES` outside of just the variable setup (which I think we do) then all of the tests suffer the same issue.

I think the answer is to discard the registry approach for `KNOWN_VARIABLES` and `RUNTIME_VARIABLES` and make these explicit variables within the model.

I've added `ve.core.variables_new.py`:

* Drops the registries.
* Drops some functions and arguments that were only there to support the creation of the `variables.rst` file.
* Merges the `Variable` and the `VariableMetadata`: they were very similar but `Variable` had the post init fields for the model vars and `VariableMetadata` used `pydantic` for validation. We now have a single pydantic dataclass that loads and validates `data_variables.toml` _and_ that we can use for tracking variable usage in models
* The `VariableMetadata` validation now handles the validation of the axes data - unique names that appear in the VALIDATORS registry.
* Adds a `load_known_variables` function that _returns_ `dict[str, VariableMetadata]`.  I've used this in the main code to populate a new `Data.known_variables` attribute which seemed like a sensible home for it. This replaces `register_all_variables`.
* Updates `setup_variables` and the various `variables._collect...` functions:
  *  I've moved the checking that variables are known into a stand alone function (`_check_model_variables_are_known`) - less repetition, separation of concerns.
  * I've updated the arguments so that they accept `known_variables` and use that as necessary to add variables to a second `runtime_variables` dictionary, which is then used to track variable usage by models.
  * The `setup_variables` function now returns a dictionary of variables with populated model usage attributes.
* I've fixed a bunch of missing variables
* I've added the metadata to NetCDF outputs.

@dalonsoa Does this seem ok to you - it seems less brittle to me and allows us to use the variable database more widely. If so I'll polish up the rough edges and take it out of draft

Fixes #1295

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+        runtime_variables[var] = known_variables[var]
+        runtime_variables[var].vars_populated_by_init.append("data")
+
+
+def _collect_vars_populated_by_init(
+    models: list[type[base_model.BaseModel]],
+    runtime_variables: dict[str, VariableMetadata],
+    known_variables: dict[str, VariableMetadata],
+) -> None:
+    """Initialise the runtime variable registry.
```
dalonsoa: Well, data has already been added in the previous function, so not really initialising the registry here.

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+        known_variables: A dictionary of known variables
+
+    Raises:
+        ValueError: If a variable required by a model is not in the known variables
+            registry or if it is already initialised by another model.
+    """
+    for model in models:
+        for var in model.vars_populated_by_init:
+            if var in runtime_variables:
+                raise ValueError(
```
dalonsoa: Maybe collect first all the things that are wrong and then raise the exception, as done in `_check_model_variables_are_known`

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+        models: The list of models to check.
+        runtime_variables: A dictionary of variables being used in this runtime
+
+    Raises:
+        ValueError: If a variable required by a model is not in the known variables
+            registry or the runtime registry.
+    """
+    for model in models:
+        for var in model.vars_required_for_init:
+            if var not in runtime_variables:
```
dalonsoa: Same here.

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+    for model in models:
+        for var in model.vars_populated_by_first_update:
+            if var in runtime_variables:
+                v = runtime_variables[var]
+                initialiser = (
+                    v.vars_populated_by_init[0]
+                    if v.vars_populated_by_init
+                    else v.vars_populated_by_first_update[0]
+                )
+                raise ValueError(
```
dalonsoa: And here.

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+        known_variables: A dictionary of known variables
+
+    Raises:
+        ValueError: If a variable required by a model is not in the known variables
+            registry or the runtime registry.
+    """
+    for model in models:
+        for var in model.vars_updated:
+            if var not in runtime_variables:
+                raise ValueError(
```
dalonsoa: And here.

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+        known_variables: A dictionary of known variables
+
+    Raises:
+        ValueError: If a variable required by a model is not in the known variables
+            registry or the runtime registry.
+    """
+    for model in models:
+        for var in model.vars_required_for_update:
+            if var not in runtime_variables:
+                raise ValueError(
```
dalonsoa: And here!

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+
+    # Check all the variables in the models are present in known variables
+    _check_model_variables_are_known(models, known_variables)
+
+    # Variables related to the initialisation step
+    _collect_initial_data_vars(
+        vars=data_vars,
+        runtime_variables=runtime_variables,
+        known_variables=known_variables,
+    )
```
dalonsoa: There's nothing wrong with this approach, but I find more pythonic and explicit to return the new variables and update the `runtime_variables` dictionary outside, rather than updating the input variable directly. Up to you:

```suggestion
    runtime_variables.update(
        _collect_initial_data_vars(
            vars=data_vars,
            runtime_variables=runtime_variables,
            known_variables=known_variables,
        )
    )
```
And the same with the others.

virtual_ecosystem/core/variables_new.py
```diff
@@ -0,0 +1,472 @@
...
+def load_known_variables() -> dict[str, VariableMetadata]:
+    """Loads the known variables from the variable database.
+
+    The pydantic classes handle validation of the input.
+    """
+
+    with open(
+        str(resources.files("virtual_ecosystem") / "data_variables.toml"), "rb"
+    ) as f:
+        known_vars = tomllib.load(f)
```
dalonsoa: Rather than hard coding this path here, make it an input argument, defaulting to this value, so that an alternative can be provided if needed, eg, via tests. Moreover, it could be useful to provide the `known_vars` object as input as well, so they can be loaded from elsewhere combining multiple variable files, if needed at some point via configuration. This will be one step to make VE extensible with new models without changing the base source code.

Something like:


```suggestion
def load_known_variables(known_vars: dict[str, Any] | None = None, var_file: str | None = None) -> dict[str, VariableMetadata]:
    """Loads the known variables from the variable database.

    The pydantic classes handle validation of the input.
    """
    if var_file is None:
        var_file = str(resources.files("virtual_ecosystem") / "data_variables.toml")
        
    if known_vars is None:
        with open(var_file) as f:
            known_vars = tomllib.load(f)
```

dalonsoa (commented):
I love it. Way cleaner and robust that what I coded back in the day. Very well done. I have a few comments and suggestions, none of them blockers.

davidorme:
Thanks @dalonsoa - I'll shift the testing from variables over to the new system and work through your comments. Once everything is passing, I'll send it back over, but if the main structure is sound then it shouldn't be a big change from this.

virtual_ecosystem/core/variables_new.py
davidorme: I can definitely see your point, but because the entries are dataclasses, each function would need to generate a dictionary of replacement versions, edited from the incoming values. It would be very much more verbose and I think that would reduce the clarity of the code.

I have however stuck warnings all over the place that `runtime_variables` is being updated by reference.

virtual_ecosystem/core/variables_new.py
davidorme: Yeah - I think this would be good. I think the easiest way to do it is to have these functions return a list of error messages. If we concatenate those at the bottom of `setup_variables` then it is either `[]` (hooray!) or not, in which case we log all the contents and raise a general error.

However, I'm going to kick that down the road!

virtual_ecosystem/core/variables_new.py
davidorme: I **know** 😄

virtual_ecosystem/core/variables_new.py
davidorme: See #1302

virtual_ecosystem/core/variables_new.py
davidorme: I've added the filepath option - but I've stopped there for now because it opens further questions about how to resolve duplication and how to add new models. That is something I think we need to do - it shouldn't just be to add them into the package source - but I don't want to get deep into the weeds second guessing myself 😄 

Or at least not _today_...

davidorme:
@dalonsoa Could you give the new module a final check?

* I've tidied up the new `variables` module and updated the docstrings.
* I've ported your original test suite for `variables` over to the new structure.
* The number of files has ballooned horribly but they are 90% changing variable names in test modules and input files, because adhoc variable names cannot now be added to the `Data` object.
* I've worked around that in a couple of tests, where it would be less clear to use real variable names.
*  I've moved `to_camel_case` and `_discover_models` into `core.base_model` - they're more at home there. 
* I've removed the old implementation.
* I've fixed the niggles in tests and docs. That includes removing the old `variables.rst` mechanism completely. That is being replaced by that new Java powered table coming in via #1248.

dalonsoa (approved):
All good from my side. Much, much nicer, indeed.

PR #1298, opened 2026-01-29
1288 add daily loop to microclimatepy

### LINKED ISSUE #1288: add daily loop to microclimate.py
he microclimate will need to run on a subdaily time step. This requires

- a function to identify the month
- a function to identify the latitude
- a function to identify the daily amplitude (potentially later, can be constant for now)
- a function to generate a diurnal cycle
- generating empty arrays to store hourly data
- a loop over 24h
- a function to calculate the daily means
- returning the daily means

vgro:
This PR adds a sub-daily loop for the microclimate and makes some of the test more general (not testing for explicit values).

There are still some problems with the energy balance and time integration (especially with the understorey), however, **I would appreciate feedback on the loop structure before making it more complicated (in microclimate.py)** - I would rather make these fixes in a separate PR.

All unit tests pass but `ve_run` returns silly values after the first month. I have a suspicion that this comes from the example data.

Fixes #1288 

**!! UPDATE !!**
I have now broken all the steps into individual functions and created a `static` dict for the variables that are updated only once, and `state` dict for the variables that changing every hour. There are still a few minor hacks to do with the understorey LAI being to large and the time integration, but this should be much easier to address now that it is more transparent.

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -413,17 +415,21 @@ def generate_diurnal_cycle_from_monthly_data(
 
     evapotranspiration_hourly = np.where(
         sw_sum > 0,
-        monthly_evapotranspiration[None, :, :] * shortwave_absorption_hourly / sw_sum,
-        monthly_evapotranspiration[None, :, :] / 24.0,
+        monthly_evapotranspiration[None, :, :]
+        / 30
```
jacobcook1995: Is this implicitly assuming a 30 day month?

jacobcook1995: Might be best to make this an argument to the function, even if you just end up feeding 30 days in for now

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -507,3 +513,74 @@ def mean_to_layers(
...
+    exclude: Iterable[str] = (),
+) -> None:
+    """Validate all variables are in update dictionary.
+
+    Args:
+        names: variable names in output
+        values: variables in hourly update
+        exclude: variable names to ignore in the comparison
+
+    Returns:
```
jacobcook1995, line 594: A "Raises:" section for the error is probably more informative here

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -191,401 +195,539 @@ def run_microclimate(
...
+        # -------------------------------------------------------------------------
+
+        # Solve energy balance for canopy temperature, [C]
+        canopy_temperature = energy_balance.solve_canopy_temperature(
+            canopy_temperature_initial=canopy_temperature,
+            air_temperature=canopy_air_temperature,
+            evapotranspiration=evapotranspiration_canopy,
+            absorbed_shortwave_radiation=shortwave_absorption_canopy,
+            absorbed_longwave_radiation=downward_longwave_radiation
+            + 0.5 * (longwave_emission_understorey * abiotic_constants.leaf_emissivity),
```
jacobcook1995: I would probably try and avoid having a calculation this complex (i.e. something that isn't just a simple addition or multiplication) in the argument of a function

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -191,401 +195,539 @@ def run_microclimate(
...
+        if not isinstance(energy_balance_canopy, dict):
+            to_raise = ValueError("The energy balance has not returned any fluxes!")
+            LOGGER.error(to_raise)
+            raise to_raise
+
+        longwave_emission_canopy = energy_balance_canopy["longwave_emission_canopy"]
+        latent_heat_flux_canopy = energy_balance_canopy["latent_heat_flux_canopy"]
+        sensible_heat_flux_canopy = energy_balance_canopy["sensible_heat_flux_canopy"]
+
+        # Update canopy air temperatures, [C], # TODO integration interval 1 hour
```
jacobcook1995: I guess this TODO is now resolved?

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -191,401 +195,539 @@ def run_microclimate(
...
+            time_interval=time_interval,
+        )
 
-    # -------------------------------------------------------------------------
-    # Soil energy balance
-    # -------------------------------------------------------------------------
+        # Conductive flux from understorey vegetation to soil, [W m-2]
+        # A positive flux is directed towards the soil
+        conductive_flux_understorey = calculate_conductive_flux_understorey(
+            soil_temperature=soil_temperature[0],
```
jacobcook1995: what is `soil_temperature[0]` and do the other values in that array get used?

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -191,401 +195,539 @@ def run_microclimate(
...
+        time_dim=24,
+        layers=layer_structure.n_layers,
+        cell_ids=data.grid.n_cells,
+    )
+
+    # -----------------------------------------------------------------------------
+    # Hourly loop
+    # -----------------------------------------------------------------------------
+
+    for hour in range(24):
```
jacobcook1995: It might be worth making a new function that contains everything within this loop, and then looping over that function. Mainly as then it will be clearer what variables are coming from outside the loop and which ones are generated within it

jacobcook1995 (commented):
The looping looks correct to me, so I can't figure out why the 2nd simulation step crashes. I had a few minor comments and suggestions but nothing that address that sorry

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -507,3 +513,74 @@ def mean_to_layers(
...
+    variables: dict[str, DataArray],
+    time_dim: int,
+    layers: int,
+    cell_ids: int,
+) -> dict[str, NDArray[np.floating]]:
+    """Create a data_record dict with a new leading time dimension.
+
+    Assumptions:
+        - 1D variables have shape (cell_ids,)
+        - 2D variables have shape (layers, cell_ids)
```
dalonsoa: Isn't there a way of not making this assumption and just checking the `DataArray`s dimension names?

virtual_ecosystem/models/abiotic/abiotic_tools.py
dalonsoa: Agree.

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -90,18 +90,29 @@ def initialise_canopy_and_soil_fluxes(
...
+    for name in (
+        "sensible_heat_flux",
+        "latent_heat_flux",
+        "longwave_emission",
+    ):
+        output[name] = base_flux.copy()
+
+    # 1D fluxes (cell-wise)
+    cell_flux = DataArray(
+        np.full(base_flux.shape[1], initial_flux_value),
```
dalonsoa: This `shape[1]` is a bit mysterious. It would be better to do something like:

```suggestion
        np.full(base_flux.sizes["cell_id"], initial_flux_value),
```
Or whatever is the appropriate name for that dimension. See [DataArray.sizes](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.sizes.html).

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -58,29 +70,43 @@ def run_microclimate(
         dictionary with updated microclimate variables
     """
 
-    output = {}
-    canopy_index = layer_structure.index_filled_canopy
-    surface_index = layer_structure.index_surface_scalar
-    atm_index = layer_structure.index_filled_atmosphere
+    output: dict[str, DataArray] = {}
+
+    idx = SimpleNamespace(
```
dalonsoa: Maybe add a one line explanation of what this is for.

virtual_ecosystem/models/abiotic/microclimate.py
dalonsoa: Most definitely this needs some breaking. This function is already enormous, so anything that can be extracted as separate functions (and the inner part of the loop is a good choice) would be welcome.

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -191,401 +195,539 @@ def run_microclimate(
...
-    # Latent heat flux understorey vegetation, [W m-2]
-    latent_heat_flux_understorey = energy_balance.calculate_latent_heat_flux(
-        evapotranspiration=evapotranspiration[surface_index].to_numpy(),
-        latent_heat_vapourisation=latent_heat_vapourisation_j[-1],
-        time_interval=time_interval,
-    )
+        if not isinstance(energy_balance_canopy, dict):
+            to_raise = ValueError("The energy balance has not returned any fluxes!")
+            LOGGER.error(to_raise)
+            raise to_raise
```
dalonsoa: Can this actually happen?

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -12,15 +14,21 @@
...
 from virtual_ecosystem.core.model_config import CoreConstants
 from virtual_ecosystem.models.abiotic import abiotic_tools, energy_balance, wind
+from virtual_ecosystem.models.abiotic.energy_balance import (
+    calculate_conductive_flux_understorey,
+)
 from virtual_ecosystem.models.abiotic.model_config import AbioticConstants
 from virtual_ecosystem.models.abiotic_simple.model_config import AbioticSimpleBounds
 
 
 def run_microclimate(
```
dalonsoa: This is a really serious, 700-long function that do a lot of stuff. While a lot of the logic has already been extracted into separate functions within `abiotic_tools` or `energy_balance`, I feel it is unmanageable and is reaching a point it is difficult to change or update its functionality.

Focusing on the code architecture, I would suggest to make this a class and break things down into smaller chunks that update its state. Something that looks like:

```python
class Microclimate:

    def __init__(...):
        # Inputs would be anything "run" independent, like
        # `data`, `layer_structure`, etc.
        # There will also be other attributes to store the transition state
        # during a run, as well as convenience shortcuts.

    def _reset_state(self):
        # Clears the state of the internal variables before a new run

    def _collect_output(self) -> dict[str, DataArray]:
        # Collects the state variables to be output after a run

    def run(...) -> dict[str, DataArray]:
        # inputs here will be run specific stuff, like `time_index` or `month`
        # This just calls other methods that update the internal state
        self._reset_state()
        ...
        # (many other methods)
        ...
        return self._collect_output()

    ...
```

I think this is a big undertaking, but I feel it will make your life - as main coder - and ours - as reviewers - way simpler.

dalonsoa (dismissed):
Just reviewed the `microclimate.py` file, and made comments there. Cannot judge the science, so just focused on the code itself.

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: I have included it in the function up to the abiotic model, how can I best include it as a ve argument?

virtual_ecosystem/models/abiotic/microclimate.py
vgro: No this is what I initially wanted to address, before I rewrote the whole thing...This problem still persists.

virtual_ecosystem/models/abiotic/microclimate.py
vgro: no longer relevant

virtual_ecosystem/models/abiotic/microclimate.py
vgro: I have rewritten this to break everything up and bring it back together. It works but is incredibly slow,probably because I chose dicts instead of the class structure.

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: done

virtual_ecosystem/models/abiotic/abiotic_tools.py
jacobcook1995: Not 100% sure, it could be included as an `abiotic` model constant or similar?

I guess the main thing for me is not having an assumption like that buried deep in the code of a function. So, so long as there's a place people are likely to look at where "number_of_days_in_a_month = 30" is explicitly written out, that basically addresses the problem I had

virtual_ecosystem/models/abiotic/microclimate.py
vgro: This is to shut up mypy, for some reason it refuses to allow two different types to be returned. I have replaced this with `# type: ignore` for now

virtual_ecosystem/models/abiotic/microclimate.py
vgro: I have structure the code like this now, although not in a class style but keeping the flow and using dicts. I assume this is why it is so slow

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: yes, I absolutely agree

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -405,17 +435,21 @@ def generate_diurnal_cycle_from_monthly_data(
 
     evapotranspiration_hourly = np.where(
         sw_sum > 0,
-        monthly_evapotranspiration[None, :, :] * shortwave_absorption_hourly / sw_sum,
-        monthly_evapotranspiration[None, :, :] / 24.0,
+        monthly_evapotranspiration[None, :, :]
+        / days
+        * shortwave_absorption_hourly
+        / sw_sum,
```
jacobcook1995: Think for readability this should either group the denominator and numerator or use brackets as `a/b*c/d` is hard to interpret

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -405,17 +435,21 @@ def generate_diurnal_cycle_from_monthly_data(
...
+        monthly_evapotranspiration[None, :, :] / days / hours_per_day,
```
jacobcook1995: Same comment as above `a/b/c` is fairly hard to interpret

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -161,317 +214,513 @@ def run_microclimate(
...
+        monthly_shortwave_absorption=data["shortwave_absorption"].to_numpy(),
+        monthly_relative_humidity=data["relative_humidity_ref"]
+        .isel(time_index=time_index)
+        .to_numpy(),
+        monthly_evapotranspiration=static["evapotranspiration"],
+        monthly_soil_evaporation=data["soil_evaporation"].to_numpy(),
+        latitude_deg=latitude,
+        month=month,
+        days=days,
+        daily_temp_amplitude=5,  # TODO abiotic_constants or input data
```
vgro:
```suggestion
        daily_temp_amplitude=5,  # TODO #1440 abiotic_constants or input data
```

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -405,17 +435,21 @@ def generate_diurnal_cycle_from_monthly_data(
...
 
     # Soil evaporation (distributed like radiation)
     soil_evaporation_hourly = np.where(
         shortwave_absorption_hourly.sum(axis=1).sum(axis=0)[None, :] > 0,
         monthly_soil_evaporation[None, :]
+        / days
         * shortwave_absorption_hourly.sum(axis=1)
         / shortwave_absorption_hourly.sum(axis=1).sum(axis=0)[None, :],
-        monthly_soil_evaporation[None, :] / 24.0,
+        monthly_soil_evaporation[None, :] / days / hours_per_day,
```
jacobcook1995: Same comment as above

virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -499,3 +526,101 @@ def mean_to_layers(
...
+) -> None:
+    """Validate all variables are in update dictionary.
+
+    Args:
+        names: variable names in output
+        values: variables in hourly update
+        exclude: variable names to ignore in the comparison
+
+    Returns:
+        None. Raises ValueError if variable mismatch is detected.
```
jacobcook1995: This should be a "Raises: " block (like the "Returns:" and "Args:" block) so that it gets picked up properly  by the automatic API documentation

virtual_ecosystem/models/abiotic/abiotic_model.py
```diff
@@ -270,12 +270,28 @@ def _update(self, time_index: int, **kwargs: Any) -> None:
...
             **kwargs: Further arguments to the update method.
         """
+        month = (
+            self.model_timing.update_datestamps[time_index]
+            .astype("datetime64[M]")
+            .astype(int)
+            % 12
+            + 1
+        )
+        # TODO The following are placeholders and should be arguments to abiotic_model
```
vgro:
```suggestion
        # TODO #1441 These placeholders should be arguments to abiotic_model
```

virtual_ecosystem/models/abiotic/microclimate.py
```diff
@@ -161,317 +214,513 @@ def run_microclimate(
...
+    # Update surface air temperatures, [C]
+    flux_from_soil = (
+        -state["sensible_heat_flux"][idx.topsoil]
+        + 0.5 * state["longwave_emission"][idx.topsoil]
+    )
+
+    # TODO The surface layer is only 10cm thick and therefore cannot absorb all energy
+    # emitted by high LAI of understorey without exploding temperatures. This scaling
+    # factor artificially expands the layer thickness to 1.5 m reference height; this
+    # needs to be addressed with location of understorey vegetation.
```
vgro:
```suggestion
    # needs to be addressed with location of understorey vegetation. #1439
```

jacobcook1995 (approved):
Can't comment on the science, but on the code side this is massively cleaner and easier to follow!

Had a few minor suggestions, but nothing major

jacobcook1995:
Also given how slow this makes the ve_run test, I think it's probably a good idea to switch the example to use `abiotic_simple` instead until the `abiotic` model performance is improved

PR #1311, opened 2026-02-03
Added a new Science top level section

### LINKED ISSUE #1265: Need to make a specific theory docs section
At the moment, both "Theory" and "Implementation" exist under the same section heading. This is a significant issue as this mixes information that general scientific readers of our papers would be interested in, with information that only people who want to tinker with specific model implementations need.

This should be fixed by changing the section to just be "Theory" (it warrants being a section as it has a definable type of reader associated, i.e. people who are reading our eventual papers).

The "Implementation" details should then be moved to a new subsection of the "Development" docs (as these details are only developer relevant)

The information in @vgro's "Implementation" pages for our theory docs (basically because physical and biological sciences talk about these things differently). So should be retained in the new section. The contents of her current "Theory" docs can be deleted

I'll make the associated PR against `develop` and then merge into the documentation branch so that if people add docs here as part of model development they are looking at the same structure

jacobcook1995:
# Description

Documentation built [here](https://virtual-ecosystem.readthedocs.io/en/1265-need-to-make-a-specific-theory-docs-section/virtual_ecosystem/science.html)

This PR changes the "Implementation" top level heading to instead be "Science".

Under this heading I have changed "Theory" and "Implementation" to instead be "Underlying theory" and "How it works"

I also moved "Project Team" to be it's own top level heading

I moved the variables table into the user documentation

I only made minimal changes to the actual content in the end as this PR is more about setting up the new structure for this section, and the content is likely to be heavily revised based on data team feedback etc.

I did make some changes that seemed like low hanging fruit:

@TaranRallings, I removed a paragraph from the animal implementation that was purely about the code structure and so only of interest to developers

@davidorme, I moved the details about PFT parameterisation out from the plant implementation docs into a new section of the Users docs as that felt like a more reasonable place

@vgro, I deleted a fair bit of content from the microclimate and hydrology theory docs as it mainly seemed to be listing things of general importance to microclimate and hydrology research. I thought it was better to point readers to the actual equations (in the implementation sections) sooner. Let me know if you think this content should be put somewhere rather than being deleted.

Fixes #1265


davidorme:
> I also moved "Project Team" to be it's own top level heading

I don't think we should have a single page with a top level heading - it pushes References into a dropdown and it doesn't make sense to have a whole navigation system point to a single item. I think References is a bit marginal there too, to be honest 😄 - it could probably live at the bottom of the Science tab (although it is used more widely, the references auto link to it).

If we remove References, we could have an "About" top level that contains details about the funding and the team and any other stuff, but I don't think having the team in the Science section is too bad.

jacobcook1995:
> I don't think we should have a single page with a top level heading - it pushes References into a dropdown and it doesn't make sense to have a whole navigation system point to a single item. I think References is a bit marginal there too, to be honest 😄 - it could probably live at the bottom of the Science tab (although it is used more widely, the references auto link to it).
> 
> If we remove References, we could have an "About" top level that contains details about the funding and the team and any other stuff, but I don't think having the team in the Science section is too bad.

That's a better idea, I just wasn't really a fan of burying the details of the team at the bottom of a section. But agree that Project Team isn't really a section in it's own right

davidorme (approved):
LGTM - can we remove `var_generator.py`, since we no longer use `generate_variable_table`?

jacobcook1995:
I've just changed this PR so that the content I was previously deleting from the abiotic and hydrology theory section is now moved to new, less prominent pages. I'm going to go ahead and merge this to the main documentation branch now (@vgro I'll follow up with you once you are back at work to check that you are happy with these changes)

PR #1315, opened 2026-02-05
Moved developer focused content out of User documentation

### LINKED ISSUE #1098: Constant removal rates from static models can lead to negative values in non-static models
When the simulation is run with all models static apart from the soil model the densities of the mycorrhizal fungi get driven to negative values.

Image

This happens because the rate at which plants take nutrients from the mycorrhizal fungi is static, so nutrients are taken from the fungi even if they have no biomass left to supply them from. When I do the same run with the other models not held static the problem goes away, as the rate at which plants take nutrients from mycorrhizal fungi reduces as nutrients become more scarce

Image

I think this is a problem that could potentially crop up anywhere where a removal rate is defined in one model and passed to another, with herbivory and detritivory being the other examples that spring to mind.

One option for resolving this bug is to make the relevant section of the soil model code more defensive, i.e. to disallow out negative mycorrhizal fungi density. I'm am not sure whether that makes sense though, as a negative value cropping up when the other models aren't held static would be a really good indicator of problems in our models, and a more defensive approach could mask this

The alternative is to just ignore the issue as it is static model issue which absolutely could be fixed through a more sensible choice of initial conditions/static model data input. We probably would want to mention the potential for this sort of error in the docs, and probably mention that the static model is intended as a developer tool so you need to be on the lookout for weirdnesses like that

Anyway (as you can probably tell from the rambling above) I'm not sure which approach makes most sense. If anyone has any thoughts please let me know

### LINKED ISSUE #1215: Review the contents of "Configuring your model"
This section needs to be reviewed, to ensure that everything is user focused. Everything else should either be more to a developer focused section or to be reference materials

### LINKED ISSUE #1216: Review the contents of "Adding data to your model"
This section needs to be reviewed, to ensure that everything is user focused. Everything else should either be more to a developer focused section or to be reference materials

### LINKED ISSUE #1217: Make an documentation section for developer focused docs
A lot of the content of "Using the Virtual Ecosystem" is developer focused rather than user focused. This content should be moved to the "Development" section, but sensible section (+ warnings to users) need to be added to allow this

jacobcook1995:
# Description

This PR mostly just moves content about. Other than that I have added a very brief landing page for the developer docs. I also added a warning to the section about configuring static models that warn users that it is very much an advanced feature.

The docs for this branch can be [found here](https://virtual-ecosystem.readthedocs.io/en/1217-make-an-documentation-section-for-developer-focused-docs/)

Fixes #1098
Fixes #1215 
Fixes #1216
Fixes #1217


davidorme (approved):
👍

PR #1323, opened 2026-02-09
Add brief section to user docs about virtual environments

### LINKED ISSUE #1310: Add brief section to User docs about virtual environments
At present we just describe how to install python in the User docs. Then we have a section on installing `poetry` in the developer docs (which is overkill for most users).

Users probably should be using a virtual environment as this is generally good practice for python. The user docs need to. describe how to do this (without going all the way to explaining how to use `poetry`)

jacobcook1995:
# Description

Another small docs PR. This one adds a brief description of how to activate a virtual environment and install VE to it (via pip). I also deleted the section that mentions `poetry` as that isn't something we want to encourage users to do. 

Here are the [relevant docs](https://virtual-ecosystem.readthedocs.io/en/1310-add-brief-section-to-user-docs-about-virtual-environments/using_the_ve/installing_ve.html) for this PR.

Fixes #1310


davidorme (approved):
👍

PR #1338, opened 2026-02-13
Adding model variable attribute validation

### LINKED ISSUE #1337: Validate claimed model variable initialisation at run time
We don't have any actual validation that the model `vars_populated_at_init` and `vars_populated_at_first_update` are correct. Currently that's entirely manual and we could validate it at runtime.

davidorme:
# Description

This PR adds a function in `main.py` that checks that the variables added to `data` during a model `__init__` or  first `_update` actually match the `vars_populated_at_init` and `vars_populated_at_first_update` attributes of the model. These currently aren't validated and this seems like the simplest way to try and keep these attributes up to date. 

I think this could be extended to verify the other `var_` attributes too, but keeping the variable population correct matters to resolving the model running order.

Fixes #1337 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
This seems like a sensible change to me!

PR #1342, opened 2026-02-16
Explain how to install the present state of develop via pip

### LINKED ISSUE #1341: Make clear that `poetry` should not be used by users
At the moment the docs suggest users use `poetry` in order to keep up to date with the most recent changes in the repo. This is massive overkill unless users are making changes to the source code, at which point they are developers rather than users.

So, the User docs should not mention poetry at all (this has already been implemented on the docs update branch). As a replacement we should suggest installing the current state of `develop` via `pip`, e.g.

```shell
# Install the lastest release into a release venv
python -m venv ve_release
source ve_release/bin/activate
pip install --upgrade virtual_ecosystem
deactivate

# Install the latest code in develop into a develop venv
python -m venv ve_develop
source ve_develop/bin/activate
pip install --upgrade git+https://github.com/ImperialCollegeLondon/virtual_ecosystem.git@develop
deactivate
```

jacobcook1995:
# Description

This PR adds @davidorme's instructions for installing the present state of `develop` via `pip`. I've been fairly paranoid about warning users away from it, mainly because while all our present users should probably be using this, we don't want to accidentally frame this as "here's how you get the newest (and therefore best) code" for future users.

Here is a [built version of these docs changes](https://virtual-ecosystem.readthedocs.io/en/1341-make-clear-that-poetry-should-not-be-used-by-users/using_the_ve/installing_ve.html#advanced-installing-the-current-state-of-develop).

Fixes #1341


davidorme (approved):
Looks great - very clear about pros and cons.

PR #1348, opened 2026-02-17
Switch cell_xy checks over to using a STRtree query

### LINKED ISSUE #1347: Cell coordinate checking is unreasonably slow
I was running the example data mentioned in #1314 and the data loading was taking _forever_.

The immediate bottleneck is the code to check that cell xy coordinates map onto grid cells correctly, which is _woeful_. To be fair to myself, there is a note saying it might be a bit crap, but it turns out to be seriously crap even with a 50x50 grid. Fortunately, the note on the code suggests an obvious improvement route.

davidorme:
# Description

This PR replaces the incredibly slow check that incoming cell coordinate data maps 1-to-1 onto the grid with a SRTtree spatial search structure (https://shapely.readthedocs.io/en/latest/strtree.html#).

I've added a test that is purely there to run speed checks for three pairs of (nx, ny) grid sizes. There might be a better place for it, but maybe that will emerge from the upcoming performance review.

Using `pytest tests/core/test_grid.py::test_map_xy_to_cell_ids_performance --durations=0 -n 0` to run this test gives the following runtimes for the old code vs this PR.

Before:

    1953.27s call test_grid.py::test_map_xy_to_cell_ids_performance[150-150]
    371.25s call  test_grid.py::test_map_xy_to_cell_ids_performance[100-100]
    23.24s call   test_grid.py::test_map_xy_to_cell_ids_performance[50-50]

After:

    1.19s call    test_grid.py::test_map_xy_to_cell_ids_performance[150-150]
    0.56s call    test_grid.py::test_map_xy_to_cell_ids_performance[100-100]
    0.13s call    test_grid.py::test_map_xy_to_cell_ids_performance[50-50]

Fixes #1347

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/core/test_grid.py
+++ b/tests/core/test_grid.py
@@ -382,40 +382,45 @@ def test_grid_dumps():
 @pytest.mark.parametrize(
     argnames=["x_coord", "y_coord", "exp_exception", "exp_message", "exp_map"],
     argvalues=[
-        (
+        pytest.param(
             [0, 1, 2],
             [[0, 1], [0, 1]],
             pytest.raises(ValueError),
             "The x/y coordinate arrays are not 1 dimensional",
             None,
+            id="coords_not_1D",
         ),
-        (
+        pytest.param(
             [0, 1, 2],
             [0, 1],
             pytest.raises(ValueError),
             "The x/y coordinates are of unequal length",
             None,
+            id="coords_not_equal_length",
         ),
-        (
+        pytest.param(
             [0, 1, 2],
             [0, 1, 2],
             does_not_raise(),
             None,
             [[], [], []],
+            id="outside_grid",
         ),
-        (
+        pytest.param(
             [500000, 500100, 500200],
             [200000, 200100, 200200],
             does_not_raise(),
             None,
             [[90], [80, 81, 90, 91], [71, 72, 81, 82]],
+            id="on_borders",
         ),
-        (
+        pytest.param(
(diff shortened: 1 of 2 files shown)


dalonsoa (approved):
Looks good to me. When the time comes (hopefully from next week, or maybe March), we can use [`pytest-benchmark`](https://pypi.org/project/pytest-benchmark/) to get some performance information out of the existing testsuite with minimum changes. We will look into it when the time comes.

dalonsoa:
Actually, if you want to give it a try, just update the existing `test_map_xy_to_cell_ids` test and use `benchmark`. It should be as easy as replacing these:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/01128a39207e3cbf7c356480964eb5298b744927/tests/core/test_grid.py#L436-L437

by these:

```python
        cell_map = benchmark(
            fixture_square_grid.map_xy_to_cell_id
            np.array(x_coord)
            np.array(y_coord)
```

And you will have the performance information for free.

Then, the new custom test won't be needed, as the performance data will be readily available.

davidorme:
Interesting - hadn't seen `pytest-benchmark`, which is annoying because we've been doing something more ersatz on `pyrealm` 😄 

It does work - but I won't put it in this PR because it's only really useful once you have a database of previous profiling that you can use to compare runs to. I agree the current test is a hack, but I'd rather leave as is until we have a profiling plan and an initial DB setup.

PR #1358, opened 2026-02-23
Delete unused variables from soil/litter testing

### LINKED ISSUE #1308: Delete now unused variables from soil/litter testing
There's a bunch of variables in the variables table that had to be added with the new system to stop testing to break. Some of these are outdated variables that I still use in testing (e.g. `soil_c_pool_microbe`) I should go through and delete these.

There's also some that are intermediate values that shouldn't be described to users as they are never output, and can't really be explained without explaining the way in which the model is evaluated (e.g. `new_fungal_fruiting_body_production`). I don't *think* the usage of these in the model is actually a problem as they are returned by the integrator as a dictionary of `numpy` arrays. I'm fairly sure the problem is that I shove them into the data object for testing. I need to check that this is actually the case and if so delete them from the testing

jacobcook1995:
# Description

This PR deletes all the variables that are not described in the `variables.toml` from the places where they are used. Primarily this involves removing places where I just added extra variables to the data object in testing for convenience's sake.

@davidorme, I've removed "leaf_lignin" from the plants model entirely, as the litter model only uses "senesced_leaf_lignin", and I don't think a difference in senesced concentrations makes sense for lignin (only for N and P). Let me know if you are happy with this change, happy to take a different approach if you object to this change.

Fixes #1308

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Key checklist

## Further checks


davidorme:
Just to note that I asked @jacobcook1995 to remove `test_PlantsModel_calculate_turnover_constant_override`. The test was very  artificial, testing model configuration but via a nasty hack to a fixture, that bizarrely turned out to propagate across tests using the fixture. It really shouldn't have, but it did.

davidorme (approved):
👍

Issue #1314, opened 2026-02-05
Maliau simulation fails


arne-exe:
I tried to run the VE using Maliau input data but the simulation ended early and gave the error below.
It seems to be related to 3 missing variables, being ```per_propagule_annual_recruitment_probability```, ```root_exudates``` and ```per_stem_annual_mortality_probability```. 

I don't think it's related to input data files, perhaps something internal to the code?

```
(ve_testing) PS C:\Users\User\OneDrive\Documents\GitHub\ve_data_science> ve_run C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\data\scenarios\maliau\maliau_1\config `
>>     --out C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\data\scenarios\maliau\maliau_1\out `
>>     --logfile C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\data\scenarios\maliau\maliau_1\logfile.log
Starting Virtual Ecosystem simulation.
Logging to: C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\data\scenarios\maliau\maliau_1\logfile.log
* Loading configuration
pyrealm\<unknown>:64: SyntaxWarning: invalid escape sequence '\s'
pyrealm\<unknown>:64: SyntaxWarning: invalid escape sequence '\s'
* Saved compiled configuration: C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\data\scenarios\maliau\maliau_1\out\ve_full_model_configuration.toml
* Built core model components
* Initial data loaded
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\ve_testing\Scripts\ve_run.exe\__main__.py", line 7, in <module>
  File "C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\ve_testing\Lib\site-packages\virtual_ecosystem\entry_points.py", line 261, in ve_run_cli
    ve_run(
  File "C:\Users\User\OneDrive\Documents\GitHub\ve_data_science\ve_testing\Lib\site-packages\virtual_ecosystem\main.py", line 186, in ve_run
    models_init = initialise_models(
                  ^^^^^^^^^^^^^^^^^^
[...]

PR #1362, opened 2026-02-23
Fixing LUE and units in pmodel

### LINKED ISSUE #1354: Handle light use efficiency predictions when P Model is undefined
Under some conditions, the light use efficiency predictions of the P Model can be undefined (`np.nan`). We need to mask this - probably substituting zero. 

In addition, the P Model inputs are being passed in using kPa for atmospheric pressure and VPD and need to be in Pa.

davidorme:
# Description

* The light use efficiency reported by the P Model can be undefined in some extreme conditions. This results in `np.nan` values that propagate down to the soil model integration, which crashes the simulation (#1314). This is now fixed by using `np.nan_to_num` to force zero LUE under those conditions.
* The PModel fitting was using VPD and PATM straight out of the data in kPa and `pyrealm.PModel` requires Pascals.  Essentially the plants were growing in an incredibly humid near vacuum.
* The PR also restructures the `estimate_gpp` - it was doing a lot of subsetting the vertical layers down to the canopy layers and it seems much more straightforward to pad the canopy layers out to the full vertical layers.

Fixes #1354

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -888,34 +891,37 @@ def estimate_gpp(self, time_index: int) -> None:
...
-                continue
-
-            # GPP for each later is estimated as (value, dimensions, units):
-            #    LUE                (n_active_layers, 1)          [gC mol-1]
-            #    * cohort fAPAR     (n_active_layers, n_cohorts)  [-]
-            #    * canopy top PPFD  scalar                        [µmol m-2 s-1]
-            #    * stem leaf area   (n_active_layers, n_cohorts)  [m2]
-            #    * time elapsed     scalar                        [s]
+            # Get cohort data into vertical structure
+            n_layers_below_canopy = (
```
vgro, line 903: What exactly is `n_layers_below_canopy`, or how is it different from canopy layers?

vgro: Could you not just replace the canopy layers with zeros by indexing below? Not a problem how it's currently done.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -888,34 +891,37 @@ def estimate_gpp(self, time_index: int) -> None:
...
             )
 
+            # Mask all nans with zero. This includes all unoccupied canopy layers, but
+            # also patches canopy conditions where GPP is not estimable using the P
+            # Model (where m <= c*, see
+            # https://pyrealm.readthedocs.io/en/stable/api/pmodel_api.html#pyrealm.pmodel.jmax_limitation.JmaxLimitationWang17)
+            per_layer_gpp = np.nan_to_num(per_layer_gpp)
+
             # Calculate and store whole stem GPP in kg C
             self.per_stem_gpp[cell_id] = per_layer_gpp.sum(axis=0) * 1e-9
```
vgro, line 935: You could also use `np.nansum` here to save yourself the setting to zero, I think

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -947,20 +953,15 @@ def estimate_gpp(self, time_index: int) -> None:
...
             # grid cell transpiration in mm m-2 since last update
-            self.per_stem_transpiration[cell_id] = per_layer_transpiration_mm.sum(
-                axis=0
+            self.per_stem_transpiration[cell_id] = np.nansum(
+                per_layer_transpiration_mm, axis=0
             )
 
             # Calculate the total transpiration per layer in m2 in mm
-            transpiration[active_layers, cell_id] = (
+            self.data["transpiration"][:, cell_id] = (
```
vgro: For consistency, it would be good if you reinstated the `nan` where the canopy layers are not true here

vgro (changes requested):
Looks like a sensible fix to me, I just have a few minor comments, most important one that the output to data needs to have the `nan` values back.

virtual_ecosystem/models/plants/plants_model.py
davidorme: > What exactly is `n_layers_below_canopy`, or how is it different from canopy layers?

Yup - not clear. The cohort canopy data is for example a block of 3 canopy layers by 4 cohorts. For the multiplication by the cell_id column array, we need it to match the vertical array structure. So we pad it with one above for the above canopy layer and then `n_layers_below_canopy` gives the remaining number of layers below. So with e.g. 14 total layers , we'd pad fAPAR and LAI along the vertical layer axis using  `(1, 10)` to give a 14 x 4 array.

davidorme: Added this to code as comments.

virtual_ecosystem/models/plants/plants_model.py
davidorme: Done - I do wonder if it would be more efficient to use zero, but that's the kind of thing the performance testing can look into.

vgro (approved):
LGTM

PR #1372, opened 2026-02-23
Add disturbance timing and tests

### LINKED ISSUE #1366: Implement disturbance timing
It should largely follow this implementation:

```python
class DisturbanceTiming:

    def __init__(
            self, 
            model_timing: ModelTiming, 
            run_first: int: None, 
            then_run_every: int | None= None, 
            run_at: list[int] | None = None,
        ) -> None:
        "Create an instance"
        # Checks the values map onto the model timing and probably coerces 
        # the run_first/run_every form into a `run_at` list

    def check_run(self, time_index) -> bool:
        return True if time_index in self.run_at else False
```

dalonsoa:
# Description

First step towards the disturbance framework: the disturbance timing. Based on some inputs, a list of time indices when the disturbance model will run is created. This is then used at runtime to decide if the disturbance should run, given the current time index.

Fixes #1366

## Type of change

## Key checklist

## Further checks


davidorme (commented):
@dalonsoa This looks good and bang on spec.

So instead, I'm going to move the target as I have a few questions about my own proposed API and the user experience 😄 

* Does it make sense to have defaults? I think the default of the disturbance happening every step is a surprising and unlikely use case. I don't think some other default is any more sensible. In retrospect, I think it is clearer to have no defaults - users _must_ specify the disturbance timing?

* I think I've overcomplicated the constructor args. 
    * The primary mode could be `run_at: int | list[int] | None`, which implicitly supports `run_once`.
    * If `run_at` is `None` then users can instead use the convenience shortcut  `run_every: tuple[int, int]`. The tuple provides a configurable start point and step `(2,5) -> [2, 7, ...]`. We _could_ allow a single int for only specifying the step and default to run first on step zero (`5 -> [0, 5, ...]`), but again that is a surprising default, so better to make it explicit?

I think those configuration options are easier to document and explain and give a simpler API without surprising behaviour.

dalonsoa:
I agree. It would be best to enforce the user to provide the timing explicitly. That would  be enforced when creating the input layer (see https://github.com/ImperialCollegeLondon/virtual_ecosystem/issues/1369).

About the input arguments, I agree with your plan. The first option (`run_at`) is clear. About the second one, I'd use exactly the same syntax as `range`, mainly because it is familiar to most people using python and widely documented. My only concern there is that the mandatory argument there is `stop`, and not `start`, so maybe we can switch to make `start` mandatory and `stop` optional? So it would be:

```python
run_every = (`start`)  # Runs to the end, with step 1
run_every = (`start`, `step`)  # Runs to the end with step `step`
run_every = (`start`, `step`, `stop`)  # Runs to `stop` with step `step`
```

davidorme:
@dalonsoa Your suggested `run_every` scheme sounds good to me. I suspect 95% of usage would by `(start, step)` but at some point someone is going to want to run some other repeating pattern.

davidorme (approved):
LGTM

dalonsoa:
@davidorme I've merged all of the PR, so this one now contains the whole of the disturbance system. Have a look, if you want, as a final review. I'll let you merge this one when convenient. Of course, first I need to solve the issue with the documentation.

dalonsoa:
Finally! All checks in green, so feel free to merge whenever you want, @davidorme

PR #1373, opened 2026-02-24
Added up to date description of the release process + a new release issue template

### LINKED ISSUE #1279: We need a template for release PRs
With release PRs there's a couple of important things to remember.

1) That the version number has been incremented (otherwise the release pipeline fails

2) That the executed notebooks in the docs have been updated recently (see #1189)

We need this checklist in a place that people we actually look at (i.e. not buried in the docs) so a template is probably the best option

### LINKED ISSUE #1280: Release docs should be updated
The release process developer docs should be updated to:

1) No long mention the git flow release strategy as we are not actually using it

2) Mention that `ipynb` files need to be updated (and how this can be done)

3) Explain the process of using the documentation template (see #1279)

jacobcook1995:
# Description

I've updated the package release docs to talk about our package release process as it actually exists, rather than as we originally planned it. As part of this I added a step for checking that `ve_run` actually runs, and for checking if any of the executed Jupyter notebooks in the docs need to be updated.

I also added a new issue template for releases. This would be better as a PR template, but [GitHub doesn't currently support switching between pull request templates (well not without using query parameters anyway)](https://github.com/orgs/community/discussions/4620)

Fixes #1279 
Fixes #1280


davidorme (approved):
🥳

PR #1263, opened 2026-01-19
1132 remove fruit from stoichiometry create the fleshseed structure

sallymatson:
# Description

Please include a summary of the change and which issue is fixed. Please also include
relevant motivation and context. List any dependencies that are required for this
change.

Fixes # (issue)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [ ] Bug fix (non-breaking change which fixes an issue)

## Key checklist

- [ ] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [ ] All tests pass: `$ poetry run pytest`

## Further checks

- [ ] Code is commented, particularly in hard-to-understand areas
- [ ] Tests added that prove fix is effective or that feature works
- [ ] Relevant documentation reviewed and updated

PR #1375, opened 2026-02-24
Develop actual biomass representation

### LINKED ISSUE #1343: Develop actual biomass representation
We have allometric expectations of biomass from the T model and fruit extensions but we do not currently have a way to track the actual realised biomass of individuals after herbivory etc. This is needed for calculating accurate stoichiometry and for tracking required allocations of carbon to replace e.g. leaf loss.

davidorme:
# Description

This PR is to develop a representation of the **realised** biomass of tissues for cohort stems into the PlantsModel. This is needed to track herbivory and maintain stoichiometry.

The new `biomasses` module is an attempt at drawing the carbon and stochiometric elements into a single framework - currently incomplete and not passing pre-commit but pushed up here as a sketch.

@sallymatson Overview:

* Added the new biomasses module with the Biomass object as a community level structure holding carbon and nutrient biomasses. 
* The biomass object contains a set of tissues that mirror the current stoichiometry tissues, but add carbon and then a dict of elements, rather than just a single element.
* The element objects then contain the current element mass in the tissue along with the ideal ratio.

* All three of those have an `.append()` method to handle recruitment. When you append a biomass onto an existing biomass, the tissues and elements all then get `.append`ed hierarchically. For this reason, I haven't implemented cohort methods to `add_cohorts`.
* I've substantially reworked the code for balancing elements into the surplus pool into a single function across all cohorts. I think it's more efficient.
* I've put in the code to generate biomasses in the model and then append recruitment.
* Testing is currently entirely at the module level rather than integration tests, but it basically plagiarises your test suite for stochiometry with the new objects.

* There are some changes to Stochiometry - this was mostly me thinking through structure and trying to align what I was trying to do to make sure I was covering the ground.

I think this makes sense but I think now is the time for a hard review before we go about ripping out the Stoichiometry and switching the integration tests over.

Fixes #1343

## Type of change

## Key checklist

## Further checks


davidorme:
@sallymatson Could you have a quick look at this initial step?

What I'm trying to do here is lay the groundwork for the biomass representation by having a definition at the PlantsModel level of the tissues in the model. That makes it easier to synchronise the Biomass tissues with the StemStochiometry tissues.

So:
* Define a list of tissues in the PlantsModel and make the tissue names marginally easier to index for use in both structures.
* Rename `Tissue` to `TissueABC` mostly to make the typing of those lists a little more obvious.
* Minor tweaks to reduce code repetition.

Do those seem like sensible tweaks? The fact that the tests all still pass without alteration gives me hope 😄

davidorme:
@sallymatson I've pushed a non-pre-commit passing and incomplete sketch of what I think biomasses might look like. I'm trying to keep exactly the same logic as Stochiometry but encapsulates it within a tissues implementation that includes carbon mass.

sallymatson:
Had a cursory look and I can look in more detail tomorrow. Everything still seems sensible to me. I followed the tissue vs element structure breakdown and I can't fully think through how easy it will be to access all the different values but I trust your thought process on the top-level implementation. The two questions that come to mind immediately: 

1. How does this handle fruit flesh vs seeds? With stoichiometry we'd discussed potentially creating them as separate tissues based on how we need to deal with the different stoichiometric ratios and pools for Tarran. Do you have thoughts on how this would work in biomass? 
2. I believe this is all a precursor to real herbivory, in which the GPP is calculated based on a (potentially) reduced foliage area. Has the implementation shed any light on how we'll be able to negotiate that along with re-growing those eaten leaves as a penalty to allocation?

davidorme:
>The two questions that come to mind immediately:
> 
> 1. How does this handle fruit flesh vs seeds? With stoichiometry we'd discussed potentially creating them as separate tissues based on how we need to deal with the different stoichiometric ratios and pools for Tarran. Do you have thoughts on how this would work in biomass?

I hadn't though about this yet - one of my two main tasks for the next couple of weeks is to square this away with the fruit stuff. I _think_ we can do this by creating two biomass tissues and have the generator function apply the partition fraction. So one will get `reproductive_mass * p_fruit_flesh` and one will get `reproductive_mass * (1- p_fruit_flesh)` (I'm completely guessing at the variable names there). That we get two tissues from one mass? But I need to look at #1263 to try and align what you have there to this structure.

> 2. I believe this is all a precursor to real herbivory, in which the GPP is calculated based on a (potentially) reduced foliage area. Has the implementation shed any light on how we'll be able to negotiate that along with re-growing those eaten leaves as a penalty to allocation?

And that's the other task. I need to think through the sequence of when we have which information available.

davidorme:
@sallymatson I've fixed up the remaining issues and transferred this onto a fresh feature branch off develop. Are you happy to approve merging this down, to then add PRs for other features.

PR #1385, opened 2026-02-26
Adds the base disturbance model

### LINKED ISSUE #1367: Implement the BaseDisturbance model
Likewise, it should follow largely this structure:

```python
class BaseDisturbance(ABC)

    disturbance_name: str
    """A name for the disturbance (eg. 'logging') that is used to key the disturbance in the DisturbanceRegistry and config."""
    disturbed_models: list[str]
    """A list of model names that this disturbance will affect."""
    data_variables_disturbed: list[str]
    """A list of data variables that will be updated."""

    def __init__(self, data: Data, models: dict[str: BaseModel], timing: DisturbanceTiming, **kwargs):
        """Creates an instance."""
        self.data = data
        self.models = models
        self.timing = timing

        # check that the self.models contains cls.disturbed_models. This is checked
        # when the class is registered (does the BaseModel exist _at all_) but also needs
        # to be checked at runtime (is the BaseModel in _this simulation__).

    def __init_subclass__(self, disturbance_name: str, disturbed_models: list[str]):
        """Checks the disturbed models and variables are all known and adds the disturbance to the registry."""

    @classmethod
    def from_config(cls, config: Config) -> BaseDisturbance:
        """Factory method to create instance from a Config instance."""

    @abstractmethod
    def disturb(self, time_index):
        if not self.timing.check_run(time_index):
            return
        # Otherwise, do stuff to self.data and self.models
```

dalonsoa:
# Description

Adds the base disturbance model that all the other disturbances will need to inherit from.

Note: No tests have been added as this will be come simpler when there are more bits and pieces put together (in particular, the model discovery and registry).

Fixes #1367 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/docs/source/virtual_ecosystem/implementation/var_generator.py
+++ b/docs/source/virtual_ecosystem/implementation/var_generator.py
@@ -2,7 +2,7 @@
 
 from pydantic import fields
 
-from virtual_ecosystem.core.base_model import _discover_models
+from virtual_ecosystem.core.base_model import discover_models
 from virtual_ecosystem.core.variables import VariableMetadata, load_known_variables
 
 # TODO - merge these into a single generate_model_variable_markdown and probably move it
@@ -16,7 +16,7 @@ def generate_variable_listing(model_name: str, var_attributes: list[str]) -> str
     variables = load_known_variables()
 
     # Find the model reference
-    models = {m.__name__: m for m in _discover_models()}
+    models = {m.__name__: m for m in discover_models()}
     if model_name not in models:
         raise ValueError("Unknown model name")
     model = models[model_name]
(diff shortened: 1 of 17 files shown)


davidorme (approved):
I'm probably overthinking this but this is set up for the top level code to do:

```python
for dist in disturbances:
    dist.disturb(time_index)
```

and letting the `disturb()` method silently skip when it doesn't run. I feel the calling code is easier to follow if those two concerns are separated:

```python
for dist in disturbances:
    if dist.runs_at_step(time_index):
        dist.disturb(time_index)
```

That's more long winded but the flow is more explicit. Of course `disturb` probably should still check it has been called correctly - but maybe raise if it has been called incorrectly.

None of this changes anything deep in the model and could be switched easily enough, so I'm approving.

dalonsoa:
I coded it that way because you coded it that way in the discussion :P . I think that if `disturb` is going to check if things have been called correctly anyway, I rather keep the time check within `disturb` and save one check. 

In summary, my view is that is not for the calling code to decide if the distubance is to run, but for the disturbance itself if it needs to run.

But I do not mind either way, obviously :)

davidorme:
> In summary, my view is that is not for the calling code to decide if the distubance is to run, but for the disturbance itself if it needs to run.

Sure - that makes sense. As I said, probably overthinking!

TaranRallings (approved):
LGTM! Straightforward to see how it would interact with the models.

PR #1391, opened 2026-03-02
1368 disturbance registry

### LINKED ISSUE #1368: Implement disturbance model registry and discovery engine
Largely copied from the normal models. In principle, it is expected they will live within the same kind of folder structure.

dalonsoa:
# Description

Adds the disturbance model discovery process. I have refactor the existing discovery engine such that it works equally for normal models and disturbance models. A disturbance testing model has been added for testing purposes within a `disturbance` folder. As the same model discovery process is used, the very same rules in terms of module name and class name conventions apply to the disturbances. New tests have been limited to just the new functionality, as the logic of the registry and discovery has already been thoroughly tested for normal models.

The `priority` attribute will be added in the final PR, when this is hooked into the main workflow and an order needs to be defined for the disturbance models. 

Fixes #1368

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/core/base_model.py
+++ b/virtual_ecosystem/core/base_model.py
@@ -104,7 +104,8 @@ from __future__ import annotations
 import pkgutil
 from abc import ABC, abstractmethod
 from importlib import import_module
-from typing import Any
+from types import ModuleType
+from typing import Any, TypeVar
 
 import pint
 
@@ -782,9 +783,11 @@ def to_camel_case(snake_str: str) -> str:
     return "".join(x.capitalize() for x in snake_str.lower().split("_"))
 
 
-def _discover_models() -> list[type[BaseModel]]:
+T = TypeVar("T")
+
+
+def _discover_models(models: ModuleType, of_type: type[T]) -> list[type[T]]:
     """Discover all the models in Virtual Ecosystem."""
-    import virtual_ecosystem.models as models
 
     models_found = []
     for mod in pkgutil.iter_modules(models.__path__):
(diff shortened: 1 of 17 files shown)


jacobcook1995 (commented):
This all looks sensible to me, I won't approve it though as I'm aware that this is an area where @davidorme has **much** stronger opinions than me and Taran

davidorme:
> This all looks sensible to me, I won't approve it though as I'm aware that this is an area where @davidorme has **much** stronger opinions than me and Taran

Ha. There's a lot written into that bolded much 😄

virtual_ecosystem/core/base_model.py
```diff
@@ -782,9 +783,11 @@ def to_camel_case(snake_str: str) -> str:
     return "".join(x.capitalize() for x in snake_str.lower().split("_"))
 
 
-def _discover_models() -> list[type[BaseModel]]:
+T = TypeVar("T")
+
+
+def _discover_models(models: ModuleType, of_type: type[T]) -> list[type[T]]:
```
davidorme, line 789: I _think_ what is going on here is that we'd "ideally" have this as:

```python
def _discover_models(models: ModuleType, of_type: type[BaseModel] | type[BaseDisturbance]) ->  list[type[BaseModel] | type[BaseDisturbance)]:
```

But because we can't import those two classes outside of the functions because of circularity (which bugs me but 🤷 ) we instead use this to tell the function that it is generically getting a type of something (not an instance etc.).

Is that right - if so, could you add a comment to that effect? It's probably pretty entry level but right now generic typing is out of my comfort zone, so it would be useful to have an explanation to hand 😄 . 

Of course, if my explanation isn't right, then it needs a comment even more!

davidorme (approved):
Neat.  As you say, makes good re-use of the existing discovery mechanisms. One question about typing!

TaranRallings (commented):
Looks sensible to me as well.

virtual_ecosystem/core/base_model.py
dalonsoa: Actually, no. We do want type[T] - i.e. to use generics - because that way I am locking the types of the inputs and outputs together. It's like saying that whatever the input, the output is a list if those same things. In your example, from the typing perspective there is no relationship between input and output. I could have `type[BaseModel]` as input and have a list of `type[BaseDisturbance]` as outputs.

Generics are really powerful, but can also be quite confusing.

davidorme: Got it. Thanks.

dalonsoa: Docstring expanded with some details.

PR #1410, opened 2026-03-06
Add disturbances to input layer

### LINKED ISSUE #1369: Implement disturbance models input layer
Largely following the this structure (where `logging` is the name of the disturbance model):

```toml
[disturbance.logging]
timing = "" # details to come
priority = 1
config = "" # details to come
```

`timing` will be such that the input requirements of the `DisturbanceTiming` class are provided, possibly via an appropriate pydantic model, like other bits of Virtual Ecosystem.

`config` will need to be discussed with the research team and required specific examples to work with. In any case, a configuration similar to the one of the other models via a pydantic model will be required.

dalonsoa:
# Description

Adds the relevant bits in the input layer to handle disturbances. Tests have been updated as appropriate.

A couple of bugs have been fixed along the way as, it would seem, `tomli_w` cannot handle `None`.

Fixes #1369

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -83,6 +83,20 @@ def reset_module_registry():
     MODULE_REGISTRY.clear()
 
 
+@pytest.fixture(autouse=True)
+def reset_disturbance_registry():
+    """Reset the disturbance registry.
+
+    The register_disturbance function updates the DISTURBANCE_REGISTRY, which persists
+    between tests. This autouse fixture is used to ensure that the registry is always
+    cleared before tests start, so that the correct registration of disturbances within
+    tests is enforced.
+    """
+    from virtual_ecosystem.core.registry import DISTURBANCE_REGISTRY
+
+    DISTURBANCE_REGISTRY.clear()
+
+
 # Shared fixtures
 
 FIXTURE_ROOT_DATA_DIR = Path(__file__).parent / "data"
--- a/tests/core/test_configuration.py
+++ b/tests/core/test_configuration.py
@@ -22,6 +22,11 @@ def test_pydantic_models(tmp_path):
         ("virtual_ecosystem.models.litter", "LitterConfiguration"),
         ("virtual_ecosystem.models.plants", "PlantsConfiguration"),
         ("virtual_ecosystem.models.soil", "SoilConfiguration"),
+        ("virtual_ecosystem.models.testing", "TestingConfiguration"),
+        (
+            "virtual_ecosystem.disturbances.disturbance_testing",
+            "DisturbanceTestingConfiguration",
+        ),
     )
 
     submodel_details = {}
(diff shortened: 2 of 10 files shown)


jacobcook1995 (commented):
This looks sensible to me, but I'll leave it to @davidorme to sign off on as he was the one who wrote the config builder code

davidorme:
> A couple of bugs have been fixed along the way as, it would seem, `tomli_w` cannot handle `None`.

TOML in general cannot handle None. It's a bit of a pain in the arse. They have reasons but inconvenient ones:
https://github.com/toml-lang/toml/issues/921

davidorme (changes requested):
I think there's a mismatch here in the TOML representation of the configuration and the internal structure of the configuration data. The TOML has:

```TOML
[core]
[plants]
[disturbances.disturbance_testing]
```

But in the code, the disturbances are stripped out and elevated to being root level members of the configuration object, which is why we get this:

```
def build_configuration_model(
    requested_modules: list[str], requested_disturbances: list[str] | None = None
) -> type[CompiledConfiguration]:
```

IIUC, that gives the resulting configuration object with a set of root configurations, some of which are models and some of which are disturbances. I think it is better to keep them explicitly separated within the configuration. My first guess at the way to do this is to create a `Disturbances` pydantic configuration that is simply a list of `DisturbanceConfigurationRoot` objects. I suspect that does mean that `build_configuration_model` then needs _two_ dynamic models built, one for a list of requested disturbances and one for the top level (including the `disturbances` root)

dalonsoa:
> I think there's a mismatch here in the TOML representation of the configuration and the internal structure of the configuration data. The TOML has:
> 
> ```toml
> [core]
> [plants]
> [disturbances.disturbance_testing]
> ```
> 
> But in the code, the disturbances are stripped out and elevated to being root level members of the configuration object

Yes, I did it this way on purpose, as I don't think there was much value on keeping the same structure as the input file. The configuration is always used as `get_configuration(...)` so its internal structure is not that relevant. Trying to match the structure meant, as you suggest, to create two dynamic models on the fly. I did not thought the extra complexity was worth the effort, but if you want it that way, I'll change it.

davidorme:
I think it's cleaner to keep the same structure. I _think_ at the moment, we'd need to reverse this collapse when writing out the compiled complete configuration - the internal representation currently can't round trip back to TOML.

I agree it makes `build_configuration_model` more complex. My feeling is keeping the original signature of having a single input of `requested_modules` and isolating the complexity of building the full structure in that one function is cleaner than pulling out/reinserting the disturbances to do something 'special' with them in multiple locations.

dalonsoa:
OK, I'll update that as soon as possible.

dalonsoa:
What about now, @davidorme ?

davidorme (approved):
Thanks @dalonsoa

PR #1412, opened 2026-03-06
Microbial constants documentation

### LINKED ISSUE #987: Microbial constants documentation
Default values for microbial constants currently aren't documented anywhere. This needs to change at some point

jacobcook1995:
# Description

This PR adds a new section to the documentation that gives the specific details needed to setup the soil/litter models. As well as the full set of configuration settings, there's also a new page that dumps the entire `toml` file of microbial functional group and enzyme parameters. This file now contains comments, which describe the sources of each value. I do this because I couldn't think of a cleaner way to automatically document the value of these constants

@hrlai, I'm adding you on this PR both so that you can give your opinion about how this looks (+ if there's anything I should add), and so that you can see the new page where I list all the microbial and enzyme constants and their sources (sorry there is an awful lot of "I just completely made this up")

This is the [relevant documentation to look at](https://virtual-ecosystem.readthedocs.io/en/987-microbial-constants-documentation/using_the_ve/model_details/decay/overview.html)

Fixes #987


docs/source/using_the_ve/model_details/decay/overview.md
```diff
@@ -0,0 +1,36 @@
...
+---
+
+# The decay models
+
+The Virtual Ecosystem includes two models that capture the decay of organic matter. The
+[`litter` model](./litter_config.md) is a simple model to capture the initial stages of
+plant matter decay, and the [`soil` model](./soil_config.md) is a much more complex
+model that captures the subsequent stages that occur within the soil. These models are
+**not** alternative implementations, so for standard runs of the Virtual Ecosystem you
+will need to include **both** models.
```
hrlai: Suggest to add "...include both models in the config file" or similar.

hrlai (approved):
LGTM and I particularly like the short notes on the level of assumption behind the parameters. This will be very helpful in the prioritisation on the data science side later.

PR #1427, opened 2026-03-11
Switch litter pools to use `ArrayResources`

### LINKED ISSUE #1235: Convert litter model to use biomass triplets
The litter model is better off tracking masses of nitrogen and phosphorus, rather than the stochiometric ratios, because averaging stoichiometric ratios is incredibly easy to screw up. If this was the case pools could be expressed as biomass triplets, which would also simplify the description of the model quite a lot.

This would be a large but fairly straightforward refactor of the litter model

This change should also change litter consumption to use the array resources implementation.

N.B. I **need** to make sure I updated the relevant documentation, as this PR will involve changing central equations that describe the litter model

jacobcook1995:
# Description

This PR replaces the existing litter pools with biomass triplets, and uses `ArrayResources` to carry out litter consumption rather than the previous special handling.

There are two main things I need feedback on:

1) The litter pools are in areal density units (in theory this could change, but I don't really want to as the soil model **has** to use density units, so it would be a load of faff just to move the conversion problem elsewhere). So in order to get the `ArrayResources` to work, I had to add a new flag `density` which if used means that pools get automatically converted from areal density units to mass units (kg m^-2 to kg). Does this seem a reasonable change to make? Also, I didn't add a conversion step for the output, and I am now wondering if that would be cleaner.

2) I'm not 100% sure if what I did with foraging is correct (I didn't want to mess around with the internals of foraging, so kept it as similar as possible). Like before I generate a `litter_list` but now it's generated by the `get_array_resources` method. So, if a cohort has two both a detritus diet and a herbivore diet, two separate lists will be generated using that method and will (I assume) be identical. I couldn't figure out whether when these lists are fed into the `forage_cohort` method there any additional steps that ensure that resources aren't foraged twice (i.e. litter being consumed both in `delta_mass_herbivory` and in `delta_mass_detritivory` if the cohort has both a herbivore and detritivore diet). Anyway if you could let me know if my fears are justified here, and suggested how it can be fixed if so, that would be great.

@hrlai, this PR is back end focused so I'm not putting down as a reviewer. There will be a further PR that changes the relevant equations in the litter model and documents these changes which I'll put you down as a reviewer on


virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1558,9 +1514,14 @@ def forage_community(self, dt: timedelta64) -> None:
...
 
+            # TODO - I went with this as it was the only way I could see to fit litter
+            # consumption into the new system without massively rewriting animal
+            # foraging. But I feel like this might expose the litter pools to detrivory
+            # and herbivory separately (if the consumer has both diets). I need to check
+            # with Taran and David
             # Plant litter detritivory
             if diet & DietType.DETRITUS:
-                litter_list = cohort.get_litter_pools(self.litter_pools)
+                litter_list = cohort.get_array_resources(self.array_resource_pools)
```
davidorme: I think ideally we pass all of the array resources down to `forage_cohort` as one set and use the filters within that stage - I think that will simplify the resource selection process - but for the moment keeping different pathways into that function is probably easier.

davidorme (approved):
This looks good to me. I think the `density` argument to the resource array definition may become the default  - I need to look at how plant array resources are scaled. I'm not sure that the filtering of which cohorts access litter is happening at the best level, but that's something to come back to.

virtual_ecosystem/models/animal/animal_model.py
jacobcook1995: Okay, I guess the issue I'm concerned about is (maybe baselessly) is that `litter_list` and `plant_list` will be identical (as far as I can tell), so this step isn't really filtering them?

jacobcook1995: `forage_cohort` filters by consumable resources, but returns a complete list of all consumable resources. So if say a beetle can eat both living leaves and ground based litter, both resources will be on both resource lists

jacobcook1995: I guess whatever I add would be a temporary fix, but the two options would be:

(a) adding an argument to `cohort.get_array_resources` to only select array resources related to diets in the list provided

(b) adding a step at the start of `cohort.forage_cohort` to split the list of array resources into a plant list and a litter list

which one sounds more reasonable to you?

davidorme: My intention was that `litter_list` and `plant_list` are replaced by a single `array_resources` arg and that the filtering then happens within the cohort foraging function, since each resource carries its dietary availability with it.

jacobcook1995: okay, so it makes most sense for me to do a naive implementation of that now to fix the immediate problem, with a note that this should be extended/improved in future?

TaranRallings: I am in favour of doing it naively right now. I will need to take a proper look at what we've got and figure out where/how to rework the filtering now that everything is going through array resources.

jacobcook1995:
Okay @davidorme and @TaranRallings, I've added the naive solution, but I had to change the type hinting quite a lot to get it past `mypy`. Can you check whether my changes look correct?

TaranRallings:
> Okay @davidorme and @TaranRallings, I've added the naive solution, but I had to change the type hinting quite a lot to get it past `mypy`. Can you check whether my changes look correct?

I think it's probably ok for now. I'm going to have to do a bunch of fiddling with the filtering afterwards and I can, at that time, have a proper think about the role of the resource protocol. It may be that it doesn't make the kind of sense now that it made when we had a dozen different resource inputs.

Issue #1234, opened 2025-12-17
Soil model should use biomass triplets


jacobcook1995:
A large number of soil model pools are elemental triplets (e.g. `lmwc`, `don`, `dop`). It would make the links between pools easier to follow if the carbon, nitrogen and phosphorus components of each of these pools were expressed as triplets, rather than three (apparently independent) pools.

This should be fairly straightforward as the soil model already uses densities (rather than elemental ratios). The biggest hurdle will be getting the integrator to play nicely with data input that has a more complex structure

PR #1432, opened 2026-03-13
Convert litter model to use biomass triplets

### LINKED ISSUE #1235: Convert litter model to use biomass triplets

jacobcook1995:
# Description

This PR follows on from #1427 and removes the "c_nutrient_ratio_x" variables entirely from the litter model.

I've tried to keep the flow as similar as possible to before to avoid the PR blowing up in size. There's probably ways to simplify the flow now that I'm no longer working with nutrient ratios, but I feel that's best addressed the next time I have reason to alter the litter model.

@hrlai there's no documentation changes to review as I still use the same equations described in [the litter docs](https://virtual-ecosystem.readthedocs.io/en/latest/virtual_ecosystem/theory/soil/litter_theory.html). For the few equations that use carbon:nutrient ratios, instead of trying to derive mass based forms, I just convert the masses to ratios and use the old forms of the equation.

N.B. When we next do a release we need to include notes that the "c_nutrient_ratio_pool_name" litter variables are now deprecated in favour of "litter_pool_name_cnp".

Fixes #1235

## Type of change

## Key checklist

## Further checks


hrlai (approved):
Looks straigthforward. So just to confirm, what this means for the initial data is to replace the old ratios with `litter_pool_*_cnp`?

jacobcook1995:
> Looks straigthforward. So just to confirm, what this means for the initial data is to replace the old ratios with `litter_pool_*_cnp`?

Yep, it should be a single variable with element as a dimension (coordinates "C", "N" and "P"). And the values in this will be masses (per area) rather than ratios.

jacobcook1995:
@hrlai also, is it easiest for you if I change the soil model to the same system (see #1234) before the next Virtual Ecosystem release, so that you only have one release where a significant number of variable have changed?

hrlai:
> @hrlai also, is it easiest for you if I change the soil model to the same system (see #1234) before the next Virtual Ecosystem release, so that you only have one release where a significant number of variable have changed?

Yes, I think so. This will probably take some time for me to keep up on the data science part, which means delays to the other members.

jacobcook1995:
> Yes, I think so. This will probably take some time for me to keep up on the data science part, which means delays to the other members.

Okay! once these PRs are merged I will go ahead and do the same for the soil variables

davidorme (approved):
Hard to review the scientific details but this all looks sensible and I think the outcome is cleaner and easier to read, maintain and document.

I think it's right to switch it over and then look to review flow improvements later.

PR #1433, opened 2026-03-13
Revert "Animal - Update territory size scaling"

TaranRallings:
Reverts ImperialCollegeLondon/virtual_ecosystem#1424


davidorme:
Thanks @TaranRallings. If we then put this code into a new subbranch then we can work on fixing the performance regression before merging down to develop.

davidorme (approved):

PR #1434, opened 2026-03-13
Animal - Predation Streamlining and Debug

TaranRallings:
# Description

This is attempt number two on the territory intersection predation update. The first attempt was doing some exciting stuff with big-O issues that blew up testing times. I think this has those times down.

This should:
- update the territory overlap system in predation to properly sort through potential prey
- bin prey cohorts of similar body sizes correctly for determining handling times
- move calls that are only needed once per encounter such that they are only called once per encounter

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


Baizurah (approved):
This looks good to me.

PR #1437, opened 2026-03-16
Improve description of the clay fraction variable

### LINKED ISSUE #1431: Check soil and litter proportions "units"
@hrlai wasn't able to find whether the clay fraction was expressed percentage or a fraction based on the variable table.

I need to go through the variable table and check that the things I have given units for as "unitless" or "proportion" clearly explain that they are fractions/percentages etc

jacobcook1995:
# Description

I've improved the description of the clay fraction variable to mention that it runs between 0 and 1, i.e. that it isn't a percentage. I didn't change the unit as I guess it is really unit-less and I couldn't think of a nice way to express "this is a proportion" in simple units notation

Let me know if this seems clear enough

Fixes #1431


hrlai (approved):
Looks good to me. I guess by size fractionalisation it is count / count, but I think it's okay in the current shape too!

PR #1454, opened 2026-03-18
Changed check for known variables to return all failed models

### LINKED ISSUE #1417: "definition contains unknown variables" error messages refers to the wrong model
When unknown variables are detected `_check_model_variables_are_known` raises an error message that links them to the last model checked, rather than the model that tries to use the unknown variables.

This is a very small issue if you have the full log, but is confusing when you are running things within `pytest`.

The code is currently

```python
for mod in models:
        for var_attr in variable_attributes:
            unknown_variables = set(getattr(mod, var_attr)).difference(known_variables)

            if unknown_variables:
                LOGGER.error(
                    f"Unknown variables in {mod.model_name}.{var_attr}: "
                    f"{', '.join(unknown_variables)}"
                )
                fail = True

if fail:
    msg = f"Model {mod.model_name} definition contains unknown variables, check log"
    LOGGER.critical(msg)
    raise ValueError(msg)
```

When it actually should be something like 

```python
for mod in models:
        for var_attr in variable_attributes:
            unknown_variables = set(getattr(mod, var_attr)).difference(known_variables)

            if unknown_variables:
                LOGGER.error(
                    f"Unknown variables in {mod.model_name}.{var_attr}: "
                    f"{', '.join(unknown_variables)}"
                )
                fail = True

    if fail:
        msg = f"Model {mod.model_name} definition contains unknown variables, check log"
        LOGGER.critical(msg)
        raise ValueError(msg)
```

jacobcook1995:
# Description

Previously it was returning the last model to be checked, even if that model was not the one to cause the failure.

I considered adding a test to show that when multiple models are evaluated only the ones that fail are returned. But this would require defining multiple fixture models, so felt like overkill for this edge case

Fixes #1417

## Type of change

## Key checklist

## Further checks


davidorme (approved):
👍

PR #1455, opened 2026-03-18
Draft/fg population density

Baizurah:
## Draft: FG Population Density Script (Tool 1)

Sharing this as a draft for review.

**Tool 1 – Population Density**

- **Input:** `animal_cohort_data.csv`  
- **Analysis:** Calculates total population density (individuals/m²) for each functional group across timesteps  
- **Output:** Line graph showing population density over time, with one line per functional group  

### What the function does
This script calculates population density (individuals/m²) for each functional group over time using `animal_cohort_data.csv`. It groups individuals by timestep and functional group, divides by total simulation area, and produces a line plot of density trends. Would really appreciate any feedback so I can refine this further 🙏 Also sharing the figures produced using the herbivore only (Maliau_FG level 1) and the FG excel file from example data. 

Figure_1_pop density testing function
Figure_2

### DIFF
--- a/fg_population_density.py
+++ b/fg_population_density.py
@@ -1,0 +1,33 @@
+""" 
+Tool 1- Calculating population density for each functional group over time. 
+Purpose: This tool is created to calculates the population density (indv/m2) for each functional group using the animal cohort data.  
+Method: Sum individuals for each FG per timestep and then divide the total individuals by total simulation area. 
+Input file: animal_cohort_data.csv
+Output: Line graph showing population density over time and each functional group is plotted as a separate line. 
+"""
+import pandas as pd
+import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


davidorme:
@Baizurah This is a data science tool - can you resubmit this as a PR to the `ve_data_science` repo. It might be worth creating a `tools/python` directory for functions like this, which you could then call from within the data science scripts.

We may very well want to pull these tools together at some point - and that might be part of the main VE module - but for the moment it is easier to keep them with the data science where they are being used.

Baizurah:
Thanks @davidorme, I’m still finding my way around the repo structure and PR process, so this is really helpful.

I will resubmit a new one.

davidorme:
No problem. It is pretty complicated. In general I think the data science team will typically raise issues in this repo when bugs occur but if you are submitting PRs that's generally going into the data science repo. There are exceptions, like your fix for the config...

Issue #1438, opened 2026-03-16
Update documentation microclimate for subdaily loop


vgro:
The `abiotic_model` has changed from a monthly equilibrium assumption to a mean day using a subdaily loop (#1288). The documentation needs to be updated accordingly.

This change requires `latitude` as an input, which is currently part of the `abiotic_model`configuration but should be determined programmatically in the future #1465 . Add this as a waring to the documentation.

PR #1460, opened 2026-03-19
1441 abiotic model need more locationtime step specific inputs

### LINKED ISSUE #1441: Abiotic model need more location/time step specific inputs
The following are placeholders and should be arguments to `abiotic_model`:
```
days = 30
latitude = 0.0
time_dim = 24
```

vgro:
This PR adds latitude to the `abiotic` model configuration (**is this sensible for now?**) and generates number of days from model time step.

Fixes #1141

## Type of change

## Key checklist

## Further checks

- [ ] Relevant documentation reviewed and updated -> TODO when all abiotic fixes are implemented


jacobcook1995 (approved):
My feeling is that this is good enough for now, as either checking that the latitude is consistent with the extents used in the rest of the model or populated the latitude based on this would be too complex this side of MVP.

Would suggest making an issue for doing this at some point though

Also, it would be good to put a warning in the documentation about this. I think it probably belongs in the `abiotic` model details section of the new docs. So maybe it's best to track that as well with an issue (until the documentation improvements are merged)

davidorme (approved):
> My feeling is that this is good enough for now, as either checking that the latitude is consistent with the extents used in the rest of the model or populated the latitude based on this would be too complex this side of MVP.

I think the way to do this is to add an [EPSG code](http://epsg.io) to the grid configuration that defines the projected coordinate system being used. That allows us to programatically get latitude and longitude data directly from the grid definition. That wouldn't actually be hard to do, but is post MVP. I think this is fine for now but we should definitely add that issue.

vgro:
> > My feeling is that this is good enough for now, as either checking that the latitude is consistent with the extents used in the rest of the model or populated the latitude based on this would be too complex this side of MVP.
> 
> I think the way to do this is to add an [EPSG code](http://epsg.io) to the grid configuration that defines the projected coordinate system being used. That allows us to programatically get latitude and longitude data directly from the grid definition. That wouldn't actually be hard to do, but is post MVP. I think this is fine for now but we should definitely add that issue.

I agree that this is the way we want to do it, I have created an issue for that #1465 and updated the issue on documentation #1438.

PR #1461, opened 2026-03-20
Replaced relevant soil pools with biomass triplets

### LINKED ISSUE #1234: Soil model should use biomass triplets

jacobcook1995:
# Description

This PR replaces 12 soil pools, with 4 elemental triplet pools. This is a change that doesn't really alter any of the science and the unit tests all still pass under the new system.

@TaranRallings, I had to change the animal model code slightly. Let me know if it looks okay to you

@hrlai, the change in the variables are listed below. I slightly changed the documentation to remove references to `DON` and `DOP` as these were confusing as the model really cares about the nitrogen and phosphorus fractions of `LMWC`, which are similar to but crucially aren't the same as `DON` and `DOP`. Let me know if the [updated documentation](https://virtual-ecosystem.readthedocs.io/en/1234-soil-model-should-use-biomass-triplets/virtual_ecosystem/theory/soil/summary.html#soil-and-litter) makes sense

Old pools:
```
"soil_c_pool_maom", "soil_c_pool_lmwc", "soil_c_pool_pom", "soil_c_pool_necromass"
"soil_n_pool_maom", "soil_n_pool_don", "soil_n_pool_particulate", "soil_n_pool_necromass"
"soil_p_pool_maom", "soil_p_pool_dop", "soil_p_pool_particulate", "soil_p_pool_necromass"
```

New pools:
`"soil_cnp_pool_maom", "soil_cnp_pool_lmwc", "soil_cnp_pool_pom", "soil_cnp_pool_necromass"`

Fixes #1234

## Type of change

## Key checklist

## Further checks


TaranRallings (approved):
If the existing tests pass, these changes to the animal code should be fine.  LGTM!

hrlai (approved):
Looks good overall! Only minor thing is that DON and DOP are still vestigial in the example generation code, but I don't think that's a problem; since the user facing item is just the netCDF file.

I think this way of specifying soil nutrients really simplify things to MAOM, POM and LMW organic matter and will avoid confusion.

jacobcook1995:
> Looks good overall! Only minor thing is that DON and DOP are still vestigial in the example generation code, but I don't think that's a problem; since the user facing item is just the netCDF file.

That's a good point! I really shouldn't assume that people won't end up looking at that code in the future (nothing lasts as long as a temporary solution and all that). I'll go ahead and change the names there as well

PR #1479, opened 2026-03-26
Fixed return units of calculate_carbon_mineralised

### LINKED ISSUE #1473: `calculate_carbon_mineralised` has the wrong return units
Return is a rate (per area) when it should be a total mass (per area)

jacobcook1995:
# Description

See title. This is a tiny PR to fix one unit. I probably could have bundled this into something else, but I thought it better to fix this while I still remembered

Fixes #1473


TaranRallings (approved):
LGTM!

PR #1480, opened 2026-03-26
Remove `sphinxcontrib-mermaid` as a dependancy

### LINKED ISSUE #1468: Remove `sphinxcontrib-mermaid` from the dependancy
`sphinxcontrib-mermaid` is listed as a documentation dependancy but isn't actually used anywhere, so we should remove it.

This is very much non-urgent and probably makes sense to bundle into another PR (e.g. #1400)

jacobcook1995:
# Description

Title. This dependancy was straightforward to remove, so not much to check here beyond all the tests passing

Fixes #1468


davidorme (approved):
👍

PR #1481, opened 2026-03-26
Removing `sphinx-hoverxref` as a dependancy

### LINKED ISSUE #1400: `sphinx-hoverxref` is deprecated
`sphinx-hoverxref` is has now been deprecated, as the [functionality now exists in ReadTheDocs](https://docs.readthedocs.com/platform/stable/link-previews.html). This isn't an immediate problem as everything works for now. But when I get the time I should remove this dependancy (which also currently blocks use from updating the `sphinx` version we use). This should be a fairly smooth process, the main thing I need to ensure is that the glossary terms highlighting that I added still works in the new system

jacobcook1995:
# Description

This PR removes `sphinx-hoverxref` as a dependency (which turned out to be much simpler than I was imagining 😅).

It also fixes some incorrect notation in the docs (which made me think that removing `hoverxref` was breaking things).

This [example page of the docs](https://virtual-ecosystem.readthedocs.io/en/1400-sphinx-hoverxref-is-deprecated/virtual_ecosystem/theory/soil/macronutrients.html) shows that the hover links still work as expected on `readthedocs`

Fixes #1400


davidorme (approved):
👍

PR #1494, opened 2026-03-31
Fixes to Biomass draft

### LINKED ISSUE #1493: Error in biomass implementation
There are some errors/issuesin the initial draft of plant biomasses:

* Turnover methods should not decrease elemental mass in tissues - it is _turnover_.
* The `account_for_growth_elements` method needs to also handle carbon - given this, it makes sense for these methods to become `apply_growth` and to increase tissue biomasses as part of their function. They still do need to return the new additional elemental biomasses so that the whole stem balancing can occur.

The upstream `Biomasses` methods then need updating to match

davidorme:
# Description

This PR:

* Fixes a stupid mistake: the new BiomassTissue classes should not extract turnover - the biomass carbon is untouched by turnover, that's the _whole point_. So the  `tissue_turnover` and `extract_turnover` methods are removed and collapse back into a single `get_turnover` method that returns the nutrient turnover losses for inclusion in the whole stem nutrient balancing.

    ETA: To clarify - I had setup the `BiomassTissue.extract_turnover` abstract method description and then the implementations in the  subclasses to _actually remove carbon and nutrient masses_ from the Tissue. This isn't what happens in turnover - a quantity of carbon is allocated to replace old tissue but the tissue mass is maintained. We need to know those masses to pass turnover into the correct pools but the tissue biomasses are unchanged. The method does also need to record the nutrient loss so that the stoichiometric balance of the individual can be resolved after all allocation has occurred.

* Updates the functionality of `elements_needed_for_growth`.  Back in Stoichiometry this just reported the elemental masses for nutrient balancing - however with Biomasses it makes more sense for this function to also be responsible for adding growth allocation to the carbon biomasses and similarly increasing nutrient allocations to tissues. This function also returns the new nutrient allocation  for inclusion in whole stem element balancing. The method is renamed `apply_growth`.

These new BiomassTissue methods are now called by the relevant functions in the Biomasses class and the tests are updated to cover the new functionality and naming.

Fixes #1493

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/plants/test_biomasses.py
+++ b/tests/models/plants/test_biomasses.py
@@ -314,8 +314,8 @@ def test_BiomassTissue_from_pft_default_ratios(
         )
 
 
-# TODO - think about how to collapse these tests, like the one above. It will need to
-#        capture the specific differences between tissues at turnover so might need a
+# TODO - think about how to collapse these tests, like the one above. It might need to
+#        capture specific differences between tissues at turnover so might need a
 #        switch/case at the end, but would remove a lot of overlap.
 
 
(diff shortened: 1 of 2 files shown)


davidorme:
@sallymatson I think there are some things about the internal representations that can be made more efficient, but this feels to me like it's shaking down into something fairly solid. I also think some of the code repetition can be reduced, but I'm going to worry about that later.

sallymatson (approved):
All looks sensible to me. 

A bit hard to get my head around everything to give a 100% full review, but I trust the details etc and I agree that it's moving in a great direction.

sallymatson:
Can you just clarify what you mean by this? "Fixes a stupid mistake: the new BiomassTissue classes should not extract turnover - the biomass carbon is untouched by turnover, that's the whole point."

davidorme:
Yup - I've updated the PR description.

PR #1495, opened 2026-04-01
Change nutrient split calculation to not use ratios

### LINKED ISSUE #1491: Change `calculate_nutrient_split_between_litter_pools` to not use ratios
For legacy reasons, `calculate_nutrient_split_between_litter_pools` still calculates C:N:P ratios to calculate litter split between pools. However, on the data science repo I've [derived expressions that don't use these ratios](https://github.com/ImperialCollegeLondon/ve_data_science/issues/255).

I think it's much cleaner to use these direct expressions, so I switch to using them. When I do this I should make sure the docstring and theory documentation are updated to explain what is going on here, as this is probably the hardest part of the litter model to follow

jacobcook1995:
# Description

This PR changes the function that calculates the division of nutrient input between litter pools to a simpler formula based on concentrations, rather than one which requires calculating C:nutrient ratios. After making this change all unit tests still passed.

I also renamed the relevant constants as I thought it was easier to describe them as a ratio of nutrient concentrations, rather than a ratio of ratios which is hard to parse

@hrlai, the [relevant documentation has been updated](https://virtual-ecosystem.readthedocs.io/en/1491-change-calculate_nutrient_split_between_litter_pools-to-not-use-ratios/virtual_ecosystem/theory/soil/litter_theory.html#split-of-nutrient-inputs-between-pools), let me know what you think

Fixes #1491

## Type of change

## Key checklist

## Further checks


hrlai (approved):
Look good. I think it currently explains what the code does well, but in the future I may need to elaborate on the $C_m / C_s$ part. We say that "pool elemental proportions cannot be determined based on empirical data" but then we also imply that $C_m$ and $C_s$ can be measured.

jacobcook1995:
> Look good. I think it currently explains what the code does well, but in the future I may need to elaborate on the C m / C s part. We say that "pool elemental proportions cannot be determined based on empirical data" but then we also imply that C m and C s can be measured.

Yes that's a good point, this needs to be written more clearly. I'm going to take another shot at writing this section

TaranRallings (approved):
LGTM!

jacobcook1995:
@hrlai, I've just updated that section slightly to make it a bit clearer, let me know what you think

More generally once this PR is merged I think we are ready on the VE side to do a new release, but I guess that needs to wait until the updated soil/litter data is on Globus. Anyway, let me know what makes sense to you as a time line on that

hrlai:
> updated that section slightly to make it a bit clearer

Nice, this is a good and quick addition that matches our data analysis too. Thanks. Let's merge this.

I think it is fine to do a release before I update Globus, since I am sure that the data will keep up. In fact, once you release I will check the data again against the release version. Anything funky is a data-side problem anyway so you don't need to wait :)

jacobcook1995:
> I think it is fine to do a release before I update Globus, since I am sure that the data will keep up. In fact, once you release I will check the data again against the release version. Anything funky is a data-side problem anyway so you don't need to wait :)

Okay great, I'll aim to do the new release tomorrow then

Issue #1164, opened 2025-11-27
Albedo is missing from radiation balance


vgro:
At some point, the albedo dropped out of the radiation balance which means that too much of the shortwave radiation enters the model. @davidorme we should discuss whether this should happen in the plant model, before light is absorbed, or in the abiotic model.

PR #1505, opened 2026-04-13
936 split radiation in par and other wavelengths

### LINKED ISSUE #936: Split radiation in PAR and other wavelengths
At the moment, `plants` absorb all shortwave radiation and the `abiotic` model uses this information for the energy balance calculations. The radiation should be split in photosynthetically active radiation for the plants npp and other wavelengths remaining for the energy fluxes.

A possible solution would be to take out a fraction of the radiation for the plants as `absorped_par` and have the residual in `shortwave_absorption`, or return `total_shortwave_absorption` and the `abiotic` model can do the subtraction internally.

This follows #721

vgro:
This PR adds the fraction of light that is not absorbed by the plants to the energy balance. @davidorme could you please check that my assumptions are correct to match the plant model.

I also moved `def compute_weights_from_absorbed_radiation` to `abiotic_tools` to avoid circular imports.

NOTE: This includes the soil layer and the understorey, maybe the naming of variables is not ideal?

Fixes #936

## Type of change

## Key checklist

## Further checks


vgro:
From the `abiotic_model` perspective, this also resolves  #1164 , is this still relevant for `plants`?

virtual_ecosystem/models/abiotic/model_config.py
```diff
@@ -177,6 +177,17 @@ class AbioticConstants(AbioticSharedConstants):
...
+    fraction_par_used_for_photosynthesis: float = 0.5
+    """Fraction of photosynthetically active radiation used for photosynthesis,
+    dimensionless."""
+
+    leaf_absorptance_non_par: float = 0.5
+    """Leaf absorptance for non-photosynthetically active radiation, dimensionless."""
+
+    par_fraction_of_shortwave_radiation: float = 0.5
+    """Fraction of shortwave radiation that is photosynthetically active,
+    dimensionless."""
```
davidorme: I think this is doing the same as 

> PlantsConstants.dsr_to_ppfd: float = 2.04

That has a unit conversion built into it - the underlying percentage is 46%

virtual_ecosystem/models/abiotic/model_config.py
```diff
@@ -177,6 +177,17 @@ class AbioticConstants(AbioticSharedConstants):
     """Maximum turbulent mixing coefficients, [m2 s-1].
     """
 
+    fraction_par_used_for_photosynthesis: float = 0.5
+    """Fraction of photosynthetically active radiation used for photosynthesis,
+    dimensionless."""
```
davidorme, line 182: Is the the extinction coefficient $k$?

davidorme (commented):
Couple of questions on the parameterisation.

virtual_ecosystem/models/abiotic/model_config.py
vgro: yes, I assume that would be the same; what is the unit here? I could call this in my function for consistency and saves us literature search for a value

virtual_ecosystem/models/abiotic/model_config.py
vgro: No, it is more an efficiency factor, I assume that not all PAR absorbed goes into NPP but some is lots as heat. If the plant model assumes that all PAR is converted to NPP I will drop this

virtual_ecosystem/models/abiotic/model_config.py
davidorme: The 2.04 converts Wm-2 DSR into µmol m2 s1 PPFD.

> Convert from downward shortwave radiation to photosynthetic photon flux density.
>
> Converting DSR in W m-2 to PPFD in µmol m-2 s-1. 1 W m-2 of sunlight is roughly 4.57 µmol m-2 s-1 of full spectrum sunlight, of which about 4.57 * 46% = 2.04  µmol m-2 s-1 is PPFD.

There was a source for the 46% but I've lost if for the moment.

virtual_ecosystem/models/abiotic/model_config.py
davidorme: The T model has a yield parameter ($y = 0.6$) that is (I think) handling this:

https://pyrealm.readthedocs.io/en/stable/api/demography_api.html#pyrealm.demography.tmodel.calculate_net_primary_productivity

That's described in the T model paper as "and y is the “yield factor” accounting for growth respiration." which doesn't sound like heat. It is sourced from here: https://onlinelibrary.wiley.com/doi/full/10.1111/j.1466-8238.2008.00442.x

virtual_ecosystem/models/abiotic/model_config.py
arne-exe: The 46% is the fraction that is useful/"visible" to plants (only part of the spectrum).

virtual_ecosystem/models/abiotic/model_config.py
arne-exe: FAPAR is how much is actually absorbed by the plant (some is lost through reflectance and transmittance). I don't know about a correction for once it is absorbed.

virtual_ecosystem/models/abiotic/model_config.py
vgro: Thanks for confirming. I will replace this constant.

virtual_ecosystem/models/abiotic/model_config.py
vgro: Thanks, this is already really helpful. I just don't want to double count/miss anything.
Here again my thought process:

- Shortwave downward radiation reaches the canopy
- part of this radiation is reflected (all wavelengths) -> does pyrealm account for this?
- part of the radiation that is not reflected is absorbed
- part of this absorbed radiation is PAR -> this should be the 46% and what the plant model used
- part of this 46% is used with a certain efficiency -> I am not sure this is the yield parameter which seems to be about respiration
- the remaining radiation (non PAR = 54% of SWD and unused PAR=PAR*0.4) should contribute to the energy balance in some form

So what I need to know is 
- is albedo accounted for in some form in the plant model?
- what _exactly_ is the `absorbed_radiation` variable?

arne-exe: > part of this 46% is used with a certain efficiency -> I am not sure this is the yield parameter which seems to be about respiration
>
The yield parameter represents growth respiration, or the fraction of GPP that is lost to respiration when constructing new tissues (so it's different from maintenance respiration, which is the carbon lost just to maintain the existing tissues).

I think the "certain efficiency" you're hinting at is not the yield parameter, because this efficiency should be taken into account before GPP is calculated, and the yield parameter is used after GPP is calculated to then calculate NPP.

The efficiency parameter is likely the light use efficiency, which represents the ratio of GPP to the amount of absorbed photosynthetically active radiation (calculated as incoming radiation that is not reflected and is useful for photosynthesis). I had a look here (https://pyrealm.readthedocs.io/en/stable/users/pmodel/pmodel_details/pmodel_overview.html#the-standard-p-model) and the P model takes this efficiency into account, so I don't think you'd need to correct the radiation passed on from the abiotic to the plants model (or you'd be correcting twice).

For absorbed radiation I looked here (https://pyrealm.readthedocs.io/en/stable/users/pmodel/shared_components/photosynthetic_environment.html#absorbed-irradiation-i-abs), and it looks like the P model already does a correction using FAPAR (although I haven't checked the actual code to verify if it does). So it looks like the abiotic model should pass PPFD, or the photosynthetically useful radiation (so using the 46%), that arrives at a specific location (i.e., at the leaves of a specific canopy layer). It looks like the P model then does most of the corrections. I guess you would still need to calculate these components in the abiotic model to convert the losses to other sources of energy, like heat.

vgro: This is really helpful, thanks @arne-exe . Just to clarify, the abiotic model does not pass radiation to the plant model, the plant model absorbed some of the radiation from the input data and the abiotic model then wants to know what part of this initial input radiation is left for the energy balance. So it seems that the `fraction_par_used_for_photosynthesis` is related to the light use efficiency. So I believe my assumptions are correct, I just need to replace my constants with the ones in the P-model

vgro: Is light use efficiency plant model output?

davidorme: Yup - with the standard P Model, the GPP is basically LUE * absorbed radiation, so it is something we can report. That varies with the environment so we would output LUE as a cells by vertical level array.

vgro: That would be brilliant. I can create a new issue for this so I can close this PR for now, would you be happy with that?

vgro: #1533

arne-exe: > Just to clarify, the abiotic model does not pass radiation to the plant model, the plant model absorbed some of the radiation from the input data and the abiotic model then wants to know what part of this initial input radiation is left for the energy balance. 
>
Oooh okay, that's very good to know. I was under the impression that radiation is first calculated for a particular timestep, but what you're saying is that the plant model uses the radiation at the start of a particular timestep (which is basically the radiation at the end of the previous timestep). Thanks for clarifying!
That simplifies this issue a lot too, I agree it makes a lot of sense to obtain LUE from the plants model and work from there.

> So it seems that the fraction_par_used_for_photosynthesis is related to the light use efficiency. So I believe my assumptions are correct, I just need to replace my constants with the ones in the P-model.
>
Agreed. If total absorbed radiation = incoming PAR * FAPAR, and heat losses = total absorbed radiation - radiation used for photosynthesis, you can then calculate heat losses as: (incoming PAR * FAPAR) * (1 - LUE).

davidorme: That seems right to me too @arne-exe - I haven't really engaged with the energy balance of photosynthesis yet!

davidorme (approved):
LGTM

PR #1517, opened 2026-04-17
Change target cohorts per functional group to 10

nickwctan:
Set target cohorts per functional group to a fixed value.


davidorme (commented):
@nickwctan This is fine for testing code locally but we wouldn't generally run an experiment like this by changing the `develop` branch.

If you want to try running the code like this you can use pip to install a package in edit mode (https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-e), or use a full poetry installation to alter VE locally.

That edit mode option might be a useful thing for us to dig into as a half way house between poetry and release versions.

nickwctan:
@davidorme my bad. I thought I was doing the commit in my own fork repo!

PR #1527, opened 2026-04-21
avoid negative flows below ground, the constant rates will require ca…

### LINKED ISSUE #1092: Negative river discharge
**Describe the bug**

When I ran ve_example to a longer time length, at the 50th month it errored
```
ValueError: The river discharge should not be negative!
```
which stemmed from
```
    subsurface_runoff_routed_plus_local = above_ground.route_horizontal_flow(
                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Any idea the river is running dry? (or soaking up water?)

**To Reproduce**

Run VE for longer by:
1. Install ve_example
2. add longer time length to config
```
[core.timing]
run_length = "10 years"
```
3. make sure the example data have matching time dimensions by changing `time` in `common.py` to
```
time = np.arange(np.datetime64("2013-01"), np.datetime64("2023-01")).astype(
    "datetime64[D]"
)
```
4. run ve_run

I ran it twice and both time it crashed at the 50th month step.

**Desktop (please complete the following information):**

- OS: Windows 10 x64
- VE dev version

vgro:
This PR makes sure that the below ground flows, and consequently the river discharge, don't go below zero.

Fixes #1092 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/hydrology/test_below_ground.py
+++ b/tests/models/hydrology/test_below_ground.py
@@ -129,3 +129,34 @@ def test_update_groundwater_storage(dummy_climate_data, fixture_hydrology_consta
     np.testing.assert_allclose(result["groundwater_storage"], exp_groundwat, rtol=1e-05)
     np.testing.assert_allclose(result["subsurface_flow"], exp_upper_flow, rtol=1e-05)
     np.testing.assert_allclose(result["baseflow"], exp_lower_flow, rtol=1e-5)
+
+
+def test_upper_flow_clamped_to_zero(fixture_hydrology_constants):
+    """Ensure negative groundwater outflow is clamped to zero."""
+
+    from virtual_ecosystem.models.hydrology.below_ground import (
+        update_groundwater_storage,
+    )
+
+    # Force a negative scenario
+    groundwater_storage = np.array(
+        [
+            [-10, -5, -1, -0.1],  # upper zone goes negative
+            [100, 100, 100, 100],  # lower zone normal
+        ]
+    )
+
+    result = update_groundwater_storage(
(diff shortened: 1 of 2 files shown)


hrlai:
Thanks @vgro , clamping it to zero sounds good. I can confirm that installing this commit works now on the maliau_2 scenario data. Only baseflow hit zero for 9 out of 100 grids, while other state variables plotted below did not really reach zero. The time at which baseflow hit zero (~midway) also explained why our run crashed around time step 70.

For curiosity, I ran for the full Maliau timeline and got the following:
image
image

Did you expect to see the runoff bounces back midway? Tagging @lelavathy also.

hrlai (approved):

vgro:
> Thanks @vgro , clamping it to zero sounds good. I can confirm that installing this commit works now on the maliau_2 scenario data. Only baseflow hit zero for 9 out of 100 grids, while other state variables plotted below did not really reach zero. The time at which baseflow hit zero (~midway) also explained why our run crashed around time step 70.
> 
> For curiosity, I ran for the full Maliau timeline and got the following: image image
> 
> Did you expect to see the runoff bounces back midway? Tagging @lelavathy also.

I am glad that this solved the problem.

Thanks for sharing the figures. We assume a constant rate for the water leaking out of the groundwater storage, and I think that was set too high causing the negative flows. This will be a key parameter for the calibration.

regarding the bouncing back, it could be the case here that as water flows down from the surface towards the groundwater, this recovers. I haven't looked into the input data, it could also be driven by higher rainfall.

This behaviour illustrates very nicely why we should run a spinup for the soil moisture...

PR #1537, opened 2026-04-23
1530 replace newton method with simpler implementation

### LINKED ISSUE #1530: replace newton method with simpler implementation

vgro:
This PR replaces the Newton method with a simpler and much faster secant method. I deleted all newton methods related functions as well as understorey explicit functions and constants that are no longer used.

Fixes #1530 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/abiotic/energy_balance.py
+++ b/virtual_ecosystem/models/abiotic/energy_balance.py
@@ -952,3 +952,47 @@ def secant_solve_cells_layers(
         return current_temperature, history
 
     return current_temperature
+
+
+def make_canopy_residual(
+    state,
+    static,
+    aerodynamic_resistance,
+    abiotic_constants,
+    core_constants,
+) -> Callable[[NDArray[np.floating]], NDArray[np.floating]]:
+    """Creates a residual function for canopy temperature to be used in root finding.
+
+    Args:
+        state: Dictionary containing state variables needed for the energy balance
+            residual.
+        static: Dictionary containing static variables needed for the energy balance
+            residual.
+        aerodynamic_resistance: Aerodynamic resistance of canopy, [s m-1]
+        abiotic_constants: Constants related to abiotic processes.
+        core_constants: Core constants.
+
+    Returns:
+        A function that takes canopy_temperature as input and returns the energy balance
+        residual.
+    """
+
+    def residual(canopy_temperature):
+        return calculate_energy_balance_residual(
+            canopy_temperature_initial=canopy_temperature,
+            air_temperature=state["air_temperature"],
+            evapotranspiration=state["evapotranspiration"],
+            absorbed_shortwave_radiation=state["shortwave_absorption"],
+            absorbed_longwave_radiation=static["absorbed_longwave_radiation"],
+            specific_heat_air=state["specific_heat_air"],
+            density_air=state["density_air"],
+            aerodynamic_resistance=aerodynamic_resistance,
+            latent_heat_vapourisation=state["latent_heat_vapourisation"],
+            leaf_emissivity=abiotic_constants.leaf_emissivity,
+            stefan_boltzmann_constant=core_constants.stefan_boltzmann_constant,
+            zero_Celsius=core_constants.zero_Celsius,
+            seconds_to_hour=core_constants.seconds_to_hour,
+            return_fluxes=False,
+        )
+
+    return residual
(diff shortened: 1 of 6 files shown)


davidorme (approved):
Looks good - well described code and clean change. 

It's solving all cells in parallel, right? Any idea how much faster it is?

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -939,3 +617,113 @@ def calculate_total_absorbed_shortwave_radiation(
...
+
+    previous_residual = residual_function(previous_temperature)
+    current_residual = residual_function(current_temperature)
+
+    for _ in range(maxiter_secant):
+        denom = current_residual - previous_residual
+
+        denom = np.where(
+            np.abs(denom) < denominator_tolerance,
+            denominator_tolerance * np.sign(denom + 1e-16),
```
jacobcook1995, line 655: Is there a specific reason for using `1e-16`. I guess this either should be explained in a comment or be an argument to the function (with `1e-16` as the default value)

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -939,3 +617,113 @@ def calculate_total_absorbed_shortwave_radiation(
...
+        next_temperature = (
+            current_temperature
+            - current_residual * (current_temperature - previous_temperature) / denom
+        )
+
+        next_residual = residual_function(next_temperature)
+
+        max_update = np.nanmax(
+            np.abs(next_temperature - current_temperature)
+            / (np.abs(current_temperature) + 1e-6)
```
jacobcook1995, line 670: Same question as above. (also should this be `1e-6` or `1e-16`?)

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -939,3 +617,113 @@ def calculate_total_absorbed_shortwave_radiation(
...
+
+    return current_temperature
+
+
+def make_canopy_residual(
+    state,
+    static,
+    aerodynamic_resistance,
+    abiotic_constants,
+    core_constants,
```
jacobcook1995, line 693: Types?

vgro:
> Looks good - well described code and clean change.
> 
> It's solving all cells in parallel, right? Any idea how much faster it is?

MUCH faster! It takes only a couple of seconds to run the `abiotic` model now

vgro:
I also found out why the values are so off and `ve_run` crashes, it's the vertical mixing, I'll do that next

davidorme:
> MUCH faster! It takes only a couple of seconds to run the `abiotic` model now

Amazing work! 

@kirkImperial - would love to see the new performance map once this hits `develop` 😄

jacobcook1995 (commented):
I had some minor comments, but agree that this looks like an improvement in general

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: Sorry I was a bit quick to merge, I'll include your comments in the next PR

virtual_ecosystem/models/abiotic/energy_balance.py
davidorme: I'm guessing that 1e-16 is shorthand for smallest representable number in float64 (https://en.wikipedia.org/wiki/Machine_epsilon)? It might be better to use a machine and float depth specific value?

```python
In [1]: import numpy as np

In [2]: np.finfo(float).eps
Out[2]: np.float64(2.220446049250313e-16)

In [3]: np.finfo(np.float32).eps
Out[3]: np.float32(1.1920929e-07)
```

I don't know if we want to make that some kind of globally available value, but using that function here for the moment and then raising an issue to make it something global might be best.

davidorme: Not that I think anyone is going to be feeding in `float32` data anytime soon, but we don't (I think) explicitly prevent it since we are using `np.floating` typing.

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: I also wonder how I get away with this despite pre-commit checks

davidorme: Huh - odd. Must be inferable from the other context? Try hovering over them in VSCode and see if it says what type they are.

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: I was just aiming for nonzero, I can change that

kirkImperial: I think naming a variable instead of having a 'magic number' is a good move, regardless if it's global or not.

hrlai:
@vgro does this affect / improve the `abiotic_simple` model too?

PR #1566, opened 2026-05-01
Better negative value handling in the soil model

### LINKED ISSUE #1559: Microbes should not be able to grow at negative nutrient concentrations
Microbial growth rates are primarily determined by their nutrient uptake rates. These rates are defined using Michaelis-Menten kinetics, e.g.

$$
q = \frac{kS}{K + S}M
$$

where $k$ is the uptake rate constant, $S$ is the substrate concentration, $K$ is a half saturation constant and $M$ is the microbial biomass.

This function is only sensibly defined when both $S\geq 0$ and $M\geq 0$. My current implementation tries to handle this by checking for cases where $q<0$ and replacing them with zero. This handles the $M < 0$ case, but isn't adequate to catch the $S < 0$ case, as if $S < 0$ **and** $|S| > |K|$ then $q>0$.

I should instead catch this by first checking if either $S<0$ or $M<0$ and returning zero if so.

There's a broader problem here that I need to track down the cause of (large) negative values in the simulation. But this change should reduce the downstream cascade of errors a bit.

(additionally, you can never 100% guard against (small) negative values when numerically integrating, so there's a value to building these protections in)

### LINKED ISSUE #1560: `calculate_water_potential_impact_on_microbes` needs to be bounded between zero and one
The `calculate_water_potential_impact_on_microbes` function is supposed to capture how much soil moisture impedes microbial activity. This is scaled between 1 (no impact on rates) and 0 (soil microbial processes halt entirely).

At present it can return negative values (which are non-sensical and break the model). In these case it should instead return zero

### LINKED ISSUE #1561: Sorption to MAOM shouldn't happen when pool is negative
`calculate_sorption_to_maom` currently has nothing to prevent negative sorption from occurring when pools are negative. This is an issue as it potentially allows small negative values (from the numerical integration) to explode

### LINKED ISSUE #1563: Organic matter removal by water has to be positive
At the moment `calculate_nutrient_removal_by_water` has steps to prevent negative removal of inorganic nutrients, but doesn't extend the same logic to organic matter. This is a clear oversight on my part which I should fix

jacobcook1995:
# Description

This PR fixes a function that could cause negative values within the soil model (`calculate_water_potential_impact_on_microbes`).

It also changes three functions that ran out of control when negative values emerged, so that they now should handle negative values properly. Fixing these is needed as the numerical integration solver can introduce small negative values when attempting to solve, which normally isn't an issue, but becomes a big problem if functions then start generating nonsense in response.

Fixes #1559 
Fixes #1560
Fixes #1561 
Fixes #1563 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/soil/test_env_factors.py
+++ b/tests/models/soil/test_env_factors.py
@@ -415,6 +415,31 @@ def test_calculate_solute_removal_by_soil_water(
     assert np.allclose(expected_rate, actual_rate)
 
 
+def test_calculate_solute_removal_by_soil_water_negative_concentrations(
+    dummy_carbon_data, fixture_core_components, fixture_soil_constants
+):
+    """Check solute removal rates handle negative values correctly."""
+
+    from virtual_ecosystem.models.soil.env_factors import (
+        calculate_solute_removal_by_soil_water,
+    )
+
+    lmwc_values = np.array([0.05, -0.02, -0.1, 0.005])
+    expected_rate = [1.07473723e-6, 0.0, 0.0, 5.25567712e-5]
+    exit_flow_per_day = np.array([0.1, 0.5, 2.5, 15.9])
+
+    actual_rate = calculate_solute_removal_by_soil_water(
+        solute_density=lmwc_values,
+        exit_rate=exit_flow_per_day,
+        soil_moisture=dummy_carbon_data["soil_moisture"][
+            fixture_core_components.layer_structure.index_topsoil_scalar
+        ],
+        solubility_coefficient=fixture_soil_constants.solubility_coefficient_lmwc,
+    )
+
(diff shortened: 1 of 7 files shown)


vgro (approved):
LGTM.
I suggest the only thing to double-check is if absolute zero works everywhere or if you need a tiny value to avoid division my zero or similar problems.

jacobcook1995:
> I suggest the only thing to double-check is if absolute zero works everywhere or if you need a tiny value to avoid division my zero or similar problems.

Ahh yes that's a good point. I've looked through and in most cases the quantities are masses (which are added/subtracted) or multiplicative factors. There is one place where division by zero could occur (if users give an impossible combination of values), which I should add some config validation to catch

PR #1568, opened 2026-05-01
Profiling feature branch

sallymatson:
# Description

Amazing work from @kirkImperial on the profiling foundation. 

The profiling package consists of `main.py` which sets the configuration for the profiling run, and `run.py` which executes the `ve_run_cli()` commands. 

I'm not sure if there is "better practice" than the may that `main.py` is formatted in terms of just being a script. Would be happy to hear form @dalonsoa and @davidorme if you have other opinions on workflow?  Otherwise, it runs well as is. 

I also added a test (which isn't really to be used a test) but that allows us to run profiling via pytest. This makes it super easy for anyone to quickly run profiling (just delete the skip decorator and run the test as any other). I included this because it is just that much simpler for anyone on the dev team to use this iteratively as they code. It also gives a prototype of using cProfile with pytest, which could be useful for other functions elswhere / targeted profiling runs. 

Fixes # (issue)

## Type of change

## Key checklist

## Further checks


pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
   "types-tabulate (>0.9)",
   "pydantic (>=2.13.4,<3.0.0)",
   "bottleneck (>=1.6.0,<2.0.0)",
+  "snakeviz (>=2.2.2,<3.0.0)",
 ]
```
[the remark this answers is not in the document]
davidorme: Agreed

davidorme:
My push doesn't _quite_ work as intended but I think is better. 

Basically `pytest.mark.skip` forces the test to be skipped - even if you explicitly use `pytest -m skip` to select them they get skipped. If we mark it as `pytest.mark.profiling` then we can also set the pytest configure to ignore that mark in day to day use (`-m "not profiling"`).

The oddness is that VSCode sees that configuration and takes it very much to heart - you **cannot** manually trigger the test within the VSCode testing tab. It's basically removed from the list of things recognised as tests.

But - without having to edit the files, you _can_ still run it from the command line using `pytest -m profiling`. Given you then want to run `snakeviz ve_run_profile.prof` from the command line, that works. It isn't massively different then from running the script from @kirkImperial but it is reasonable. Let's see how it goes!

pyproject.toml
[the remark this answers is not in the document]
sallymatson: Fixed

davidorme (approved):

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+# Set custom variables.
+"""Python version, can be written as "3.14" or "3_14"."""
+ver = "3.12"
+ver = ver.replace(".", "_")
+
+"""How many steps to run (negative values means no truncation)."""
+truncation = 3
+
+"""The OS you are running the code on. Options: "windows", "linux", "mac"."""
+user_os = "mac"
```
dalonsoa: This is not needed. See below.

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+    "windows": f".\\.venv\\Python{ver}\\Scripts\\python.exe",
+    "linux": f"./.venv/Python{ver}/bin/python",
+}
+command_options["mac"] = command_options["linux"]
+
+command = command_options.get(user_os, None)
+if command is None:
+    raise ValueError(
+        "Invalid user_os value. Please choose from 'windows', 'linux', or 'mac'."
+    )
```
dalonsoa: To get the python executable, just ask Python itself!

```suggestion
import sys

command = sys.executable
```

This is cross platform and more robust than trying to put together the path to the executable manually. Then you can remove the `user_os` variables.

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+
+import datetime
+import os
+import shutil
+import subprocess
+
+# Set custom variables.
+"""Python version, can be written as "3.14" or "3_14"."""
+ver = "3.12"
+ver = ver.replace(".", "_")
```
dalonsoa: Just ask python itself for its version


```python
"""Python version, for file naming purposes."""
import sys

ver = f"{sys.version_info.major}.{sys.version_info.minor}"
```

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+)
+ve_version = (
+    subprocess.check_output(command_ve_version, shell=True)
+    .decode("utf-8")
+    .strip()
+    .replace(".", "_")
+)
+print(f"Virtual Ecosystem v{ve_version}")
+# Generate terminal command to run `ve_run_cli()` via cProfile.
+output_name = f"VE_{ve_version}__py{ver}__truncated_at_step_{truncation}"
```
dalonsoa: This is very convoluted. Why not just:


```suggestion
import virtual_ecosystem as ve

output_name = f"VE_{ve.__version__}__py{ver}__truncated_at_step_{truncation}"
```

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+
+# The terminal command if you want to view the results table and/or visual breakdown.
+print(f"python -m snakeviz {profiler_folder}/{output_name}.prof")
+
+# Copy over the output to a nested folder sorted by Python version for historic runs.
+nested_profiler_folder = f"{profiler_folder}/python{ver}"
+if not os.path.exists(nested_profiler_folder):
+    os.makedirs(nested_profiler_folder)
+shutil.copy(
+    f"{profiler_folder}/{output_name}.prof",
```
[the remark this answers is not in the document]
dalonsoa: I broadly agree with these suggestions.

tests/test_profiling.py
```diff
@@ -0,0 +1,58 @@
...
+                    "core.debug.truncate_run_at_update=1",
+                ]
+            )
+
+            pr.disable()
+            pr.dump_stats("ve_run_profile.prof")
+
+        except Exception:
+            # If the code above fails then tidy up the logger to restore normal
+            # stream logging rather than leaving all other tests logging to the file.
```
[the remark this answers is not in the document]
dalonsoa: I agree with this.

tests/test_profiling.py
```diff
@@ -0,0 +1,58 @@
...
+                    "--logfile",
+                    str(logfile),
+                    "--config",
+                    "core.debug.truncate_run_at_update=1",
+                ]
+            )
+
+            pr.disable()
+            pr.dump_stats("ve_run_profile.prof")
+
```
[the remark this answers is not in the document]
dalonsoa: This is a good point. I suggest to gitignore the `.prof` files, instead, if not done already, or to dump the output into a gitignored directory.

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+
+# Create a folder for the cProfile outputs.
+profiler_folder = f"{path}/cProfile_outputs"
+if not os.path.exists(profiler_folder):
+    os.makedirs(profiler_folder)
+
+# Create a folder for the VE outputs.
+out_folder = f"{path}/out"
+if not os.path.exists(out_folder):
+    os.makedirs(out_folder)
```
dalonsoa: In modern python it is better to use `pathlib` to handle paths and then transform to a string at the end, if needed. So this would look like this (you might need to change other bits):


```suggestion
from pathlib import Path

path = Path.cwd() / "data" / "ve_example"

# Check example data exists to start the simulation.
if not (config_path := path / "config").exists():
    raise FileNotFoundError(
        f"Config folder not found at {config_path}. Please ensure the path variable is"
        " correctly set and the config folder exists."
    )

# Create a folder for the cProfile outputs.
profiler_folder = path / "cProfile_outputs"
profiler_folder.mkdir(parents=True, exist_ok=True)

# Create a folder for the VE outputs, removing it first if it exists already.
if (out_folder := path / "out").exists():
    shutil.rmtree(out_folder)
out_folder.mkdir(parents=True)
```

The last bit also clear the `out` folder in the same go, saving some code below.

profiling/main.py
```diff
@@ -0,0 +1,107 @@
...
+for filename in os.listdir(out_folder):
+    file_path = os.path.join(out_folder, filename)
+    try:
+        if os.path.isfile(file_path) or os.path.islink(file_path):
+            os.unlink(file_path)
+        elif os.path.isdir(file_path):
+            shutil.rmtree(file_path)
+    except Exception as e:
+        print(f"Failed to delete {file_path}. Reason: {e}")
+
```
dalonsoa: After the above, none of this is needed.

profiling/run.py
```diff
@@ -0,0 +1,34 @@
...
+    path = args.path if args.path else "."
+    truncate = args.truncate
+
+    ve_run_args = [
+        f"{path}/config",
+        "--out",
+        f"{path}/out",
+        "--logfile",
+        f"{path}/logfile{ver}.log",
+    ]
```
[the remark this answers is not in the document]
dalonsoa: Indeed! It seems Copilot knows VE better than you!

profiling/run.py
```diff
@@ -0,0 +1,34 @@
...
+    if truncate >= 0:
+        ve_run_args.extend(
+            ["--config", f"core.debug.truncate_run_at_update={truncate}"]
```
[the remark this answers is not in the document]
dalonsoa: This should do:

```suggestion
    parser.add_argument("--truncate", required=False, type=int)
    args = parser.parse_args()

    ver = args.ver
    path = args.path if args.path else "."
    truncate = args.truncate

    ve_run_args = [
        f"{path}/config",
        "--out",
        f"{path}/out",
        "--logfile",
        f"{path}/logfile{ver}.log",
    ]
    if truncate is not None and truncate >= 0:
        ve_run_args.extend(
            ["--config", f"core.debug.truncate_run_at_update={truncate}"]
```

dalonsoa (changes requested):
OK, this looks great to keep the profiling always at hand. I've made a bunch of comments to improve the styling and robustness of the code, as well as agreeing with Copilot in a few others.

profiling/main.py
kirkImperial: No, that's not what this line is about. It's not about grabbing the version of Python currently running, but the user saying what version of Python they want to run the VE. This was added due to the mention that VE had different run times across different Python versions. 

I understand what I just said was kinda confusing lol, I tried documenting it atop the document but I understand how it could have slipped past. This code is in the "custom variables" section for a reason 😁

profiling/main.py
kirkImperial: Because we don't want the version of VE being run from `main.py` but the version of VE being run in the user-specified version of Python. I could be running `main.py` in Python 3.14 but specify for VE to be run with Python 3.12 thus I'd not want to grab the VE version saved in 3.14's library via `ve.__version__`. I FULLY agree it's convoluted and ugly as heck, but it works for what I was looking for.

dalonsoa: Got it. In that case, rather than hard-coding variables that are meant for the user to tweak, use input arguments for the script. It does not need to be anything sophisticated or super fool-proof, just [argparse](https://docs.python.org/3/library/argparse.html) will do, but we do not want to have to edit the script every time we want to run it.

davidorme: If this is running within the VE development environment (which is pretty much a given right now) it will need to be using the `poetry` managed virtual environment that has all of the requirements installed. A user could create a different `venv` in the hard coded locations but currently that is assumed to exist.

I think the working practice right now would be that a user activates a poetry env and then runs profiling. It would then just use that activated Python version - which the script can pick up and record. If we wanted to run a different Python version, we'd activate a different poetry managed venv using the other Python version.

So, TLDR, I think it's easier to assume the user is handling version switching and just scrape that version data than handling switching within the profiling script.

davidorme: So - the code detects VE version, Python version and OS and puts those in the filename/metadata for the results, and everything else is controlled through activating the required venv first. That would also support VE version comparisons - can compare  two venvs with default vs branch code installed.

kirkImperial: I made it exactly the opposite -- so the script handles the version switch -- because I find that WAY easier, even if it is unfortunately uglier 😅 If you want to change it so the user to handle the version switching, go for it! However, please keep in mind that you need to do the development and testing for that because it's not the use-case I was working with.

profiling/run.py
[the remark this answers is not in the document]
sallymatson: Ahaha - it seems this is a semi-widespread issue throughout our docs. It seems to work fine as is though... odd !!!

sallymatson:
I think I've now made all the requested changes. After discussing with @davidorme we decided to move away from pre-defined python/VE versions in the script. Since we have such heavy adoption of poetry it'll be easier to run it like this for now and experiment with python versions manually. This just makes the script clearer - but tradeoffs noted @kirkImperial  and definitely acknowledge that you were working toward a brief which required that!! Thank you again for this work, it's already allowed us a 7x increase in speed (!!!) and will be invaluable going forward as well.

dalonsoa (approved):
LGTM!

davidorme (approved):
👍

PR #1573, opened 2026-05-06
Bumping version number

### LINKED ISSUE #1572: New release with faster abiotic model

vgro:
New release with faster abiotic model.


vgro:
this is postponed

PR #1576, opened 2026-05-07
1575 atmospheric humidity update needs revisit

### LINKED ISSUE #1575: atmospheric humidity update needs revisit
There are currently too many untraceable steps in this function, would be better to split in small testable functions. One big problem is the condensation, which causes more problems further down the line.

vgro:
This PR makes the update of atmospheric humidity a bit more clean with smaller functions and more tests.

Fixes #1575 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/abiotic/energy_balance.py
+++ b/virtual_ecosystem/models/abiotic/energy_balance.py
@@ -52,7 +52,6 @@ from xarray import DataArray
 
 from virtual_ecosystem.core.core_components import LayerStructure
 from virtual_ecosystem.core.model_config import CoreConstants
-from virtual_ecosystem.models.abiotic import wind
 from virtual_ecosystem.models.abiotic.abiotic_tools import (
     compute_weights_from_absorbed_radiation,
     set_unintended_nan_to_zero,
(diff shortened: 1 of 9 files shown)


virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -525,19 +523,24 @@ def update_humidity_vpd(
...
-        specific_humidity_updated - saturation_specific_humidity,
-        limits_specific_humidity[0],
+    # Negative values mean supersaturation
+    specific_humidity_deficit = saturation_specific_humidity - specific_humidity_mixed
+
+    # Excess humidity available for condensation, [kg kg-1]
+    excess_specific_humidity = np.where(
+        specific_humidity_deficit > 0,
+        0.0,
+        specific_humidity_deficit,
```
jacobcook1995, line 533: Just to quickly sense check, `specific_humidity_deficit` should never be positive, so this step catches any values that are and replaces them with zero?

jacobcook1995 (approved):
One quick sense check, but otherwise LGTM!

virtual_ecosystem/models/abiotic/energy_balance.py
vgro: The idea here is that if the air is oversaturated, specific humidity is larger than saturation specific humidity, so the deficit gets negative and we have this excess which turns into condensation. If the humidity is less than saturation, the value should be positive and no excess produced.

jacobcook1995: that makes sense, just thought I should double check!

PR #1580, opened 2026-05-08
Feed herbivory back into plant model

### LINKED ISSUE #1550: Feed foliage herbivory back into productivity
Feed foliage herbivory back into productivity.

This issue needs to:

1. Read the contents of `canopy_foliage_cnp_consumed`
2. Apply a carbon cost to the plants of replacing that foliage
3. Update the canopy representation to reflect herbivory - probably need to express as the average herbivory penalty over the time period?

davidorme:
# Description

This PR:

* Moves the generation of herbivory consumption pools from the animal model into the plants model and updates the `ResourceArray` implementation to allow existing consumption pools (ultimately - I think it makes more sense for source models to handle consumption pool creation, rather than the current mixed model @jacobcook1995).

* Adds biomass methods:
    * `BiomassTissueABC.get_relative_carbon_biomass_by_pft`  which allows the herbivory (which happens at a PFT level) to be distributed across cohort biomasses in proportion to their relative biomass within their PFT in a cell.
    * `BiomassTissueABC.apply_herbivory` which removes CNP from the cohorts within a tissue instance.

* Adds `PlantsModel.apply_herbivory` which:
    * Loops over the canopy foliage, seed and fruit consumption pools and removes biomasses from the tissue biomasses.
    * Adjusts the LAI and foliage turnover traits so that calculation of canopy absorption and carbon allocation to foliage turnover accounts for foliage loss and replacement.

* Adds `PlantsModel.update_allometry` which:
   * Separates the StemAllometry update from inside `allocate_gpp`
   * Adds code to restore LAI and tau_f to maintain the theoretical expectations of biomass and to avoid compounding herbivory effects over time steps.

* Adds tests for the two new `PlantsModel` methods that demonstrate they do what they claim.
 
Fixes #1550

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -658,29 +687,55 @@ def _update(self, time_index: int, **kwargs: Any) -> None:
         # Get the canopy top shortwave downwelling radiation for the current time slice
         self.set_canopy_top_radiation(time_index=time_index)
 
-        # Update the canopy layers and subcanopy and then set the shortwave absorption
+        # Apply herbivory effects - this modifies the stem LAI and tau_f parameters, and
+        # care needs to be taken about the sequencing of using the modified values - see
+        # comments on downstream methods.
+        self.apply_herbivory()
```
sallymatson, line 693: Was it intentional to call this function and the biomasses function both `apply_herbivory`?

sallymatson: The biomasses version should really be called `get_eaten`


/s

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -808,6 +863,111 @@ def set_canopy_top_radiation(self, time_index: int) -> None:
...
+                # biomass of cohorts to distribute the herbivory between cohorts.
+                herbivory_by_cohort = (
+                    self.data[f"canopy_{herbivory_tissue}_cnp_consumed"].sel(
+                        cell_id=cell_id, pft=community.cohorts.pft_names
+                    )
+                    * relative_herbivory[:, np.newaxis]
+                )
+
+                # Apply the per cohort herbivory to the tissue
+                tissue.apply_herbivory(herbivory_by_cohort)
```
sallymatson, line 938: Okay I see why you've used these naming conventions :)

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -808,6 +863,111 @@ def set_canopy_top_radiation(self, time_index: int) -> None:
...
+        .. math::
+
+            \tilde{\tau_f} = \frac{W_f \tau_f}{H_f \tau_f + W_f}
+
+
+        .. todo::
+
+            There is a timing issue here - the effects of herbivory in the last time
+            step are being used to modify the light gathering and leaf turnover costs
+            for the current plants timestep.
```
sallymatson, line 910: Does this mean that the LAI used at the start of the next timestep will be this modified value? 

Because in that case I think we just have to accept that. We can't do everything instantaneously, so as long as the effects of X herbivory are impacting the model, even if it's on a slight delay, I think that's just something we accept.

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -1157,9 +1317,40 @@ def allocate_gpp(self) -> None:
             # BALANCE THE NUTRIENTS WITHIN BIOMASSES
             biomasses.balance_elements()
 
+    def update_allometry(self) -> None:
```
sallymatson, line 1320: I really like this use of allometry. It makes a lot of sense and works perfectly because we otherwise never recalculate allometry, so this will not be over-ridden.

sallymatson (approved):
LGTM. There's still something I need to fully internalise about the timing / replacement, but I know you've worked through those.

virtual_ecosystem/models/plants/plants_model.py
davidorme: It felt like keeping that chain down from model -> biomasses -> tissues helped trace the process.

virtual_ecosystem/models/plants/plants_model.py
davidorme: No - the `update_allometry` method restores LAI and $\tau_f$ but conceptually the herbivory being applied at time t+1 was generated by the animals in the previous time step. I haven't quite settled on whether that is a bug or not. I'm leaning towards not because we don't apply growth until the end of the update so what we are modelling here is what happened between the last allometry being calculated and the new one - and that includes the herbivory in between. 

So actually, I think it's correct!

virtual_ecosystem/models/plants/plants_model.py
davidorme: Yup - it felt cleaner to have it as a distinct step anyway and since I was bolting some other code onto it, it just felt right to extract it from `allocate_gpp`

PR #1584, opened 2026-05-12
Animal - Heterotroph Biomass Normalization

### LINKED ISSUE #1523: Animal - Heterotroph biomass normalization
Madingley uses a 'total heterotroph biomass' normalization on animal initialization densities. This keep the total het biomass scaled appropriately regardless of total functional group number. We should implement this.

TaranRallings:
# Description

Implements the Madingley heterotroph normalization system to constrain total heterotroph biomass density at initialization regardless of how many functional groups are defined. Previously, each functional group independently contributed biomass, meaning total heterotroph density scaled linearly with the number of groups defined.

A fixed target total heterotroph biomass density `total_heterotroph_biomass_density_kg_m2` is introduced as a config-level constant in `AnimalConstants`. At initialization, each functional group receives a share of this budget proportional to its raw allometric biomass density. A single normalization factor is computed across all groups and applied uniformly, so the sum of normalized biomass densities always approximates the target. Both empirical density overrides `density_individuals_m2` in the CSV and allometric groups participate in the budget.

`raw_biomass_density_kg_m`2 - returns a functional group's raw biomass density in kg m⁻², routing through the empirical or Madingley allometric

`heterotroph_normalization_factor` - computes the single cross-group scaling factor as target / sum(raw biomass densities)

`biomass_density_to_individuals` - scaling-law-agnostic conversion from normalized biomass density to a total individual count

New field `total_heterotroph_biomass_density_kg_m2` on `AnimalConstants`, defaulting to the Madingley terrestrial median (0.151 kg m⁻²)

Fixes #1523 

## Type of change

## Key checklist

## Further checks


nickwctan:
I only had a quick skim through it, but this seems promising for reducing the insanely high population number of each FG which we are having right now.

sallymatson (approved):
Conceptually this all makes a lot of sense and was quite easy to follow. 

Can't say I was doing dilligent mathematically checking though, but I trust your values & testing!

PR #1592, opened 2026-05-13
Animal - Foraging Speed-Up

### LINKED ISSUE #1588: Animal - Foraging Speed-up
There are a few straight-forward changes that can be made to the foraging loop that should speed it up. 

- within `calculate_total_handling_time_for_herbivory`, `H_i_k` is constant at the cohort level but being recomputed per plant resource
- `alpha` and `total_handling_time_herbivory` - calculate once per cohort inside within `forage_resource_list`

Moving `calculate_total_handling_time_for_herbivory` outside of the loop is the most important thing, this is what is blowing up calculation times.

TaranRallings:
# Description

Profiling identified the herbivory path (~200s per update) as the main bottleneck.

`calculate_total_handling_time_for_herbivory`  was being called once per resource inside `forage_resource_list` unnecessarily. `alpha` and `H_i_k` were similarly recomputed redundantly. All three are now precomputed once per foraging bout.

The only science that changes here is fixing a small arithmetic bug (a sum that should be a product), the rest is simply aligning helper methods and tests with the above structural change.

edit - I added some clamping on cnp noise to this PR as it was straightforward to do and was producing errors. 

Fixes #1588 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -1212,18 +1129,29 @@ def forage_resource_list(
...
+        )
+        A_cell = self.grid.cell_area
+        dt_days = float(adjusted_dt / timedelta64(1, "D"))
+        conv_eff = self.functional_group.conversion_efficiency
+
         for resource in resources:
-            requested = calculate_consumed_mass(resources, resource, adjusted_dt)
+            potential_biomass_consumed = sf.k_i_k(alpha, resource.mass_current, A_cell)
+            F = self.F_i_k(resource, potential_biomass_consumed, total_handling_t)
+            requested = resource.mass_current * (1.0 - exp(-F * dt_days))
```
davidorme, line 1182: Could you add some comments here, just to help follow the process?

davidorme (approved):
Looks good - it's a clear calculation saving. A few more comments is my only plea 😄

virtual_ecosystem/models/animal/animal_cohorts.py
TaranRallings: This is a good point. I was speeding a bit and got lazy.

After putting this up, I realized that I have a floating point error problem that is breaking ve_run. I'm going to add a bit of clamping to handle that and I'll put some more comments at the same time.

PR #1589, opened 2026-05-13
canopy temperature added to abiotic_simple init

vgro:
This PR adds `canopy_temperature` and `diurnal_temperature_range` to the `abiotic_simple` `init`. @TaranRallings   could you please double-check that the `canopy_temperature` is actually what you are looking for in the animal model, or if you want the air temperature surrounding the canopy.

Fixes #1585

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Optimization (back-end change that speeds up the code)
- [x] Bug fix (non-breaking change which fixes an issue)

## Key checklist

- [x] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`
- [x] All tests pass: `$ poetry run pytest`

## Further checks

- [x] Code is commented, particularly in hard-to-understand areas
- [x] Tests added that prove fix is effective or that feature works
- [x] Relevant documentation reviewed and updated

PR #1602, opened 2026-05-18
Add integration tests to publication workflow

### LINKED ISSUE #1587: Make `test_ve_run` an integration test
`test_ve_run` is already a very slow test. It also fails to catch any simulation errors that emerge after the 1st time step, or with alternative model combinations (which at present is only really swapping `abiotic` with `abiotic_simple`). We don't need to check these things with every release/PR (as we already have extensive unit testing), but we really need to check that the model can run using the example data before we make a new release

We already have an automated release pipeline, so we could separate out a suite of integration tests (which would contain `test_ve_run` an a `abiotic_simple` variant). Then the existing CI could be altered to exclude this suite, and an extra step could be added to the release pipeline to run this suite.

At some point in the future this suite could then be extended to run with the Maliau data (or other real world scenarios) instead, but that introduces a lot of additional complexity, and just verifying that the model runs with the example data before releasing would already be a big improvement

In terms of separating out the suites there's two (non exclusive) options.

1) labelling tests using `@pytest.mark.xyz`. We already mark `test_ve_run` as `slow`
2) Moving integration tests into a separate folder, i.e. so the unit tests live in `tests/unit` and the integration tests live in `tests/integration`

Both of these options can be made to work for the CI. I guess it more comes down to what people think would work better with how they run tests locally.

Option 1, would be run as `pytest -m "not slow"`

Option 2, would be run as `pytest tests/unit`

@dalonsoa, if you have any thoughts/suggestions on this they would be very much appreciated!

jacobcook1995:
# Description

This PR adds integration tests, and includes them as part of the publication workflow. The actual tests are currently simple. They are:
1. does the standard model combo run with the example data run successfully for 24 months.
2. does running an alternative model set (`abiotic` replaced with `abiotic_simple`)  with the example data run successfully for 24 months.

It is definitely worth thinking about using more realistic example data at some point, but I think this is a good starting point, particularly given that both integration tests currently fail (though (2) should be fixed by #1589).

I also changed how the example config is split up, so that each model has its own configuration file. The main impact of this change is that you can't just point `ve_run` at the configuration folder as it contains a config for both `abiotic` and `abiotic_simple`. Instead you have to point it to every config file you want to include.

I couldn't think of anyway to test the new publication pipeline (aside trying to use it in anger). If anyone had any suggestions of how to do that they would be very welcome

Fixes #1578

## Type of change

## Key checklist

## Further checks


davidorme (dismissed):
LGTM - I don't think there is any way around the CI setup testing apart from trying to push the new version!

davidorme (changes requested):
It is good practice to document new markers in the pytest config (you get warnings otherwise). I think it also might be worth turning off running integration tests in the day to day test suite?

We've had a similar thing over in the `pyproject.toml` for the profiling:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1568/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

One downside is that VSCode takes those project settings very seriously - you can't run the tests from the normal testing pane if they are removed in the config. But you can still run them manually using `pytest -m integration`. This basically just saves us from having bare `pytest` calls (which I do a lot) running the integration test by default.

sallymatson:
> It is good practice to document new markers in the pytest config (you get warnings otherwise). I think it also might be worth turning off running integration tests in the day to day test suite?
> 
> We've had a similar thing over in the `pyproject.toml` for the profiling:
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1568/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711
> 
> One downside is that VSCode takes those project settings very seriously - you can't run the tests from the normal testing pane if they are removed in the config. But you can still run them manually using `pytest -m integration`. This basically just saves us from having bare `pytest` calls (which I do a lot) running the integration test by default.

Hm... is there no way around this in VSCode? I may be the only one but I exclusively run tests from VSCode so if it explicitly disallows, that is a bit of a nuisance.

jacobcook1995:
> It is good practice to document new markers in the pytest config (you get warnings otherwise).

Ahh good catch I'll document the new marker

> I think it also might be worth turning off running integration tests in the day to day test suite?

Doing this would certainly make my life easier (I run bare `pytest` quite a lot). I guess it comes down to a combo of (a) how many people run mainly via the VSCode pane (b) who would have the easier adapting their test running

davidorme:
From my experiments, if a marked test appears in `pyproject.toml` as `-m "not markname"` then the run test icon in the margin disappears and it drops out of the list of tests in the Testing sidebar. 

I can't see an obvious way of fixing that _apart_ from altering the config and I guess what I am after here is a thing that runs all the things we normally want as standard and we don't have to change file contents to do other things. But... I use `pytest` from the shell a lot: I tend to use the built in Testing when debugging and doing test-lead development, but use the command line when checking a PR passes.

With this setup, all the standard tests are there in VSCode, we just have to swap to the command line for running `profiling` or `integration` tests.

dalonsoa:
I've just found this. Might be useful, as it is or on a similar form. https://stackoverflow.com/a/78937721

jacobcook1995:
> I've just found this. Might be useful, as it is or on a similar form. https://stackoverflow.com/a/78937721

Hmmm yes, the bit that confuses me there though is

`In the conftest.py where you setup the "skip", ...`

As far as I can tell we don't do anything to setup the skip, we just use the `-m "not profiling"` as an addopts argument for `pytest.ini_options` in `pyproject.toml`. It's not particularly clear to me how to setup the skip within conftest instead (I could be missing something major though)

davidorme (approved):
LGTM

PR #1604, opened 2026-05-18
Set variable paths dynamically 

### LINKED ISSUE #1603: Set variable paths dynamically
In https://github.com/ImperialCollegeLondon/ve_data_science/pull/321 there is a usage issue where the data science team want to iterate over input variable files.

This can't currently be set using `ve_run --config ...` overrides, because the override needs to target entries in the `[[variable]]` table. We might be able to write a special case to handle this, but we can also inject paths dynamically in the config.

From that PR:

For what it is worth, I think a technical fix to this would to be to allow a data config TOML file like this:

```toml
[[core.data.variable]]
file_path = "$SOIL_DATA_PATH"
var_name = "fungal_fruiting_bodies"
```

And then a run could do:

```bash
export SOIL_DATA_PATH=../../data/sobol/soil_sobol_1.nc
ve_run ...
```

We'd need to extend VE to allow it to detect that environment variable and then look for and sub in the value, but with this approach, ve_run can dynamically change input variables but use a single config file.

_Originally posted by @davidorme in https://github.com/ImperialCollegeLondon/ve_data_science/issues/321#issuecomment-4449532964_

davidorme:
# Description

> [!IMPORTANT]
> Rewritten to document shift in approach from environment variables to command line

New functionality:

* Extends the validation of file and directory paths in config file to allow file markers that can be populated dynamically at run time. Markers have the format "$FILE_MARKER" - for example:

```TOML
[[core.data.variable]]
file_path = '$FILE_MARKER'
var_name = "prec"
```

* File markers can be set at the command line by using the `--data-path` or `-p` argument to `ve_run`, for example `-p FILE_MARKER=/path/to/file.nc`
*  ~Alternatively, file markers can also be set as environment variables before running `ve_run`, eg. `export FILE_MARKER=/path/to/file.nc`~

Implementation

* In `entry_points`: 
    * Added the  `--data-path` argument to `ve_run_cli`
    * Adds the `_parse_cli_paths` helper function to compile the command line arguments to `--data-path`.
    * Also added the `--validate-config-only` to `ve_run_cli` .
    * Updated docstring/command line help.
    
* In `main`:
    *  Updated `ve_run` to add `cli_paths` and `validate_only` argument.
    * The `cli_paths` argument is a dictionary mapping markers to paths:
    
    ```python
    {"FILE_MARKER": "/path/to/file.nc"}
    ```
    This is passed in to the pydantic validation of the configuration through the `context` argument, which is passed to all downstream validation.

* In `core.config_builder`:
    * The `_resolve_config_paths` function does not try and resolve paths starting with `$` to avoid breaking file markers.
    * The `generate_configuration` object now accepts `context` to pass down to validation through `pydantic`. This is *incredibly* convenient - it is an already existing mechanism to pass extra information down throughout the validation tree.

* In `core.configuration`:
    *  The `placeholder_validation` validator - which runs on file and directory fields, now accepts context and looks for marker definitions in ~both the environment and~ the context. 

* Adds a new doc page giving some approaches to model modification for running scenarios and sensitivity testing.

* Adds tests:
    * New test of the argument parsing function `_parse_cli_paths`
    * Updates testing of `placeholder_validation` to check it handles ~env and~ context marker definitions
    * Adds a new `test_cli_integration.test_ve_run_cli_cli_paths` that checks that the arguments are passed down through `ve_run` to the field validation correctly.

Fixes #1603 

## Type of change

## Key checklist

## Further checks


docs/source/using_the_ve/modifying_a_model.md
jacobcook1995, line 1: Would be good to link to this file from `docs/source/using_the_ve/overview.md` as all the other how to guides/tutorials have links there

jacobcook1995 (changes requested):
I think the new documentation looks really good, just had one suggestion to better highlight the new page

The new way of handling paths also looks sensible to me

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+language_info:
+  codemirror_mode:
+    name: ipython
+    version: 3
+  file_extension: .py
+  mimetype: text/x-python
+  name: python
+  nbconvert_exporter: python
+  pygments_lexer: ipython3
+  version: 3.11.9
```
hrlai, line 23: Just for curiosity, there is no python code in this md, but these YAML configs are still required for a Jupytext?

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+# Modifying a model
+
+Once you have set up a model for your site, you are almost certainly going to want to
+modify the initial configuration to explore model behaviour under different conditions.
+
+There are a number of different ways you can do this and we describe some options below.
+In general, we strongly advise against creating many copies of configuration files and
+altering each set. This is hard to do systematically and creates a lot of file
+duplication and it is generally easier to use one of the solutions below to run
+alternative models based on the same initial site configuration.
```
hrlai:
```suggestion
alternative models based on the same initial core set of config files.
```

Suggest to reuse the same phrase as L43.

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Overriding a configuration parameter
+
+The `ve_run` command accepts a `--config` option that can be used to override one or
+more of the settings from the configuration file. This makes it fairly easy to have a
+single core set of config files and permute a small number of settings.
+
+For example, say you wanted to rerun the model for your site but using a different
```
hrlai:
```suggestion
For example, you wanted to rerun the model for your site but using a different
```

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+maximum soil depth for microbial activity and leaf albedo then you could run:
+
+```bash
+ve_run \
+  config/data_variables.toml \
+  config/soil_config.toml \
+  ...
+  config/animal_config.toml \
```
hrlai, line 53: I probably have missed it, but this way of specifying individual config files has always been available? It isn't clear to us, and we thought that one needs to put the files in a single folder and then specify a single path in the args, i.e., we thought inputting individual configs is an upcoming feature. If we knew that then we might have set `maliau_1` and `maliau_2` differently to reduce file duplication. For example, the animal, soil microbial, plant constants, and core.timing are basically the same between these scenarios.

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+
+There are a number of different ways you can do this and we describe some options below.
+In general, we strongly advise against creating many copies of configuration files and
+altering each set. This is hard to do systematically and creates a lot of file
+duplication and it is generally easier to use one of the solutions below to run
+alternative models based on the same initial site configuration.
+
+## Overriding a configuration parameter
+
+The `ve_run` command accepts a `--config` option that can be used to override one or
```
hrlai, line 41: Should we emphasise the overriding property? E.g., add a line somewhere below like "...`core.constants.max_depth_of_microbial_activity` will override the existing value in the config files"

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+  config/soil_config.toml \
+  ...
+  config/animal_config.toml \
+  --config core.constants.max_depth_of_microbial_activity=0.5 \
+  --config abiotic.constants.leaf_albedo=0.2
+```
+
+## Using an alternative configuration file
+
+The approach only works for explicitly named configuration settings and some parts of
```
hrlai, line 63:
```suggestion
The approach above only works for explicitly named configuration settings. However, some parts of
```

I hope I haven't changed the original meaning.

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+  config/rcp_45_climate_variables.toml \
+  config/soil_config.toml \
+  ...
+  config/animal_config.toml \
+```
+
+## Iterating over many inputs
+
+If you are trying to explore model sensitivity to parameter variation then it is fairly
+easy to loop over combinations of configuration settings using `--config`. However, this
```
hrlai, line 98:
```suggestion
easy to loop over combinations of explicitly named configuration settings using `--config`. However, this
```

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+doesn't work if you need to iterate over different input data files. For example, if you
+have 1000 simulations of climatic patterns from El Niño-Southern Oscillation events then
+you might want to change the abiotic conditions to use each one of those files.
```
hrlai:
```suggestion
you might want to change the abiotic conditions to use each one of those input data files.
Below we offer two possible solutions:
```

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+
+### File substitution
+
+One approach that works if the models are running in series or on different computers
+(such as on an HPC cluster) is to generate a configuration model that points to a
+placeholder data file location and then copy the appropriate real data to that location
+before running the file.
+
+For example, your might have a data configuration file `climate_data_variables.toml`
```
hrlai:
```suggestion
For example, your might have a data configuration file `config/climate_data_variables.toml`
```

This is very not essential, your call!

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+```
+
+### Dynamic file definition
+
+File substitution does not work well if you are running models in parallel on the same
+system: the shared location can only point to one file at a time. An alternative
+approach is to set the file path dynamically in the configuration using environment
+variables.
+
+In this approach, your climate configuration file `climate_data_variables.toml` would
```
hrlai:
```suggestion
In this approach, your climate configuration file `config/climate_data_variables.toml` would
```

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+look like this:
+
+```toml
+# Climate data
+[[core.data.variable]]
+file_path = "$ENSO_FILE"
```
hrlai: We need to explain that this is called an "environment variable" to the layperson. I would personally appreciate some hand-holding here. 

Maybe something like:

...where `$ENSO_FILE` is an environment variable that can be defined in the terminal prior to calling `ve_run`. An advantage of setting `$ENSO_FILE` external to `ve_run` is that we can vary it in a bash job array script <blah blah>

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -0,0 +1,168 @@
...
+export ENSO_FILE=enso_simulations/enso_0001.nc
+
+ve_run \
+  config/climate_data_variables.toml \
+  config/soil_config.toml \
+  ...
+  config/animal_config.toml \
+```
+
+```{important}
```
hrlai: Sorry I am not sure what this callout box means... Are you trying to explain the scope/limitation of `os.environ.get()` in `virtual_ecosystem/core/configuration.py`?

hrlai (approved):
Looks good, this is really useful! 

I only have minor suggestions that do not change it overall, so I opted to also approve this.

I really wish I had known this a year ago 😢 It would change how we do the Maliau scenario. Things are a bit set in stone there but I will try to discuss with the group about refactoring the two Maliau scenario files. I've always been feeling that there are too many file duplication, which had been hard to maintain on Globus.

docs/source/using_the_ve/modifying_a_model.md
davidorme: Yeah - they aren't required here it is basically just a template we tend to copy between pages 😄

docs/source/using_the_ve/modifying_a_model.md
davidorme: Yes - it has always been available - hence my detailed comments on https://github.com/ImperialCollegeLondon/ve_data_science/pull/321. More on Teams

docs/source/using_the_ve/modifying_a_model.md
davidorme: Yeah - needs more unpacking.

hrlai:
We don't really mention it here, but it occurs to me that this may also be useful not only for sensitivity analysis, but also calibration.

dalonsoa (approved):
LGTM, although environment variables are not my preferred option, in general. They are magically available, set somewhere, somehow and not explicitly defined when calling a tool. I see their value, obviously, but I think they tend to make things more obscure.

davidorme:
> LGTM, although environment variables are not my preferred option, in general. They are magically available, set somewhere, somehow and not explicitly defined when calling a tool. I see their value, obviously, but I think they tend to make things more obscure.

I agree with your concerns completely. This might only be a temporary fix while we figure a better way to replace files dynamically in a config. It feels "correct" in a technical sense, and I wanted a sense check on that, but it certainly isn't the friendliest approach.

davidorme:
@dalonsoa OK, so I think the alternative would be to setup a new command line parameter:

```bash
ve_run \
  config/climate_data_variables.toml \
  config/soil_config.toml \
  ...
  config/animal_config.toml \
  -p ENSO_FILE=/path/to/enso_01.nc
```

We can pass that information down into the pydantic validation and use it swap in paths. I mean that could do:

```bash
ve_run \
  config/climate_data_variables.toml \
  config/soil_config.toml \
  ...
  config/animal_config.toml \
  -p /old/path/enso.nc=/path/to/enso_01.nc
```

but I'm inclined to stick to explicit `$DYNAMIC_PATH` indicators rather than just substitute any old path?

dalonsoa:
> @dalonsoa OK, so I think the alternative would be to setup a new command line parameter:
> 
> ```shell
> ve_run \
>   config/climate_data_variables.toml \
>   config/soil_config.toml \
>   ...
>   config/animal_config.toml \
>   -p ENSO_FILE=/path/to/enso_01.nc
> ```
> 
> We can pass that information down into the pydantic validation and use it swap in paths. 

I think this would be more explicit and clear, rather than trusting on environment variables, and it serves the same purpose.

davidorme:
@jacobcook1995 and @dalonsoa 

I've reworked this a lot to handle it as a command line argument to `ve_run` and `ve_run_cli` - worth reading the top description as I've rewritten it with the details of the updated content.

I've kept environment variables as an alternative for now, but could just rip that out to have one approach.

jacobcook1995 (approved):
Addition to the docs reads well!

dalonsoa (approved):
This looks nicer :) About the environment variables, unless you have a specific use case where that approach is better/the only option, I would ditch that entirely.

davidorme:
> About the environment variables, unless you have a specific use case where that approach is better/the only option

I can't think of one...

Issue #1549, opened 2026-04-27
Deal with no canopy case


vgro:
The model cannot handle 'no canopy' grid cells well, especially in the vertical mixing. This causes the temperatures to go mad.

vgro:
resolved in #1608

PR #1608, opened 2026-05-19
1591 sign needs fixing in vertical mixing

### LINKED ISSUE #1591: Sign needs fixing in vertical mixing

vgro:
This PR turned into quite a bit change, fixing a number of bugs that lead to the extreme temperatures. Now the abiotic model runs at least sensibly in static mode. When adding the plants, there is still a problem when the canopy layers get very small. With soils it runs into a loop and gets very slow. I didn't want to squeeze more changes in here but will continue to work on this.

Main changes:
- corrected calculation of absorbed longwave radiation
- accounted for canopy layers lower than the surface layer
- simplified update of surface air temperature
- proofed against division by zero due to zero canopy

Fixes #1591 #1549

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
Happy to approve as the general code quality looks good, and if there are any problems with the science they are going to be at a level I can't spot

PR #1612, opened 2026-05-20
Prevent negative organic nutrient uptake

### LINKED ISSUE #1611: `calculate_maximum_uptake_rates` doesn't handle negative concentrations properly
To prevent negative values from spiralling out of control in the soil model, I need to ensure that microbes don't attempt to uptake negative nutrients.

This is already handled for nutrients that are explicitly taken up by microbes (i.e. uptake halts if concentration goes below zero).

What isn't handled is the uptake of  organic nutrients, where the uptake is implicit 

Currently, `calculate_maximum_uptake_rates` does the following

```python
lmwc_c_n_ratio = soil_c_pool_lmwc / soil_n_pool_don
lmwc_c_p_ratio = soil_c_pool_lmwc / soil_p_pool_dop
organic_nitrogen_uptake_rate_max = carbon_uptake_rate_max / lmwc_c_n_ratio
organic_phosphorus_uptake_rate_max = carbon_uptake_rate_max / lmwc_c_p_ratio
```

while `carbon_uptake_rate_max` is guaranteed to be >= 0, that is not true for the pools themselves. If either `soil_c_pool_lmwc` or `soil_n_pool_don` is negative, then organic nitrogen uptake should halt

jacobcook1995:
# Description

Adds a step to catch cases where the nutrient components of organic matter are negative, and prevents (negative) uptake by microbes in this case (this should stop the soil model from providing the plants with negative mycorrhizal nutrient supplies)

Fixes #1611

## Type of change

## Key checklist

## Further checks

- [N/A] Relevant documentation reviewed and updated

### DIFF
--- a/tests/models/soil/test_uptake.py
+++ b/tests/models/soil/test_uptake.py
@@ -184,6 +184,54 @@ def test_calculate_maximum_uptake_rates(
             )
 
 
+def test_calculate_maximum_uptake_rates_negative_values(
+    dummy_carbon_data, environmental_factors, averaged_soil_temp, functional_groups
+):
+    """Test that calculate_maximum_uptake_rates handles negative values sensibly."""
+    from virtual_ecosystem.models.soil.uptake import calculate_maximum_uptake_rates
+
+    lmwc_values = dummy_carbon_data["soil_cnp_pool_lmwc"].sel(element="C").to_numpy()
+    don_values = dummy_carbon_data["soil_cnp_pool_lmwc"].sel(element="N").to_numpy()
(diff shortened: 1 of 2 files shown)


vgro (approved):
LGTM. The only point to double check is if the model is happy with zero, or if you need a small non-zero value.

jacobcook1995:
> The only point to double check is if the model is happy with zero, or if you need a small non-zero value.

That's a good point. Thinking about it, this shouldn't be a problem in this case as the rates get used additively and multiplicatively in downstream code

PR #1619, opened 2026-05-21
Stop array resources from overwriting the data object litter pools

### LINKED ISSUE #1617: The array resources implementation overwrites densities pools
Currently, the array resource initialisation inadvertently overwrites the resource pool it is drawing data from, which then screws everything downstream

```python
mass_data = self.data[self.resource.pool_array]
if self.density:
      mass_data *= self.data.grid.cell_area
```

I should replace this with a deep copy (need to check this function is repeatedly called, in which case I need to do something more complex than replacing with a deep copy)

jacobcook1995:
# Description

Basically title.

@davidorme can you check this fix looks sane/efficient to you (it does stop the Maliau data from stalling at timestep 15, it does fix my immediate issue)

Fixes #1617 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/animal/array_resources.py
+++ b/virtual_ecosystem/models/animal/array_resources.py
@@ -339,11 +339,8 @@ class ResourcePool:
         the local array tracking consumed total biomass.
         """
 
-        # Needs to collapse down to a single mass and element ratio per cell and to
-        # convert to mass units in the density case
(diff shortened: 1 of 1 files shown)


davidorme (commented):
LGTM

davidorme (approved):

PR #1629, opened 2026-05-27
1626 handle canopy layer heights below surface layer

### LINKED ISSUE #1626: Handle canopy layer heights below surface layer
If the canopy height falls below the surface layer, the mixing layer thickness becomes negative and the canopy temperature calculation crashes.

vgro:
This PR keep the model from crashing in the full setup (with `abiotic`) when the plants shrink. There is still an issue with the correct selection of indices - the air temperature exists in all initial canopy layers after the canopy becomes `nan` in some, but this needs some more digging and checking the order of models to make sure all layer values are provided at the right time.

Fixes # 1626

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/abiotic/abiotic_tools.py
+++ b/virtual_ecosystem/models/abiotic/abiotic_tools.py
@@ -377,16 +377,20 @@ def update_profile_from_reference(
 
 
 def calculate_atmospheric_layer_geometry(
-    data: Data, idx: SimpleNamespace, lowest_canopy_layer_correction: float
+    data: Data, idx: SimpleNamespace, minimum_mixing_depth: float
 ) -> dict[str, NDArray[np.floating]]:
     """Calculate heights, layer thickness, and midpoints for atmospheric layers.
 
     The midpoint values are distances in metres above ground for each cell.
+    For layer heights below the surface layer, a minimum mixing depth is introduced.
+    This is to prevent unrealistically low mixing depths and therefore high temperatures
+    in the lowest canopy layer when the layer height is very low. Note that this is an
+    artificial inflation of the mixing depth.
 
     Args:
         data: Data object
         idx: SimpleNamespace containing layer indices
-        lowest_canopy_layer_correction: Correction factor for lowest canopy layer
+        minimum_mixing_depth: Minimum depth for lowest canopy layer
 
     Returns:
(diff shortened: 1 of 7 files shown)


virtual_ecosystem/models/abiotic/abiotic_tools.py
```diff
@@ -377,16 +377,20 @@ def update_profile_from_reference(
...
+    For layer heights below the surface layer, a minimum mixing depth is introduced.
+    This is to prevent unrealistically low mixing depths and therefore high temperatures
+    in the lowest canopy layer when the layer height is very low. Note that this is an
+    artificial inflation of the mixing depth.
 
     Args:
         data: Data object
         idx: SimpleNamespace containing layer indices
-        lowest_canopy_layer_correction: Correction factor for lowest canopy layer
+        minimum_mixing_depth: Minimum depth for lowest canopy layer
```
jacobcook1995: what's the unit of this?

jacobcook1995 (approved):
LGTM (just had a small docstring suggestion)

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro:
```suggestion
        minimum_mixing_depth: Minimum depth for lowest canopy layer, [m]
```

virtual_ecosystem/models/abiotic/model_config.py
```diff
@@ -191,9 +191,16 @@ class AbioticConstants(AbioticSharedConstants):
...
+    minimum_mixing_depth: float = 1.5
+    """Minimum mixing depth for lowest canopy layer, [m].
+    
+    This is to prevent unrealistically low mixing depths and therefore high temperatures
+    in the lowest canopy layer when the layer height is very low. Note that this is an
+    artificial inflation of the mixing depth.
+    """
+
+    min_leaf_area_index_for_mixing: float = 0.5
+    """Minimum leaf area index required for turbulent mixing to occur, dimensionless."""
```
vgro:
```suggestion
    """Minimum leaf area index required for turbulent mixing to occur, [m m-1]."""
```

virtual_ecosystem/models/abiotic/abiotic_tools.py
vgro: It's meters, I added it

PR #1631, opened 2026-05-28
Animal - Resource Pool Exporter

### LINKED ISSUE #1627: Animal - Resource Pool Exporter
The resource pool classes that are foraged on by animals are not actually exported by anyone. 

An exporter is needed that follows the structure of the animal cohort exporter, dumping the resource data into its own csv file.

TaranRallings:
# Description

This PR introduces an exporter for the resource pool objects that are found inside the animal model. The structure of the exporter closely follows the existing exporter for animal cohort data. 

New things:
- the `ResourcePoolDataExporter` class
- the `ResourcePoolExportConfig` class
- a new on/off config option for the animal model

Things that are true:
- structurally, a bunch of this is redundant and could (LATER) be collapsed into a single reusable structure both in terms of exporting specific classes and with the other cohort data exporter
- a bunch of the exported data is also exported in some fashion as part of the continuous data
- we discussed it and this is a workable solution to the animal data team's needs right now

Fixes #1627

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/animal/exporter.py
+++ b/virtual_ecosystem/models/animal/exporter.py
@@ -19,7 +19,17 @@ import pandas as pd
 from virtual_ecosystem.core.exceptions import ConfigurationError
 from virtual_ecosystem.core.logger import LOGGER
 from virtual_ecosystem.models.animal.animal_cohorts import AnimalCohort
-from virtual_ecosystem.models.animal.model_config import AnimalExportConfig
+from virtual_ecosystem.models.animal.array_resources import ResourcePool
+from virtual_ecosystem.models.animal.decay import (
+    CarcassPool,
+    ExcrementPool,
+    FungalFruitPool,
+    SoilPool,
+)
+from virtual_ecosystem.models.animal.model_config import (
+    AnimalExportConfig,
+    ResourcePoolExportConfig,
+)
 
 
 class AnimalCohortDataExporter:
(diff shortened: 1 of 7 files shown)


virtual_ecosystem/models/animal/exporter.py
```diff
@@ -433,3 +443,374 @@ def _build_trophic_rows(
...
+    @classmethod
+    def from_config(
+        cls,
+        output_directory: Path,
+        config: ResourcePoolExportConfig,
+    ) -> ResourcePoolDataExporter:
+        """Create an exporter from a ResourcePoolExportConfig instance.
+
+        If the config has ``enabled=False``, returns an inactive exporter that
+        silently no-ops on all ``dump`` calls.
```
jacobcook1995, line 506: what does "no-ops" mean?

virtual_ecosystem/models/animal/exporter.py
```diff
@@ -433,3 +443,374 @@ def _build_trophic_rows(
...
+        rows = []
+        for pool in resource_pools:
+            pool_name = pool.resource.pool_array
+            pft = pool.pft or ""
+            for cell_id, (C, N, P) in enumerate(pool.elemental_masses):
+                rows.append(
+                    {
+                        "time": time,
+                        "time_index": time_index,
+                        "pool_type": "plant_array",
```
jacobcook1995: Don't think "plant_array" is the best name as this already also covers litter (and will be expanded to cover other resources with time)

jacobcook1995 (commented):
Looks like a sensible approach to the soil/litter pools to me. I'm not approving as I don't feel sufficiently familiar with how the animal exporter works to say whether this is a sensible extension or not

virtual_ecosystem/models/animal/exporter.py
davidorme: Clear to me 😆 🤣 

But yes, could be clearer...

davidorme (approved):
LGTM 

I'd argue that this could be part of the existing CSV exporter for animals, rather than a separate class (in the same way that plants exports multiple things from one class), but that is something we can finesse later.

virtual_ecosystem/models/animal/exporter.py
TaranRallings: changed it to resource_array

virtual_ecosystem/models/animal/exporter.py
TaranRallings: Haha, fair, it just means "no operations"

PR #1636, opened 2026-05-29
1623 parsing data variable units to include or exclude element

### LINKED ISSUE #1623: Parsing data variable units to include or exclude element
Hi there, I would like to discuss the possibility of changing the way data variables' units are entered to facilitate parsing on the data science side. I will focus on units with elements (i.e., C, N and P) in them.

Currently, we have `kg C m^-3`. I want it to be `kg m^-3` for a more generic unit conversion in R / Python. Otherwise, I will need separate conversions from a non-canonical unit to `kg C m^-3`, `kg N m^-3` and `kg P m^-3`, even when they are the same operation.

I can remove the elements using spaces and/or the character "C", "N", and "P", but this does not feel safe because we also have something like `kg N m^-2 day^-1`. I mean, there must be a safe way of parsing this, but I would like to discuss a more explicit option, for example with curly braces `kg{C} m^-3` following the [Unified Code for Units of Measure](https://ucum.org/ucum). My colleagues and I do this for a trait database. 

However, this entails a lot of work to the data variables TOML. What do you think?

hrlai:
# Description

This PR changes the unit of measurements in the soil and litter modules to follow:
- Temperature units are spelled in full and consistently as Celsius or Kelvin; this follows `scipy.constants.convert_temperature()`'s argument options. Spelling them in full makes it easier to detect or parse the units in downstream analyses. I decided to forgo "degree" in Celsius for brevity.
- Explicit element in mass units now uses the curly bracket comment convention following [UCUM](https://ucum.org/ucum), e.g., kg{C}, kg{N}, kg{P} and kg{lignin C}. This also minimises parsing error. So generally, the idea is to use curly brackets for custom units like 1{apple}, 136{dinosaur} etc.

As I went through the scripts, I feel that it is better to not break a line of text about units into separate lines. Keeping them in the same line makes it easier to Ctrl + F to find them.

Fixes #1623

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/data_variables.toml
+++ b/virtual_ecosystem/data_variables.toml
@@ -616,7 +616,7 @@ variable_type = "float"
 
 [[variable]]
 axis = ["spatial"]
-description = "Soil low molecular weight carbon pool. Carbon, nitrogen and phosphorous content"
+description = "Soil low molecular weight organic pool. Carbon, nitrogen and phosphorous content"
 name = "soil_cnp_pool_lmwc"
 unit = "kg m^-3"
 variable_type = "float"
@@ -632,28 +632,28 @@ variable_type = "float"
 axis = ["spatial"]
 description = "Soil bacterial biomass (carbon) pool"
 name = "soil_c_pool_bacteria"
-unit = "kg C m^-3"
+unit = "kg{C} m^-3"
 variable_type = "float"
 
 [[variable]]
 axis = ["spatial"]
 description = "Soil saprotrophic fungal biomass (carbon) pool"
 name = "soil_c_pool_saprotrophic_fungi"
-unit = "kg C m^-3"
+unit = "kg{C} m^-3"
 variable_type = "float"
 
 [[variable]]
 axis = ["spatial"]
 description = "Soil arbuscular mycorrhizal fungal biomass (carbon) pool"
 name = "soil_c_pool_arbuscular_mycorrhiza"
-unit = "kg C m^-3"
+unit = "kg{C} m^-3"
 variable_type = "float"
 
 [[variable]]
 axis = ["spatial"]
 description = "Soil ectomycorrhizal fungal biomass (carbon) pool"
 name = "soil_c_pool_ectomycorrhiza"
-unit = "kg C m^-3"
+unit = "kg{C} m^-3"
 variable_type = "float"
 
 [[variable]]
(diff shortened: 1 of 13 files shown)


jacobcook1995:
> As I went through the scripts, I feel that it is better to not break a line of text about units into separate lines. Keeping them in the same line makes it easier to Ctrl + F to find them.

The reason for splitting them over lines is the line length linter, which is why the QA/CI is failing now. What you can do instead is push the whole unit to the line below, e.g. 

```python
reference_temperature: The reference temperature of the Arrhenius equation [Celsius]
```

should become

```python
reference_temperature: The reference temperature of the Arrhenius equation
    [Celsius]
```

For function docstrings the new line (`[Celsius]`) needs to be indented otherwise the automatic docstring reader treats it as a new variable (for attribute docstrings e.g. `above_structural` you don't need an indent)

jacobcook1995 (approved):
Thanks for doing this, units look way better now!

(let me know if you need any help fixing the QA tests, I've given some pointers above, but docstring formatting is far from intuitive)

hrlai:
Cool thanks, I also went through Ruff's tooltip explanation and managed to understand the python conventions :)

hrlai:
@jacobcook1995 now the test is stuck at building the docs, probably due to:
```
writing output... [ 99%] virtual_ecosystem/theory/soil/summary
/home/runner/work/virtual_ecosystem/virtual_ecosystem/docs/source/api/models/soil/pools.md:: WARNING: py:class reference target not found: kg{C} m^-3 [ref.class]
writing output... [100%] virtual_ecosystem/theory/underlying_theory
/home/runner/work/virtual_ecosystem/virtual_ecosystem/docs/source/api/models/soil/pools.md:: WARNING: py:class reference target not found: kg{C} m^-3 [ref.class]
```

Any idea?

hrlai:
Ah thanks I saw your edits. Those indents don't get picked up by Ruff?

jacobcook1995:
> @jacobcook1995 now the test is stuck at building the docs, probably due to:
> 
> ```
> writing output... [ 99%] virtual_ecosystem/theory/soil/summary
> /home/runner/work/virtual_ecosystem/virtual_ecosystem/docs/source/api/models/soil/pools.md:: WARNING: py:class reference target not found: kg{C} m^-3 [ref.class]
> writing output... [100%] virtual_ecosystem/theory/underlying_theory
> /home/runner/work/virtual_ecosystem/virtual_ecosystem/docs/source/api/models/soil/pools.md:: WARNING: py:class reference target not found: kg{C} m^-3 [ref.class]
> ```
> 
> Any idea?

This ones a nightmare to track down if you aren't already familiar with the docs build process, so I fixed it for you. (It's basically that in docstring "Args:" blocks if something is on a new line it gets treated as a function argument and the autodoc tries to document it, you need a 4 space indent to prevent this)

jacobcook1995:
> Ah thanks I saw your edits. Those indents don't get picked up by Ruff?

Yeah Ruff doesn't pick this up as a problem, it's only when it hits the documentation build step that the error crops up, which makes it a lot harder to catch/debug

PR #1639, opened 2026-06-04
Animal - Replaced ceiling mortality with stochastic rounding

### LINKED ISSUE #469: Animal - review the use of ceiling for number_dead
In inflict_non_predation_mortality in animal_cohorts.py the ceil function is used to turn the continuous rate function into an integer number of dead individuals. This is a problem because the rate will always be > 0, meaning at least one individual will die every time-step. For large animals this is an unsound assumption. 

Find a better solution than ceil for making that value an integer. For numbers below one, it may make sense to convert it to a probability of one individual dying.

TaranRallings:
The non-predation mortality method was using a `ceil` function with non-integer mortality rates. This produced unreasonably high total mortality for large bodied animals. 

Changes:
- ceiling function has been replaced with a floor function + stochastic rounding
- tests for non_predation_mortality have been updated

Fixes #469 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/animal/animal_cohorts.py
+++ b/virtual_ecosystem/models/animal/animal_cohorts.py
@@ -1680,8 +1680,6 @@ class AnimalCohort:
     ) -> None:
         """Inflict combined background, senescence, and starvation mortalities.
 
-        TODO: Review the use of ceil in number_dead, it fails for large animals.
-
         Args:
             dt: The time passed in the timestep (days).
             carcass_pools: The local carcass pool to which dead individuals go.
@@ -1717,8 +1715,12 @@ class AnimalCohort:
         )  # starvation mortality
         u_t = u_bg + u_se + u_st
 
-        # Calculate the total number of dead individuals
-        number_dead = ceil(pop_size * (1 - exp(-u_t * dt)))
+        # Calculate the total number of dead individuals w/ stochastic rounding
+        expected_dead = pop_size * (1 - exp(-u_t * dt))
+        floor_dead = int(expected_dead)
+        remainder = expected_dead - floor_dead
+        stochastic_extra = 1 if random.random() < remainder else 0
+        number_dead = floor_dead + stochastic_extra
 
         # Remove the dead individuals from the cohort
         self.die_individual(number_dead, carcass_pools)
--- a/virtual_ecosystem/models/animal/animal_cohorts.py
+++ b/virtual_ecosystem/models/animal/animal_cohorts.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import random
 import uuid
 from _collections_abc import Callable, Mapping
-from math import ceil, exp, log, sqrt
+from math import ceil, exp, floor, log, sqrt
 from typing import Literal, TypeVar, cast
 
 from numpy import mean, timedelta64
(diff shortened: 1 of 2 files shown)


virtual_ecosystem/models/animal/animal_cohorts.py
```diff
@@ -1717,8 +1715,12 @@ def inflict_non_predation_mortality(
...
         u_t = u_bg + u_se + u_st
 
-        # Calculate the total number of dead individuals
-        number_dead = ceil(pop_size * (1 - exp(-u_t * dt)))
+        # Calculate the total number of dead individuals w/ stochastic rounding
+        expected_dead = pop_size * (1 - exp(-u_t * dt))
+        floor_dead = int(expected_dead)
+        remainder = expected_dead - floor_dead
+        stochastic_extra = 1 if random.random() < remainder else 0
+        number_dead = floor_dead + stochastic_extra
```
davidorme: This is more efficient:

```suggestion
        # Calculate the total number of dead individuals w/ stochastic rounding
        expected_dead = pop_size * (1 - exp(-u_t * dt))
        number_dead = int(floor(expected_dead + random.random()))
```

See: https://stackoverflow.com/a/40921597

davidorme (changes requested):
I think there's a more efficient solution, but the underlying rationale looks good.

virtual_ecosystem/models/animal/animal_cohorts.py
TaranRallings: Oh nice! Thanks.

davidorme (approved):
👍

nickwctan (approved):
This makes sense. 

It seems like the the stochastic rounding only takes effect when expected dead is <1 with an increase probability of one individual dying when expected dead approaches 0. Does that sound right?

hrlai:
Just curious, couldn't we skip the need to arbitrarily choose ceiling or floor by generating `number_dead` as a random variate from a [Binomial distribution](https://numpy.org/doc/stable/reference/random/generated/numpy.random.binomial.html)?

$\text{number dead} = \text{Binomial}(N = \text{pop size}, p = 1 - \exp(-u_t * dt)) \,,$

where `pop_size` is the trial size and `1 - exp(-u_t * dt)` the probability? Sorry if I missed something and jumping in like this.

This generative process is inherently stochastic, and ensures `number dead` < `pop_size`.

davidorme:
The stochastic rounding ensures that the average realised number dead equals the target expected dead.

```python
import numpy as np
realised_dead = np.floor(6.1 + np.random.random(1000))
realised_dead[0:10]
# array([6., 6., 6., 6., 6., 6., 7., 6., 6., 6.])
realised_dead.mean()
# np.float64(6.105)
```

It's a neat trick - adding a random number in $(0,1)$ means that floor proportionally generates the expectation. If the expected is 6.5 then 50% of random numbers are pushed up above 7. If it is 6.1 then only 10% of random numbers do.

davidorme:
> Just curious, couldn't we skip the need to arbitrarily choose ceiling or floor by generating `number_dead` as a random variate from a [Binomial distribution](https://numpy.org/doc/stable/reference/random/generated/numpy.random.binomial.html)?
> 
> number dead = Binomial ( N = pop size , p = 1 − exp ⁡ ( − u t ∗ d t ) ) , ,
> 
> where `pop_size` is the trial size and `1 - exp(-u_t * dt)` the probability? Sorry if I missed something and jumping in like this.
> 
> This generative process is inherently stochastic, and ensures `number dead` < `pop_size`.

Interesting point, @hrlai. That is what we do in the plants model, now that I think about it!

```python
In [109]: realised_dead = np.random.binomial(n=20, p=0.305, size=1000)

In [110]: realised_dead[0:10]
Out[110]: array([6, 9, 9, 4, 7, 6, 3, 6, 9, 4])

In [111]: realised_dead.mean()
Out[111]: np.float64(6.067)
```

TaranRallings:
> Just curious, couldn't we skip the need to arbitrarily choose ceiling or floor by generating `number_dead` as a random variate 

This is a good idea! I jumped straight to the more complex hybrid without even really considering the most reasonable solution. I'll try it out.

PR #1648, opened 2026-06-09
Handle cells with unusual cohort configuration

### LINKED ISSUE #1645: The  `BiomassTissueABC.get_relative_carbon_biomass_by_pft method` does not handle missing PFTs
The function loops over PFT types to calculate the relative carbon biomass within each PFT for cohorts. But... it loops over the total set of PFT types and so breaks if a PFT is missing from the cell.

davidorme:
# Description

This started with #1645 but blew up to provide a bunch of fixes to handle edge cases in cohort data. There are three issues:

* A cell with _no cohort_ data can get an empty Community, but you can't create a Canopy from no cohorts, so the changes make sure empty communities get created and changes PlantsModel.canopies to `dict[int, Canopy | None]` and then handles `None` in canopy code.
* A cell with an _empty cohort_ (that is a tree diameter but no individuals) can have both a community _and_ Canopy - but the max canopy height is bogus and above canopy should be 2m.
* A cell can not contain all of the PFTs, so needed to take more care when iterating over the flora.

The patched code isn't elegant - we're basically waiting on a better version of `pyrealm.demography` here.

To check what is happening, I've added parameterisation to the `plants_cohort_data` fixture to yield an  alternative set of cohort data with the above edge cases. It's finicky, because that means all tests that access the fixture (mostly indirectly) need to provide the parameterisation even though only one test (`test_PlantsModel_update_canopy_layers`) actually does detailed tests. The `fixture_canopy_layer_data` fixture has also been rewritten to generate the expectations for both standard and tricky cohort data.

Fixes #1645 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/plants/conftest.py
+++ b/tests/models/plants/conftest.py
@@ -58,7 +58,12 @@ def fixture_exporter(tmpdir, fixture_configuration):
 
 @pytest.fixture
 def plants_cohort_data(tricky_plant_cohorts):
-    """Construct a simple initial cohort dataframe."""
+    """Construct a simple initial cohort dataframe.
+
+    The tricky plant cohorts switch can be passed down through tests to switch from
+    a normal set of cohorts and a set with edge cases (no cohorts at all, empty cohorts,
+    only one of 2 PFTs.)
+    """
 
     data = pd.DataFrame(
         {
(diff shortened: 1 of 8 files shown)


sallymatson (approved):
LGTM !

Issue #1649, opened 2026-06-12
Finalising fruit/seed input to litter


jacobcook1995:
At the moment the litter model does not receive any fruit or seed inputs. This is something that should be addressed. I'm not quite sure what links need to be added, so I think this needs to be discussed in person.

(I'm making this issue mainly as a place to record things)

jacobcook1995:
The basic conclusion we reached was that only fruit and seeds should hang around on the ground for animals to eat, all other plant tissues should go straight to the litter model.

N.B. the fruit and seed "pools" are essentially two aspects of a singular propagule pool.

We also decided that the fruit should have some decay rate, so that they don't accumulate if left uneaten. When this decay happens it should be straight to the soil (otherwise effectively fruit decay twice, once as fruit and once as litter).

We didn't think it made sense for seeds to decay. They are either eaten by animals, or they propagate (in which case the plant model should handle the biomass transfers). This means there's no direct role for the soil/litter models to play in this case

arne-exe:
> We didn't think it made sense for seeds to decay. They are either eaten by animals, or they propagate (in which case the plant model should handle the biomass transfers). This means there's no direct role for the soil/litter models to play in this case
>
Thanks for writing these decisions down here. I don't agree with the idea of seeds not decaying though. Dipterocarp seeds are recalcitrant, they lack dormancy and are sensitive to desiccation. They must germinate almost immediately upon dropping onto the forest floor - usually within days or a few weeks - or they quickly lose viability and die. Consequently, they do not form a persistent soil seed bank.

jacobcook1995:
> Thanks for writing these decisions down here. I don't agree with the idea of seeds not decaying though. Dipterocarp seeds are recalcitrant, they lack dormancy and are sensitive to desiccation. They must germinate almost immediately upon dropping onto the forest floor - usually within days or a few weeks - or they quickly lose viability and die. Consequently, they do not form a persistent soil seed bank.

That's interesting. I think the decision is probably better phrased as "the plant recruitment functionality should decide what happens to the biomass in seeds, rather than an arbitrary decay function". It's complex because that recruitment functionality doesn't exist yet, but I feel like the decision needs to be made within the plant model proper rather than in a decay function that I have tacked onto the model

davidorme:
[...]

PR #1656, opened 2026-06-17
Litter flows refactor (sub-canopy, herbivory waste, lignin) 

### LINKED ISSUE #1229: Litter model should use the inputs from the sub-canopy vegetation
The sub-canopy vegetation in the plants model currently creates litter inputs (`subcanopy_vegetation_litter_cnp`) which the litter model doesn't use. This needs to change

### LINKED ISSUE #1361: 'HerbivoryWaste` objects should populate their lignin proportions based on actual plant matter consumed
At present they just use a fixed constant, when really this needs to vary based on the plant matter they have consumed. This fix can only be made once the animal consumption of plants is fully implemented

jacobcook1995:
# Description

This PR was originally intended to just add flows from the sub-canopy vegetation into the litter model.

It's grew in scope though, so it also updates the `herbivory_waste` pools. There were two main changes here

1) The waste pools now populate their lignin proportions based on the consumed biomass (previous there was just a fixed dummy value for this). Doing this necessitated extending the ArrayResources to (optionally) accept an array of lignin values.

2) I realised that having a herbivory waste pool for each type of animal consumed plant matter was going to grow out of controll, particularly as we only really care about chemistry (already tracked by the pools) and strata. So now there's one waste pool for above ground and one for below. (I think it's fair to treat any plant matter that has been chewed up by animals as pretty much the same, elephants knocking down trees doesn't count as herbivory imho) 

I also removed the `foliage_turnover_cnp_consumed` variable as I don't think it makes sense for animals to be able to eat fallen leaves as well as being able to eat litter separately.

(There's still more to sort in terms of fully connecting biomass flows into the litter model, see #1649, but I thought that I had already added enough for one PR)

Fixes #1229 
Fixes #1361 

## Type of change

## Key checklist

## Further checks


TaranRallings (approved):
Unless I'm missing something, the changes to the real animal model code look straightforward. The only thing I had questions about was the foraging change but it's clear what is going on.

jacobcook1995:
> Unless I'm missing something, the changes to the real animal model code look straightforward. The only thing I had questions about was the foraging change but it's clear what is going on.

Yep the change within the foraging code amounts to "when herbivory waste is generated by foraging also pass along lignin information as well as the C:N:P information". So a very minor change, but still thought I should get your sign-off because it's close to the foraging logic

virtual_ecosystem/models/animal/protocols.py
```diff
@@ -54,7 +54,7 @@ def get_eaten(
         self,
         consumed_mass: float,
         consumer: Consumer,
-    ) -> tuple[dict[str, float], dict[str, float]]:
+    ) -> tuple[dict[str, float], dict[str, float], float]:
```
davidorme, line 57: Is there any mileage in making this type a dataclass? Just to centralise definitions?

davidorme: Given it is used in several places?

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -150,27 +173,35 @@ def combine_input_sources(data: Data, update_interval: float) -> dict[str, DataA
         cell_area=data.grid.cell_area,
         update_interval=update_interval,
     )
```
davidorme: Use a loop over the calls to `convert_to_input_masses_to_rates_per_area` - five identical calls - can just swap in variable names and make the code shorter.

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -212,9 +243,36 @@ def calculate_metabolic_proportions_of_input(
...
+        split_sensitivity_phosphorus=constants.metabolic_split_phosphorus_sensitivity,
+    )
+
     return {
         "leaves_meta_split": leaves_metabolic_split,
         "roots_meta_split": roots_metabolic_split,
+        "subcanopy_meta_split": subcanopy_metabolic_split,
+        "herbivore_waste_above_meta_split": herb_waste_above_metabolic_split,
+        "herbivore_waste_below_meta_split": herb_waste_below_metabolic_split,
     }
```
davidorme, line 237:
```suggestion
# Calculate split of each input biomass type

split_variables = (
    ("leaves_meta_split", "leaf_mass", "leaf_lignin"),
    ("roots_meta_split", "root_mass", "root_lignin"),
    ("subcanopy_meta_split", "subcanopy_mass", "subcanopy_lignin"),
    ("herbivore_waste_above_meta_split", "herbivore_waste_above_mass", "herbivore_waste_above_lignin"),
    ("herbivore_waste_below_meta_split", "herbivore_waste_below_mass", "herbivore_waste_below_lignin"),
)

return {var:  split_pool_into_metabolic_and_structural_litter(
    input_masses=total_input[mass],
    lignin_proportion=total_input[lignin],
    max_metabolic_fraction=constants.max_metabolic_fraction_of_input,
    split_sensitivity_nitrogen=constants.metabolic_split_nitrogen_sensitivity,
    split_sensitivity_phosphorus=constants.metabolic_split_phosphorus_sensitivity,
    )
    for var, mass, lignin in split_variables
}
```

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -580,12 +621,51 @@ def calculate_litter_input_nutrient_masses(
...
+        find_nutrient_split_between_litter_pools(
+            input_carbon_rate=litter_inputs.herbivore_waste_below_mass.sel(element="C"),
+            input_nutrient_rate=litter_inputs.herbivore_waste_below_mass.sel(
+                element=element_symbol
+            ),
+            metabolic_split=litter_inputs.herbivore_waste_below_meta_split,
+            meta_to_struct_nutrient_ratio=meta_to_struct_nutrient_ratio,
+        )
+    )
+
```
davidorme, line 592: You could also loop here to avoid five calls - pass in the tuples of like `(litter_inputs.herbivore_waste_below_mass, litter_inputs.herbivore_waste_below_meta_split)`. You would then have to store them in something to unpack them for the summations in the return value, but I think it would be cleaner?

davidorme (changes requested):
Looks good - I haven't looked at the tests in detail, but we need the new structure.

My main issues with the code are that there are a lot of repeated calls with the same structure and I think the code is shorter and more maintainable if those are a single calls with looped inputs. That's more of a style thing than correctness, so feel free to push back!

virtual_ecosystem/models/animal/protocols.py
jacobcook1995: maybe, but I'm hesitant to make that change given that this is a fairly core part of Taran's code. Though maybe we could ask him? @TaranRallings would you be comfortable with this output becoming a data class?

jacobcook1995:
@davidorme thanks for the pointers on removing those repeated calls, very much worth doing as I will need to add more input streams for fruit etc, so doing that now will save work down the line

virtual_ecosystem/models/animal/protocols.py
TaranRallings: I don't have an inherent objection to replacing the protocol with something else. _However_, the resource protocol is used widely in my foraging system. If you're going to change it, you're going to need to come up with a structure that regularly computes mass from stoichiometric mass and come up with a structure for get_eaten, that works across all resources without changing anything on the consumer end. 

Perhaps this goes smoothly but perhaps it doesn't. If the only gain is elegance, perhaps it's better as a tidy-up issue for down the line.

TaranRallings (approved):

virtual_ecosystem/models/animal/protocols.py
davidorme: Yup - I more meant creating an explicit type for the return value so that the protocol returns a clearly identified thing. It makes it easier to structure and follow the code if this returns an `EatenStuff` instance (with all the same properties as before).

But yes, this is could be largely cosmetic, so parking it makes sense.

jacobcook1995: My vote is for `GotEaten`. But agree it's probably best to park (and merge this PR), is it worth me making an issue for this?

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -129,48 +153,48 @@ def combine_input_sources(data: Data, update_interval: float) -> dict[str, DataA
...
+        "subcanopy_mass": rates["subcanopy_vegetation_litter"],
+        "herbivore_waste_above_mass": rates["herbivory_waste_above"],
+        "herbivore_waste_below_mass": rates["herbivory_waste_below"],
+        "leaf_lignin": data["senesced_leaf_lignin"],
+        "root_lignin": data["root_lignin"],
+        "stem_lignin": data["stem_lignin"],
+        "subcanopy_lignin": data["subcanopy_vegetation_litter_lignin"],
+        "herbivore_waste_above_lignin": data["herbivory_waste_above_lignin"],
+        "herbivore_waste_below_lignin": data["herbivory_waste_below_lignin"],
     }
```
davidorme, line 195: I think it would be cleaner (and more auto-extending if other variables get added) to use this approach:

```python
In [4]: a = {"n":1, "o":2}
In [5]: b = {"p":1, "q":2}
In [6]: c = {"r": 1, **a, **b}
In [7]: c
Out[7]: {'r': 1, 'n': 1, 'o': 2, 'p': 1, 'q': 2}
```

So: 

```python
    return {"leaf_mass": leaf_rates, **rates, **data}
```

virtual_ecosystem/models/litter/inputs.py
```diff
@@ -566,26 +571,40 @@ def calculate_litter_input_nutrient_masses(
...
+        meta, struct = find_nutrient_split_between_litter_pools(
+            input_carbon_rate=mass.sel(element="C"),
+            input_nutrient_rate=mass.sel(element=element_symbol),
+            metabolic_split=getattr(litter_inputs, f"{pool_name}_meta_split"),
+            meta_to_struct_nutrient_ratio=meta_to_struct_nutrient_ratio,
+        )
+        nutrient_splits[pool_name] = {
+            "meta": meta,
+            "struct": struct,
+        }
```
davidorme, line 591: Hmmm. I was going to complain about not using an explicit statement of the attributes, but it is unwieldy, isn't it 😄 I doubt the `getattr` calls add much to runtime...

```python
    split_variables = [
        (
            "leaf",
            litter_inputs.leaf_mass,
            litter_inputs.leaf_mass_meta_split,
        ),
        (
            "root",
            litter_inputs.root_mass,
            litter_inputs.root_mass_meta_split,
        ),
        (
            "subcanopy",
            litter_inputs.subcanopy_mass,
            litter_inputs.subcanopy_mass_meta_split,
        ),
        (
            "herbivore_waste_above",
            litter_inputs.herbivore_waste_above_mass,
            litter_inputs.herbivore_waste_above_mass_meta_split,
        ),
        (
            "herbivore_waste_below",
            litter_inputs.herbivore_waste_below_mass,
            litter_inputs.herbivore_waste_below_mass_meta_split,
        ),
    ]
    nutrient_splits = {}

    for pool, mass, metabolic_split in split_variables:
        meta, struct = find_nutrient_split_between_litter_pools(
            input_carbon_rate=mass.sel(element="C"),
            input_nutrient_rate=mass.sel(element=element_symbol),
            metabolic_split=metabolic_split,
            meta_to_struct_nutrient_ratio=meta_to_struct_nutrient_ratio,
        )
        nutrient_splits[pool] = {
            "meta": meta,
            "struct": struct,
        }

```

davidorme (approved):
Looks good - one minor suggestion and one "find out that my way is clunkier".

virtual_ecosystem/models/litter/inputs.py
jacobcook1995: > Hmmm. I was going to complain about not using an explicit statement of the attributes, but it is unwieldy, isn't it 😄 I doubt the `getattr` calls add much to runtime...

Yeah this is a function called once per simulation time step and the `getattr` returns all cells at once, so I think it makes more sense to prioritise clean code here (happy to revisit though if profiling flags this as a problem though)

virtual_ecosystem/models/litter/inputs.py
jacobcook1995: I can't do that with `data` (or at least it will need extra steps) because it is the entire data object, but that's a good point about the rates dictionary

Issue #1634, opened 2026-05-28
`abiotic` integration test is breaking


sallymatson:
**Describe the bug**
The integration test does not run with the `abiotic` - only `abiotic_simple`. The `abiotic` test is commented out right now but this should be undone when it is fixed. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Additional context**

PR #1667, opened 2026-06-29
Change output format to Zarr

### LINKED ISSUE #1558: More efficient way to merge continuous output data
@jacobcook1995 and I discussed the speed at which VE merges the continuous data files after a simulation has completed. On the `maliau_2` dataset with 20x20 grids and 132 time steps, merging the 132 continuous data files takes a very long time at tens of minutes, putting it at the same order of magnitude of the simulation runtime itself --- a very significant time spent when the number of time steps is large. This is an expected result of `xarray.open_mfdataset` handling so many netCDFs.

@kirkImperial would you be able to help us get a sense of the performance of this step in the profiling? Correct me if wrong, the corresponding code should be https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/a27237c387e10396bedf5bd11010cd5a725146d2/virtual_ecosystem/main.py#L484-L492
and `merge_continuous_data_files()` uses `open_mfdataset()` in https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/a27237c387e10396bedf5bd11010cd5a725146d2/virtual_ecosystem/core/data.py#L550-L556

Thanks!

### LINKED ISSUE #1665: Move initial state into Zarr store
At the moment, we write out the initial state to a NetCDF file and then (if #1558 is fixed and merged) the remaining data is written to a Zarr store (which is better suited to continuous data output).

It would be much neater if all of the data was written to a single output, but the short term goals of #1558 are to avoid data compilation time costs during a simulation - and that is mostly about appending outputs, and this is trickier.

The main issue is that there are three classes of variable that might be stored:

* Forcing data, loaded as a time series to run the model and which have 0..T values along the `time_index` axis.
* Variables created during initialisation and which therefore have two values "at time step 0": the initialised value and then the value at the end of update 0 and so have length T+1.
* Update variables, which have one value for each time step 0..T

These last two types are (currently) not held with a time index in the Data object - this is appended as they are written out. If we use `time_index` then the single time slot variables get broadcast to the full length of the time series, so there's a need for something like  a separate `time_step` index that can take a value (-1?) for the init step and then fill in 0..T. However this also means padding out the initial -1 time step for Update variables.

The code below is an attempt to get something as a model, but it doesn't work - the variables end up with inconsistent lengths on the axes.

```python
import xarray
import numpy as np

initial_state = xarray.Dataset(
    data_vars={
        "forcing_var_1": (("cell_id", "time_index"), np.ones((9, 20))),
        "forcing_var_2": (("cell_id", "time_index"), np.ones((9, 20))),
        "init_var_1": (("cell_id",), np.ones(9)),
        "init_var_2": (("cell_id",), np.ones(9)),
    },
    coords={"cell_id": np.arange(9), "time_index": np.arange(20)},
)

for v, k in initial_state.data_vars.items():
    if "time_index" not in k.coords:
        initial_state[v] = (
            initial_state[v].expand_dims({"time_step": 1}).assign_coords(time_step=[0])
        )

initial_state.to_zarr("testzarr2.zarr", consolidated=False, zarr_format=2, mode="w")


pre_update_fill = xarray.Dataset(
    data_vars={
        "update_var_1": (("cell_id",), np.full(9, fill_value=np.nan)),
        "update_var_2": (("cell_id",), np.full(9, fill_value=np.nan)),
    },
    coords={"cell_id": np.arange(9), "time_index": np.arange(20)},
)

for v, k in pre_update_fill.data_vars.items():
    if "time_index" not in k.coords:
        pre_update_fill[v] = (
            pre_update_fill[v]
            .expand_dims({"time_step": 1})
            .assign_coords(time_step=[0])
        )

pre_update_fill.to_zarr(
    "testzarr2.zarr", append_dim="time_step", consolidated=False, zarr_format=2
)


update_state = xarray.Dataset(
    data_vars={
        "forcing_var_1": (("cell_id", "time_index"), np.ones((9, 20))),
        "forcing_var_2": (("cell_id", "time_index"), np.ones((9, 20))),
        "init_var_1": (("cell_id",), np.ones(9)),
        "init_var_2": (("cell_id",), np.ones(9)),
        "update_var_1": (("cell_id",), np.ones(9)),
        "update_var_2": (("cell_id",), np.ones(9)),
    },
    coords={"cell_id": np.arange(9), "time_index": np.arange(20)},
)

for v, k in update_state.data_vars.items():
    if "time_index" not in k.coords:
        update_state[v] = (
            update_state[v].expand_dims({"time_step": 1}).assign_coords(time_step=[1])
        )


update_state[
    [
        "init_var_1",
        "init_var_2",
        "update_var_1",
        "update_var_2",
    ]
].to_zarr("testzarr2.zarr", append_dim="time_step", consolidated=False, zarr_format=2)

d = xarray.load_dataset("testzarr2.zarr")
```

davidorme:
# Description

This PR:

* Changes the model output data from a sequence of per-time step NetCDF files to a single Zarr format store that is appended to at each time step. The simplifies the model workflow when saving data after model updates: the Zarr format allows the update variables to be appended to the existing store along the time axis (#1558). This has a couple of immediate advantages:
    1. We can drop the compilation of per-update NetCDF files into a single file at the end of the simulation, which is extremely slow and memory hungry.
    2. If a simulation crashes, the output directory contains a single Zarr store that contains all the variable updates up to the crash, rather than leaving a pile of unconsolidated `continuous_data_000N.nc` files lying around.

* Actually implements setting `variables_to_save` in the configuration - this was present in downstream code but never actually handled in `ve_run` so all variables were always written.
* Updates saving model initial states (forcing data and post `__init__` variables) into the same Zarr store, removing the need for a separate `initial_state` NetCDF file (#1665). This isn't completely simple because the data dimensions are not trivial to align during the simulation. For this reason, Zarr "groups" (basically subdirectories) are used to partition the output data:
    * the forcing data is complete across timesteps from the start, whereas we want model updates to be appended, so forcing data is written to the `inputs` data group.
    * variables created during `__init__` have a confusing time identity: they are sort of at timestep 0, but at the beginning of it, not the end. Most are then updated and saved as an output at timestep 0. To keep them separate, the values of these variables after initialisation are written to the `init` group. These values are less likely to be used, and to align them with the rest of the output data, we'd need to create an extra `time_index` (e.g. -1) to prepend them onto the model outputs. I think it is clearer to keep these values in a separate group to make it explicit that they differ from output data.
    * Updated variables are written to the `outputs` group.

* Simplifies the model data output configuration to simply:

```toml
out_path = "output/dir/path"
output_data_file_name = "model_data.zarr"
save_compiled_configuration = true
compiled_configuration_file_name = "compiled_configuration.toml"
variables_to_save = []
```

Fixes #1558
Fixes #1665 

## Type of change

## Key checklist

## Further checks


davidorme:
@dalonsoa Could you have a look at this and see if anything sticks out? I haven't updated the docs in case there are major changes needed.

I don't get why the Windows runners are failing. Windows can be picky about path names in TOML (needs raw strings in single quotes), but these are being generated using: `out_path = shared_datadir / "test_output.zarr"`. I can't see what Windows would object to particularly.

The zarr specific tests fail with:

```
=========================== short test summary info ===========================
FAILED tests/core/test_data.py::test_save_to_zarr[False-vars] - PermissionError: [Errno 13] Permission denied: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\popen-gw1\\test_save_to_zarr_False_vars_0\\data\\test_output.zarr'
```

virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -763,28 +763,20 @@ def calculate_initial_symbiotic_supply(self) -> dict[str, DataArray]:
...
-                0.0,
-            ),
-            "arbuscular_mycorrhizal_p_supply": where(
-                DataArray(initial_arbuscular_p) >= 0.0,
-                self.to_total_mass(initial_arbuscular_p),
-                0.0,
-            ),
-        }
+        # TODO - this could just write to self.data rather than returning a dict
+        var_dict = {}
```
davidorme, line 767: This is fixing an issue with an unnamed dimension (`dim_0` -> `cell_id`) in these variables.

davidorme:
@hrlai Can you have a look at this PR - essentially there is now just a single Zarr store for data coming out of the simulation. I think it may be possible to get a slightly cleaner internal structure (try and get rid of the internal groups), but if the transition to Zarr works OK, then that is a smaller issue later.

annarallings:
So the input data will still be NetCDF (where applicable), but all the outputs that were NetCDF will now be zarr, correct? Can you recommend any packages and tools for viewing and manipulating zarr in R and Python? Would you be able to share a example output file package to the team to play with while you're sorting out the VE update?

davidorme:
> So the input data will still be NetCDF (where applicable),

Yup - although in practice `xarray` can handle other formats so we could use something else (like zarr)

> but all the outputs that were NetCDF will now be zarr, correct? 

Yup

> Can you recommend any packages and tools for viewing and manipulating zarr in R 

@hrlai has looked into this: `stars` and `pizarr` look promising.

> and Python?

The `xarray` package that we already use handles them just fine. The only thing that changes is users need to set (e.g.) `xr.open_dataset(..., group="outputs")` to access the different data groups.

> Would you be able to share a example output file package to the team to play with while you're sorting out the VE update? 

Yup - good plan!

hrlai:
Thanks @davidorme , will test run this branch asap. 

>  we'd need to create an extra time_index (e.g. -1) to prepend them 

Nice. This has been how I index the initial data as well when manually pasting them with the continuous data.

> So the input data will still be NetCDF (where applicable),

Good point @annarallings ... So maybe a less confusing way to demonstrate VE for the wider user base (in R) is to simply convert Zarr outputs to netCDF and then stick with netCDF thereof, but I'll think about it more.

dalonsoa:
This looks like agreat move - and a lot of work, to be honest. It seems that using Zarr brings a lot of advantages, at least from the point of view of the software. Having said that, I had never heard of it before this PR, while NetCDF is widely known in multiple disciplines. I wonder if end users might not know how to handle or used the output files for further processing and analysis. When updating the documentation, I think it will be useful to include a section on handling the output files in some common scenarios.

> I don't get why the Windows runners are failing. Windows can be picky about path names in TOML (needs raw strings in single quotes), but these are being generated using: `out_path = shared_datadir / "test_output.zarr"`. I can't see what Windows would object to particularly.
> 
> The zarr specific tests fail with:
> 
> ```
> =========================== short test summary info ===========================
> FAILED tests/core/test_data.py::test_save_to_zarr[False-vars] - PermissionError: [Errno 13] Permission denied: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-runneradmin\\pytest-0\\popen-gw1\\test_save_to_zarr_False_vars_0\\data\\test_output.zarr'
> ```

No idea. I seem to remember we have found this on other occasions, but I do not know how we solved it. Has anyone using Windows reproduced it locally? If not yet, I think it will be useful to see what happens. 

I've also found [that issue in another package](https://github.com/openai/codex/issues/19791). No solution, yet, but it seems you're not alone. Maybe try some of the solutions that did not work for them, just in case they work for you. In particular, not using `pytest-xdist`.

hrlai:
@davidorme I wasn't able to fully test this out on the example data because I'm getting an error:
```
ValueError: Soil model integration cannot proceed because the following variables contain invalid values (e.g. NaN or Inf): {'soil_temperature'}
```

I have not seen this before, and I'm not sure where this PR branched off initially so hard for me to tell why.

I am testing it out on the data science with a [bash script](https://github.com/ImperialCollegeLondon/ve_data_science/blob/adf5baad188b27669c9c935ea83e6fdd1900da41/analysis/troubleshoot/zarr_test/setup.sh), which may help reproduce the problem.

davidorme:
> ```
> ValueError: Soil model integration cannot proceed because the following variables contain invalid values (e.g. NaN or Inf): {'soil_temperature'}
> ```
> 
> I have not seen this before, and I'm not sure where this PR branched off initially so hard for me to tell why.

I've been seeing the same thing - you should get a single Zarr store with the updates until the crash, but yup we need to fix this to help evaluate.

virtual_ecosystem/core/data.py
```diff
@@ -391,80 +377,87 @@ def load_data_config(self, config: CoreConfiguration) -> None:
...
-        # Save and close new dataset
-        time_slice.to_netcdf(Path(output_file_path))
-        time_slice.close()
+        # Save the variables to the zarr store, appending along time index after the
+        # first time step. Zarr format 2 is used here because format 3 doesn't currently
+        # handle fixed length strings, such as the PFT coords.
+        #
+        # TODO - will need to do something cleverer if we aren't writing all time steps
+        #        and potentially skipping zero. Create a flag on data that records if
+        #        any data has been written by this method
```
dalonsoa, line 444: Is not writing data every time step a use case? Just curious.

virtual_ecosystem/main.py
```diff
@@ -380,38 +403,36 @@ def ve_run(
...
-    # Container to store paths to continuous data files
-
-    continuous_data_files = []
+    # Identify which variables should be saved to the different zarr store groups:
+    # - the 'inputs' group contains variables provided to the model
+    # - the 'init' group contains variable states after model initialisation.
+    # - the 'outputs' group contains the model values at each time step
+
+    # Identify variable groups to save
+    data_vars_to_save = [
```
dalonsoa: Maybe rename this to match the groups above:

```suggestion
    input_vars_to_save = [
```

virtual_ecosystem/main.py
```diff
@@ -380,38 +403,36 @@ def ve_run(
...
+        for k in variables_to_save
+        if runtime_variables[k].vars_populated_by_init == ["data"]
+    ]
+    init_vars_to_save = [
+        k
+        for k in variables_to_save
+        if runtime_variables[k].vars_populated_by_init
+        and runtime_variables[k].vars_populated_by_init != ["data"]
+    ]
+    update_vars_to_save = [
```
dalonsoa: Same here:

```suggestion
    output_vars_to_save = [
```

dalonsoa (approved):
I finally reviewed it and it looks really good. The code is much simpler and cleaner.

I have a couple of comments, but my only concern is what I mentioned before: how many people really knows about zarr format and how to handle it. Docs associated with this will need to be really good to help the end user.

virtual_ecosystem/core/data.py
davidorme: @dalonsoa Yup - at the moment we are writing data for every step as the only option. We might want to take less common snapshots - in which case we might not be able to use time_index == 0 as our diagnostic on when to create vs append.

davidorme:
* I've fixed the windows bug - we've bought a Windows laptop at home and I've immediately installed the dev environment on it 😄 It's an odd thing - `xarray` tries to read a file to check what kind of data `model_data.zarr` is, and for some reason that fails. If I explicitly set `engine = "zarr"` in the tests, then all is good. 🤷 

* I've updated the docs that loaded data (in the Using the VE section). I did swap back to the `abiotic_simple` model because of the current #1634

I'm going to check in with the data science team state of readiness for this change before merging to `develop`.

PR #1671, opened 2026-07-01
1670 replace vertical mixing limits

### LINKED ISSUE #1670: replace vertical mixing limits
The limits for mixing are current set to a suboptimal value which does not represet canopy conditions.

vgro:
Resolve old TODO with simplified limits and defaults. This needs to be calibrated properly, but for now we get much closer to the observed SAFE data (black dots, OGF only) with the troubleshooting setup. This will hopefully identify bugs across all model.

The drift is interesting, it could be a leak in the energy balance or a spinup problem. Tomorrows investigation...

image

Fixes #1670 

## Type of change

## Key checklist

## Further checks


sallymatson (approved):
LGTM!

PR #1682, opened 2026-07-13
1677 animal   ectotherm activity window placeholders

### LINKED ISSUE #1677: Animal - Ectotherm activity window placeholders
The ectotherm activity window system uses placeholder annual values for mean temperature and standard deviation. These are then used to form the optimum temperature of the ectotherm cohorts. The problem is that these values produce zero activity windows for many/all of the ectothermic cohorts. 

We need a new placeholder value and it may need to be parsed by vertical occupancy.

TaranRallings:
# Description

This PR addresses two problems with the activity window system for ectotherms. Firstly, that temperature was not being averaged over vertical strata for cohorts living over multiple strata. Secondly, that the placeholder reference temperature mean and standard deviations being used were out of step with the example climate being used, leading to no activity for ectotherms. 

This PR:
- adds a vertical strata averaging system for the temperature experienced by cohorts
- changes the placeholder values to work at init with the example data
- adds temperature and activity window data to the exporter

We are still going to need a proper addition of these means and sd's from the abiotic model but this should get us going to now. Temperature is still unstable in the model and ectotherms will quickly stop foraging after the first time step. 

Fixes #1677 

## Type of change

## Key checklist

## Further checks


davidorme (approved):
Looks ok to me. I vaguely worry about the ramifying tree like structure of the animals config, but that's not unique to this PR 😄.

PR #1686, opened 2026-07-14
Animal - Euclidean Dispersal

### LINKED ISSUE #1683: Animal - Euclidean Dispersal
The dispersal system needs to be updated to run on euclidean distance.

TaranRallings:
# Description

The thermal dispersal and territory system is going to come as three PRs, this is the first.

This PR fixes some unit conversion bugs in the dispersal system and changes the dispersal system to run on euclidean distance. It looks like a lot of code but it isn't actually changing very much.

Changes:
- fixed unit bug
- renamed some methods for a more general use case
- new 'cells_within_distance' method
- updated testing

Fixes #1683  (issue)

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/scaling_functions.py
```diff
@@ -627,27 +627,125 @@ def H_i_j(
...
+    return speed_km_month * 1000.0 * (dt_days / 30.0)
+
+
+def cells_within_distance(
+    centroid_key: int,
+    distance_m: float,
+    cell_side: float,
+    cell_nx: int,
+    cell_ny: int,
+) -> list[int]:
```
davidorme, line 668: I think you can replace this entire function using the `Grid` object. It should be much faster.


```suggestion
def cells_within_distance(
    grid: Grid,
    centroid_key: int,
    distance_m: float,
) -> list[int]:

    # Pin small distances to the centroid to centroid distance
    distance_m = max(distance_m, np.sqrt(grid.cell_area)
    
    # Find reachable cells
    reachable = np.where(grid._distances[centroid_key, :] < distance_m)[0].tolist()
    
    # Remove centroid from list 
    reachable.remove(centroid_key)
    
    return reachable

```

davidorme (changes requested):
Looks sensible, but I think we can use the `Grid` object to speed up the process.

virtual_ecosystem/models/animal/scaling_functions.py
TaranRallings: So I guess the question here is which problem do we prefer:
1) populating the distances matrix and carrying that in memory for the simulation
2) using get_distances for each dispersal event and losing time to it

Is populate_distances already called somewhere?

davidorme: Honestly don't know if it is called already. I think having all the distances calculated once at startup is going to save a huge amount of time. We'll worry about the scaling issue later...

TaranRallings: Fair enough! I've put it in the animal model setup.

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -303,6 +304,9 @@ def _setup(
 
         self.model_constants = model_constants
 
+        # initialize grid distances for use by dispersal methods
+        self.data.grid.populate_distances()
+
```
davidorme: Honestly, I'd put this in the main setup between these two lines:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/05c125fc0cbf30a17daebdf1af709076dddeb8d1/virtual_ecosystem/main.py#L331-L332

As:

`core_components.grid.populate_distances()`

virtual_ecosystem/models/animal/scaling_functions.py
```diff
@@ -627,27 +628,77 @@ def H_i_j(
...
+    # Clamp to at least one cell side. Uses <= below, matching set_neighbours, so the
+    # orthogonal neighbours at exactly sqrt(cell_area) are retained.
+    distance_m = max(distance_m, np.sqrt(grid.cell_area))
+
+    # get_distances returns shape (1, n_cells); the reachable cell ids are the column
+    # indices of the single row, hence [1].
+    reachable = np.where(grid.get_distances(centroid_key, None) <= distance_m)[1]
+
+    # The centroid's self-distance is zero, so drop it from the destination set.
+    return reachable[reachable != centroid_key].tolist()
```
davidorme: Honestly, I would stick with the original suggestion:

```python
    # Find reachable cells
    reachable = np.where(grid._distances[centroid_key, :] < distance_m)[0].tolist()
    
    # Remove centroid from list 
    reachable.remove(centroid_key)
    
    return reachable
```

I see why you might use `get_distances`, but it's a more general function and the only advantage for this use case is that it copes when  `_distances` isn't populated. But if we are explicitly populating it, then all `get_distances` does is complicate taking a 1D slice out of the array.  `list.remove()` is also 3x faster than running the numpy logical subset.


```
In [27]: 
    ...: import numpy as np
    ...: import timeit
    ...: 
    ...: focal_cell = 10
    ...: within_distance = np.arange(20)
    ...: 
    ...: def get_cells_1(focal_cell, within_distance):
    ...: 
    ...:     w = within_distance.tolist()
    ...:     w.remove(focal_cell)
    ...:     return w
    ...: 
    ...: def get_cells_2(focal_cell, within_distance):
    ...: 
    ...:     return within_distance[within_distance != focal_cell].tolist()
    ...: 
    ...: assert (
    ...:     get_cells_1(focal_cell, within_distance) == get_cells_2(focal_cell, within_distance)
    ...: )
    ...: 
    ...: globals = dict(
    ...: focal_cell = focal_cell,
    ...: within_distance = within_distance,
    ...: get_cells_1 = get_cells_1,
    ...: get_cells_2=get_cells_2
    ...: )
    ...: timeit.timeit("get_cells_1(focal_cell, within_distance)", globals=globals)
    ...: 
Out[27]: 0.2896722920704633

In [28]: timeit.timeit("get_cells_2(focal_cell, within_distance)", globals=globals)
Out[28]: 1.065800875192508

**Review by davidorme — CHANGES_REQUESTED** [review]
Couple of quick things:

File: virtual_ecosystem/models/animal/scaling_functions.py (hunk 3 above)
Comment (davidorme) [reply virtual_ecosystem/models/animal/scaling_functions.py]: Although admittedly I don't know how that scales, but I'd expect it to be faster across the board and the number of cells is likely to be small much of the time.

Comment (TaranRallings) [reply virtual_ecosystem/models/animal/scaling_functions.py]: I'm generally fine with this but I need to make a couple of small changes - clamping to minimum one cell move and doing <= instead of a strict less than. Whether to move is covered elsewhere, so no-move would be a kind of negative double-counting and I need the = for edge cases.

Comment (davidorme) [reply virtual_ecosystem/models/animal/scaling_functions.py]: Yup - I'd handled the minimum one cell in `distance_m` but changing the `np.where` to `<=` makes sense. 

Having said though, using the _exact_ centroid to centroid distance is a bit of a hostage to floating point errors, so I might be tempted to add a 1 mm buffer 😄 

```python
def cells_within_distance(
    grid: Grid,
    centroid_key: int,
    distance_m: float,
) -> list[int]:
    # Pin small distances to the centroid to centroid distance
    distance_m = max(distance_m, np.sqrt(grid.cell_area) + 0.001)  # 1 mm greater to avoid floating point issues
    
    # Find reachable cells
    reachable = np.where(grid._distances[centroid_key, :] <= distance_m)[0].tolist()
    
    # Remove centroid from list 
    reachable.remove(centroid_key)
    
    return reachable
```

davidorme (approved):

PR #1687, opened 2026-07-15
Update traits system to pyrealm 3

### LINKED ISSUE #1685: Update traits to new pyrealm 3.0.0
The `pyrealm` package has had a substantial revision of the `demography` module (mostly aimed at improving workflows in VE!).

The aim is to try and bring this in bit by bit. The first step is to get rid of `ExtraPFTTraits` and replace it with the extensible trait class now implemented in `pyrealm`.

davidorme:
# Description

This PR starts to move the PlantsModel over to pyrealm 3.

* It updates some usage of pyrealm functions where names have changed.
* It creates the new `VEFlora` class, which is an extended version of the `pyrealm` `Flora` class containing all of the extra traits that were previously in `ExtraTraitsPFT`. It also adds some VE specific calculated classes: the reference values for `tau_f` and `lai` and the fruit flesh fraction. 
* It adds a temporary `Community` class that replaces the `pyrealm` `Community` class:
    * Eventually I think we want to drop Plant Communities and Community, but this is a half way house to limit the scope of this PR.
    * The new `Community` has the same attributes as the old `Community` _except_ that the `stem_traits` attribute no longer exists, because the PFT traits from `stem_traits` are now fields in the `Community.cohorts` dataframe. (I could have recreated a `stem_traits` attribute, but it would probably be more code than just going with the new structure).
    * As a result, all references to `extra_pft_traits` and `stem_traits` can be replaced with `cohorts["attr_name"]` (and should generally be used with `cohorts["attr_name"].to_numpy()` to avoid subbing in `pd.Series` for `np.ndarray` values.
* It removes `ExtraTraitsPFT` completely across the whole model.
* It adds a new cohort id generator to the model attributes.
* It fixes up changes in location of `stem_traits` -> `cohorts` and `delta_xxx` from `StemAllocation` to `GrowthIncrements`
* More serious rewiring of `allocate_gpp` to move fruit and symbiote C supply between allocation to NPP and calculation of growth increments.

One thing that is currently hack-adjacent is that we passed information into Biomasses using using StemAllocation and StemAllometry. Those no longer have fruit attributes, but rather than engineer in a whole new argument and pathway I have stapled fruit data attributes back onto objects. Apparently `mypy` doesn't care about declaring new attributes on classes. I think we can do this more explicitly and cleanly by subclassing those pyrealm classes and adding documented new slots that are only ever written to externally, but I mostly just want to get this one over the line.

Fixes #1685 

## Type of change

- [x] New feature (**Breaking** change which adds functionality)

## Key checklist

## Further checks


virtual_ecosystem/models/plants/biomasses.py
```diff
@@ -318,32 +313,23 @@ class FoliageBiomass(BiomassTissueABC):
...
-                    for name in pft_names
-                ]
-            )
+            ideal_ratio = community.cohorts[
+                f"foliage_c_{elem.lower()}_ratio"
+            ].to_numpy()
+
+            turnover_ratio = community.cohorts[
+                f"leaf_turnover_c_{elem.lower()}_ratio"
+            ].to_numpy()
```
sallymatson, line 339: much nicer

sallymatson (commented):
This all looks great to me so far. I think it is a lot cleaner this way!

sallymatson (commented):
From what I can tell, this is looking good. Bit hard to do a complete pick through, but nothing stood out to me. Thanks David!

sallymatson (approved):
From what I can tell, this is looking good. Bit hard to do a complete pick through, but nothing stood out to me. Thanks David!

PR #1688, opened 2026-07-15
1657 documentation needs to make clear that the soil model is a topsoil model

### LINKED ISSUE #1514: LayerStructure documentation figure does not match default setup
The vertical structure figure in the docs has a different number of layers to the actual implementation which is a bit confusing see [here](https://teams.microsoft.com/l/message/19:meeting_NjU4ZjNiODktNjFmOS00MDZmLWE2OTktYWE3MDJlZDMwY2Q0@thread.v2/1776327535851?context=%7B%22contextType%22%3A%22chat%22%7D) .

The total number of layer needs to be 14, divided into the following:

- 1 above
- 10 canopy
- 1 surface
- 1 topsoil
- 1 subsoil

I think @davidorme generated the figure in the first place, could you update this or share the file so I or @jacobcook1995 can do it please?

### LINKED ISSUE #1657: Documentation needs to make clear that the soil model is a topsoil model
Basically the title. Also should reframe the "maximum depth of microbial activity" as soil scientists do not like the implication that microbial activity halts below a certain depth

I don't think there's any way we move the soil model away from being a topsoil model in the short to medium future (doing so would add a huge number of hard to estimate parameters). But rebranding it makes the assumptions and limitations of the model clearer

jacobcook1995:
# Description

This PR adds content to the documentation talking about how the soil model is in reality a topsoil model, why this is justified (most microbial activity and thus carbon turnover is near the surface). It also removes references to a "maximum depth of microbial activity" (very much a loaded term), and tries to explain why we use two separate definitions of topsoil (so that you don't have to completely change the hydrology model parameters every time you change what the soil model considers topsoil)

The docs are built [here](https://virtual-ecosystem.readthedocs.io/en/1657-documentation-needs-to-make-clear-that-the-soil-model-is-a-topsoil-model/)

@hrlai asking you to review this as you are probably the most familiar with the soil model docs in general, let me know if it all makes sense

@vgro asking you to review as there's clear overlap with the abiotic models here, so wanted to check you agree with how I've described the situation

Fixes #1657 
Fixes #1514

## Type of change

## Key checklist

- [] Make sure you've run the `pre-commit` checks: `$ pre-commit run -a`

## Further checks


docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
 fit into any of the cases mentioned already, and so it will be discussed in a separate
 section.
 
+## Finding soil and litter layer temperatures
+
+Before we discuss the different classes of environmental response, we need to mention
+how the environment itself is determined. The temperatures and soil waster potentials
```
vgro:
```suggestion
how the environment itself is determined. The temperatures and soil water potentials
```

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+definitions that do not correspond to the depth of the biotically active topsoil.
+
+:::{note}
+
+The Virtual Ecosystem uses two different definitions of "topsoil", the uppermost layer
+of the hydrology model and the entire simulated depth of the soil model. This allows the
+depth of soil that is considered to be (microbially active) topsoil to be altered by
+users, without forcing them to update the entire setup of the hydrology model to match.
+In places in the documentation where both definitions of topsoil are pertinent the soil
+model specific definition of topsoil is referred to as the "biotic topsoil".
```
vgro: I find this a bit confusing. Would there be a way around the 'topsoil', maybe 'biotic soil zone'? That would keep it clearly distinct from the model layer approach

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+For the above-ground litter, things are pretty straightforward as the air temperature of
+the layer immediately above the soil surface is just used. (As this litter is
+above-ground, soil water potential is not relevant).
+
+The below-ground litter is assumed to be spread out evenly over the topsoil, so the
+process for calculating the relevant soil temperatures is identical to that of the
+(biotic) topsoil (which we discuss below).
+
+The (biotic) topsoil is simulated as a single stratum (of user configurable) depth. This
+stratum uses a homogeneous environment, e.g. a single temperature is used for its entire
```
vgro: here you could also use 'zone' ?

tests/core/test_core_components.py
```diff
@@ -251,16 +257,16 @@ def test_CoreComponents(config, expected_layers, expected_timing, expected_const
             (
                 (
                     ERROR,
-                    "Maximum depth of soil layers is less than the maximum depth "
-                    "of microbial activity",
+                    "Maximum depth of soil layers is less than the biological topsoil "
```
vgro: If you use the topsoil term, it would be good to be consistent. You sometime use 'biological' and then 'biotic'

virtual_ecosystem/core/core_components.py
```diff
@@ -185,13 +185,13 @@ class LayerStructure:
...
 
         1. The ``active_soil`` role indicates soil layers that fall even partially above
-           the configured `max_depth_of_microbial_activity`. The `soil_layer_thickness`
-           attribute provides the thickness of each soil layer - including both top- and
-           sub-soil layers - and the `soil_layer_active_thickness` records the thickness
-           of biologically active soil within each layer. Note that the ``soil_layers``
-           provides the sequence of depths of soil horizons relative to the surface and
-           these values provide the thickness of individual layers: the default
-           ``soil_layers`` values of ``[-0.25, -1.00]`` give thickness values of
+           the configured `biotic_topsoil_depth`. The `soil_layer_thickness` attribute
```
vgro: In the bigger picture, I am not sure how much sense the specific naming of layers has; the topsoil is clear but if we had multiple soil layers, what would the subsoil layers be named? maybe we can use first, second soil layer?

vgro (changes requested):
I generally like the direction, but I am not sure this is much clearer. Although you explained it well, maybe a more distinct name would help? I think this is a general problem with the naming of the sub-components and what different people imagine they contain (e.g. the abiotic model should be microclimate, the soil does not actually do soil conditions, etc).

Also a minor comment on the figure; the surface layer is also a flux layer.

docs/source/virtual_ecosystem/theory/soil/summary.md
```diff
@@ -54,6 +54,17 @@ soils. Explicitly represented microbes presents major data challenges, as soils
 generally poorly characterised compared to the other constituent parts of terrestrial
 ecosystems.
 
+:::{important}
+
+The soil model is a topsoil model, and does not represent subsoil processes. Our
+justification for doing this is that the topsoil is far more microbially active, so the
+majority of changes in total soil carbon storage are due to changes in this layer. While
+changes in subsoil absolutely do effect total soil carbon storage, we did not believe
```
hrlai:
```suggestion
changes in subsoil absolutely do affect total soil carbon storage, we did not believe
```

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
hrlai: Instead of zone, maybe just consistently "layer".

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
hrlai: maybe just "layer"

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+## Finding soil and litter layer temperatures
+
+Before we discuss the different classes of environmental response, we need to mention
+how the environment itself is determined. The temperatures and soil waster potentials
+are calculated using the abiotic and hydrology models which use their own layer
+definitions that do not correspond to the depth of the biotically active topsoil.
+
+:::{note}
+
+The Virtual Ecosystem uses two different definitions of "topsoil", the uppermost layer
```
hrlai: Can you refer to `docs/source/_static/images/layer_structure.svg` here?

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+of the hydrology model and the entire simulated depth of the soil model. This allows the
+depth of soil that is considered to be (microbially active) topsoil to be altered by
+users, without forcing them to update the entire setup of the hydrology model to match.
+In places in the documentation where both definitions of topsoil are pertinent the soil
```
hrlai:
```suggestion
In places in the documentation where both definitions of topsoil are pertinent, the soil
```

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+
+:::
+
+For the above-ground litter, things are pretty straightforward as the air temperature of
+the layer immediately above the soil surface is just used. (As this litter is
+above-ground, soil water potential is not relevant).
+
+The below-ground litter is assumed to be spread out evenly over the topsoil, so the
+process for calculating the relevant soil temperatures is identical to that of the
+(biotic) topsoil (which we discuss below).
```
hrlai: I find "(biotic) topsoil" still a bit confusing, is it the BS layer in the layer figure?

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+
+The (biotic) topsoil is simulated as a single stratum (of user configurable) depth. This
+stratum uses a homogeneous environment, e.g. a single temperature is used for its entire
+extent. The environmental variables for this stratum are calculated as follows:
+
+$$
+T = \sum^N_{i=1} f_i *\frac{V_{i-1} + V_{i}}{2}
+$$
+
+where $N$ is the number of soil layers, $f_i$ is the fraction of soil layer $i$ that
```
hrlai, line 75: The number of soil layers, is it 2, or 3?

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+lies within the (biotic) topsoil, and $V_{i}$ is the value of the environmental variable
+of interest at depth $i$. No values are generated for "layer zero", so for each variable
```
hrlai: I don't quite get the "layer zero" sentence here.

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+we need to choose a value for it. For temperature, we chose $T_0$ to be the air
```
hrlai:
```suggestion
we need to choose a value for it. For temperature, we chose layer-zero temperature $T_0$ to be the air
```

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,49 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+temperature of the layer immediately above the soil surface. For soil water potential,
+we set $\psi_0 = \psi_1$, which amounts to assuming that the soil water potential does
```
hrlai:
```suggestion
we set layer-zero soil water potential to be the same as that of the first layer, $\psi_0 = \psi_1$, to assume that the soil water potential does
```

hrlai (commented):
I can get the gist of it, but some small parts still lose me. Perhaps more linking back to the layer figure would help!

virtual_ecosystem/core/core_components.py
jacobcook1995: I was wondering about that, because if e.g. the first soil layer is 1cm, then the model is basically saying that subsoil starts at 1cm of depth. It doesn't matter massively, because there aren't any big mechanical differences between the layers (I assume), but it's still a weird claim

(I don't think this is something we can fix now, I'll try to think of a better name for the "active soil zone" or whatever)

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
jacobcook1995: "layer" has a specific meaning in the model layer structure (which is used by abiotic, plants and hydrology), litter and soil sit on top of that structure but don't use the exact layers. So I can't really use the word layer as it introduces confusion

jacobcook1995: > I find this a bit confusing. Would there be a way around the 'topsoil', maybe 'biotic soil zone'? That would keep it clearly distinct from the model layer approach

Completely agree that I need a different name, issue I have with "biotic soil zone" is kind of the same as I had with "maximum depth of microbial activity", in that it sounds like we are making a strong claim about where there is microbial activity in the soil, rather than choosing to only simulate a subset of the total activity (because we can't simulate it all and the upper soil makes the biggest contribution)

I was thinking of going with something like "microbial simulation depth" instead, as that explains the situation more clearly (we are making a decision not to simulate below a certain depth)

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
jacobcook1995: going to use "zone" as it isn't already in use for another purpose ("layer" is used for `abiotic/hydrology/plants`, "stratum" is used by `animal`)

hrlai: Actually zone could work since we do say "root zone".

jacobcook1995:
Okay I've tried to clarify the terminology by calling it "microbial_simulation_depth" (which is what it fundamentally is), hopefully that works better!

@vgro I've also corrected the error you mention in the layer structure figure

@hrlai thanks for the feedback on the layer 0 stuff, looking at that again I think it only made sense to people who've spent too long starting at the VE layer structure diagram 🫠 I've tried to add more explanation about how the averaging of layers works, let me know if that makes more sense!

vgro (approved):
This is much clearer, great job!

docs/source/virtual_ecosystem/theory/soil/environmental_links.md
```diff
@@ -38,6 +38,50 @@ is the the rate at which nutrients leach from the soil. As such, this process do
...
+For the above-ground litter, things are pretty straightforward as the air temperature of
+the layer immediately above the soil surface is just used. (As this litter is
+above-ground, soil water potential is not relevant).
+
+The below-ground litter is assumed to be spread out evenly over the simulated depth of
+the soil-microbial model, so the process for calculating the relevant soil temperatures
+is identical to that of the soil microbes (which we discuss below).
+
+The zone the soil-microbial model is simulated for is of user configurable depth and is
+assumed to be homogeneous. This means that for each environmental variable and average
```
hrlai:
```suggestion
assumed to be homogeneous. This means that for each environmental variable an average
```

hrlai (approved):
Thanks, much clearer now! All good except a potential typo. 😄

PR #1694, opened 2026-07-17
Changing typo on functional_group_data.md

nickwctan:
Just found a tiny typo on the readthedocs. Please let me know if it was intentional.


jacobcook1995 (approved):
Yep that's definitely a typo on my part 🫠 

Thanks for spotting + fixing it!

PR #1733, opened 2026-07-23
Example fertilisation addition disturbance

### LINKED ISSUE #1693: example fertilisation disturbance
As a demonstration of the disturbance model approach, I should make a fertilisation disturbance. This is primarily to demonstrate/test what can be done using the disturbance model. So, I'll start with a relatively simple fertilisation regime, we can then expand on this existing disturbance if we need more complex fertilisation regimes for future applications

jacobcook1995:
# Description

This PR adds a basic disturbance to simulate the addition of inorganic nitrogen fertiliser (relevant for plantation scenarios).

The main point of doing this is to test the process of making a disturbance, so @dalonsoa if you could let me know whether or not my usage seems correct/sensible that would be great!

Fixes #1693 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/disturbances/add_fertiliser/add_fertiliser_model.py
```diff
@@ -0,0 +1,137 @@
...
+from virtual_ecosystem.disturbances.add_fertiliser.model_config import (
+    AddFertiliserConfiguration,
+    AddFertiliserConstants,
+)
+
+
+class AddFertiliserModel(
+    BaseDisturbance,
+    model_name="add_fertiliser",
+    disturbed_models=("soil",),
```
dalonsoa: You are not really disturbing the `soil` model directly, only variables in the `data` object that are used by the `soil` model, so this argument should be an empty tuple:

```suggestion
    disturbed_models=(),
```

dalonsoa (approved):
Yes, almost right. See my comment below. 

Nice to see disturbances in action!

virtual_ecosystem/disturbances/add_fertiliser/add_fertiliser_model.py
jacobcook1995: ahh right! what would disturbing the `soil` model directly entail?

jacobcook1995: would that be basically running the disturbance model instead of the `soil` model? Or altering (non-`data` object)  parts of the model state?

dalonsoa: Changing the model state, essentially. Anything that the model needs in order to run and that is not kept in the data object.

PR #1736, opened 2026-07-23
Brief user guide to setting up disturbance models

### LINKED ISSUE #1734: Document user usage of disturbances
There currently isn't any user documentation of how to run disturbances.  I've created an example disturbance (see #1733), once that has been approved I should write user documentation on how to run it (I may have got the correct usage wrong so I should wait until it is signed off/merged)

jacobcook1995:
# Description

This PR adds a [brief user guide](https://virtual-ecosystem.readthedocs.io/en/1734-document-user-usage-of-disturbances/using_the_ve/modifying_a_model.html#configuring-disturbances) to setting up disturbance models. There's documentation already in the API, the aim is really just to have something in the user documentation that explains how to include disturbances via the config system.

(it also fixes a small bug I noticed on the last PR, but forgot to push before I merged the branch to main 🫠)

@dalonsoa, if you could confirm that the information is correct that would be great!

Otherwise I've tagged a bunch of people as reviewers and I'm basically looking to see if people think it's in a sane location in the user docs and fits with the general style (no need for everyone to weigh in)

Fixes #1734

## Type of change

## Key checklist

## Further checks


docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -172,3 +172,43 @@ ve_run \
...
+increases the concentration of inorganic nitrogen in the relevant soil pools, in order
+to mimic the impacts of repeated fertiliser addition in agricultural landscapes.
+
+To setup a disturbance, you need to include it within a configuration file. You then
+need to provide any specific configuration settings that the disturbance requires. In
+the case of the fertiliser addition disturbance, these are the total amount of inorganic
+nitrogen added (per disturbance event) and the fraction of this addition that is nitrate
+(rather than ammonium).
+
+The final thing you need to decide is how frequently the disturbance should occur. Each
```
dalonsoa: This bit on the timing is common to all disturbance models and I feel that the format and meaning of the different options might be lost in the text. I'd suggest to clearly identify this par onwards as a sub section (###, three levels) and use bullet points to describe the options.

dalonsoa (changes requested):
This looks very nice. I've left a comment, but feel free to ignore.

In many tutorials, the example is shown first and then the different bits and pieces it is made of explained. I think that format works well, but again feel free to ignore if doesn't align with how it is done in other parts of the docs.

Something that is not mentioned is the 'priority' of the disturbance, that defaults to zero. Disturbance models will run in order of priority and no two disturbance models (in a specific simulation) can have the same priority, so as soon as a user includes two disturbances, they will get an error if they have not changed the default priority of one of them. That should be explained clearly here. It is the only other option that is common to all disturbances, alongside the timing.

docs/source/using_the_ve/modifying_a_model.md
davidorme: Yeah - I agree. I'd pop this out into a whole new page. That could start with the general settings at the top (and show the root config as TOML as we have on the model config pages). Then it can give the specific example of the fertiliser disturbance and again use the TOML to make the settings clear. We should use the auto-documentation where possible.

davidorme: Also - I'm torn on the fertiliser disturbance setting a default frequency and so locally overriding `run_every`. Different disturbances probably do have different natural frequencies, but it is a lot easier to document if we just make the root timing settings mandatory with no defaults and then all disturbances explicitly _have_ to set them.

jacobcook1995:
@dalonsoa, thanks for the feedback! I think your suggested structure is a significant improvement, so I have tried to implement it.

I didn't initially include anything on `priority` because I wasn't sure how it worked. I've added in the explanation you suggested, let me know if that seems sufficient or if I should add more detail

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -172,3 +172,67 @@ ve_run \
...
+  `run_every = [start]`, e.g. `run_every = [2]` runs the disturbance for every timestep
+  from the second timestep onwards.
+* Run the disturbance at a set frequency after a specified initial time step using
+  `run_every = [start, step]`, e.g. `run_every = [2, 3]` first runs the disturbance at
+  the second timestep and runs the disturbance once every three timesteps from then on.
+* Run the disturbance at a set frequency after a specified initial time step until a cut
+  off point using `run_every = [start, step, stop]`, e.g. `run_every = [2, 3,
+  13]` first runs the disturbance at the second timestep and from then runs the
+  disturbance once every three timesteps up until the 13th timestep beyond which the
+  disturbance doesn't occur. N.B. a disturbance only occur on the "final" timestep if
```
vgro: What is N.B.? The sentence that follow is a bit convoluted, would it be 'occur**s**'? 'if would be expected to based on' seems like a few words got mixed up :-)

vgro (approved):
I like it! I only have a small comment on one sentence that is a bit unclear

docs/source/using_the_ve/modifying_a_model.md
jacobcook1995: My understanding (@dalonsoa correct me if I am wrong) is that a default (either for `run_at` or `run_every`) has to be set for every disturbance model. If one isn't then the following error gets emitted from `build_configuration_model` "Value error, Timing options 'run_at' and 'run_every' cannot be both ()." This occurs even if `run_at` or `run_every` is included as part of the configuration.

I'm not sure if that is intended behaviour or a bug though?

davidorme: One of the two has to be supplied when creating a config _instance_, but not when defining the subclass. We should probably update that error message to "Provide one of `run_at` or `run_every` in your disturbance configuration" (and we should probably check that one of the two is empty).

davidorme: I also think there's an argument for setting a class attribute so that config error messages can be located to the disturbance? Something like:

```
class DisturbanceConfigurationRoot(Configuration):
    """Root configuration class for disturbance Virtual Ecosystem models.
    ...
    """
   disturbance_name: ClassVar[str]
```

And hence:

```
class AddFertiliserConfiguration(DisturbanceConfigurationRoot):
    """Configuration class for the fertiliser addition disturbance."""

    disturbance_name: ClassVar[str] = "add_fertiliser"
    
    constants: AddFertiliserConstants = AddFertiliserConstants()
    """Constants values for the fertiliser addition disturbance."""
```

Annoyingly, it looks like the `__init_subclass__` pattern we've used elsewhere to set class variables is not available on `pydantic` at present: https://github.com/pydantic/pydantic/issues/6024

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -172,3 +172,67 @@ ve_run \
...
+disturbance, so are generally use case specific. In the example above, the total amount
+of inorganic nitrogen added (per disturbance event) and the fraction of this addition
+that is nitrate (rather than ammonium) both need to be specified by the user.
+
+### Disturbance timing
+
+Users also need to specify how often a disturbance will occur. There are five different
+options for this:
+
+* Use the default behaviour for the disturbance by not adding any timing details to the
```
dalonsoa:
```suggestion
* Use the default behaviour for the disturbance, if any, by not adding any timing details to the
```

docs/source/using_the_ve/modifying_a_model.md
```diff
@@ -172,3 +172,67 @@ ve_run \
...
+* Run the disturbance at a set frequency after a specified initial time step using
+  `run_every = [start, step]`, e.g. `run_every = [2, 3]` first runs the disturbance at
+  the second timestep and runs the disturbance once every three timesteps from then on.
+* Run the disturbance at a set frequency after a specified initial time step until a cut
+  off point using `run_every = [start, step, stop]`, e.g. `run_every = [2, 3,
+  13]` first runs the disturbance at the second timestep and from then runs the
+  disturbance once every three timesteps up until the 13th timestep beyond which the
+  disturbance doesn't occur. N.B. a disturbance only occur on the "final" timestep if
+  would be expected to based on the step size, so in the previous example the last
+  disturbance occurs on the 11th timestep.
```
dalonsoa: I'd add at the bottom:


```suggestion
  disturbance occurs on the 11th timestep.
  
Mind that if the disturbance does not implement a default timing, you will
need to explicitly set it in the configuration file using one of the options indicated above. Failing to do so will result in an error being raised. 
```

docs/source/using_the_ve/modifying_a_model.md
dalonsoa: Yep. Don't know what N.B. is, either, and the sentence is not fully clear to me, either.

dalonsoa (approved):
I've left a couple of comments, but it looks good to me.

I do not see a fundamental problem with specific disturbances setting a default timing, as long as it is properly documented and it is a somewhat reasonable value. But up to you what you want to do.

docs/source/using_the_ve/modifying_a_model.md
dalonsoa: > One of the two has to be supplied when creating a config _instance_, but not when defining the subclass. We should probably update that error message to "Provide one of `run_at` or `run_every` in your disturbance configuration" (and we should probably check that one of the two is empty).

I'm not following this. You cannot create a disturbance configuration instance where both timing options are empty (in particular, both are `()` which is the default of the `DisturbanceConfigurationRoot`). If a subclass define a non-empty value for one of them, that's OK - the user won't need to provide a value for it in the config file. If it does not, that is also OK, but the user will need to provide it.

What it is not there is a check to ensure that both are NOT provided at the same time. As it is now, [`run_at` is prioritised](https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/00128ab8b05d29052311193f0c558da22d91337e/virtual_ecosystem/core/core_components.py#L648), silently ignoring `run_every` if provided, which is probably not Ok. There should be a warning, at least, when not a complete failure.

But that's work for a separate PR.

About the `disturbance_name`, I feel is redundant because you can always get such a name out of the class name as we are enforcing a specific naming convention for the configuration and model classes. Something like:

```python
@property
def disturbance_name(self):
    snake_case = re.sub(r'(?<!^)(?=[A-Z])', '_', self.__class__.__name__)
    return snake_case.lower()
```

davidorme: > I'm not following this. You cannot create a disturbance configuration instance where both timing options are empty (in particular, both are `()` which is the default of the `DisturbanceConfigurationRoot`).

Sorry - wasn't clear. This is exactly what I meant - we don't need to code a default into a specific disturbance and the root class enforces that at least one is set.

jacobcook1995: Okay, based on how you have described the system @dalonsoa, I'm fairly sure the requirement to define a default is a bug. I've filed a bug report here #1748.

I'm going to go ahead and finish off these docs now, as I don't think it makes sense to fix that bug as part of a docs PR

docs/source/using_the_ve/modifying_a_model.md
jacobcook1995: it's Nota Bene, but yeah probably best to avoid abbreviated latin (expect the really commonly used ones like "e.g." or "i.e."). I'll remove it and try to make the sentence clearer

Issue #1742, opened 2026-07-27
Update docs for new air temperature update


vgro:
The update of air temperature is now integrated in the secant method for canopy temperature which is running until both canopy and air temperatures converge #1531 . This needs to be updated in the documentation.

PR #1743, opened 2026-07-27
1531 integrate air temperature update

### LINKED ISSUE #1531: integrate air temperature update

vgro:
This PR integrates the update of air temperature in the secant method for canopy temperature which is running until both canopy and air temperatures converge.

I also replaced some random tests with existing fixtures.

Fixes #1531 

## Type of change

## Key checklist

## Further checks

- [ ] TODO Relevant documentation reviewed and updated (separate issue #1742)

### DIFF
--- a/virtual_ecosystem/models/abiotic/energy_balance.py
+++ b/virtual_ecosystem/models/abiotic/energy_balance.py
@@ -934,13 +934,11 @@ def solve_canopy_temperature_with_air_coupling(
         tuple of canopy temperature, air temperature, and final fluxes
     """
 
-    canopy_temperature = state["canopy_temperature"].copy()
-    air_temperature = state["air_temperature"].copy()
+    state_local = state.copy()
+    air_temperature = state_local["air_temperature"]
+    canopy_temperature = state_local["canopy_temperature"]
 
     for _ in range(maxiter_air):
-        state_local = state.copy()
-        state_local["air_temperature"] = air_temperature
-
         residual_function = make_canopy_residual(
             state=state_local,
             static=static,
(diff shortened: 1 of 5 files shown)


jacobcook1995 (approved):
The added code is well documented and explained. The for loop within `solve_canopy_temperature_with_air_coupling` seems like it will be a pretty large processing step, but I couldn't see any easy wins related to it.

vgro:
> The added code is well documented and explained. The for loop within `solve_canopy_temperature_with_air_coupling` seems like it will be a pretty large processing step, but I couldn't see any easy wins related to it.

There might be a way of using a solver but i think for now that should at least be better than the previous version

virtual_ecosystem/models/abiotic/energy_balance.py
```diff
@@ -878,16 +881,151 @@ def residual(canopy_temperature):
...
+
+    for _ in range(maxiter_air):
+        residual_function = make_canopy_residual(
+            state=state_local,
+            static=static,
+            aerodynamic_resistance=state_local["aerodynamic_resistance_canopy"],
+            abiotic_constants=abiotic_constants,
+            core_constants=core_constants,
+            idx=idx,
+        )
```
davidorme, line 958: @vgro Late to the party here, but what do you want to happen when the algorithm does not converge within the number of iterations? Because at the moment it returns normally with the last guess. That might be intentional (but needs documenting) but if not then should we raise on convergence failure?

```suggestion
    canopy_change = air_change = np.inf
    iter_number = 0

    while  max(canopy_change, air_change) >= air_temperature_tolerance::
    
        iter_number += 1
        if iter_number > maxiter_air:
            raise RuntimeError("Canopy temperature solver did not converge")
           
        residual_function = make_canopy_residual(
            state=state_local,
            static=static,
            aerodynamic_resistance=state_local["aerodynamic_resistance_canopy"],
            abiotic_constants=abiotic_constants,
            core_constants=core_constants,
            idx=idx,
        )
```

vgro: Good point. I decided to log an info instead of raising an error as discussed earlier, as part of #1742 . I also added a bit more info about the cells that are affected

PR #1749, opened 2026-07-29
Fix creating instance of config

### LINKED ISSUE #1748: Disturbances can't be setup unless a default timing option is defined
**Describe the bug**
If you try to setup a disturbance model without giving a default value for `run_at` *or* `run_every` the simulation fails during configuration loading

**To Reproduce**

1. Go to 'disturbances/add_fertiliser/model_config.py'
2. Comment out the default timing setting (which is `run_every: tuple[int, ...] = (0,)`)
3. Run `ve_run` with the following config file included:
```toml
[disturbance.add_fertiliser]
run_at=[0,1]
```
```log
* Starting Virtual Ecosystem simulation using v0.2.1.
* Logging to: out/ve_run.log
* Loading configuration
  File "/Users/jc2017/Library/Caches/pypoetry/virtualenvs/virtual-ecosystem-TdEglf9z-py3.14/bin/ve_run", line 6, in <module>
    sys.exit(ve_run_cli())
             ~~~~~~~~~~^^
  File "/Users/jc2017/Repositories/virtual_ecosystem/virtual_ecosystem/entry_points.py", line 332, in ve_run_cli
    ve_run(
    ~~~~~~^
        cfg_paths=args.cfg_paths,
        ^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<4 lines>...
        progress=progress,
        ^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/jc2017/Repositories/virtual_ecosystem/virtual_ecosystem/main.py", line 302, in ve_run
    configuration: CompiledConfiguration = generate_configuration(
                                           ~~~~~~~~~~~~~~~~~~~~~~^
        config_data.data, context={"cli_paths": cli_paths}
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/jc2017/Repositories/virtual_ecosystem/virtual_ecosystem/core/config_builder.py", line 621, in generate_configuration
    ConfigurationModel = build_configuration_model(
        requested_modules=requested_modules,
        requested_disturbances=list(data.get("disturbance", {}).keys()),
    )
  File "/Users/jc2017/Repositories/virtual_ecosystem/virtual_ecosystem/core/config_builder.py", line 565, in build_configuration_model
    **{fname: (cname, cname()) for fname, cname in subdisturbance_models},
                      ~~~~~^^
  File "/Users/jc2017/Library/Caches/pypoetry/virtualenvs/virtual-ecosystem-TdEglf9z-py3.14/lib/python3.14/site-packages/pydantic/main.py", line 263, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for AddFertiliserConfiguration
  Value error, Timing options 'run_at' and 'run_every' cannot be both (). [type=value_error, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/value_error
```

You can see the same bug in the `tests/disturbances/test_add_fertiliser.py::test_generate_add_fertiliser_disturbance` test, where the `id=run_at_config` should still pass (as it provides a valid timing option) when the default option is commented out but it fails with a similar message as above

@davidorme might have more to add as he looked into this bug in a bit more detail

dalonsoa:
# Description

Uses `Field` to define the default values of a configuration, rather than the actual default value.

Fixes #1748

## Type of change

## Key checklist

## Further checks


dalonsoa:
It seems that tests are failing because the core configuration used in test does not have `grid`, which is a bit weird. I do not think I've made any change affecting that, so no idea why it is failing.

dalonsoa:
Sorted!

tests/disturbances/add_fertiliser/test_add_fertiliser_model.py
```diff
@@ -12,17 +12,17 @@
     "cfg_string,nitrate_fraction",
     [
         pytest.param(
-            "[disturbance.add_fertiliser]\n",
+            "[core]\n[disturbance.add_fertiliser]\n",
```
jacobcook1995, line 15: I think it might be a good idea to remove the default timing setting from the `AddFertiliser` disturbance (I only added a default because I thought one was required) and then rework these tests so that they all provide timing settings. As that would test that disturbances can be setup without defaults set

jacobcook1995: I guess alternatively you could delete the default `run_at` option from `DisturbanceTestingModel`, as that would achieve the same purpose

jacobcook1995 (approved):
Just had a small suggestion about testing, imagine @davidorme will have more thoughts about the fix itself

dalonsoa:
In the absence of any further comment from @davidorme , I'm going to implement @jacobcook1995 suggestions and merge this one, if that's all-right.

jacobcook1995 (approved):
LGTM!

davidorme (approved):
Sorry for the slow response - meant to do it before going on leave.

I don't fully understand the difference between using `cname()` directly and creating a new field (correct ?) using `Field(cname)` but if this solves the problem then I'm happy.

dalonsoa:
`cname()` create and instance of the configuration right there with no input arguments, at that time, while `Field(cname)` postpone creating the instance with the right input arguments when the whole configuration is created.

davidorme:
> `cname()` create and instance of the configuration right there with no input arguments, at that time, while `Field(cname)` postpone creating the instance with the right input arguments when the whole configuration is created.

Thanks for clarifying, @dalonsoa. Really helpful.

Issue #1744, opened 2026-07-27
PlantsModel - moving to clear cohort / cell calculations


davidorme:
The current implementation nests cohort data within communities, but actually most calculations can be done in parallel across all cohorts, with cell_id recorded in `self.cohorts`. This should give clearer and faster code by restricting the amount of time spent in `cell_id` loops. 

* The `communities` attribute disappears.
* The following attributes go from `dict[int, XYZ]` to just `XYZ`
    * `self.biomasses` (This change probably has the most engineering and, since we want to rework the internals of Biomasses anyay, probably best done as a separate issue, although the temporary glue holding the join together would be finicky)
    * `self.stem_allocations`
    * `self.growth_increments`
    * `self.per_stem_gpp` (probably becomes a field in `self.cohorts`)
    * `self.per_stem_transpiration` (also probably becomes a field in `self.cohorts`)
* The `self.canopies` attribute stays as `dict[cell_id, Canopy]`

The update flow would be:

1. Mortality - applied across all cohorts (drop  current `cell_id` loop)
2. Recruitment - applied across all cohorts (drop  current `cell_id` loop)
3. Update allometry - a single StemAllometry object across cohorts. (drop  current `cell_id` loop updating community objects).
4. Herbivory - applied across all cohorts
5. Calculate canopies - CELL level calculations (store cohort indices on canopy object).
6. Update canopy layers - CELL level calculations.
7. Subcanopy light capture - already in parallel across cells
8. Calculate light use efficiency (already at cell by layer level).
9. Estimate GPP - CELL level calculations, extract cohorts in each cell,
10. Calculate nutrients - applied across all cohorts
11. Allocate GPP - applied across all cohorts
12. Subcanopy dynamics - already in parallel

davidorme:
Note that it is markedly faster to update an array and then insert into a `pd.DataFrame` than it is to update a column in place:

```python
import pandas as pd
import numpy as np
import timeit

rng = np.random.default_rng()

cells = np.arange(10)
cohorts = pd.DataFrame(
    dict(cell_id = rng.choice(cells, 200), data=rng.random(200), cell_val=np.nan)
)

def cell_function(cohorts):

    return (cohorts["data"] + cohorts["cell_id"])

def update_with_pandas(cohorts):

    for cell in cells:

        idx = cohorts['cell_id'] == cell
        cohorts.loc[idx, "cell_val"] = cell_function(cohorts.iloc[idx])

    return cohorts

[...]

PR #1752, opened 2026-07-30
Biomass update

### LINKED ISSUE #1747: Update biomasses to cohort based representation
As note in #1744, removing the community structure has most impact on Biomasses. There's also a lot of existing biomasses commentary and TODO about moving from a dictionary representation to a faster and cleaner array based representation.

This issue is to develop code to move to this new architecture.

davidorme:
# Description

This PR updates the `biomasses` module, replacing it with `biomasses_new`. All of the model functionality has been shifted over to `biomasses_new` but the old model has been kept for now for comparison. 

The main changes are:

* BiomassTissueABC subclasses no longer have carbon mass and then dictionaries of elemental masses and properties. Each tissue now just has a single elemental biomass array (n_cohorts as row, CNP as columns) and then two similar arrays containing per cohort ideal and turnover element ratios. These ratio arrays _include_ carbon - the first column is always 1.

* The calculation of fruit and seed carbon biomasses (allometry, allocation and growth) are calculated outside of biomasses and added as specific attributes on incoming Cohorts, StemAllometry, StemAllocation and GrowthIncrements. 

* This change means that the tissue specific derived classes can be defined simply by setting class attributes that identify the specific tissue attribute names on the various inputs. All of the functional machinery is then provided by the base class, giving a much cleaner interface.

The biomasses class is then updated to use those changed internals of the individual tissues. The plants model code is then also updated to match and the tests updated.

Note that the previous version generated biomass arrays for use in calculations that were structured as (CNP as row, n_cohorts as columns). For consistency with cohorts and other structures, biomass arrays are now transposed and so indexing and axis sums have been reversed in several places.

> [!IMPORTANT]
> Despite the branch name, this PR still retains the nesting of biomasses within communities. It would be easy to swap over to pure cohorts (#1744)
 
Fixes #1747

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/plants/biomasses_new.py
```diff
@@ -0,0 +1,691 @@
...
+          the tissues are at their ideal ratio, then the elemental masses will not
+          change. However, if the nutrient inputs within a timestep are not sufficient
+          to maintain the current ratios of replacement tissue, then the tissue
+          elemental masses should decrease. This is handled by recording the lost
+          elemental masses in the stem-wide element surplus pools and balancing the
+          accumulated deficits and gains to those pools at the end of the allocation
+          process.
+
+        TODO: This needs to account for herbivory - proportional to herbivory loss?
+
```
sallymatson, line 569: I'm a bit lost on whether this is happening or not? Does this TODO confirm not? I thought last time we chatted you said it has been incorporated. :)

sallymatson (approved):
Finally had a chance to go through and looks good to me. One clarification in there.

virtual_ecosystem/models/plants/biomasses_new.py
davidorme: Yup - this is out of date.

PR #1754, opened 2026-07-30
Fix warnings in `BiomassTissueABC.Cx_ratio`

### LINKED ISSUE #1753: Better handling of zero divides in BiomassTissueABC.Cx_ratio
At the moment, `ve_run` generates a lot of warnings about zero divides in biomass calculations. This is because we are using `np.where` to trap zero divides in calculating Cx ratios. This is something that actually happens - saplings with no fruit for example. We can switch this to use `np.divide(... where=..., out=...)` which does not evaluate the masked cases and so does not trigger the warning.

davidorme:
# Description

This swaps out the numpy handling of calculating Cx ratios to a form that doesn't generate warnings when handling zero divides. The previous code handled the case but was gobby about it, because it had to evaluate the zero divides even though they were not passed on

Fixes #1753

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/plants/biomasses.py
+++ b/virtual_ecosystem/models/plants/biomasses.py
@@ -229,9 +229,8 @@ class BiomassTissueABC(ABC):
     def Cx_ratio(self) -> dict[str, NDArray[np.floating]]:
         """Get the carbon to element ratio for the tissue type.
 
-        This has to handle cases where a tissue has no biomass at all or no actual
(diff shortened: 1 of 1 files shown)


tannerdulay (approved):
This looks good to me - initiates an array of the same shape as carbon_mass populated with infs and replaces them with the actual quotients where they exist, leaving infs where a divide-by-zero error would have occurred.

PR #1758, opened 2026-07-31
Fixes zero nutrient edge case

### LINKED ISSUE #1757: Plant Model generating `np.nan` element masses.
We're getting pathological behaviour in the plants model where all the tissues in a plant have no P at all. Obviously there's a wider problem but the immediate issue is that the element balancing across tissues in plant biomasses calculates the relative masses of elements in each tissue:

```python
> v = np.array([0,0,0,0,0])
> v / v.sum()
np.array([np.nan, np.nan, np.nan, np.nan, np.nan,])
```

So - we shouldn't get to zero nutrients but if we do, it isn't a state that should crash the model. The calculation fails in the bit that calculates how much tissues can be depleted to balance overall tree deficits, so we can simply set these values to zero: do not take elemental masses from tissues that have none.

davidorme:
# Description

This PR just replaces a np.array division with safer version using `np.divide` to catch divide by zero when trying to estimate the relative masses of an array of zero values. The change stops `np.nan` values appearing in plant biomass elemental masses and the zero value assignment prevents tissues with no elemental mass being further depleted (to negative masses!) to balance nutrient deficits.

Fixes #1757

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/plants/biomasses.py
+++ b/virtual_ecosystem/models/plants/biomasses.py
@@ -710,6 +710,7 @@ class StemBiomass(BiomassTissueABC):
                 actual_element_mass=carbon_mass / ideal_ratio,
                 turnover_ratio=turnover_ratio,
             )
+            pass
 
         return cls(
             carbon_mass=carbon_mass,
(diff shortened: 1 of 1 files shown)


tannerdulay (approved):
Fixed with same method as #1754 - looks good to me!

PR #1771, opened 2026-08-06
Add water limitation

### LINKED ISSUE #1770: Add initial model of water limitation to PlantsModel.
Implement water limitation following the discussion below.

Orme, David
05/08/2026 14:27

> So, for the moment, what I think we have is an average limitation over the time step, given the mean soil moisture
> 
> Data provides "soil_moisture" in mm, the hydrology configuration provides "soil_moisture_residual" (This looks like it might become a cell specific array at some point, the function is set up for it, but currently not).
Plants estimates the total cell demand as the sum of per stem transpiration across all cohorts in mm.
We can then calculate GPP moisture limitation factor (GMLF) as: min(1, (soil_moisture - soil_moisture_residual) / total_demand). That will be 1 under well watered conditions but < 1 when available soil moisture is insufficient.
We can then simply scale GPP and actual Transpiration by GMLF.
Vivienne and Jacob - do I need to worry about soil layers? I can simply take the sum across layers as the total available water. I'm feeding back vertically structured transpiration in canopy layers as before rather than explicitly taking water from the soil? Should I handle the residual soil moisture per layer - presumably the residual mm varies with the layer thickness?
>
> Oh bugger it. Need to factor in subcanopy respiration too. 

Groner, Vivienne P
05/08/2026 14:33
> 'soil_moisture_residual' is a constant that does not change, it's a function of soil type and currently assumed to be the same everywhere
>
> The hydrology model assumed that the plants take water out of the second soil layer, the first one is really just for exchange with the atmoshere

Orme, David
05/08/2026 14:53
> Ok, so I should ignore the e.g. 125mm in topsoil and just limit on the e.g. 375mm - residual in the subsoil layer(s)?
>
> That doesn't feel totally realistic, especially for the subcanopy, but the immediate aim is to add limitation without screwing up other models

Groner, Vivienne P
05/08/2026 14:57
> it is probably different for the subcanopy because these take water from the topsoil, but I don't think we assume trees taking water out of the top 10cm? This should probably be even thinner...
>
> it helps with the distribution/ separation of processes
>
> the water runs from the topsoil to the next soil layers, then the plants come, so it is the same water just a bit delayed
>
> which you won't notice 😉

Orme, David
05/08/2026 14:59
> Does it screw things up if subcanopy depletes topsoil water?
>
> Because that actually helps if I don't have to aggregate water demand
>
> If it does, no problem, just more complex

Groner, Vivienne P
05/08/2026 15:02
> it should be fine, it just needs to all happen in a sensible order. at the moment it's just the sum over all layers
>
> it will need some fiddling 
> so maybe in the first instance try all from the same layer

Orme, David
05/08/2026 15:04
> That's probably where we'd end up anyway, needing to at least consider total water demand from canopy and subcanopy even if they end up targeting different layers

Cook, Jacob
05/08/2026 15:07
> I don't think there's any extra considerations from a soil model perspective
>
> basically if whatever you implement works okay for the hydrology model, the soil model calculations (which are all nutrient tracking really) are downstream of that so should still work fine

davidorme:
# Description

This PR adds water limitation:

* The `Subcanopy.calculate_dynamics` method has been split into `Subcanopy.estimate_gpp` and `Subcanopy.calculate_dynamics`, allowing extra steps to be positioned between the calculation of GPP and allocation of that GPP.
* The `PlantsModel` has two new methods: `calculate_daily_water_demand()` and `apply_water_limitation()`. These could be collapsed into one method, but it is useful in testing to be able to calculate demand and then manipulate inputs before applying limitation.
* It adds a new test that does just that - calculates the water demand of the fixture plant model used in testing and then sets up the soil moisture to result in some simple expectations. It then checks that all of the resulting quantities have been penalised following those simple expectations: subcanopy and canopy GPP and transpiration and the reported transpiration used by hydrology.

I have a feeling that that actual quantities aren't scaled correctly but I'd rather fix this in a separate PR.

Fixes #1770

## Type of change

## Key checklist

## Further checks


davidorme:
@vgro I think it would be cleaner if `data["transpiration"]` used mm/day not total mm for the update but we can patch that later.

sallymatson (commented):
This looks good to me from the plants side - I like the way you've done the implementation. I'll leave it for Vivi to explicitly approve, given that it is highly connected & important for hydrology. :)

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -559,6 +579,15 @@ def _setup(
             1 - model_constants.per_propagule_annual_recruitment_probability
         ) ** (1 / self.model_timing.updates_per_year)
 
+        # Calculate the residual soil water limit in mm(currently a global value) as the
+        # product of the total depth of subsoil layers in mm and the hydrology soil
+        # moisture residual constant.
+        self.soil_water_residual = (
```
vgro, line 589: The water taken up by plants is removed from the second soil layer only, not any layers below that, so I would suggest replacing ` self.layer_structure.soil_layer_thickness[1:].sum()` with `self.layer_structure.soil_layer_thickness[1]`

virtual_ecosystem/models/plants/plants_model.py
```diff
@@ -1191,6 +1218,69 @@ def estimate_gpp(self, time_index: int) -> None:
...
+        calculates a soil moisture limitation factor as the simple ratio of
+        available water over total demand. This factor is then applied to per stem
+        GPP and transpiration estimates and to subcanopy GPP and transpiration.
+        """
+
+        water_limitation_factor = np.minimum(
+            1,
+            (
+                self.data["soil_moisture"]
+                .sel(layers=self.layer_structure.index_subsoil)
```
vgro: See above, this should only access the second soil layer

vgro (changes requested):
This approach looks good to me, my only request is to change the layers that the water is takes from. In reality, plants might go to deeper levels but with the current bucket hydrology model, the root water uptake is by definition from soil layer 2.
Is the returned transpiration unit per day or is it aggregated over the month in `data`?

davidorme:
> This approach looks good to me, my only request is to change the layers that the water is takes from. In reality, plants might go to deeper levels but with the current bucket hydrology model, the root water uptake is by definition from soil layer 2. 

@vgro: done could you check and approve

> Is the returned transpiration unit per day or is it aggregated over the month in `data`?

It is aggregated but then you already scale back to daily in `hydrology`:

https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/f0100f33bc10f073dc6a94399a0f91c327700670/virtual_ecosystem/models/hydrology/hydrology_tools.py#L203-L205

It might be better to make this explicitly mean daily transpiration in the outputs, rather than the more obscure 'per timestep' unit?

vgro:
> > This approach looks good to me, my only request is to change the layers that the water is takes from. In reality, plants might go to deeper levels but with the current bucket hydrology model, the root water uptake is by definition from soil layer 2.
> 
> vgro: done could you check and approve
> 
> > Is the returned transpiration unit per day or is it aggregated over the month in `data`?
> 
> It is aggregated but then you already scale back to daily in `hydrology`:
> 
> https://github.com/ImperialCollegeLondon/virtual_ecosystem/blob/f0100f33bc10f073dc6a94399a0f91c327700670/virtual_ecosystem/models/hydrology/hydrology_tools.py#L203-L205
> 
> It might be better to make this explicitly mean daily transpiration in the outputs, rather than the more obscure 'per timestep' unit?

That's a good point but I think I prefer the totals per timestep as the rainfall, runoff etc are all sums per timestep, it helps with the water balance.

vgro (approved):
LGTM

PR #1775, opened 2026-08-11
Move fungal fruiting bodies into the soil model

### LINKED ISSUE #1528: Move "fungal_fruiting_bodies" into the soil model
At the moment the "fungal_fruiting_bodies" variable is initialised in the animal model. This has caused confusion for @annarallings and others as it means the animal model is listed as requiring something that's clearly soil data to setup.

Fixing this isn't a top priority, but I should think about moving this variable across to the soil model to make the situation clearer. When I make this migration I should remove the special handling for fungal fruiting bodies and replace it with the `ArrayResources` interface

### LINKED ISSUE #1600: Fungal fruiting bodies should be accessible to below ground animal taxa
Two reasons for this:

1) Fungal fruiting bodies generally grow out of the ground, so it's slightly arbitrary to restrict below ground animals from consuming them just because they are mostly above ground
2) Truffles exists (though I have no idea how common they are in Borneo)

Because of (1) and because parameterising a separate "truffle" pool for below ground fruiting bodies, my plan is to retain a single pool which is accessible to any taxa that can occupy either the "ground" or "below-ground" strata

jacobcook1995:
# Description

This PR moves the fungal fruiting bodies pool from the animal model into the soil model (which is a much more natural home for this). The customs classes in the animal model to handle fungal fruiting body consumption have now been deleted and replaced with the standard `ArrayResources` interface. Finally, I changed the fungal pools so that they are accessible to both ground dwelling and below-ground taxa

@TaranRallings, sorry this is an absolutely massive PR to review. I guess where I specifically need your input is with the changes to the foraging functions. Most of the other changes are moving clunky code from the animal model to the soil model, the code is less clunky (because the soil model is a more natural place for it), but remains a bit clunky. But I don't think fixing that clunkiness is relevant for this PR

@hrlai, also tagging you in, because there are some changes for you to be aware of:

1) The `fungal_fruiting_bodies` pool is now `fungal_fruiting_bodies_cnp` which is a biomass triplet, so once this PR is merged and a release is made (unlikely to happen for a while because of school summer holidays), the variable will need to be renamed and estimates of N and P contents will have to also be included with the data (best place to see the changes is `virtual_ecosystem/example_data/generation_scripts/soil_example_data.py`)

2) I've deleted the two `fungal_fruiting_bodies_c_nutrient_ratio` constants, as the ratios are now explicitly tracked, and retaining these constants felt like unnecessary bloat. The downstream consequence of this is that the stoichiometric ratio of the fungal fruiting bodies is determined by the stoichiometric ratios of the soil fungi. This is somewhat inaccurate (fruiting bodies tend to be somewhat nutrient enriched relative to soil mycelium), but seems like an assumption we can get away with to reduce the number of parameters

Fixes #1528 
Fixes #1600 

## Type of change

## Key checklist

## Further checks


hrlai:
Thanks for the heads-up. Changing to `fungal_fruiting_bodies_cnp` in the soil input data should be straightforward, and I'll probably use fungal fruiting body stoichiometry to convert N and P from C. I'll also take note of the sporocarp vs. mycelium stoichiometry. I think I have come across both before, with sporocarp having more published stoichiometric datasets.

TaranRallings (approved):
There is a lot of changes but, ultimately, what you're doing to the animal model here seems pretty straight forwardly a shift from bespoke shroom management to following the plant system. What from I can see, that was done correctly. 

It's harder for me to tell with the internal soil stuff but everything seemed reasonable enough, at least with my general ignorance of your internals.

PR #1786, opened 2026-08-17
Add fruit + seed breakdown dynamics

### LINKED ISSUE #1649: Finalising fruit/seed input to litter

jacobcook1995:
# Description

This PR makes a number of changes to the flows between the plant/animal/soil/litter models:

1) The `"subcanopy_seedbank_litter_cnp"` now gets added to the litter model, this flow represents the turnover in biomass from the subcanopy seedbank. (I think this change is fairly uncontroversial)

2) Instead of animals consuming `seed_turnover_cnp` and `fruit_turnover_cnp`  they now consume the `fallen_seeds_cnp` and `fallen_fruit_cnp` pools

3) The plant model now populates the `fallen_seeds_cnp` and `fallen_fruit_cnp` based on the rates of animal consumption of each pool and the turnovers rates `seed_turnover_cnp` and `fruit_turnover_cnp`. Seeds do not have a decay term, but fruits do, and this decay is applied to the fruit left over after animal consumption

4) The decay from the `fallen_fruit_cnp` pool is then added directly to the soil model (to the low molecular weight carbon pools). This is to prevent a situation where fruit have to decay twice (i.e. once to become litter and once to become soil)

Along with the substantial changes above I also retired the following variables as they were included in the registry but not actually used for anything: `fallen_seeds_per_fruit`, `fallen_fruit_n`, `canopy_seeds_per_fruit`, and `canopy_fruit_n`. I'm happy to reverse this if there's a strong reason to

Further, I deleted all code related to the `plant_reproductive_tissue_turnover_cnp` variable as a combined "reproductive tissue" biomass input is no longer needed by the soil model.

(this branch also contains the changes in #1775, so can't be merged before that is merged)

@arne-exe, I'm tagging you in to this PR as it changes the constants used by the plants model and I thought you should be aware.

Firstly, I got rid of the `plant_reproductive_tissue_lignin` constant, as this is a weird hangover from when the litter model expected a combined input of plant "reproductive tissues" (i.e. fruits, seeds and flowers). That's an absurd thing to try to estimate from data, so it makes a lot of sense to get rid of it

Secondly, I've added a `fallen_fruit_decay_rate`. This represents the rate at which fruit on the forest floor decays, it is currently a constant across all PFTs (but it could be made PFT specific if there's a strong argument for doing so). The rate is expressed in per degree day units (relative to zero C as a basis). So, decay is linear with both time and temperature (above zero C). This is obviously a very crude model of fruit decay, but hopefully is sufficient to capture the main pattern we want to capture (i.e. the fruit that falls in the UK in October sticks around a few months as a resource to support animals, but fruit that drops in a tropical rainforest is gone in a matter of days)

Let me know if that makes sense to you, if not I am happy to explain further and/or change the implementation!

Fixes #1649 

## Type of change

## Key checklist

## Further checks

- [N/a] Relevant documentation reviewed and updated


davidorme (approved):
Great work @jacobcook1995, thanks for leading this.

I've pushed a commit to move the fruit function into its own module (avoids bulking out `PlantModel`,  anticipation of more fruit function in future, easier to do deeper testing).

Issue #1782, opened 2026-08-13
Fix high subcanopy evaporation


vgro:
The canopy evaporation in the subcanopy is very high which causes the latent heat flux to be strongly negative and eventually air temperatures to drop way below zero. This is actually a consequence of the LAI being close to 30 which is a bit unrealistic. As an intermediate solution, there should be a cap, but in the long term the LAI needs to be sorted.

vgro:
The cap is implemented and needs to be removed when the subcanopy LAI is corrected.

davidorme:
@vgro @tannerdulay @arne-exe 

OK - I've stepped through this using the dummy data out to the third time step. The dummy data shows extreme behaviour because the tree communities are _extremely_ sparse and to a good approximation all the light reaches the subcanopy. Maybe we changed the area of the cells without upscaling the number of individuals[^1]? I don't think that is terribly relevant to the bigger picture issue though. Grasslands should grow without exploding the leaf area index.

I think the main issue here is the subcanopy specific leaf area, which defaults to 14. This leads to the following sequence in the subcanopy layer.

Step 1:

- 0.07 kg m2 vegetation mass -> LAI = ~1
- FAPAR ~ 0.39
- VPD = 1 (!?), Tc = 23
- GPP in month ~ 2 kg / m2
- IWUE ~ 3 µmol / mol
- Transpiration ~ 480 µmol / m2 = 8.89995340e-06 mm

Step 2:

- ~ 0.97 kg m2 vegetation mass -> LAI ~ 14 
- FAPAR ~ 1
- VPD = 180, Tc = 21
- GPP in month ~ 4 kg / m2
- IWUE ~ 38 µmol / mol
- Transpiration ~ 14000 µmol / m2 = ~ 0.00025 mm

Step 3:

- ~ 3.1 kg m2 vegetation mass -> LAI ~ 44 
- FAPAR ~ 1
- VPD = 950, Tc = 30
- GPP in month ~ 3.5 kg / m2
- IWUE ~ 48 µmol / mol
- Transpiration ~ 14000 µmol / m2  = ~ 0.00025 mm

The next step would get ~ 4.6 kg m2 vegetation mass -> LAI ~ 64

If we keep running the simulation, the time series (well, until it crashed) for the vegetation biomass is:

```python
[...]

PR #1790, opened 2026-08-19
1779 new replace dummy data branch

### LINKED ISSUE #1779: replace dummy climate data with maliau outputs

vgro:
This PR updates the `conftest.py` for `abiotic` and `hydrology` model. I deleted the `dummy_climate_data` which was the very first data with all layers filled, and kept only the `dummy_climate_varying_canopy` which has some np.nan canopy layers and is now named `dummy_climate_data`.

I tried to update the values to be more realistic based on the Maliau 2 experiment but it is tricky to get everything consistent for the model specific integration tests. I therefore removed the testing for values from the `hydrology` and `abiotic` model tests.

I also put an intermediate solution in place to avoid the latent heat flux to explode when the understorey canopy gets unrealistically high, described in #1782.

Also a small fix on the surface precipitation, there was one argument not returned properly.

Fixes #1779 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/abiotic/test_abiotic_model.py
+++ b/tests/models/abiotic/test_abiotic_model.py
@@ -2,6 +2,7 @@
 
 from contextlib import nullcontext as does_not_raise
 from logging import ERROR, INFO
+from unittest.mock import patch
 
 import numpy as np
 import pytest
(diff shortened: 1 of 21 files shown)


tests/models/abiotic/test_abiotic_model.py
```diff
@@ -275,83 +281,32 @@ def test_setup_and_update_abiotic_model(
...
-        (valid_values_rel_hum_clean >= 0.0) & (valid_values_rel_hum_clean <= 100.0)
-    ).all()
-
-
-def test_update_warns_for_fractional_days(
-    fixture_abiotic_init_data,
-    dummy_climate_data_varying_canopy,
-    fixture_core_components,
-):
-    """Test warning raised if days are not a whole number of days."""
```
jacobcook1995, line 335: Why is this test being removed?

jacobcook1995 (approved):
The changes to the testing seem sensible to me, had a small question about one of the units tests but that's it

tests/models/abiotic/test_abiotic_model.py
vgro: That was an accident, I aimed to delete all the numbers from the previous test and overlooked this one at the end of the file. I will add it back in.

PR #1796, opened 2026-08-24
np.abs wind speed at reference height

### LINKED ISSUE #1795: handle negative input wind speeds
The reference wind speed is positive or negative depending on the wind direction. Since we do not take direction into account, and to ensure correct computation of wind profiles, the wind speed should be always positive going into the equations.

vgro:
The reference wind speed is positive or negative depending on the wind direction. Since we do not take direction into account, and to ensure correct computation of wind profiles, the wind speed should be always positive going into the equations.

Fixes #1795 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
LGTM! Both the change and the reason for the change are very clearly documented

PR #1809, opened 2026-09-01
Deleted repeated function for soil moisture response

### LINKED ISSUE #1803: `calculate_soil_water_effect_on_litter_decomp` is a repeated function
`calculate_soil_water_effect_on_litter_decomp` and `calculate_water_potential_impact_on_microbes` are the same function one of which is in the litter model and one of which is in the soil model. I should make a generic version of this function that can be imported by both

(I should base this generic function on the soil model version as this has better handling of thresholds)

jacobcook1995:
# Description

This is a small PR to delete a function that is repeated between the soil and litter models. It makes it easier to maintain if only a single copy is kept. The motivation for doing this is that I (slightly) improved the soil model version and forgot to also update the litter model version, something that switching to only defining a single function should prevent this from happening in future.

I'm  tagging the whole code development team as reviewers as this basically just needs a "this isn't a terrible idea" thumbs up rather than any specific and/or detailed feedback

Fixes #1803

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/tests/models/litter/test_env_factors.py
+++ b/tests/models/litter/test_env_factors.py
@@ -26,28 +26,6 @@ def test_calculate_temperature_effect_on_litter_decomp(
     assert np.allclose(actual_factor, expected_factor)
 
 
-def test_calculate_soil_water_effect_on_litter_decomp(
-    dummy_litter_data, fixture_core_components, fixture_litter_constants
-):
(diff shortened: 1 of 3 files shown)


davidorme (approved):
👀 ✅

PR #1810, opened 2026-09-02
Allow data in object to go through during variables check

### LINKED ISSUE #1313: Trying to run static mode without setup (or update) crashes the model
**Describe the bug**
If you follow the [instructions given in the docs for running a model in static mode](https://virtual-ecosystem.readthedocs.io/en/latest/using_the_ve/configuration/config.html#static-models) and take the approach that is supposed to skip the model setup (or update) process the simulation run crashes during initialisation

**To Reproduce**

1. Set up a virtual_ecosystem example simulation
2. Configure a model (e.g. soil) to be static
3. Provide all variables in that models `_vars_populated_by_init` as input data
4. Run the simulation
5. See error, i.e. "ValueError: Variable dissolved_nitrate initialised by soil already initialised by ['data']."

**Expected behavior**
Setup (or update) is skipped

**Additional context**
The question here is really whether this is a bug in the static model that should be fixed, or just a documentation problem (which was my assumption with #1218). An alternative option to make a model static, is to delete all references from it from the config and instead just provide the outputs that it should generate. This gives us a way to skip the update and setup process, but not the setup process by itself

dalonsoa:
# Description

This fixes the static models system, which was broken because the variables system was not allowing for variables meant to be populated by a model to be passed as input data. Now, we weaken that control during the variables setup, just logging a warning, and let the model initialisation process to do the final validation of the configuration.

Fixes #1313 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/core/variables.py
+++ b/virtual_ecosystem/core/variables.py
@@ -368,15 +368,30 @@ def _collect_vars_populated_by_init(
             data .
     """
     for model in models:
+        provided_by_data: list[str] = []
         for var in model.vars_populated_by_init:
             if var in runtime_variables:
-                raise ValueError(
-                    f"Variable {var} initialised by {model.model_name} already "
-                    f"initialised by {runtime_variables[var].vars_populated_by_init}."
-                )
+                # If var is provided by 'data', we let it through with a warning.
+                if "data" in runtime_variables[var].vars_populated_by_init:
+                    provided_by_data.append(var)
+                else:
+                    raise ValueError(
+                        f"Variable {var} initialised by {model.model_name} already "
+                        "initialised by"
+                        f"{runtime_variables[var].vars_populated_by_init}."
+                    )
 
-            runtime_variables[var] = known_variables[var]
-            runtime_variables[var].vars_populated_by_init.append(model.model_name)
+            # We only add to the runtime variables if not already provided by data.
+            if var not in provided_by_data:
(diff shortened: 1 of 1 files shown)


virtual_ecosystem/core/variables.py
```diff
@@ -368,15 +368,30 @@ def _collect_vars_populated_by_init(
...
+            if var not in provided_by_data:
+                runtime_variables[var] = known_variables[var]
+                runtime_variables[var].vars_populated_by_init.append(model.model_name)
+
+        if provided_by_data:
+            LOGGER.warning(
+                f"Variable(s) {provided_by_data} populated by {model.model_name} init "
+                "are provided by the input data instead. This is allowed only if all "
+                "are provided and the model is 'static'"
+            )
```
jacobcook1995, line 394: If the user doesn't followed the allowed procedure here is it caught by something downstream?

jacobcook1995 (approved):
The changes look sensible to me, I just had one query

virtual_ecosystem/core/variables.py
dalonsoa: It will be caught during model initialisation, where the static nature of a model is assessed.

davidorme (approved):
Yup - think the logic here works. We want to postpone model failure until we can assess it in the context of the models and static setup and I think this does that.

PR #1813, opened 2026-09-02
1782 fix high subcanopy evaporation

### LINKED ISSUE #1782: Fix high subcanopy evaporation

davidorme:
# Description

This PR makes some simple changes to control subcanopy LAI:

* It adds a subcanopy leaf fraction config setting (default of 0.5) that removes a fraction of sub canopy vegetation biomass representing stems and structural material before calculating LAI.
* Adds a more hacky cap on maximum subcanopy LAI to avoid abiotic model blowups, controlled by a new config setting. We expect to replace this with more biological models of LAI limitation, but this is a sensible safeguard for now to stop a known issue.
* It removes the existing cap on this that is only implemented in abiotic and makes the cap a general model feature. 

Of note - the example data makes this problem very acute by overestimating the SWD by about an order of magnitude and having essentially open grassland as the scenario.

Fixes #1782 

## Type of change

## Key checklist

## Further checks

### DIFF
--- a/virtual_ecosystem/models/plants/model_config.py
+++ b/virtual_ecosystem/models/plants/model_config.py
@@ -56,9 +56,9 @@ class PlantsConstants(Configuration):
     """The specific leaf area of subcanopy vegetation (m2 kg-1)."""
 
     subcanopy_maximum_leaf_area_index: float = 5
-    """A cap on the maximum leaf area index from the subcanopy. This is an temporary
+    """A cap on the maximum leaf area index from the subcanopy. This is a temporary
     safeguard to prevent subcanopy dynamics from driving unrealistic abiotic conditions
-    and should be replaced by biological limitations on subcanopy growth. (m2 m)"""
+    and should be replaced by biological limitations on subcanopy growth (m2 m-2)."""
 
     subcanopy_respiration_fraction: float = 0.1
     """The fraction of gross primary productivity used in respiration (unitless)."""
--- a/virtual_ecosystem/models/plants/model_config.py
+++ b/virtual_ecosystem/models/plants/model_config.py
@@ -48,8 +48,17 @@ class PlantsConstants(Configuration):
     subcanopy_extinction_coef: float = 0.5
     """The extinction coefficient of subcanopy vegetation (unitless)."""
 
+    subcanopy_leaf_fraction: float = 0.5
+    """The fraction of subcanopy vegetation biomass allocated to leaf tissue rather than
+    to stems and other structural tissue (unitless)."""
+
     subcanopy_specific_leaf_area: float = 14
     """The specific leaf area of subcanopy vegetation (m2 kg-1)."""
+
+    subcanopy_maximum_leaf_area_index: float = 5
+    """A cap on the maximum leaf area index from the subcanopy. This is an temporary
+    safeguard to prevent subcanopy dynamics from driving unrealistic abiotic conditions
+    and should be replaced by biological limitations on subcanopy growth. (m2 m)"""
 
     subcanopy_respiration_fraction: float = 0.1
     """The fraction of gross primary productivity used in respiration (unitless)."""
--- a/virtual_ecosystem/models/plants/subcanopy.py
+++ b/virtual_ecosystem/models/plants/subcanopy.py
@@ -543,11 +543,15 @@ class Subcanopy:
     def set_light_capture(self, below_canopy_light_fraction: NDArray) -> None:
         r"""Calculate the leaf area index and absorption of subcanopy vegetation.
 
-        The subcanopy vegetation is represented as pure leaf biomass (:math:`M_{SC}`, kg
-        m-2), with an associated extinction coefficient (:math:`k`) and specific leaf
-        area (:math:`\sigma`, kg m-2) set in the model constants. These can be used to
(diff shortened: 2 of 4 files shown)


tannerdulay (approved):
All changes look good to me for incorporating temporary subcanopy SLA cap

virtual_ecosystem/models/plants/model_config.py
```diff
@@ -48,9 +48,18 @@ class PlantsConstants(Configuration):
...
+    """The fraction of subcanopy vegetation biomass allocated to leaf tissue rather than
+    to stems and other structural tissue (unitless)."""
+
     subcanopy_specific_leaf_area: float = 14
     """The specific leaf area of subcanopy vegetation (m2 kg-1)."""
 
+    subcanopy_maximum_leaf_area_index: float = 5
+    """A cap on the maximum leaf area index from the subcanopy. This is an temporary
+    safeguard to prevent subcanopy dynamics from driving unrealistic abiotic conditions
+    and should be replaced by biological limitations on subcanopy growth. (m2 m)"""
```
arne-exe:
```suggestion
    and should be replaced by biological limitations on subcanopy growth. (m2 m-2)"""
```

virtual_ecosystem/models/plants/model_config.py
```diff
@@ -48,9 +48,18 @@ class PlantsConstants(Configuration):
...
 
+    subcanopy_leaf_fraction: float = 0.5
+    """The fraction of subcanopy vegetation biomass allocated to leaf tissue rather than
+    to stems and other structural tissue (unitless)."""
+
     subcanopy_specific_leaf_area: float = 14
     """The specific leaf area of subcanopy vegetation (m2 kg-1)."""
 
+    subcanopy_maximum_leaf_area_index: float = 5
+    """A cap on the maximum leaf area index from the subcanopy. This is an temporary
```
arne-exe:
```suggestion
    """A cap on the maximum leaf area index from the subcanopy. This is a temporary
```

virtual_ecosystem/models/plants/subcanopy.py
```diff
@@ -543,11 +543,15 @@ def calculate_dynamics(self) -> None:
...
-        m-2), with an associated extinction coefficient (:math:`k`) and specific leaf
-        area (:math:`\sigma`, kg m-2) set in the model constants. These can be used to
-        calculate the leaf area index (:math:`L`) and hence the absorption fraction
-        (:math:`f_{a}`) of  the subcanopy vegetation layer via the Beer-Lambert law: 
+        The subcanopy vegetation is represented as a mixture of structural and leaf
+        biomass, where the
+        :attr:`virtual_ecosystem.models.plants.model_config.PlantsConstants.subcanopy_leaf_fraction`
+        defines the fraction associated with subcanopy structural tissue. The remaining
+        leaf biomass (:math:`M_{SC}`, kg m-2) has an associated extinction coefficient
+        (:math:`k`) and specific leaf area (:math:`\sigma`, kg m-2) set in the model
```
arne-exe:
```suggestion
        leaf biomass (:math:`M_{SC}`, kg C m-2) has an associated extinction coefficient
        (:math:`k`) and specific leaf area (:math:`\sigma`, m2 kg-1 C) set in the model
```

virtual_ecosystem/models/plants/subcanopy.py
```diff
@@ -543,11 +543,15 @@ def calculate_dynamics(self) -> None:
...
 
-        The subcanopy vegetation is represented as pure leaf biomass (:math:`M_{SC}`, kg
-        m-2), with an associated extinction coefficient (:math:`k`) and specific leaf
-        area (:math:`\sigma`, kg m-2) set in the model constants. These can be used to
-        calculate the leaf area index (:math:`L`) and hence the absorption fraction
-        (:math:`f_{a}`) of  the subcanopy vegetation layer via the Beer-Lambert law: 
+        The subcanopy vegetation is represented as a mixture of structural and leaf
+        biomass, where the
+        :attr:`virtual_ecosystem.models.plants.model_config.PlantsConstants.subcanopy_leaf_fraction`
+        defines the fraction associated with subcanopy structural tissue. The remaining
```
arne-exe: Some more details on how this change affects (or doesn't affect) the components in the subcanopy model would help with interpretation here. It looks like this PR is mainly intended to cap LAI and scale the photosynthetically active tissue mass, rather than introduce a new structural tissue pool with downstream effects on turnover, stoichiometry or allocation.

arne-exe (approved):
A few suggestions on units and description.

virtual_ecosystem/models/plants/subcanopy.py
davidorme: Yes - that's correct. There is still only one pool with associated stoichiometry but not all of the pool is necessarily photosynthetically active. That's partly driven by your comments here:

https://github.com/ImperialCollegeLondon/ve_data_science/issues/547

Having one global fraction probably doesn't make a lot of sense (grassland vs understory shrubs) but we'll worry about that later.

PR #1814, opened 2026-09-03
1651 timeseries

### LINKED ISSUE #1651: Support microclimate time-series input as an alternative to internally derived vertical profiles
Currently, the `abiotic` model uses above-canopy climate at each time step to calculate vertical microclimate profiles internally. We would like to support runs where users already have observed microclimate data and want to use those values rather than deriving profiles internally, for example for `soil` model experiments (@jacobcook1995). To do so, we need an option to instead provide measured microclimate time-series data directly.

This should:

- work with the existing time index system,
- allow users to supply microclimate data at model time steps,
- overwrite/replace the internal microclimate in the `data` object when provided,
- let downstream model components use the supplied microclimate values,
- preserve the current workflow when external microclimate input is absent.

In the long-term, this is probably not limited to the `abiotic` model, but we might want to provide time series of other variable such as those related to vegetation structure.

dalonsoa:
# Description

Makes the `Data` object `time_index` aware, so it can do slicing internally.

While the actual changes are small, the ramifications are huge, and I had used Copilot (Claude Sonnet 5) to help me tracking the changes required to make this work. As a brief summary:

- Where explicit `time_index` selection is needed, the code has been replaced with `get_time_slice`. E.g. `data["downward_longwave_radiation"].isel(time_index=time_index)` becomes `data.get_time_slice("downward_longwave_radiation", time_index)`.
- Where the whole timeseries data is needed, the underlying DataArray object is used, so no implicit slicing takes place. E.g.`init_data[var] = dummy_climate_data[var]` becomes `init_data[var] = dummy_climate_data.data[var]`
- Some fixtures - and tests - were amended to include `time_index` coordinates, now that the validator is actually doing something.
- New unit tests have been added to test the new functionality.

Fixes #1651 

## Type of change

## Key checklist

## Further checks


jacobcook1995 (commented):
The changes to the documentation and the core code look good to me. However, I'll leave it to @vgro and @davidorme  to give the approval thumbs up as it is their science model code that is being changed by this PR

virtual_ecosystem/models/abiotic_simple/abiotic_simple_model.py
```diff
@@ -152,8 +152,8 @@ def _setup(
 
         # calculate vapour pressure deficit at reference height for all time steps
         vapour_pressure_and_deficit = calculate_vapour_pressure_deficit(
-            temperature=self.data["air_temperature_ref"],
-            relative_humidity=self.data["relative_humidity_ref"],
+            temperature=self.data.data["air_temperature_ref"],
+            relative_humidity=self.data.data["relative_humidity_ref"],
```
davidorme, line 156: Do we want to make a special method (`Data.get_all_times`)  for this use case as well? Functionally no different but again makes the code explicit about _why_ this is being used here. Could do the same with comments, but I think there's clarity value in only ever using `self.data.explicit_method()` or `self.data['var']`?

@vgro The issue here is that `self.data["air_temperature_ref"]` now unavoidably returns only a single time slice for the current time step. So @dalonsoa is using `self.data.data` to bypass our home-rolled `__get_item__` that does that time slicing and use the `xarray.Dataset.__get_item__` that gets the whole array.

davidorme: Asides from this, no problems

dalonsoa: Could very well implement that. As you say, it is more explicit. However, if we do that, I would call it `get_time_series` (by comparison to `get_time_slice`) and check that there is indeed a `time_index`, otherwise the method might be misleading.

davidorme: Sounds good to me - and I had given about 1 Planck time's worth of thought to the method name, so very happy to go with a better suggestion 😄

dalonsoa: That's a very nerdy unit of time :laughing:

davidorme: Literally _couldn't_ have given it any less thought 😄

davidorme: I suppose that's not _technically_ true, but I would need to propose a functional theory of quantum gravity to define a smaller amount of time.

davidorme: And that's probably a task for a new feature branch.

vgro: I agree this would be a cleaner way of accessing the timeseries and could be part of the next step/branch

vgro (approved):
This looks good to me 👍

virtual_ecosystem/models/abiotic_simple/abiotic_simple_model.py
dalonsoa: The time series thing, I'm doing it in this branch, as it is related with these changs. The new functional theory of quantum gravity, that's probably better for a separate branch, yes. And it's on David.

Issue #1820, opened 2026-09-07
Explore how to add descriptions and units to plant export data.


davidorme:
All of these will need to have a description and units too, similar to the input data and constants.
Not critical to include in this PR, but we'll need to address this before doing validation of the simulation outputs. I will look into this under https://github.com/ImperialCollegeLondon/ve_data_science/issues/441

_Originally posted by @arne-exe in https://github.com/ImperialCollegeLondon/virtual_ecosystem/pull/1816#discussion_r3946122350_

davidorme:
To add

- need to document the biomass output options
- would be useful to have a shortcut for "ALL" (partly for actually figuring out what it currently can export!)

arne-exe:
Thanks a lot @davidorme, the updated docs are very much appreciated!

PR #1827, opened 2026-09-09
Improve biomasses exportability

### LINKED ISSUE #1826: Improve Biomasses exportability
The Biomasses object currently

* has hardcoded array attributes
* does not have a built in export method
* does not record cohort_id

All of these things are useful for documenting and improving the plant exporter functionality.

davidorme:
# Description

This PR makes usability changes to the Biomasses API:

* Defines the simulation complete set of tissues as a global variable in Biomasses rather than nesting it within PlantsModel. This is partly to expose that list within the model objects, but mostly it is just a more logical place to define the set.
* The class now includes `cohort_id` to allow consistent mapping of rows to cohorts.
* Adds a `to_dataframe` method that exports the elemental mass pools for all tissues _and_ surpluses.
* Removes `Biomasses._array_attrs` - unless we pin `Biomasses` to always use the complete set of tissues [^1], then the actual set tissues in an instance is variable and so class _does not really have_ a constant set of array fields. We'll have to fix the only use of that attribute (in the exporter) some other way - upcoming in #1820.
* Updates tests and docstrings. 

Fixes #1826

## Type of change

## Key checklist

## Further checks

[^1]: We do in `from_config` which makes sense but we don't through `__init__` at present and this is really handy for testing element balancing on a more manage set of 2 tissues, not all 5.

### DIFF
--- a/tests/models/plants/test_biomasses.py
+++ b/tests/models/plants/test_biomasses.py
@@ -301,6 +301,24 @@ def test_Biomasses_from_cohorts(fixture_biomass_components, fixture_biomasses):
         )
 
 
+def test_Biomasses_to_dataframe(fixture_biomass_components):
+    """Test the biomass to_dataframe method works."""
+    from virtual_ecosystem.models.plants.biomasses import Biomasses
+
+    cohorts, allometry, _, _ = fixture_biomass_components
+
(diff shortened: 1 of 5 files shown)


sallymatson (approved):
LGTM :)

PR #1828, opened 2026-09-09
Litter environmental factors documentation update

### LINKED ISSUE #1801: Litter environmental factors documentation
When looking through the documentation for the litter model I've noticed a few things that are incorrectly documented. I need to update the docs to fix these things

- [ ] Carbon use efficiency of litter decay is not affected by temperature. This is a major assumption that isn't currently documented
- [x] the "responsiveness" parameter of the litter temperature response function, is better described as setting the low and high temperature limits of the function.
- [x] litter decay moisture response function is incorrect as it doesn't provide the behaviour above and below the thresholds (these are currently only described in the text)

jacobcook1995:
# Description

Two of the changes I have made a pretty minor. They are:

1. Giving a proper explanation of what a "response" parameter means in terms of the shape of the function
2. Correcting the limits given for various environmental functions

The substantive change I've made is adding the equations for the calculation of litter mineralisation (of carbon, nitrogen and phosphorus), litter respiration and lignin loss from litter. This entails adding a fair amount of documentation, so if you can see anyway that it could be made clearer let me know!

This is [relevant branch of the docs to look at](https://virtual-ecosystem.readthedocs.io/en/1801-litter-environmental-factors-documentation/virtual_ecosystem/theory/soil/summary.html).

Fixes #1801

### DIFF
--- a/docs/source/virtual_ecosystem/theory/soil/litter_theory.md
+++ b/docs/source/virtual_ecosystem/theory/soil/litter_theory.md
@@ -171,15 +171,14 @@ rates.
 The decay of all litter pools are assumed to follow linear kinetics, with the rate of
 change in the carbon mass, $P$, of litter pool $j$ being given by
 
-$$\frac{dP_j}{dt} = I_j - K_j(T,\psi,L)P_j,$$
+$$\frac{dP_j}{dt} = I_{C,j} - K_j(T,\psi,L)P_j,$$
 
-where $I_j$ is the rate of input to pool $j$ (in carbon terms), $T$ is the temperature,
-$\psi$ is the soil water potential and $K_j(T,\psi,L)$ is the decay rate of the pool.
-The decay rate of the pool will always depend on temperature, but only for the
-below-ground pools will it be affected by soil water potential. Further, lignin content
-only affects the decay rates of pools which contain lignin (i.e. structural and woody
-pools). As an example, the decay rate of the below-ground structural litter pool is
-calculated as
+where $I_{C,j}$ is the rate of carbon input to pool $j$, $T$ is the temperature, $\psi$
+is the soil water potential and $K_j(T,\psi,L)$ is the decay rate of the pool. The decay
+rate of the pool will always depend on temperature, but only for the below-ground pools
+will it be affected by soil water potential. Further, lignin content only affects the
+decay rates of pools which contain lignin (i.e. structural and woody pools). As an
+example, the decay rate of the below-ground structural litter pool is calculated as
 
 $$K_{bs}(T,\psi,L) = k_{bs} * f_t(T) * A(\psi) * f_l(L),$$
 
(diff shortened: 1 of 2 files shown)


docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
-decay of litter exceeds the initial litter pool size. Total loss of the chemicals is
-then found either using the initial pool chemistry (if total decay is less than the
-initial pool size), or a weighted average of the initial pool chemistry and the
-chemistry of the input biomass.
+where $P_{j,0}$ is the size of litter pool $j$ at the start of the update interval.
+
+## Litter mineralisation
+
+We now want to find the amount of carbon that is mineralised into the soil, the amount
+of carbon respired from the litter, the amount of nitrogen and phosphorus mineralised
```
hrlai: N and P mineralisation can be combine with C mineralisation above to be less mouthful in this sentence?

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+into the soil, and the amount of lignin lost by the litter pools. To calculate all these
+things we first have to find the total carbon loss from each litter pool, which is given
+by
+
+$$
+L_j(\tau) = P_{j,0} + \Theta_j(\tau) - P_j(\tau).
```
hrlai: It took me a while to understand this. I had to rearrange it first to

$$
P_j(\tau) - P_{j,0} = \Theta_j(\tau) - L_j(\tau)
$$

I think it is fine as is, but if space allow then perhaps consider adding this line first?

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+$$
+
+where $\Theta_j(\tau)$ is the total input to litter pool $j$ over the time step $\tau$,
```
hrlai: Do you usually call things like $\tau$ "time step" or "time interval"? I tend to do the latter because an interval may consist of multiple time steps in some simulations. But if you have been calling it time step then please just retain it.

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+which is calculated as
+
+$$
+\Theta_j(\tau) = \tau * I_{C,j}.
+$$
+
+### Carbon mineralisation
+
+With the carbon loss from each pool found, it is trivial to calculate the mass of carbon
+mineralised over the simulation time step, which is given by.
```
hrlai:
```suggestion
mineralised over the simulation time step, which is given by:
```

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+
+$$
+M_C(\tau) = \sum^5_{j=1} \epsilon_j * L_j(\tau),
+$$
+
+where $\epsilon_j$ is the carbon use efficiency of litter breakdown for litter pool $j$.
```
hrlai, line 241:
```suggestion
where $\epsilon_j$ is the microbial carbon use efficiency of litter breakdown for litter pool $j$.
```
Correct?

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+
+### Respiration
+
+Any carbon that is not mineralised by definition has to be respired, so total
```
hrlai:
```suggestion
Any remaining carbon that is not mineralised by definition has to be respired, so total
```

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+respiration is calculated as
+
+$$
+R(\tau) = \sum^5_{j=1} (1 - \epsilon_j) * L_j(\tau).
+$$
+
+As a simplifying assumption we assume that these carbon use efficiencies ($\epsilon_j$)
+**do not** vary with temperature. However, litter respiration will still vary with
```
hrlai:
```suggestion
, and hence litter mineralisation, **do not** vary with temperature. However, litter respiration will still vary with
```

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
```diff
@@ -197,19 +196,108 @@ litter pool size at the end of the model update interval ($\tau$). This solution
...
+### Macronutrient mineralisation
+
+By definition all macronutrient loss (nitrogen and phosphorus) from the litter is
+mineralised to the soil. Unfortunately, there is no exact solution for the macronutrient
+losses. Instead, the loss of each macronutrient has to be estimated based on the loss of
+carbon. The estimate is based on the assumption that old litter (i.e. litter that is
+there at the start of the time step) decays first and that litter added during the
+update only decays if the total decay of litter exceeds the initial litter pool size.
+Using this assumption the total mineralisation of macronutrient $k$ over the time step
+length ($\tau$) can be estimated as
```
hrlai: Ah reading to this point, I think $\tau$ as "time step" is better because we call it so in ve_run, and each update corresponds to a time step.

hrlai (approved):
Overall looks good to me, this is a great addition. Only a few minor comments.

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
jacobcook1995: As you mention this in a comment below, the reason I use time step is because it refers to the overall Virtual Ecosystem simulation time step. I'll keep it as "time step" but I'll try and make it a bit clearer what I am referring to

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
jacobcook1995: no, as annoyingly carbon use efficiency has multiple definitions. All of these definitions centre are basically "how much carbon is respired vs how much goes elsewhere", but they differ in where that "elsewhere" is. For microbial CUE (honestly the only sensible definition) "elsewhere" is microbial biomass.

For litter decay what can be measured is how much carbon enters the soil vs how much is respired. So the "CUE" definition is based off of those destinations.

Short of developing a whole new microbial model for the litter (I would enjoy that but there very much isn't the time), we are stuck using these "litter decay CUEs", but I probably should spend a bit more effort explaining them

docs/source/virtual_ecosystem/theory/soil/litter_theory.md
jacobcook1995: Both litter mineralisation and respiration can vary with temperature (because litter decay increases with temp), what doesn't change is the ratio of the two (as that is governed by the fixed CUEs). I'll add a bit of text clarify this

PR #1831, opened 2026-09-11
Adding NetCDF postprocessing

### LINKED ISSUE #1830: Add NetCDF post-process option
The Zarr outputs are efficient to write and can be read as time series, but the format contains a lot of files because data are written as separate files containing chunks per time step and variable. Also, handling Zarr stores isn't quite as widely supported outside Python/xarray.

We can provide a post-processor option to compile the store to a single monolithic NetCDF file, retaining the same group input/init/outputs group structure. Not sure if this will become standard use but it is simple to add and we'll see.

davidorme:
# Description

This PR:

* Defines a new function `ve.core.data.convert_zarr_outputs_to_netcdf` that takes a Zarr store and converts it to NetCDF file.
* The conversion automatically rewraps `cell_id` to `XY` and reverses the data along the Y coordinates to give a map/graph Y axis not an array Y axis.
* Adds a new argument to the CLI `--to-netcdf`/`-n` and feeds that into `ve_run`.
* Adds a final step in `ve_run` that converts the data. It does not delete the Zarr in case something goes wrong.
* Adds a test that generates a Zarr store containing cell_id flattened data and checks the function reconstructs the inputs.
* Updates the `virtual_ecosystem_in...` notebooks to show the option in use. This also updates all the plotting code to use, XY flattening to cell id as necessary.

It would be better for this to skip the zarr step and append to NetCDF but `xarray` does not yet support this (https://github.com/pydata/xarray/issues/1672). Although it is technically possible to use other python netcdf libraries to do this, it's a lot more heavy lifting than just using Zarr and converting.

Fixes #1830

## Type of change

## Key checklist

## Further checks


jacobcook1995 (approved):
Changes all look sensible. My feeling is that we are only going to know whether this solves the overarching problem by testing it in practice, so it makes sense to merge and iterate from there.

My only suggestion (other than the user docs that you've already mentioned) is adding the new entry point intto the existing entry point testing

hrlai (commented):
Overall looks good to me, just a few comments/questions:
1. What is the default setting of `--to-netcdf` ?
2. How hard is it to move the cell-id to xy conversion to happen before both Zarr and netCDF writing? I guess for it to happen before Zarr write/append you'll need to fundamentally revise the plant output function, which is non-trivial? I ask because some of the data science workflow have migrated to Zarr so it'd be great if we could have xy in Zarr as well, but again not crucial. We do have some crude functions to convert cell_id to xy, so when this PR merges we can deprecate those functions, but only if we use netCDF outputs.

Like @jacobcook1995 said we will need to let time tell whether Zarr remains an optional output format or get demoted to an intermediate output.

PR #1836, opened 2026-09-14
Animal - Thermal Response Dispersal

### LINKED ISSUE #1692: Animal - Thermal Response Dispersal
Dispersal needs to be thermally responsive. 

This system should involve a thermal habitat suitability check, running off `sigma_f_t`, prior to dispersal. The highest sigma_f_t option should be selected or randomly select from equal sigmas.

TaranRallings:
# Description
The thermal dispersal and territory system is coming as three PRs; this is the second. It builds the thermal habitat suitability machinery and wires it into dispersal. Territory construction will use the same machinery in a smaller third PR.

This functionality has a config option which defaults to True.  

New things:

- `StratumClimate` - a small dataclass bundling the per-cell, per-stratum climate arrays, built once per timestep by _build_stratum_climate
- `stratum_mean_climate` - averages climate over the strata a functional group occupies, per cell
- `thermal_suitability` - the per-cell analogue of sigma_f_t; delegates to activity_window cell by cell, so the two can't drift. Endotherms get 1.0 everywhere, so the system is inert for them
- `_update_thermal_suitability` and the self.thermal_suitability cache - one per-cell suitability array per functional group, rebuilt each timestep
- `_select_destination` - weights the reachable set by suitability (floored, exponent-tunable) when the toggle is on, falls back to uniform choice when off
- a thermal-escape migration trigger in migrate_community for cohorts whose activity window drops below threshold
- four new config options on AnimalConstants

Things that are true:

- the update order changed: intra-grid migration now runs at the end of the in-territory block (after foraging, birth, metabolism, mortality), so a cohort experiences a whole timestep in one territory and only then relocates for the next. This dissolves a stale-sigma_f_t problem rather than patching it. External migration stays last
_update_thermal_suitability recomputes stratum_mean_climate per functional group rather than deduplicating by occupancy pattern
- the exporter runs after migration, so an exported row shows the post-migration centroid alongside pre-migration process values (there's already a TODO about this in _update)

Fixes #1692 

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/animal/scaling_functions.py
```diff
@@ -1019,3 +1022,122 @@ def activity_window(
...
+                annual_mean_temp=annual_mean_temp,
+                annual_temp_sd=annual_temp_sd,
+                t_opt=t_opt,
+                t_max_crit=t_max_crit,
+                t_min_crit=t_min_crit,
+                constants=constants,
+            )
+            for t, d in zip(temperature, diurnal_temp_range)
+        ]
+    )
```
davidorme, line 1143: Currently all the functions in this chain (`activity_window` -> `t_opt_/min/max` functions) are defined as using float inputs but it looks like they could all be `NDArray` and then drop this iteration over the `zip`? That's likely a big performance bump?

davidorme (changes requested):
Really only got one major comment, which is more of a performance thing than a science thing.

virtual_ecosystem/models/animal/scaling_functions.py
TaranRallings: This sounds like a good idea but I think vectorizing all of the math will be a delicate operation on an already large PR. If its not a performance bump and not a core issue, I'd rather make an issue and do it in a new PR. Would that work for you?

davidorme (approved):
Happy to postpone arrayalization. Or whatever we want to call it...

virtual_ecosystem/models/animal/scaling_functions.py
davidorme: I _think_ all the calculations vectorise automatically, but pushing it to a new PR is fine.

PR #1838, opened 2026-09-16
Add soil/litter respiration outputs

### LINKED ISSUE #1769: Add soil gas fluxes as data object variables
Gas fluxes from the soil are something we do have validation data for, so should be predicted by the model.

For CO2, we already have all the information we need to predict soil and litter respiration. The only weirdness is that the observed soil respiration will have two components (soil + below ground litter). I need to decide whether it is worth outputting "below ground litter respiration" as a thing in its own right, or whether I should use a single variable. While a single variable would be simpler, it would need to be updated by two models (which we have tried to avoid with the models so far).

The other gases fluxes that we can predict are nitrogen compounds. In this case we need to add functionality to convert our existing fluxes (which bundle together multiple compounds) into release rates for specific compounds. There are existing models for this, it will just require a bit of searching

This isn't high priority (yet) as we are still working on the basics of the calibration/validation pipeline, but it is something I should start looking into when I get the time

### LINKED ISSUE #1835: Soil and litter respiration flux
Needs to be added, and is fairly trivially to implement as the soil + litter models already calculate these quantities

jacobcook1995:
# Description

I decided to make a smaller PR just adding soil and litter respiration outputs, rather than all gas fluxes in one, as some more urgent tasks came up. I will return to the other fluxes at a later date.

@hrlai , this PR adds three new output variables, "soil_respiration", "above_ground_litter_respiration" and "below_ground_litter_respiration" (the rationale for this is explained in the additions to the docs). Something I was wondering is whether I should also add "dead_wood_respiration" as an output? This would be an estimate of total respiration from dead wood over the grid cell. Is this something that would be a useful validation target?

[These are the updated docs](https://virtual-ecosystem.readthedocs.io/en/1769-add-soil-gas-fluxes-as-data-object-variables/virtual_ecosystem/theory/soil/summary.html).

Fixes #1835

## Type of change

## Key checklist

## Further checks


virtual_ecosystem/models/soil/soil_model.py
```diff
@@ -169,10 +169,9 @@ class SoilModel(
         "arbuscular_mycorrhizal_p_supply",
         "ectomycorrhizal_n_supply",
         "ectomycorrhizal_p_supply",
+        "soil_respiration",
     ),
-    # TODO - If anything gets added to this section the implementation docs will need to
-    # be updated
-    vars_populated_by_first_update=(),
```
vgro, line 175: Did you do the docs update?

vgro (approved):
Looks sensible to me 👍

virtual_ecosystem/models/soil/soil_model.py
jacobcook1995: This part of the docs referred to had already been changed, so to turned out that this TODO was outdated

PR #1842, opened 2026-09-18
Animal - Thermal Territories

### LINKED ISSUE #1841: Animal - Thermal Territories
A thermal suitability system has already been put in place for dispersal. This system needs to be utilized for adding thermal suitability to territory construction such that a a forest-edge territory would follow the edge instead of including thermally unsuitable areas.

TaranRallings:
# Description

The thermal dispersal and territory system is coming as three PRs; this is the third and last. It generalises bfs_territory into suitability-aware growth and wires it into the one seam that builds every territory the model uses.

Changes:

- new `thermal_territory` - a priority-queue generalisation of bfs_territory; grows into the most suitable adjacent cell rather than the oldest discovered one
- `get_territory_cells and update_community_occupancy` updated to pass the suitability cache through, with None falling back to plain BFS

Things that are true:
- initial cohorts get breadth-first territories, because the climate pass hasn't run yet at setup; they become suitability-aware on first migration. An initial cohort that never migrates keeps a BFS territory for life — acceptable given spinup, but worth knowing
- under a strong gradient, greedy growth can produce a thin tendril along the suitability ridge rather than a compact blob - area is preserved, shape isn't. Watch for this in calibration; if it matters, a compactness term in the priority is a small follow-up

Fixes #1841 

## Type of change

## Key checklist

## Further checks


nickwctan (approved):
At first glance, the science behind this sounds good. Just one question, this mainly only affects ectotherms?

virtual_ecosystem/models/animal/animal_model.py
```diff
@@ -1112,12 +1113,28 @@ def update_community_occupancy(
...
+        breadth-first. They become suitability-aware the first time they are
+        rebuilt, which is to say on the cohort's first migration.
+
         Args:
             cohort: The animal cohort being updates.
             centroid_key: The grid cell key of the anchoring grid cell.
         """
 
-        territory_cells = cohort.get_territory_cells(centroid_key)
+        suitability = (
```
Baizurah: Just checking, so if the thermal suitability changes later, does it means that the territory only get updated when the cohort moves again? @TaranRallings

Baizurah (approved):
LGTM. Finally something that I can relate to :D

 I will give further read this week, just because it's very interesting!

TaranRallings:
> At first glance, the science behind this sounds good. Just one question, this mainly only affects ectotherms?

Correct! This only effects ectotherms.

It's built off the activity window system, which is only in place for ectotherms. Getting thermal response for endotherms is a whole other kettle of fish and would take some significant research and work.

virtual_ecosystem/models/animal/animal_model.py
TaranRallings: That's right. Territory will only be reconstructed after a move. This shouldn't be too much of a problem because a declining thermal suitability will trigger a movement even anyway.

davidorme (changes requested):
Both the new `thermal_territory` and existing `bfs_territory` recalculate cell neighbours for every cohort and every timestep. That can be sidestepped entirely by using the `grid.neighbours` property.

There is an issue here about coordinating between models with setting those neighbours  but that isn't hard to solve - at the moment it is only used by `hydrology` and by `animals` and you and @vgro _both_  use rook move neighbours. So - we can set those neighbours once in the core start up and then they are defined (models might want to check the maximum number of neighbours!).  We'd need to revisit this if you wanted to move to something different (like queen or knight move neighbours).

I think this is an import performance issue - the whole reason for having those neighbours calculated once is to support functions like this. I have prototypes of Grid based functions and one quick test - I can push them to this PR or to a new PR onto it?

